PR #6, opened 2024-03-03
Adding helm chart resources

rmvangun:
Very much a draft Helm chart for now. I created this to get it running on my own home cluster and decided to take it all the way as we need this at work too. Was a great little exercise, no hard feelings if you all are handling this differently or have your own standards here!

Something kind of neat (or maybe a hack) is that this auto-generates the account ID and the GPG key using a little utils Docker image in a pre-install hook.

### DIFF
--- a/charts/omni/Chart.yaml
+++ b/charts/omni/Chart.yaml
@@ -1,0 +1,13 @@
+apiVersion: v2
+appVersion: v0.30.0
+description: Siderolabs Omni for SaaS-simple deployment of Kubernetes
+home: https://github.com/siderolabs/omni
+icon: https://www.siderolabs.com/wp-content/uploads/2021/12/image-8-1.png
+keywords:
+- omni
+- talos
+kubeVersion: '>=1.27.0-0'
+name: omni
+sources:
+- https://github.com/siderolabs/omni
+version: 0.1.0
--- a/charts/omni/README.md
+++ b/charts/omni/README.md
@@ -1,0 +1,3 @@
+omni
+
+Draft helm chart for Omni. Includes the deployment and service. Additionally, this chart automatically generates the account ID and a GPG key.
--- a/charts/omni/templates/_helpers.tpl
+++ b/charts/omni/templates/_helpers.tpl
@@ -1,0 +1,47 @@
+{{/*
+Expand the name of the chart.
+*/}}
+{{- define "omni.name" -}}
+{{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" }}
+{{- end }}
+
+{{/*
+Generate basic labels
+*/}}
+{{- define "omni.labels" -}}
+helm.sh/chart: {{ include "omni.chart" . }}
+{{ include "omni.selectorLabels" . }}
+{{- if .Chart.AppVersion }}
+app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
+{{- end }}
+app.kubernetes.io/managed-by: {{ .Release.Service }}
+{{- end }}
+
+{{/*
+Selector labels
+*/}}
+{{- define "omni.selectorLabels" -}}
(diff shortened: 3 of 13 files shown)


rmvangun:
If you'd like me to continue in this direction, some things I would include:

- Full test suite
- Ingress
- PV & PVC for the internal etcd cluster
- Figure out something to do with that Docker image, would be interested in feedback on that mechanism

I think that's all I would add in the scope of this PR. Pulling in the nginx-ingress chart as a subchart might make sense in order to serve UDP, though I'm not a fan of sub-charts.

Also there are some Omni internals and direction I don't know about, but given a little discussion could probably hone the configuration to be more appropriate.

TimJones:
Thanks for the initial contribution.

Ideally we wouldn't want a Helm chart that does anything dynamically for ease of integration with other tools like Docker Compose, Kustomize, ArgoCD etc so the hooks to generate UUIDs, GPG keys and the like should be replace with config in the values file to be supplied by the operator. We certainly wouldn't want to have to build an entire new container to do this one-time setup.

Probably need to separate out the Services too as the Siderolink/Wireguard can often need completely different settings/traffic policy than the rest of the services.

Ingresses can be complex as we have a lot of controller-specific annotations for timeouts, gRPC settings etc that wouldn't carry over to other controllers.

Happy to work with you on this, but I worry that Omni being a complex service to host is going to require an even-more-complex chart to maintain.

rmvangun:
@TimJones Absolutely agree. This was the result of setting up some initial manifests to launch Omni on my home lab management cluster, for personal use and to prepare to discuss hosting Omni at my day job, and I scoped it to what I could get done in a weekend. I hesitated to share it at all but it started to feel almost useful so I thought, why not?

I have various feelings on Helm in general but have found it very useful especially when incorporated as a Flux [HelmRelease](https://fluxcd.io/flux/components/helm/helmreleases/). In my setup I wrap collections of related manifests in a local Helm chart so I can leverage the `dependsOn` field to build a dependency graph, and to abstract some aspects of the configuration in to `values`. My preferred style is more Kustomize oriented, but in my setup the two work harmoniously together. So, again, Helmifying this was more in line with my own needs and not really meant for general distribution in its current form.

Addressing a few of your points, as I would absolutely still get this in to proper shape based on a discussion and a clear understanding that this aligns with your goals with the Omni project.

#### Auto-generating accountId and GPG key
- The auto-generated stuff doesn't happen if you set `accountId.create: false` and `gpg.create: false`
- I also don't like the idea of a utility Docker image for obvious reasons. I initially did it all entirely with scripts in the job run, but that also doesn't feel very nice
- This mechanism is maybe stepping in to "over-automation" anyway. I think it's a tasteful decision to defer to the operator to set these up. But, for my own in-house setup, if I can automate it, I will

#### Single-service vs. Multi-service
I also initially used a separate service for siderolink/wireguard and fell back to a single service since I wasn't yet sure about the reference architecture. This is an area I want to learn more about and will need to better understand.

#### Ingress
Yes, agreed on the problem with the ingresses and not playing nicely with every controller. What do you all use as your ingress? I imagined using nginx-ingress as it handles UDP.

#### Omni Complexity
At first, it seems as if this is entirely contained in a single Docker image, how simple! But it became apparent how much of the complexity lies in the infrastructure and operation. Which, on one hand, was why a Helm Chart approach seemed attractive.

#### Conclusion
I hear you on the maintenance aspects of this chart. I understand that every line of code is a line of risk and maintenance, so unless you've already considered an Omni Helm chart in your future, I don't want to continue barking up this tree. I mentioned elsewhere that operating this ourselves (at work, not home lab stuff) is likely in our future so thought this might be a good opportunity to collaborate. I'll defer to you all and when you're ready we can discuss the right approach to self-hosting Omni in production.

rmvangun:
Going to wait on further announcements concerning production reference architecture.

PR #50, opened 2024-03-04
fix: properly generate labels meta in the UI

Unix4ever:
Make the UI use proper data format for the machine labels meta value when generating the image on the `DownloadInstallationMedia` page.


smira:
Can we update the schematic at least in the integration test to include the labels and assert on them in the test?

smira (approved):

Unix4ever:
> Can we update the schematic at least in the integration test to include the labels and assert on them in the test?

Yeah, can do that.

Issue #13, opened 2023-12-14
Support Omni join config via machine config


smira:
Talos since 1.5.0 support partial machine config submitted as user-data, which includes machine configuration for SideroLink join and event sink.

When Talos runs on the cloud platform with user-data support, we can use that to use "vanilla" image of Talos with join config submitted via user-data.

The flow goes like that:

1. Vanilla Talos boots on the cloud.
2. Talos discovers join config as partial machine config in the cloud user-data.
3. Talos joins Omni.
4. Omni pushes full machine config to the machine, which **should** include join configuration.
5. Omni resets the STATE partition, wiping the full machine config.
6. Talos reboots, and re-loads the join config from userdata, cycle repeats.

* [SideroLinkConfig](https://www.talos.dev/v1.6/reference/configuration/siderolink/siderolinkconfig/)
* [EventSinkConfig](https://www.talos.dev/v1.6/reference/configuration/runtime/eventsinkconfig/)
* [KmsgLogConfig](https://www.talos.dev/v1.6/reference/configuration/runtime/kmsglogconfig/)

First goal: make Omni respect join config, that is submit it to the machine with the other bits of the machine config (see also siderolabs/omni-archive#1459). 

> Note: in maintenance mode, there's no way to get machine configuration from Talos, so Omni should synthesize and submit join config as one of the "config patches" if it discovers a machine joined to Omni e.g. without proper kernel args.

Second goal: make Omni publish expected join config in the UI/as a resource.

utkuozdemir:
Testing the full cycle requires a release with this fix: https://github.com/siderolabs/talos/commit/1bb6027ccd7c63ae3a012eb310d1e05027ec1f80

Currently, with a partial machine config, the initial install works, but when the machine is reset and tried to be re-added to a new cluster, the triggered maintenance upgrade causes Talos to crash due to the partial config it holds.

smira:
The fix was merged, so it should be unblocked?

utkuozdemir:
> The fix was merged, so it should be unblocked?

Yep, I'm back at this, but now the way I'm doing, the integration test I'm adding won't even require the fix.

PR #68, opened 2024-03-12
feat: preserve maintenance machine configs

utkuozdemir:
When a machine has a partial machine config (e.g., `SideroLinkConfig` or `KmsgLogConfig`) while being in maintenance mode, detect and store it on Omni side.

When this machine has a config applied to it (i.e., when it joins to a cluster) or when its config is updated, always preserve these detected configs in addition to the main (`v1alpha1`) config.

This allows machines to join Omni by using partial configs (for example, by using cloud user data) instead of having kernel args for the SideroLink and kernel log sync configuration.

We do not fail in the cases where we are not authorized to access the COSI resource API, as the full access to the resources over SideroLink is only available in Talos v1.6.5 and above.

Modify the Talos upgrade integration tests to apply a partial config during maintenance mode, and later assert that it was preserved through the upgrade/revert flow.

Additionally, bump the default Talos version and the version used in the integration tests to 1.6.6.

Closes siderolabs/omni#13.


internal/backend/runtime/omni/controllers/omni/internal/task/machine/poll.go
```diff
@@ -340,3 +346,48 @@ func pollExtensions(ctx context.Context, c *client.Client, info *Info) error {
...
 }
+
+func pollMaintenanceConfig(ctx context.Context, c *client.Client, info *Info) error {
+	if !info.MaintenanceMode {
+		return nil
+	}
+
+	conf, err := safe.StateGetByID[*config.MachineConfig](ctx, c.COSI, config.V1Alpha1ID)
+	if err != nil {
+		// if we successfully probed the config and found out that it is not set, we mark it as ready by setting it to a non-nil value by empty config.
```
frezbo, line 357: comment doesn't match as this is check on non-nil error

internal/backend/runtime/omni/controllers/omni/internal/task/machine/poll.go
```diff
@@ -340,3 +346,48 @@ func pollExtensions(ctx context.Context, c *client.Client, info *Info) error {
...
+			return fmt.Errorf("unexpected v1alpha1.Config document in maintenance config: %w", err)
+		}
+	}
+
+	configBytes, err := conf.Container().Bytes()
+	if err != nil {
+		return fmt.Errorf("failed to read maintenance config bytes: %w", err)
+	}
+
+	info.MaintenanceConfig = &specs.MachineStatusSpec_MaintenanceConfig{
```
frezbo, line 388: if we're looping over multiple documents, aren't we setting the last one only?

internal/backend/runtime/omni/controllers/omni/internal/task/machine/poll.go
utkuozdemir: Actually it is intended, but I changed the wording. The idea here is: if the error is a NotFound, it is not an unexpected error, our call actually succeeded in a way - we just found out that there is no config set at all. In this case, we know what the maintenance config is - it is explicitly empty. So we set the value to that.

internal/backend/runtime/omni/controllers/omni/internal/task/machine/poll.go
utkuozdemir: The loop is to make sure that there is no `v1alpha1` document in the config. Then, we get the whole config we found and set it on the `MachineStatus`. This is an example of what we could find:
```
apiVersion: v1alpha1
kind: SideroLinkConfig
apiUrl: grpc://omni.omni-local.utkuozdemir.org:8090?jointoken=w7uVuW3zbVKIYQuzEcyetAHeYMeo5q2L9RvkAVfCfSCD
---
apiVersion: v1alpha1
kind: EventSinkConfig
endpoint: "[fdae:41e4:649b:9303::1]:8090"
---
apiVersion: v1alpha1
kind: KmsgLogConfig
name: omni-kmsg
url: "tcp://[fdae:41e4:649b:9303::1]:8092"
```

The loop is to fail fast if we see a `version: v1alpha1` doc among them (which shouldn't actually happen).

frezbo: oh, nevermind, I missed the loop closure :sweat_smile:

internal/backend/runtime/omni/controllers/omni/internal/task/machine/poll.go
frezbo: i got the intention, just that the comment was confusing since it said if it was successful

frezbo (approved):

Issue #55, opened 2024-03-06
Calculate additional schematic ID in the machine status using extensions list only


Unix4ever:
Then in the `TalosUpgradeStatus` controller compare this stripped schematic ID with desired schematic.
Trigger the upgrade only if they don't match.

PR #74, opened 2024-03-15
feat: compute schematic id only from the extensions

utkuozdemir:
When determining the schematic ID of a machine, instead of relying the ID on the schematic ID meta-extension, compute the ID by gathering the extensions on the machine. This way, the extension ID will not contain the META values, labels or the kernel args.

This ID is actually the ID we need, as when we compare the desired schematic with the actual one during a Talos upgrade, we are only interested in the changes in the list of extensions.

This does not cause the kernel args, labels, etc. to disappear, as they are used at installation time and preserved afterward (e.g., during upgrades).

Additionally:
- Remove the list of extensions from the `Schematic` resource, as it relied upon the schematics always being created through Omni. This is not always the case - i.e., when a partial join config is used. Therefore, instead of relying on it, we store the list of extensions by directly reading them from the machine and storing them on the `MachineStatus` resource.
- Skip setting the schematic META section at all if there are no labels set on Download Installation Media screen.

Closes siderolabs/omni#55.


internal/backend/runtime/omni/controllers/omni/talos_extensions.go
```diff
@@ -30,7 +29,7 @@ import (
 type TalosExtensionsController = qtransform.QController[*omni.TalosVersion, *omni.TalosExtensions]
 
 // NewTalosExtensionsController instantiates the TalosExtensions controller.
-func NewTalosExtensionsController() *TalosExtensionsController {
+func NewTalosExtensionsController(imageFactoryClient *imagefactory.Client) *TalosExtensionsController {
```
utkuozdemir, line 32: Switch to using the new wrapper, so the only point where we create a new image factory GRPC client is actually in `magefactory.Client`

frontend/src/views/omni/Modals/DownloadInstallationMedia.vue
```diff
@@ -311,7 +311,7 @@ const createSchematic = async () => {
       meta_values: {},
     };
 
-    if (labels.value) {
+    if (labels.value && Object.keys(labels.value).length > 0) {
```
utkuozdemir, line 314: If no labels were set, we were setting here a YAML with empty content for the labels key. Now we do not set any META information if there are no labels.

internal/backend/runtime/omni/controllers/omni/cluster_machine_config.go
```diff
@@ -284,7 +284,7 @@ func (clusterMachineConfigControllerHelper) generateConfig(
 		return nil, fmt.Errorf("talos version is not set on the resource %s", clusterConfigVersion.Metadata())
 	}
 
-	installImage, err := buildInstallImage(clusterMachineTalosVersion, machineStatus, talosVersion)
+	installImage, err := buildInstallImage(clusterMachineTalosVersion, machineStatus, talosVersion) // here should come from the full schematic id, with args
```
Unix4ever, line 287: this comment should be dropped I guess

internal/backend/runtime/omni/controllers/omni/cluster_machine_config_status.go
```diff
@@ -345,17 +345,18 @@ func (h *clusterMachineConfigStatusControllerHandler) syncTalosVersionAndSchemat
...
+
 	if actualVersion == expectedVersion && actualSchematic == expectedSchematic {
 		configStatus.TypedSpec().Value.TalosVersion = actualVersion
 		configStatus.TypedSpec().Value.SchematicId = actualSchematic
 
 		return true, nil
 	}
 
-	image, err := buildInstallImage(talosVersion, machineStatus, expectedVersion)
+	image, err := buildInstallImage(talosVersion, machineStatus, expectedVersion) // this should come "plain", without kernel args
```
Unix4ever, line 359: this comment too

internal/backend/runtime/omni/controllers/omni/internal/talos/schematic.go
```diff
@@ -20,30 +20,60 @@ import (
...
-// extension, or calculates vanilla schematic ID if there's none.
-func GetSchematicID(ctx context.Context, c *client.Client) (string, error) {
-	extensions := map[resource.ID]*runtime.ExtensionStatus{}
+// SchematicInfo contains the information about the schematic - the plain schematic ID and the extensions.
+type SchematicInfo struct {
+	ID         string
+	Extensions []string
+}
+
+// GetSchematicInfo uses Talos API list all the schematics, and computes the plain schematic ID,
```
Unix4ever, line 29:
```suggestion
// GetSchematicInfo uses Talos API to list all the schematics, and computes the plain schematic ID,
```

Unix4ever (approved):

frontend/src/api/resources.ts
```diff
@@ -150,7 +150,7 @@ export const KubernetesUsageType = "KubernetesUsages.omni.sidero.dev";
 export const PermissionsID = "permissions";
 export const PermissionsType = "Permissions.omni.sidero.dev";
 export const SecureBoot = "secureboot";
-export const DefaultTalosVersion = "1.6.4";
+export const DefaultTalosVersion = "1.6.6";
```
frezbo, line 153: 1.6.7

frezbo (approved):

utkuozdemir:
🙏🏻

PR #77, opened 2024-03-18
fix: properly handle upgrades for the machines with invalid schematics

Unix4ever:
Simple fix to not skip updating Talos if the machine schematic is empty.

Verified that on a cluster running on the nodes with the invalid schematics. Upgrades work.

### DIFF
--- a/internal/backend/runtime/omni/controllers/omni/cluster_machine_config_status.go
+++ b/internal/backend/runtime/omni/controllers/omni/cluster_machine_config_status.go
@@ -129,7 +129,7 @@ func NewClusterMachineConfigStatusController() *ClusterMachineConfigStatusContro
 					machineStatus.TypedSpec().Value.Schematic.Id != expectedSchematic
 
 				// don't run the upgrade check if the running version and expected versions match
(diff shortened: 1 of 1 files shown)


internal/backend/runtime/omni/controllers/omni/cluster_machine_config_status.go
```diff
@@ -129,7 +129,7 @@ func NewClusterMachineConfigStatusController() *ClusterMachineConfigStatusContro
 					machineStatus.TypedSpec().Value.Schematic.Id != expectedSchematic
 
 				// don't run the upgrade check if the running version and expected versions match
-				if versionMismatch && talosVersion.TypedSpec().Value.TalosVersion != "" && talosVersion.TypedSpec().Value.SchematicId != "" {
+				if versionMismatch && talosVersion.TypedSpec().Value.TalosVersion != "" {
```
smira, line 132: how does it work with "invalid" schematics? is it going to wipe it?

Unix4ever: No, it's going to keep using the vanilla installer.
This code is in place, but this check made it skip the upgrade function as a whole.

smira (approved):

Issue #45, opened 2023-11-08
Specifying schematic in a cluster template


smira:
```yaml
Machine:
   systemExtensions:
       -tailscale
       - zfs
...
Cluster:
   systemExtensions:
       -tailscale
       - zfs
...
MachineSet (ControlPlane/Workers):
   systemExtensions:
       - zfs
```  

(systemExtensions are not merged, they override each other)

`systemExtensions` is mapped to schematic ID, and `SchematicConfiguration` with proper labels is generated.

PR #85, opened 2024-03-25
feat: support setting extensions list in the cluster template

Unix4ever:
Fixes: https://github.com/siderolabs/omni/issues/45

Introduced new resource type `ExtensionsConfiguration` that allows
setting machine extensions list.

`SchematicConfiguration` is now readonly and is created by
`SchematicConfigurationController` from `ExtensionsConfiguration`
resource. It also ensures that schematic exists in the image factory by
calling the API.

This change is required to simplify the flow in the cluster templates
(no need to call `CreateSchematic` for each resource).

Export command support added as well.
Added cleanup hooks for the `ExtensionsConfiguration` for machine set, machine and cluster levels.

Changed the resource format to use `labels` instead of `target`. Now
it's the same as for config patches, except it doesn't merge several
resources, but gets the first one.

Signed-off-by: Artem Chernyshev <artem.chernyshev@talos-systems.com>


client/pkg/template/order.go
```diff
@@ -16,10 +16,11 @@ import (
 
 // Canonical order of resources in the generated list.
 var canonicalResourceOrder = map[resource.Type]int{
-	omni.ClusterType:        1,
-	omni.ConfigPatchType:    2,
-	omni.MachineSetType:     3,
-	omni.MachineSetNodeType: 4,
+	omni.ClusterType:                 1,
+	omni.ExtensionsConfigurationType: 2,
```
frezbo, line 20: would this affect for an existing cluster going from no extensions to extensions?

internal/backend/runtime/omni/controllers/omni/machine_status_test.go
```diff
@@ -306,7 +309,8 @@ func (suite *MachineStatusSuite) TestMachineSchematic() {
 				},
 			},
 			expected: &specs.MachineStatusSpec_Schematic{
-				Id: "7d79f1ce28d7e6c099bc89ccf02238fb574165eb4834c2abf2a61eab998d4dc6",
+				Id:               "7d79f1ce28d7e6c099bc89ccf02238fb574165eb4834c2abf2a61eab998d4dc6",
```
frezbo, line 312: probably add a comment showing the yaml data used to produce this input

frezbo (approved):

client/pkg/omni/resources/omni/extensions_configuration_status.go
```diff
@@ -0,0 +1,48 @@
...
+// ExtensionsConfigurationStatusExtension provides auxiliary methods for ExtensionsConfigurationStatus resource.
+type ExtensionsConfigurationStatusExtension struct{}
+
+// ResourceDefinition implements [typed.Extension] interface.
+func (ExtensionsConfigurationStatusExtension) ResourceDefinition() meta.ResourceDefinitionSpec {
+	return meta.ResourceDefinitionSpec{
+		Type:             ExtensionsConfigurationStatusType,
+		Aliases:          []resource.Type{},
+		DefaultNamespace: resources.DefaultNamespace,
+		PrintColumns:     []meta.PrintColumn{},
```
utkuozdemir, line 46: maybe let's add the extensions as the print column?

client/pkg/template/internal/models/machine.go
```diff
@@ -130,14 +131,22 @@ func (machine *Machine) Translate(context TranslateContext) ([]resource.Resource
...
 		return nil, err
 	}
 
-	return append(resourceList, patches...), nil
+	resourceList = append(resourceList, patches...)
+
+	schematicConfigurations := machine.SystemExtensions.translate(
+		ctx,
+		string(machine.Name),
+		pair.MakePair(omni.LabelClusterMachine, string(machine.Name)),
```
utkuozdemir, line 146: should this also have the cluster label, like the config patches?

Unix4ever: it already has it. It adds it inside `translate` func

client/pkg/template/internal/models/models.go
```diff
@@ -30,6 +34,35 @@ type TranslateContext struct {
...
+type SystemExtensions struct {
+	SystemExtensions []string `yaml:"systemExtensions,omitempty"`
+}
+
+func (s *SystemExtensions) translate(ctx TranslateContext, nameSuffix string, labels ...pair.Pair[string, string]) []resource.Resource {
+	if len(s.SystemExtensions) == 0 {
+		return nil
+	}
+
+	slices.Sort(s.SystemExtensions)
```
utkuozdemir, line 47: I'm not sure if we should do this (anywhere) - different order produces a different output (as the extensions are layered on top of each other). So I think it makes sense to not do this anywhere, or do it everywhere.

IMO not doing makes a bit more sense as the order actually matters. In the frontend, we list them in an ordered manner anyway, and for the templates, we can communicate the fact that order matters in the docs or so.

client/pkg/template/internal/models/machine.go
utkuozdemir: oh, I see, all good

client/pkg/template/internal/models/models.go
Unix4ever: That's a great question. I would rather expect it to be order independent.

Unix4ever: OK, gotcha.

internal/backend/runtime/omni/controllers/omni/machine_status_test.go
Unix4ever: It's not yaml, but the definition of the schematic is right above these lines.

frezbo: i missed the definitions, all good here :+1:

PR #101, opened 2024-04-01
fix: remove freshly created `siderolink.Link` if PeerEvent ended with error

DmitriyMV:
Current code assumes if `WireguardHandler.PeerEvent` fails on the first connection attempt, it will never try to do `PeerEvent` it again. That is because on the second Talos provision attempt, the check `spec.NodePublicKey != req.NodePublicKey` will return `false`, and `dirty` will be also set to `false`. So Omni will just happily return the `link` and be done with it. That means if (for some reason) `WireguardHandler.PeerEvent` failed on first connection from Talos - it will never configure Wireguard on Omni side until Omni restarts, and you can actually see why `PeerEvent` fails in `pollWireguardPeers`.

Fixing this by deleting freshly created `siderolink.Link` if `PeerEvent` failed.

### DIFF
--- a/internal/pkg/siderolink/manager.go
+++ b/internal/pkg/siderolink/manager.go
@@ -631,10 +631,12 @@ func (manager *Manager) getLink(ctx context.Context, req *pb.ProvisionRequest, i
 func (manager *Manager) Provision(ctx context.Context, req *pb.ProvisionRequest) (*pb.ProvisionResponse, error) {
 	ctx = actor.MarkContextAsInternalActor(ctx)
 
-	link, dirty, err := manager.getLink(ctx, req, req.NodeUuid)
+	link, created, err := manager.getLink(ctx, req, req.NodeUuid)
 	if err != nil {
 		return nil, err
 	}
+
+	var updated isDirty
 
 	spec := link.TypedSpec().Value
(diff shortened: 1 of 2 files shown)


internal/pkg/siderolink/manager.go
```diff
@@ -654,11 +656,31 @@ func (manager *Manager) Provision(ctx context.Context, req *pb.ProvisionRequest)
...
 		}
 
-		dirty = true
+		updated = true
 	}
 
-	if dirty {
+	if created || updated {
 		if err = manager.wgHandler.PeerEvent(ctx, spec, false); err != nil {
+			if !created {
```
smira, line 664: this feels still like a partial fix to me.... ? as updated will try only once same way

I wonder if we should somehow tie the resource create/update with the PeerEvent error?

I don't see the full picture though

Unix4ever: I would agree with Andrey here.
Probably the right way would be to move create/update logic after we successfully do `PeerEvent`, that will naturally fix all issues with this code.

So:
1. get the link
2. if it exists or the key hasn't changed: return.
3. update the spec in memory, send peer event.
4. do create or update call with the created spec.

Issue #20, opened 2023-09-01
bundle Discovery Service with Omni


smira:
Talos now allows insecure scheme, so we could create a cluster and point towards the discovery service in Omni over SideroLink API.

See siderolabs/talos#7340 (Talos >= 1.5.0)

utkuozdemir:
The feature is mostly implemented, but putting it on hold until we do #54.

PR #105, opened 2024-04-02
feat: run embedded discovery service in Omni

utkuozdemir:
Depends on https://github.com/siderolabs/discovery-service/pull/56

Run a discovery service instance inside Omni, which can be disabled but is enabled by default. When enabled, it will listen only on the SideroLink interface on port 9093. When enabled, all new clusters will use the embedded discovery service instead of the public one.

Existing clusters will remain unaffected by this change.

Closes siderolabs/omni#20.


Makefile
```diff
@@ -20,13 +20,13 @@ REGISTRY_AND_USERNAME ?= $(REGISTRY)/$(USERNAME)
 PROTOBUF_GRPC_GATEWAY_TS_VERSION ?= 1.2.1
 TESTPKGS ?= ./...
 NODE_BUILD_ARGS ?=
-PROTOBUF_GO_VERSION ?= 1.33.0
+PROTOBUF_GO_VERSION ?= 1.34.1
 GRPC_GO_VERSION ?= 1.3.0
-GRPC_GATEWAY_VERSION ?= 2.19.1
+GRPC_GATEWAY_VERSION ?= 2.19.1 # manually reverted to 2.19.1, as 2.20.0 breaks the frontend
```
utkuozdemir, line 25: This is the culprit, otherwise `PROTOBUF_GO_VERSION` 1.34.1 is fine.

TimJones:
Just to note: make sure we also update omin-docs to reflect the new feature option in cluster templates.

utkuozdemir:
> Just to note: make sure we also update omin-docs to reflect the new feature option in cluster templates.

Thanks for the reminder, sure will do.

utkuozdemir:
> I think (after the discussion in the meeting) we need to default this feature to 'disabled' (i.e. use discovery.talos.dev by default).

Changed the implementation to be that way, greatly simplified.

client/api/omni/specs/omni.proto
```diff
@@ -251,10 +251,18 @@ message TalosConfigSpec {
 // Cluster describes a Talos cluster.
 message ClusterSpec {
   message Features {
+    enum EmbeddedDiscoveryServiceConfig {
```
Unix4ever, line 254: Not used anywhere?

frontend/src/views/omni/Clusters/EmbeddedDiscoveryServiceCheckbox.vue
```diff
@@ -0,0 +1,32 @@
...
+Use of this software is governed by the Business Source License
+included in the LICENSE file.
+-->
+<template>
+  <tooltip placement="bottom">
+    <template #description>
+      <div class="flex flex-col gap-1 p-2">
+        <p>Configure the cluster to use the discovery service embedded in Omni instead of the public one.</p>
+        <p>This will only available if:</p>
+        <p>- This Omni has the feature enabled.</p>
```
Unix4ever, line 13:
```suggestion
        <p>- This Omni instance has the feature enabled.</p>
```

internal/backend/runtime/omni/controllers/omni/discovery_service_config_patch.go
```diff
@@ -0,0 +1,96 @@
...
+			TransformFunc: func(_ context.Context, _ controller.Reader, _ *zap.Logger, cluster *omni.Cluster, configPatch *omni.ConfigPatch) error {
+				configPatch.Metadata().Labels().Set(omni.LabelSystemPatch, "")
+				configPatch.Metadata().Labels().Set(omni.LabelCluster, cluster.Metadata().ID())
+
+				enabledForCluster := cluster.TypedSpec().Value.GetFeatures().GetUseEmbeddedDiscoveryService()
+
+				if !enabled || !enabledForCluster {
+					configPatch.TypedSpec().Value.Data = ""
+
+					return nil
```
Unix4ever, line 80: I think we should use the destroy tag here to drop the patch or skip the creation of the patch.

Unix4ever:
I guess `ClusterMachineTeardown` controller should check the cluster features and pick the right discovery service instance depending on which one is enabled for it.

internal/backend/runtime/omni/controllers/omni/discovery_service_config_patch.go
utkuozdemir: Yes, that was the reason I added it, but we decided to hold it for now: https://github.com/cosi-project/runtime/pull/452#discussion_r1622292823

Unix4ever: Well, I don't think it's just a quality of life improvement :)
So I would suggest no to hold-off and add destroy tag support.

Unix4ever: there's another place like that, which might benefit from that as well.
For now I did in a bit hacky way:
https://github.com/siderolabs/omni/blob/c6358272732628b01e38efdccbcfd766205ed7da/internal/backend/runtime/omni/controllers/omni/maintenance_config_patch.go#L57

utkuozdemir:
> I guess `ClusterMachineTeardown` controller should check the cluster features and pick the right discovery service instance depending on which one is enabled for the cluster.

Oh that's a good point, we need 2 discovery clients living in the same time in fact. I'll fix that.

internal/backend/runtime/omni/controllers/omni/discovery_service_config_patch.go
utkuozdemir: Not sure if teardown & check if ready & destroy is the right way either atm.

Unix4ever: Only weird thing here is that we don't remove the finalizer from the main input even though we didn't create the resource, otherwise it works

utkuozdemir: will use https://github.com/cosi-project/runtime/pull/452 to address this, when I get that right and get it merged.

internal/backend/runtime/omni/controllers/omni/discovery_service_config_patch_test.go
```diff
@@ -0,0 +1,64 @@
...
+	})
+	suite.Require().NoError(err)
+
+	// assert that the endpoint is removed from the config patch
+	rtestutils.AssertResource[*omni.ConfigPatch](suite.ctx, suite.T(), suite.state, patchID, func(r *omni.ConfigPatch, assertion *assert.Assertions) {
+		assertion.Empty(r.TypedSpec().Value.Data)
+	})
+}
+
+func TestDiscoveryServiceConfigPatchSuite(t *testing.T) {
```
Unix4ever, line 62:
```suggestion
func TestDiscoveryServiceConfigPatchSuite(t *testing.T) {
	t.Parallel()
```

client/api/omni/specs/omni.proto
utkuozdemir: removed

frontend/src/views/omni/Clusters/EmbeddedDiscoveryServiceCheckbox.vue
utkuozdemir: done

internal/backend/runtime/omni/controllers/omni/discovery_service_config_patch.go
utkuozdemir: replaced the logic both here and in the `maintenance_config_patch.go` with the destroy output feature merged in cosi: https://github.com/cosi-project/runtime/commit/15e9d678159d409649a1afe95f4d225e0f0bcef8

internal/backend/runtime/omni/controllers/omni/discovery_service_config_patch_test.go
utkuozdemir: done

utkuozdemir:
> I guess `ClusterMachineTeardown` controller should check the cluster features and pick the right discovery service instance depending on which one is enabled for the cluster.

Done. The logic there was not working properly anyway - was probing an already deleted `ClusterMachineIdentity` and returning without actually calling `AffiliateDelete`. I fixed the logic there and tested both with and without embedded discovery service being enabled.

Unix4ever (approved):

PR #106, opened 2024-04-02
fix: prevent link and clustermachine deletion from getting stuck

### LINKED ISSUE #97: [feature] Support force-deleting a cluster
### Problem Description

Currently a cluster can be torn down either using `omnictl cluster delete`, or `omnitcl delete Clusters.omni.sidero.dev`, and with the latter, all clusters can be deleted by passing the `--all` flag.

However, if a cluster is stuck in an unrecoverable state, while still having machines attached to it, the cluster cannot be deleted. The `MachineSetNodes` type can be deleted, the `Links` however, not. (How I got a cluster in a stuck state is still a mystery, but possibly via trying to destroy it while it was being scaled up, and all nodes being disconnected).

It would be neat if one could run more destructive operations, possibly even delete all cluster-related resources (while keeping users, etc.) to force-delete a cluster and release machines from it.

It is OK if the machines would be left with the old config, or in whichever state they were (possibly disconnected.)

### Solution

_No response_

### Alternative Solutions

_No response_

### Notes

_No response_

### LINKED ISSUE #100: 🐛 cluster got stuck destroying
The controller is constantly failing on:
```
{"level":"error","ts":1711984723.117565,"caller":"qruntime/qruntime.go:269","msg":"reconcile failed","component":"omni_runtime","controller":"MachineSetStatusController","namespace":"default","type":"MachineSets.omni.sidero.dev","id":"guinea-pig-cluster-1-control-planes","job":"reconcile","error":"failed to get: resource Machines.omni.sidero.dev(default/00000000-0000-0000-0000-7CC25551B74A@undefined) doesn't exist","interval":75.88332874,"busy":0.023714839}
```

utkuozdemir:
This PR contains fixes for the following issues:
- When a `Link` resource was deleted, `MachineCleanupController` was unable to remove finalizers on the `MachineSetNode` resource, as it did not pass the correct owner `""` when calling `Teardown`.
- `ClusterMachineStatusController` not handling the cases where the matching `Machine` might be missing.
- `MachineSetController` not handling the cases where the `Machine` resource was missing for any of the `MachineSetNode`s that were tearing down.

The latter two fixes prevent clusters from getting stuck if they are in a state where there is a missing machine in their machine sets.

Closes #100, closes #97.

### DIFF
--- a/internal/backend/runtime/omni/controllers/omni/machine_cleanup.go
+++ b/internal/backend/runtime/omni/controllers/omni/machine_cleanup.go
@@ -49,7 +49,7 @@ func (h *sameIDHandler[I, O]) FinalizerRemoval(ctx context.Context, r controller
 		return nil
 	}
 
-	ready, err := r.Teardown(ctx, md)
+	ready, err := r.Teardown(ctx, md, controller.WithOwner(""))
 	if err != nil {
 		return err
 	}
--- a/internal/backend/runtime/omni/controllers/omni/cluster_machine_status.go
+++ b/internal/backend/runtime/omni/controllers/omni/cluster_machine_status.go
@@ -40,7 +40,7 @@ func NewClusterMachineStatusController() *ClusterMachineStatusController {
 			},
 			TransformFunc: func(ctx context.Context, r controller.Reader, _ *zap.Logger, clusterMachine *omni.ClusterMachine, clusterMachineStatus *omni.ClusterMachineStatus) error {
 				machine, err := safe.ReaderGet[*omni.Machine](ctx, r, resource.NewMetadata(resources.DefaultNamespace, omni.MachineType, clusterMachine.Metadata().ID(), resource.VersionUndefined))
-				if err != nil {
+				if err != nil && !state.IsNotFoundError(err) {
 					return err
 				}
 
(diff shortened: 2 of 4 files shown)


internal/backend/runtime/omni/controllers/omni/machine_cleanup.go
```diff
@@ -49,7 +49,7 @@ func (h *sameIDHandler[I, O]) FinalizerRemoval(ctx context.Context, r controller
 		return nil
 	}
 
-	ready, err := r.Teardown(ctx, md)
```
smira, line 52: can we add a unit-test for it please?

smira (approved):
I still don't understand how we got into a state that `ClusterMachine` exists, while `Mahcine` doesn't - it should not be ever possible...

utkuozdemir:
> I still don't understand how we got into a state that `ClusterMachine` exists, while `Mahcine` doesn't - it should not be ever possible...

Yes, that's really confusing, but it happened and we need to recover from it atm.
I have couple of ideas to reproduce that, will give them a try. Probably another bug.

internal/backend/runtime/omni/controllers/omni/machine_cleanup.go
utkuozdemir: done

PR #115, opened 2024-04-06
chore: support Auth0 client playing nicely with other OAuth2/OIDC providers

sherif-fanous:
Omni doesn’t use a generic OAuth2/OIDC client/SDK instead it uses the Auth0 SDK for Vue and it’s using a pretty old version,  v1.0.2 as can be seen by inspecting the Omni package.json [here](https://github.com/siderolabs/omni/blob/7fb5d2b20a9372e1a0906b9384696daf93a45c51/frontend/package.json#L13). This SDK in turn uses [auth0-spa-js](https://github.com/auth0/auth0-spa-js) v1.22.1 which can be seen by inspecting its package.json [here](https://github.com/auth0/auth0-vue/blob/bb3bc817d18b8b6d68f3292fe6fadb31f28320db/package.json#L80).

**This has significant implications as the v1 of the SDK is not compliant with OAuth2 in 1 critical area.**

OAuth2 mandates the use of  the `application/x-www-form-urlencoded` content type for grant messages sent to the token endpoint and that sending JSON request bodies will result in a 400 error.

Unfortunately the v1 of the SDK sends the request payload as JSON which means that IdPs such as Authentik rightfully returns a 400 error and this results in an infinite loop of requests from Omni to Authentik.

The behavior can be confirmed by looking at the comment in the Auth0 SDK code [here](https://github.com/auth0/auth0-spa-js/blob/371e5a82a6da3be24a2f89b7a3a4473f01156c02/src/global.ts#L251). Interestingly the default for the `useFormData` was changed to `true` in v1.22.6 of the SDK.

This PR introduces a new Omni flag called `--auth-auth0-use-form-data`. By default the flag is set to `false` to maintain backwards compatibility. If the flag is set to `true` then the Auth0 client is created with the `useFormData` set to `true`


Unix4ever (approved):
LGTM. I will squash the commits and create a branch in origin as github actions don't work with forks for Omni right now.

Unix4ever:
Close in favor of https://github.com/siderolabs/omni/pull/126 
PRs from forks won't build

PR #132, opened 2024-04-10
fix: ignore not existing cluster in `MachineSet` teardown flow

Unix4ever:
If the cluster is not found it's not critical for teardown flow, so we shouldn't skip reconcile in that case.

### DIFF
--- a/internal/backend/runtime/omni/controllers/omni/internal/machineset/reconciliation_context.go
+++ b/internal/backend/runtime/omni/controllers/omni/internal/machineset/reconciliation_context.go
@@ -7,14 +7,13 @@ package machineset
 
 import (
 	"context"
+	"errors"
 	"fmt"
 
 	"github.com/cosi-project/runtime/pkg/controller"
-	"github.com/cosi-project/runtime/pkg/controller/generic/qtransform"
 	"github.com/cosi-project/runtime/pkg/resource"
 	"github.com/cosi-project/runtime/pkg/safe"
 	"github.com/cosi-project/runtime/pkg/state"
-	"github.com/siderolabs/gen/xerrors"
 	"github.com/siderolabs/gen/xslices"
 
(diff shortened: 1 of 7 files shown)


smira (commented):
how is the `Cluster` gone before a `MachineSet`?

Unix4ever:
> how is the Cluster gone before a MachineSet?

Well, nothing stops you from creating a `MachineSet` before creating a cluster. Then you lock yourself out without the ability to get rid of it.

Unix4ever:
This is a wrong API usage, but it shouldn't break Omni

smira:
> > how is the Cluster gone before a MachineSet?
> 
> Well, nothing stops you from creating a `MachineSet` before creating a cluster. Then you lock yourself out without the ability to get rid of it.

ah, got you, makes sense.

smira (approved):

internal/backend/runtime/omni/controllers/omni/internal/machineset/reconciliation_context.go
```diff
@@ -345,9 +347,18 @@ func (rc *ReconciliationContext) GetOutdatedMachines() Set[string] {
 	return rc.idsOutdated
 }
 
-// GetCluster reads the related cluster resource.
-func (rc *ReconciliationContext) GetCluster() *omni.Cluster {
-	return rc.cluster
+// GetKubernetesVersion reads kubernetes version from the related cluster if the cluster exists.
+func (rc *ReconciliationContext) GetKubernetesVersion() (string, error) {
+	if rc.cluster == nil {
+		return "", fmt.Errorf("failed to get kubernetes version for the machine set as the cluster couldn't be found")
```
DmitriyMV, line 353:
```suggestion
		return "", errors.New("failed to get kubernetes version for the machine set as the cluster couldn't be found")
```

DmitriyMV (approved):

PR #133, opened 2024-04-11
chore: rekres with latest kres

### LINKED ISSUE #131: [feature] Add Version Number to Home page for Omni On-Prem
### Problem Description

For on-prem deployments, the Omni home page shows the currently assigned image tag as the ```Backend Version```.  If you are running with the "latest" image tag, the version simply shows as "latest"

For those of us who don't mind living on the bleeding edge, it would be nice to see the actual version number in use. 

### Solution

Instead of using the image tag for ```Backend Version```, show the actual backend version currently in use.

### Alternative Solutions

I've considered NOT using the "latest" image tag, but I like to live dangerously.

### Notes

_No response_

smira:
Fixes #131

### DIFF
--- a/client/.golangci.yml
+++ b/client/.golangci.yml
@@ -1,21 +1,20 @@
 # THIS FILE WAS AUTOMATICALLY GENERATED, PLEASE DO NOT EDIT.
 #
-# Generated on 2024-02-09T09:47:03Z by kres decf506.
+# Generated on 2024-04-11T11:10:37Z by kres e4c6337.
 
 # options for analysis running
 run:
   timeout: 10m
   issues-exit-code: 1
   tests: true
-  build-tags: []
-  skip-dirs: []
-  skip-dirs-use-default: true
-  skip-files: []
+  build-tags: [ ]
   modules-download-mode: readonly
 
 # output configuration options
 output:
-  format: colored-line-number
+  formats:
+  - format: colored-line-number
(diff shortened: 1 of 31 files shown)


client/.golangci.yml
```diff
@@ -1,21 +1,20 @@
...
 # options for analysis running
 run:
   timeout: 10m
   issues-exit-code: 1
   tests: true
-  build-tags: []
-  skip-dirs: []
-  skip-dirs-use-default: true
-  skip-files: []
+  build-tags: [ ]
```
utkuozdemir, line 10: is the change in formatting ok/intended?

utkuozdemir (approved):

client/.golangci.yml
smira: not me, it's kres :)

utkuozdemir: I know, was wondering if it was expected from kres... But anyways, no big deal.

DmitriyMV:
@smira @utkuozdemir  since we download and build `golangci-lint` ourselves, maybe it makes sense to just temporary bind to the Go pseudo-version of it? Like `v1.52.3-whatevergooutputsforthelatesthash`.

smira:
but on a second thought, I could only bump in Omni

Unix4ever (approved):

Issue #19, opened 2023-12-04
Discovery Service Affiliate Audit Controller


Unix4ever:
Would rather not put that to the machine set status controller.
Need to clean up the destroyed machines members from the discovery service.

Otherwise things like cluster recreation from etcd backups fails due to the conflicting ip addresses.

Ilustration of the problematic kubespan peer statuses:
```
192.168.88.81   kubespan    KubeSpanPeerStatus   2CDZosuUNmNzoap0oi7AjuzHTIQ8T9Aze9w/KHxatCw=   4         cp-1                192.168.89.129:51820   unknown   0       1480
192.168.88.81   kubespan    KubeSpanPeerStatus   6BSE6Jtx/HQM8SJLrt51Bf8n25NZV/Y6I7QYXnlcXRU=   6         ruch-vbox-storage   77.108.97.220:51820    up        20060   12948
192.168.88.81   kubespan    KubeSpanPeerStatus   FxDwBDQphUHdYWRH59RcHiw7EsqEGx1p4oolHzeCcxI=   3         rockpi4-storage     192.168.88.10:51820    up        19676   12500
192.168.88.81   kubespan    KubeSpanPeerStatus   L6bqRdXqh70HGdy8I2XkhnU69mtIMBo/P2x/eEKLvQ4=   4         cp-0                77.108.97.212:1099     unknown   0       1480
192.168.88.81   kubespan    KubeSpanPeerStatus   WzDJ23MuoE8JBH48T95PZIL5lcwNwyfgVVxLHlBW4Gg=   3         proxmox-storage     192.168.89.131:51820   up        124     212
192.168.88.81   kubespan    KubeSpanPeerStatus   d8LQYcAIOs32ir7Rmq2vyp5/O8PqT9OVOaT6M5CMKXs=   4         rpi4-8gb            77.108.97.212:1096     unknown   0       1480
192.168.88.81   kubespan    KubeSpanPeerStatus   fzhSZATNLcXKETCe7QvjHlUvZ51FK1upiYk8qPcFHQg=   4         rpi4-8gb            77.108.97.212:1096     unknown   0       1480
192.168.88.81   kubespan    KubeSpanPeerStatus   gxsuUsUd06cq0JTL8zcDLWdrX7P+rbfY+7wscp7WQwg=   4         cp-1                77.108.97.212:1098     unknown   0       1480
[...]

PR #134, opened 2024-04-11
feat: implement discovery service audit controller

utkuozdemir:
Clean up invalid affiliates from the discovery service if they are no longer part of a cluster from the perspective of Omni.

Introduce a new intermediate resource, `ClusterIdentity`, containing a cluster's unique identifier in the discovery service and all node IDs as they appear ni discovery service as affiliate IDs.

Create the resource, `DiscoveryAuditResult`, to provide a summary on the latest deleted affiliates deleted from the discovery service.

Closes siderolabs/omni#19.


internal/backend/runtime/omni/controllers/omni/discovery_audit.go
```diff
@@ -0,0 +1,298 @@
...
+				logger = logger.With(zap.String("cluster_id", clusterID))
+
+				discoveryAffiliateIDs, err := auditor.listAffiliateIDs(ctx, discoveryClient, clusterID)
+				if err != nil {
+					return err
+				}
+
+				var localNodeIDs []string
+
+				if clusterIdentity.Metadata().Phase() != resource.PhaseTearingDown { // if the cluster is being torn down, we consider all nodes to be invalid
```
smira, line 76: can we do potentially something more simple, i.e. drop affiliates for ClusterIdentities which are being torn down?

utkuozdemir: Done, and reworked the whole thing quite a bit. This was wrong anyway - it needed to be in `FinalizerRemoval`.

(still a draft, will add a couple of more things before it's ready for review).

internal/backend/runtime/omni/controllers/omni/discovery_audit.go
```diff
@@ -0,0 +1,303 @@
...
+
+// CreateDiscoveryClientFunc is a function that creates a new DiscoveryClient.
+type CreateDiscoveryClientFunc = func(ctx context.Context) (DiscoveryClient, error)
+
+// DiscoveryAuditController creates the system config patch that contains the maintenance config.
+type DiscoveryAuditController = qtransform.QController[*omni.ClusterIdentity, *omni.DiscoveryAuditResult]
+
+// NewDiscoveryAuditController initializes DiscoveryAuditController.
+func NewDiscoveryAuditController(createDiscoveryClient CreateDiscoveryClientFunc, deleteOlderThan time.Duration) *DiscoveryAuditController {
+	requeueAfter := deleteOlderThan + 1*time.Second
```
Unix4ever, line 42:
```suggestion
	requeueAfter := deleteOlderThan + time.Second
```

internal/backend/runtime/omni/controllers/omni/discovery_audit.go
```diff
@@ -0,0 +1,303 @@
...
+				logger.Info("found affiliates ready to be deleted")
+
+				deleted, requeue, err := auditor.deleteAffiliates(ctx, discoveryClient, logger, clusterIdentity, cluster, readyToDelete)
+				if err != nil {
+					return err
+				}
+
+				discoveryAuditResult.TypedSpec().Value.DeletedAffiliateIds = deleted
+
+				if requeue {
```
Unix4ever, line 113: Why don't we just remove affiliates that do not have corresponding `ClusterMachines` immediately?

Unix4ever, line 113: Here I thought about something a bit simpler:
just watch the list of `ClusterMachines` and cleanup the affiliates which don't have a corresponding machine without requeue.

Or even with the used approach: why do we need destroy grace period?

utkuozdemir:
The issue #19 was instead resolved by #149.

This evolved into the https://github.com/siderolabs/omni/pull/150 with some changes.

PR #137, opened 2024-04-12
release(v0.33.0): prepare release

Unix4ever:
This is the official v0.33.0 release.


smira:
I don't want to derail this particular release, but I think we should a beta release first on `main`, as right now `main` has no tags at all, and using `:latest` gives you a weird version as well.

And as we publish the release, we have no way to test it in staging before calling it "done".

smira (approved):

Unix4ever:
Yeah, will do that too

utkuozdemir (approved):

Issue #143, opened 2024-04-15
add overlays for sbc profiles when generating installation media


Unix4ever:
If the profile generates the image for SBC and Talos version is 1.7.0, automatically pick the overlay.

PR #147, opened 2024-04-15
feat: support generating installation media with overlays for Talos 1.7+

Unix4ever:
Fixes: https://github.com/siderolabs/omni/issues/143

This is crucial if we want to support SBCs in Omni.

Automatically detect which overlay we need to install when any SBC type is selected on the backend.
Move some of filename generation to the backend, as it's now Talos version dependent.


client/pkg/schematics/schematics.go
```diff
@@ -0,0 +1,26 @@
...
+import (
+	"strings"
+
+	"github.com/blang/semver"
+)
+
+var supportedVersion = semver.MustParse("1.7.0")
+
+// SupportsOverlays checks if Talos version supports overlays.
+func SupportsOverlays(version string) (bool, error) {
```
smira, line 17: fyi, we already have this in Talos: https://github.com/siderolabs/talos/blob/e23af33b59f0ac020c77dd44ce0059cb9f93e375/pkg/imager/quirks/quirks.go#L57-L65

smira: but I guess it's not in machinery

smira (approved):

client/pkg/schematics/schematics.go
Unix4ever: oh nice. Thanks for pointing out :+1:

client/api/omni/specs/installation_media.go
```diff
@@ -0,0 +1,33 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+package specs
+
+// GenerateFilename gets the final part of the image factory URL.
+func (media *InstallationMediaSpec) GenerateFilename(legacy, secureBoot, withExtension bool) string {
+	var filename string
```
DmitriyMV, line 9: I propose `var builder strings.Builder` here.

internal/backend/factory/factory.go
```diff
@@ -179,15 +180,12 @@ func parseRequest(r *http.Request, st state.State) (*ProxyParams, error) {
 		return nil, err
 	}
 
-	// filename is the last part of the proxy URL which is parsed by the image factory
-	// so it should have the appropriate format
-	filename := media.TypedSpec().Value.SrcFilePrefix
-
-	if secureBoot {
-		filename += "-secureboot"
+	supportsOverlays, err := schematics.SupportsOverlays(segments[2])
```
DmitriyMV, line 183:
```suggestion
	supportsOverlays, err := schematics.SupportsOverlays(srcFilename)
```

internal/backend/grpc/schematics.go
```diff
@@ -70,11 +74,51 @@ func (s *managementServer) CreateSchematic(ctx context.Context, request *managem
...
+		var overlays []client.OverlayInfo
+
+		overlays, err = s.imageFactoryClient.Client.OverlaysVersions(ctx, request.TalosVersion)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get overlays list for the Talos version: %w", err)
+		}
+
+		var found bool
+
+		for _, overlay := range overlays {
```
DmitriyMV, line 101: `slices.IndexFunc` maybe?

DmitriyMV (approved):

client/pkg/schematics/schematics.go
Unix4ever: But it's not in the `machinery` :(

Unix4ever: I can extract it though

internal/backend/runtime/omni/controllers/omni/installation_media.go
```diff
@@ -295,15 +302,16 @@ var installationMedia = []installationMediaSpec{
...
 		SBC:          true,
-
-		ContentType: "application/x-xz",
+		Overlay:      "rockpi4c",
+		ContentType:  "application/x-xz",
 	},
 	{
 		Name:         "Raspberry Pi 4 Model B",
 		Architecture: arm64Arch,
 		Profile:      constants.BoardRPiGeneric,
```
frezbo, line 311: do we use this anymore?

Unix4ever: only for the output file name

Issue #9, opened 2023-12-11
Kubernetes Node Audit Controller


utkuozdemir:
Use the machine ID provided by Kubelets to compare Kubernetes nodes with the node identity data collected via Omni.
Remove stale Kubernetes nodes via Kubernetes API.

(See how machine ID is generated: https://github.com/siderolabs/talos/blob/4f9d3b975fa689dc9eea4e44ff453d8b68ae54ef/internal/app/machined/pkg/controllers/cluster/node_identity.go#L110-L122)

Similar to siderolabs/omni#19.

PR #150, opened 2024-04-16
feat: implement kubernetes node audit controller

utkuozdemir:
Clean up invalid nodes (from the Kubernetes perspective) from the cluster if they are no longer part of a cluster from the perspective of Omni.

Introduce a new intermediate resource, `ClusterKubernetesNodes`, containing the ordered list of nodes in a cluster.

Create the resource, `KubernetesNodeAuditResult`, to provide a summary on the latest deleted nodes from the Kubernetes cluster.

Closes siderolabs/omni#9.


internal/backend/runtime/omni/controllers/omni/cluster_kubernetes_nodes.go
```diff
@@ -0,0 +1,74 @@
...
+
+					return nil
+				}
+
+				nodes := make([]string, 0, identityList.Len())
+
+				for iter := identityList.Iterator(); iter.Next(); {
+					identity := iter.Value()
+
+					if identity.TypedSpec().Value.Nodename != "" {
```
Unix4ever, line 59: I wonder if we should skip audit if this controller finds the nodes without a nodename.
Normally it should never be empty and if it is we might want to hold off the audit until it gets fixed.

internal/backend/runtime/omni/controllers/omni/omni_test.go
```diff
@@ -347,7 +347,7 @@ func (suite *OmniSuite) newServer(suffix string, opts ...grpc.ServerOption) (*ma
 }
 
 func (suite *OmniSuite) SetupTest() {
-	suite.ctx, suite.ctxCancel = context.WithTimeout(context.Background(), 20*time.Second)
+	suite.ctx, suite.ctxCancel = context.WithTimeout(context.Background(), 20*time.Hour) // todo: temp, revert
```
Unix4ever, line 350: Don't forget to revert :)

smira (approved):
I think we should have an integration test for it.

Also we need to skip deleting nodes which are Ready?

Unix4ever (approved):

internal/backend/runtime/omni/controllers/omni/omni_test.go
utkuozdemir: good point :D

utkuozdemir:
> I think we should have an integration test for it.
> 
> Also we need to skip deleting nodes which are Ready?

already doing the `Ready` check here:
https://github.com/siderolabs/omni/pull/150/files#diff-f856c47e45cecba4018f26b99f4529186e842968b8a8a6acf6201f6bbaec9d61R230-R234

Will add an integration test.

Unix4ever:
Just remembered: we also need to drop `MachineSet` dependency in the cluster machine teardown controller.
But please run all e2e tests after that: especially e2e scaling.

utkuozdemir:
> Just remembered: we also need to drop `MachineSet` dependency in the cluster machine teardown controller. But please run all e2e tests after that: especially e2e scaling.

I see, but why? - We still remove the node in that controller as well. This audit controller is just additional.

internal/backend/runtime/omni/controllers/omni/cluster_kubernetes_nodes.go
utkuozdemir: done - we don't skip empties here anymore, and the other controller skips reconciliation if there's an empty one.

PR #152, opened 2024-04-16
add basic helm chart

Btijmen:
This PR adds a basic helm chart to deploy Omni in an existing Kubernetes cluster. It does not take care of gpg keys, certificates, etcetera, leaving that to the user's own devices.

### DIFF
--- a/deploy/helm/README.md
+++ b/deploy/helm/README.md
@@ -1,0 +1,233 @@
+# Omni Helm chart
+
+## Prerequisites
+
+To run Omni using this Helm chart the following additional components are required:
+
+- An ingress controller e.g. [ingress-nginx](https://github.com/kubernetes/ingress-nginx)
+- SSL certificates e.g. created via [cert-manager](https://cert-manager.io/)
+- An Authentication provider. e.g. See [Configure Authentication](https://omni.siderolabs.com/how-to-guides/self_hosted/index#configure-authentication) for using Auth0.
+- A GPG key. See [Create Etcd Encryption Key](https://omni.siderolabs.com/how-to-guides/self_hosted/index#create-etcd-encryption-key)
+  The GPG key can be added to a Secret like:
+
+  ```sh
+  kubectl create secret -n omni generic gpg --from-file=omni.asc=./omni.asc
+  ```
+
+- A PV for etcd storage. See [Local Storage](https://www.talos.dev/v1.8/kubernetes-guides/configuration/local-storage/) for a simple storage setup or [Storage](https://www.talos.dev/v1.8/kubernetes-guides/configuration/storage/) for more advanced options.
+
+## Example Walkthrough
+
+This example assumes 3 separate DNS entries will be used:
+
+| Example URL                 | Purpose          |
+| --------------------------- | ---------------- |
+| omni.example.com            | UI and API       |
+| kubernetes.omni.example.com | Kubernetes proxy |
+| siderolink.omni.example.com | SideroLink       |
+
+Assuming the above URLs and an Auth0 account the only values that need
+to be modified are:
+
+```yaml
+domainName: omni.example.com 
+accountUuid: <UUID> # This can be created using uuidgen
+name: "MyOmniInstance"
+auth:
+  auth0:
+    clientId: <AUTH0_CLIENT_ID>
+    domain: <AUTH0_DOMAIN>
+initialUsers:
+  - <EMAIL>
+service:
+  siderolink:
+    domainName: siderolink.omni.example.com
+    wireguard:
+      # The Helm chart currently assumes that Omni will be bound on a particular Node.
+      address: <IP address of worker node>
+  k8sProxy:
+    domainName: kubernetes.omni.example.com
+```
+
+### Ingresses
+
+Example Ingresses are given below for ingress-nginx. N.B. They also require SSL certs to exist.
+
+#### api
+
+```yaml
+apiVersion: networking.k8s.io/v1
+kind: Ingress
+metadata:
+  annotations:
+    nginx.ingress.kubernetes.io/backend-protocol: GRPC
+    nginx.ingress.kubernetes.io/proxy-body-size: 32m
+    nginx.ingress.kubernetes.io/service-upstream: "true"
+  labels:
+    app.kubernetes.io/name: omni
+  name: api
+  namespace: omni
+spec:
+  ingressClassName: nginx
+  rules:
+  - host: omni.example.com
+    http:
+      paths:
+      - backend:
+          service:
+            name: omni-internal-grpc
+            port:
+              number: 8080
+        path: /cosi.resource.State
+        pathType: ImplementationSpecific
+      - backend:
+          service:
+            name: internal-grpc
+            port:
+              number: 8080
+        path: /management.ManagementService
+        pathType: ImplementationSpecific
+      - backend:
+          service:
+            name: internal-grpc
+            port:
+              number: 8080
+        path: /machine.MachineService
+        pathType: ImplementationSpecific
+      - backend:
+          service:
+            name: internal-grpc
+            port:
+              number: 8080
+        path: /cluster.ClusterService
+        pathType: ImplementationSpecific
+      - backend:
+          service:
+            name: internal-grpc
+            port:
+              number: 8080
+        path: /inspect.InspectService
+        pathType: ImplementationSpecific
+      - backend:
+          service:
+            name: internal-grpc
+            port:
+              number: 8080
+        path: /resource.ResourceService
+        pathType: ImplementationSpecific
+      - backend:
+          service:
+            name: internal-grpc
+            port:
+              number: 8080
+        path: /storage.StorageService
+        pathType: ImplementationSpecific
+      - backend:
+          service:
+            name: internal-grpc
+            port:
+              number: 8080
+        path: /time.TimeService
+        pathType: ImplementationSpecific
+      - backend:
+          service:
+            name: internal-grpc
+            port:
+              number: 8080
+        path: /auth.AuthService
+        pathType: ImplementationSpecific
+      - backend:
+          service:
+            name: internal-grpc
+            port:
+              number: 8080
+        path: /oicd.
+        pathType: ImplementationSpecific
+  tls:
+  - hosts:
+    - omni.example.com
+    secretName: omni.example.com-tls
+```
+
+#### kubernetes-proxy
+
+```yaml
+apiVersion: networking.k8s.io/v1
+kind: Ingress
+metadata:
+  name: kubernetes-proxy
+  namespace: omni
+spec:
+  ingressClassName: nginx
+  rules:
+  - host: kubernetes.omni.example.com
+    http:
+      paths:
+      - backend:
+          service:
+            name: internal
+            port:
+              number: 8095
+        path: /
+        pathType: ImplementationSpecific
+  tls:
+  - hosts:
+    - kubernetes.omni.example.com
+    secretName: kubernetes.omni.example.com-tls
+```
+
+#### siderolink
+
+```yaml
+apiVersion: networking.k8s.io/v1
+kind: Ingress
+metadata:
+  annotations:
+    nginx.ingress.kubernetes.io/backend-protocol: GRPC
+  name: siderolink
+  namespace: omni
+spec:
+  ingressClassName: nginx
+  rules:
+  - host: siderolink.omni.example.com
+    http:
+      paths:
+      - backend:
+          service:
+            name: internal-grpc
+            port:
+              number: 8090
+        path: /
+        pathType: ImplementationSpecific
+  tls:
+  - hosts:
+    - siderolink.omni.example.com
+    secretName: siderolink.omni.example.com-tls
+```
+
+#### ui
+
+```yaml
+apiVersion: networking.k8s.io/v1
+kind: Ingress
+metadata:
+  name: ui
+  namespace: omni
+spec:
+  ingressClassName: nginx
+  rules:
+  - host: omni.example.com
+    http:
+      paths:
+      - backend:
+          service:
+            name: internal
+            port:
+              number: 8080
+        path: /
+        pathType: Prefix
+  tls:
+  - hosts:
+    - omni.example.com
+    secretName: omni.example.com-tls
+```
--- a/deploy/helm/omni/templates/deployment.yaml
+++ b/deploy/helm/omni/templates/deployment.yaml
@@ -1,0 +1,120 @@
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ include "omni.name" . }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "omni.labels" . | nindent 4 }}
+    app.kubernetes.io/component: omni
+    {{- with .Values.deployment.labels }}
+    {{- toYaml . | nindent 4 }}
+    {{- end }}
+  annotations:
+    {{- toYaml .Values.deployment.annotations | nindent 4 }}
+spec:
+  strategy:
+    type: Recreate
+  replicas: {{ .Values.deployment.replicaCount }}
+  selector:
+    matchLabels: {{- include "omni.selectorLabels" . | nindent 6 }}
+  template:
+    metadata:
+    {{- if .Values.deployment.podAnnotations }}
+      annotations:
+      {{- toYaml .Values.deployment.annotations | nindent 6 }}
+    {{- end }}
+      labels: 
+      {{- include "omni.labels" . | nindent 8 }}
+      app.kubernetes.io/component: omni
+      {{- with .Values.deployment.labels }}
+      {{- toYaml . | nindent 8 }}
+      {{- end }}
+    spec:
+      containers:
+      - name: omni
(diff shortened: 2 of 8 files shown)


deploy/helm/omni/values.yaml
```diff
@@ -0,0 +1,72 @@
+global: {}
+
+omni:
```
TimJones, line 3: I don't see the need for the nesting. This is the Omni chart so all values should just be at the "base" i.e. `.Values.debug` etc...

deploy/helm/omni/values.yaml
```diff
@@ -0,0 +1,72 @@
...
+  nameOverride: omni  # optional, if ommitted the Chart name will be used
+  debug: false
+  domainName: omni.makomedia.nl
+  accountUuid: D5C2A21E-7162-4122-B923-A87ED7827F5B
+  privateKeySource: "/omni.asc"
+  namespace:
+    name: omni
```
TimJones, line 10: Namespace should be set via the Helm args (`helm install --namespace ...`) and templated in the chart as `{{ .Release.Namespace }}`.
```suggestion
```

deploy/helm/omni/templates/service.yaml
```diff
@@ -0,0 +1,40 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{ include "omni.name" . }}
+  namespace: {{ .Values.omni.namespace.name }}
```
TimJones, line 5:
```suggestion
  namespace: {{ .Values.omni.namespace.name }}
```

deploy/helm/omni/templates/_args.tpl
```diff
@@ -0,0 +1,43 @@
+{{- define "omni.args" -}}
+{{- $omni := .Values.omni -}}
+- --account-id="{{ $omni.accountUuid }}"
+- --advertised-api-url="{{ printf "https://%s:%s/" $omni.domainName (toString $omni.service.api.targetPort) }}"
```
TimJones, line 4:
```suggestion
- --advertised-api-url="{{ printf "https://%s:%d/" .Values.domainName .Values.service.api.targetPort }}"
```

deploy/helm/omni/templates/_args.tpl
```diff
@@ -0,0 +1,43 @@
...
+{{- end }}
+- --event-sink-port={{ $omni.service.eventSink.port }}
+- --image-factory-address="{{ $omni.imageFactory.address }}"
+{{- if $omni.imageFactory.pxeAddress }}
+- --image-factory-pxe-address="{{ $omni.imageFactory.pxeAddress }}"
+{{- end }}
+{{- if and $omni.initialUsers (gt (len $omni.initialUsers) 0) }}
+- --initial-users={{ join "," $omni.initialUsers }}
+{{- end }}
+- --k8s-proxy-bind-addr="{{ printf "0.0.0.0:%d" (int $omni.service.k8sProxy.port) }}"
```
TimJones, line 36:
```suggestion
- --k8s-proxy-bind-addr="{{ printf "0.0.0.0:%d" .Values.service.k8sProxy.port }}"
```

deploy/helm/omni/templates/deployment.yaml
```diff
@@ -0,0 +1,77 @@
...
+  replicas: {{ .Values.omni.deployment.replicaCount }}
+  selector:
+    matchLabels: {{- include "omni.selectorLabels" . | nindent 6 }}
+  template:
+    metadata:
+    {{- if .Values.omni.deployment.podAnnotations }}
+      annotations:
+      {{- range $key, $value := .Values.omni.deployment.podAnnotations }}
+        {{ $key }}: {{ $value | quote }}
+      {{- end }}
```
TimJones, line 25:
```suggestion
      {{- toYaml .Values.omni.deployment.annotations | nindent 6 }}
```

deploy/helm/omni/templates/deployment.yaml
```diff
@@ -0,0 +1,77 @@
...
+            containerPort: {{ .Values.omni.service.siderolink.wireguard.port }}
+            protocol: UDP
+          - name: event-sink
+            containerPort: {{ .Values.omni.service.eventSink.port }}
+            protocol: TCP
+          - name: k8s-proxy
+            containerPort: {{ .Values.omni.service.k8sProxy.port }}
+            protocol: TCP
+        env:
+        - name: OMNI_ACCOUNT_ID
```
TimJones, line 59: Any reason this is set as an environment variable when everything else is set as args?

deploy/helm/omni/values.yaml
Btijmen: The consideration here was that we might want to include other components down the line, such as image factory or discovery service. Nesting by `omni` would allow the variables for those apps to be included in the same valuesfile.

deploy/helm/omni/values.yaml
Btijmen: Agreed, will change that.

deploy/helm/omni/templates/deployment.yaml
Btijmen: No, this is no longer relevant. This is a remnant of a different approach I was trying at first, but moved away from. The other env references have all been removed but evidently I forgot to remove it here. Will update.

deploy/helm/omni/values.yaml
TimJones: I would have them as separate charts and pull them in as chart dependencies. This automatically creates a "nested" Values field for the external charts.

Btijmen: As discussed, I removed the nesting

deploy/helm/omni/templates/deployment.yaml
Btijmen: removed

Btijmen (commented):
Removed all of the `.omni.` nesting from the valuesfile and the templates.

deploy/helm/omni/values.yaml
```diff
@@ -0,0 +1,68 @@
+
+nameOverride: omni  # optional, if ommitted the Chart name will be used
```
TimJones, line 2: Comment it out, if it's optional it shouldn't be set as a default value.
```suggestion
# nameOverride: omni  # optional, if ommitted the Chart name will be used
```

deploy/helm/omni/values.yaml
```diff
@@ -0,0 +1,68 @@
...
+debug: false
+domainName: omni.makomedia.nl
```
TimJones, line 4: I would almost prefer `{{ required "You must define a domainName" .Values.domainName }}` where it is consumed, but it's not a requirement.
At the very least the default values should be something more like `omni.example.com`.

```suggestion
domainName: 
```

deploy/helm/omni/values.yaml
```diff
@@ -0,0 +1,68 @@
...
+accountUuid: D5C2A21E-7162-4122-B923-A87ED7827F5B
```
TimJones, line 5: Same as domainName...
```suggestion
accountUuid:
```

deploy/helm/omni/templates/deployment.yaml
```diff
@@ -0,0 +1,72 @@
...
+      labels: 
+      {{- include "omni.labels" . | nindent 8 }}
+      app.kubernetes.io/component: omni
+      {{- with .Values.deployment.labels }}
+      {{- toYaml . | nindent 8 }}
+      {{- end }}
+    spec:
+      containers:
+      - name: omni
+        image: {{ .Values.deployment.image }}
```
TimJones, line 34:
```suggestion
        image: {{ .Values.deployment.image }}:{{ .Values.deployment.tag | default .Chart.AppVersion }}
```

deploy/helm/omni/values.yaml
```diff
@@ -0,0 +1,68 @@
...
+privateKeySource: "/omni.asc"
+deployment:
+  image: ghcr.io/siderolabs/omni:latest
```
TimJones, line 8: This needs to be split, and for `tag` to be populated by default from the 
```suggestion
  image: ghcr.io/siderolabs/omni
  tag: ""
```

deploy/helm/omni/values.yaml
```diff
@@ -0,0 +1,68 @@
...
+  replicaCount: 1
+  annotations:
+    amazing: Annotation
```
TimJones, line 11:
```suggestion
  annotations: {}
```

deploy/helm/omni/values.yaml
```diff
@@ -0,0 +1,68 @@
...
+auth:
+  auth0:
+    enabled: true
+    clientId: 123456
+    domain: "https://www.auth0.example"
+  saml:
+    enabled: false
+    url: "https://www.saml.example"
+initialUsers:
+ - foo@bar.com
```
TimJones, line 22: This will actually cause the first deploy of Omni to be locked out of the user deploying it. Once an initial use is set, that is the only user that can access Omni :sweat_smile:
```suggestion
initialUsers: []
```

deploy/helm/omni/values.yaml
```diff
@@ -0,0 +1,68 @@
...
+      bindAddress: 0.0.0.0
+      port: 8090
+      targetPort: 8090
+    wireguard: 
+      port: 50180
+      targetPort: 50180
+      address: 192.168.68.107
+  eventSink:
+    port: 8091
+    targetPort: 8091
```
TimJones, line 39: As far as I know this isn't configurable.
```suggestion
```

deploy/helm/omni/values.yaml
```diff
@@ -0,0 +1,68 @@
...
+    targetPort: 443
+  siderolink:
+    api:
+      bindAddress: 0.0.0.0
+      port: 8090
+      targetPort: 8090
+    wireguard: 
+      port: 50180
+      targetPort: 50180
+      address: 192.168.68.107
```
TimJones, line 36: Where does this address come from?

deploy/helm/omni/values.yaml
Btijmen: fixed

deploy/helm/omni/values.yaml
Btijmen: Ok, I changed this in both the valuesfile and the service that consumes the values

deploy/helm/omni/values.yaml
Btijmen: Agreed, this is unclear. Removed the address and inserted a comment to help the user figure out what to put here.

TimJones:
@Btijmen Awesome work, if you can squash all commits down to a single one and rebase from master, I will sign the commit and we can merge! :D

Btijmen:
@TimJones  Nice, should be done like this!

deploy/helm/omni/templates/deployment.yaml
```diff
@@ -0,0 +1,72 @@
...
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "omni.labels" . | nindent 4 }}
+    app.kubernetes.io/component: omni
+    {{- with .Values.deployment.labels }}
+    {{- toYaml . | nindent 4 }}
+    {{- end }}
+  annotations:
+    {{- toYaml .Values.deployment.annotations | nindent 4 }}
+spec:
```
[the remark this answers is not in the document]
TimJones: Good catch, I change it now :pray:

utkuozdemir:
Do we want to add a PVC for the log storage in this iteration, or ever? @TimJones

TimJones:
> Do we want to add a PVC for the log storage in this iteration, or ever? @TimJones

Not in this iteration, an external etcd is required. We will no doubt add it in the future.

aspekt112:
Hi! Thanks the initiative! When do you plan to merge? Will you create a chart registry?

TimJones:
@aspekt112 Apologies for the late reply but @utkuozdemir was right and we're missing a PVC for the boltdb storage if we're not supporting external etcd in this first revision.

deploy/helm/README.md
```diff
@@ -0,0 +1,233 @@
...
+- SSL certificates e.g. created via [cert-manager](https://cert-manager.io/)
+- An Authentication provider. e.g. See [Configure Authentication](https://omni.siderolabs.com/how-to-guides/self_hosted/index#configure-authentication) for using Auth0.
+- A GPG key. See [Create Etcd Encryption Key](https://omni.siderolabs.com/how-to-guides/self_hosted/index#create-etcd-encryption-key)
+  The GPG key can be added to a Secret like:
+
+  ```sh
+  kubectl create secret -n omni generic gpg --from-file=omni.asc=./omni.asc
+  ```
+
+- A PVC for etcd storage.
```
TimJones, line 17:
```suggestion
- A PV for etcd storage.
```

deploy/helm/omni/values.yaml
```diff
@@ -0,0 +1,59 @@
...
+    url: "https://www.saml.example"
+includeGenericDevicePlugin: true # 
+initialUsers: []
+name: "My Omni instance"
+service:
+  type: ClusterIP # currently only ClusterIP is supported by this Helm Chart
+  siderolink:
+    domainName: omni.siderolink.example.com
+    wireguard:
+      address: #<ip address of the host running Omni>
```
TimJones, line 28:
```suggestion
      address: "" #<ip address of the host running Omni>
```

deploy/helm/omni/values.yaml
```diff
@@ -0,0 +1,59 @@
...
+resources:
+  requests:
+    cpu: 100m
+    memory: 128Mi
+  limits:
+    cpu: 200m
+    memory: 256Mi
+etcd:
+  # embedded: false not yet supported by this helm chart
+  embedded: true
```
TimJones, line 41: If it's not supported, drop the option entirely.
```suggestion
```

deploy/helm/omni/values.yaml
```diff
@@ -0,0 +1,59 @@
...
+imageFactory:
+  address: ""
+  pxeAddress: ""
```
TimJones, line 44: Just drop this and support `extraArgs` appended to the pod command. It's way more flexible and means we don't need to explicitly support every single possible combination.
```suggestion
imageFactory:
  address: ""
  pxeAddress: ""
```

deploy/helm/omni/templates/_args.tpl
TimJones, line 1: I *really* dislike this entirely. Is it feasible to just drop in the args in the Deployment template directly, and use `extraArgs` for users to add their specific environment configs?

deploy/helm/omni/templates/_args.tpl
```diff
@@ -0,0 +1,43 @@
...
+- --cert=/etc/omni/tls/tls.crt
+- --key=/etc/omni/tls/tls.key
+{{- end }}
+{{- if .Values.debug }}
+- --debug=true
+{{- end }}
+{{- if .Values.etcd.embedded }}
+- --etcd-embedded=true
+{{- end }}
+{{- if .Values.imageFactory.pxeAddress }}
```
TimJones, line 26:
```suggestion
{{- if .Values.imageFactory.address }}
```

deploy/helm/omni/templates/_args.tpl
```diff
@@ -0,0 +1,43 @@
...
+- --image-factory-address="{{ .Values.imageFactory.address }}"
+{{- end }}
+{{- if .Values.imageFactory.pxeAddress }}
+- --image-factory-pxe-address="{{ .Values.imageFactory.pxeAddress }}"
+{{- end }}
+{{- if and .Values.initialUsers (gt (len .Values.initialUsers) 0) }}
+- --initial-users={{ join "," .Values.initialUsers }}
+{{- end }}
+{{- if .Values.name }}
```
TimJones, line 35: `--name` is a required parameter, it should not be optional

deploy/helm/omni/templates/_args.tpl
```diff
@@ -0,0 +1,43 @@
...
+- --name={{ .Values.name}}
+{{- end }}
+{{- if .Values.privateKeySource }}
+- --private-key-source={{ .Values.privateKeySource }}
```
TimJones, line 39: Same, required parameter.

TimJones (approved):

alongwill:
@Btijmen could you please sign the commit with your GPG key?

```sh
git commit --amend --signoff

git push --force
```

utkuozdemir (approved):

smira:
> @Btijmen could you please sign the commit with your GPG key?

@alongwill this won't work - you need to sign with Sidero Labs employee key.

alongwill:
> @alongwill this won't work - you need to sign with Sidero Labs employee key.

@smira I have done that. In the commits tab, the commit is showing as "Partially verified"

"We cannot verify signatures from co-authors, and some of the co-authors attributed to this commit require their commits to be signed."

smira:
> > @alongwill this won't work - you need to sign with Sidero Labs employee key.
> 
> @smira I have done that. In the commits tab, the commit is showing as "Partially verified"
> 
> "We cannot verify signatures from co-authors, and some of the co-authors attributed to this commit require their commits to be signed."

That's true, but the core issue is different, I'll reach out to you in private.

Btijmen:
@alongwill @smira I'd be happy to sign off, but I'll let you talk privately first. Just let me know if I need to do anything!

PR #156, opened 2024-04-16
feat: read overlays from the machine, preserve them during updates

Unix4ever:
Change the layout of resources:
Now `SchematicConfiguration` is always defined per machine level. `SchematicConfiguration` controller creates these resources for each `ClusterMachine` by reading `ExtensionsConfiguration` and merging it with the overlay information that was read from the `MachineStatus`.

`MachineStatus` controller now reads overlay information from extraInfo of the schematic pseudo-extensions of the Talos nodes.


client/pkg/omni/resources/omni/machine_extensions.go
```diff
@@ -0,0 +1,57 @@
...
+		DefaultNamespace: resources.DefaultNamespace,
+		PrintColumns: []meta.PrintColumn{
+			{
+				Name:     "Extensions",
+				JSONPath: "{.extensions}",
+			},
+			{
+				Name:     "Phase",
+				JSONPath: "{.phase}",
+			},
```
utkuozdemir, line 54:
```suggestion
```
guess this field does not exist?

internal/backend/runtime/omni/controllers/omni/machine_extensions.go
```diff
@@ -0,0 +1,241 @@
...
+				Kind:      controller.InputQPrimary,
+			},
+			{
+				Namespace: resources.DefaultNamespace,
+				Type:      omni.ClusterMachineType,
+				Kind:      controller.InputQMapped,
+			},
+			{
+				Namespace: resources.DefaultNamespace,
+				Type:      omni.MachineExtensionsType,
```
utkuozdemir, line 59: should it take it's output also as input (even as `DestroyReady`)?

client/pkg/omni/resources/omni/machine_extensions.go
```diff
@@ -0,0 +1,57 @@
...
+	)
+}
+
+const (
+	// MachineExtensionsType is the type of the MachineExtensions resource.
+	// tsgen:MachineExtensionsType
+	MachineExtensionsType = resource.Type("MachineExtensions.omni.sidero.dev")
+)
+
+// MachineExtensions describes desired machine extensions list for a particular machine, machine set or cluster.
```
utkuozdemir, line 31:
```suggestion
// MachineExtensions describes the desired extensions list for a particular machine.
```

internal/backend/runtime/omni/controllers/omni/internal/talos/schematic.go
```diff
@@ -43,28 +45,38 @@ func GetSchematicInfo(ctx context.Context, c *client.Client) (SchematicInfo, err
...
 	)
 
-	items.ForEach(func(status *runtime.ExtensionStatus) {
+	err = items.ForEachErr(func(status *runtime.ExtensionStatus) error {
 		name := status.TypedSpec().Metadata.Name
 		if name == constants.SchematicIDExtensionName { // skip the meta extension
 			schematicID = status.TypedSpec().Metadata.Version
 
-			return
+			if status.TypedSpec().Metadata.ExtraInfo != "" {
```
utkuozdemir, line 58: Should we also grab the extensions-only ID (the `id` below) from this ExtraInfo? (not sure if it contains it, but if it does, it could be more reliable than computing it ourselves)

internal/backend/runtime/omni/controllers/omni/omni.go
```diff
@@ -97,6 +118,10 @@ func (rt *resourceTracker) cleanup(ctx context.Context) error {
 		}
 	}
 
+	if destroyReady && opts.destroyReadyCallback != nil {
+		return opts.destroyReadyCallback()
+	}
```
utkuozdemir, line 123: This runs only once if all items are destroy-ready right? It is a bit confusing this way imo, as it is passed differently (via opts) and behaves differently from `beforeDestroyCallback`.

Instead maybe we can return a `cleanupResult` struct from this `cleanup` function, containing the actually destroyed IDs for example, and on the caller site we can check and act on that. Wdyt?

Unix4ever: I wanted to keep the API without changes. Also most of the other places don't need this `cleanupResult`

internal/backend/runtime/omni/controllers/omni/internal/talos/schematic.go
Unix4ever: it doesn't contain it

internal/backend/runtime/omni/migration/migration_test.go
```diff
@@ -1268,6 +1268,60 @@ func (suite *MigrationSuite) TestClearEmptyConfigPatches() {
 	suite.Assert().Empty(cp2After.TypedSpec().Value.Patches)
 }
 
+func (suite *MigrationSuite) TestCleanupDanglingSchematicConfigurations() {
+	ctx := context.Background()
+
+	version := system.NewDBVersion(resources.DefaultNamespace, system.DBVersionID)
+	version.TypedSpec().Value.Version = 23
+	suite.Require().NoError(suite.state.Create(ctx, version))
```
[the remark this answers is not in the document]
Unix4ever: just copy pasted it from the previous test :upside_down_face: 
Surely can use filter too.

internal/backend/runtime/omni/controllers/omni/machine_extensions.go
Unix4ever: It needs to catch the moment when finalizers are moved from all produced `MachineExtensions`, so yes.
It doesn't work without this input.

utkuozdemir (approved):

Issue #145, opened 2024-04-15
detect overlays for legacy 1.6 Talos


Unix4ever:
Do it only for the machines running in a cluster.
Use `talosctl read /proc/cmdline` to get `talos.board` value.

PR #164, opened 2024-04-19
feat: generate overlay info by extracting board kernel args

Unix4ever:
This is compatibility code for SBCs running Talos < 1.7.x. With this change it's possible to do seamless upgrade of SBCs running without overlays.

Fixes: https://github.com/siderolabs/omni/issues/145

### DIFF
--- a/internal/backend/runtime/omni/controllers/omni/internal/task/machine/poll.go
+++ b/internal/backend/runtime/omni/controllers/omni/internal/task/machine/poll.go
@@ -9,14 +9,18 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"io"
 	"maps"
 
 	"github.com/cosi-project/runtime/pkg/safe"
 	"github.com/cosi-project/runtime/pkg/state"
 	"github.com/siderolabs/gen/value"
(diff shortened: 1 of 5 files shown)


internal/backend/runtime/omni/controllers/omni/internal/task/machine/poll.go
```diff
@@ -327,6 +332,28 @@ func pollMeta(ctx context.Context, c *client.Client, info *Info) error {
...
+	data, err := io.ReadAll(reader)
+	if err != nil {
+		return nil, err
+	}
+
+	cmdline := procfs.NewCmdline(string(data))
+
+	value := cmdline.Get(constants.KernelParamBoard)
+
+	if value == nil || value.Get(0) == nil {
```
frezbo, line 350: 1.7+ doesn't set the `talos.board` args, so after upgrading how's it going to handle it?

Unix4ever: it'll get overlay from the schematic info in extraData

frezbo (approved):

PR #168, opened 2024-04-23
feat: compress machine logs before writing to disk

### LINKED ISSUE #118: [bug] Pushing of machine logs seems expensive

- [X] I have searched the existing issues

### Current Behavior

For customers with lots of nodes, in monitoring we see the 10m timer where we push machine logs to the ceph volumes seems to heavily saturate the I/O.

### Expected Behavior

This is maybe an exploration of how we might make this less heavy (if we even determine it's something we should be worried about).

### Steps To Reproduce

N/A

_No response_

### Anything else?

Screenshot from 2024-04-08 11-05-59

^ An example, but I've seen quite a bit higher percentage than this. Saw over 300% last week.

utkuozdemir:
Compress the logs before writing to disk. Implement the feature in a manner that if the compression setting is enabled or disabled, it be able to do the switch as expected.

Additionally, skip flushing the log to the disk if there was no log activity since the last flush.

Addresses 2 of the 3 ideas in [this comment](https://github.com/siderolabs/omni/issues/118#issuecomment-2049397856):
- compress logs (less I/O) ✅ 
- push less aggressively ❌ 
- don't push if no new logs came in the last period ✅ 

Closes #118.


internal/pkg/siderolink/logstorage/logstorage.go
```diff
@@ -0,0 +1,299 @@
+// Copyright (c) 2024 Sidero Labs, Inc.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+
+// Package logstorage implements storage for storing logs of machines on the filesystem.
+package logstorage
```
utkuozdemir, line 7: Moved log storage module into a separate package for better code organization.

go.mod
```diff
@@ -215,6 +215,7 @@ require (
 	github.com/spf13/pflag v1.0.5 // indirect
 	github.com/stretchr/objx v0.5.2 // indirect
 	github.com/tmc/grpc-websocket-proxy v0.0.0-20220101234140-673ab2c3ae75 // indirect
+	github.com/ulikunitz/xz v0.5.12 // indirect
```
frezbo, line 218: xz :expressionless:

utkuozdemir: can easily replace, what would you prefer instead?

frezbo: i'm not sure either, maybe zstd:?

utkuozdemir: When choosing I also try to use a format which users could potentially go to the file system and open extract the logs easily. `xz` was compressing our logs a bit better than `gz`, so I simply picked it (and since the backdoor story is already over). I am fully open to suggestions - can even go with gzip.

smira: I would propose to take a look at https://github.com/klauspost/compress, this has way better compression methods.

Let's skip xz and the library

internal/pkg/siderolink/logstorage/logstorage.go
```diff
@@ -0,0 +1,299 @@
...
+
+		return fmt.Errorf("invalid log buffer hash in file: expected %s, got %s", hashHexExpected, hashHexActual)
+	}
+
+	_, err = io.Copy(writer, bytes.NewReader(bufferData))
+
+	return err
+}
+
+func (l *Storage) readLogFile(id string, compressed bool) (optional.Optional[[]byte], string, error) {
```
smira, line 124: reading log files should support both compressed an uncompressed, otherwise we would not be able to make a switch

go.mod
utkuozdemir: will do

smira (commented):
can we add some jitter to the timer which fires to flush the logs?

internal/pkg/siderolink/logstorage/logstorage.go
utkuozdemir: Yep, already doing that at the call site of this: https://github.com/siderolabs/omni/pull/168/files#diff-1601a497b172d27a79fe9d4e7b1c6718d5269c89637b589edbeaa2ae155a5889R77-R90

I implemented the feature with the switch in mind, so turning it on and off (as many times as you want) works as expected - nothing gets lost.

utkuozdemir:
> can we add some jitter to the timer which fires to flush the logs?

Added a configurable jitter with the default being 2 minutes - it'd flush 10(+-2) mins.

internal/pkg/siderolink/logstorage/logstorage.go
smira: Hmmm... I think it's trivial to detect whether the file is compressed or not: https://github.com/facebook/zstd/blob/dev/doc/zstd_compression_format.md#zstandard-frames

So we probably don't need to try to read it twice to understand that? 

We could also encode the compression format in our file format if we need to. 

I also wonder if we could compress logs in memory and/or flush them from memory, as we certainly eat some RAM by simply having (active) logs in RAM all the time. From this point of view, we might benefit from compressing files in chunks vs. trying to compress it all, or even having compressed/uncompressed chunks.

So if we go with compression today, what is our path to reduce RAM usage?

utkuozdemir: > So we probably don't need to try to read it twice to understand that?
The compressed file has a different extension (`.zst` in this case), that's why we try to read it twice. I did not want to leave the extension as `.log` if it was compressed, as `.log` kinda hints plaintext.

So only one of `a.log` or `a.log.zst` would exist on the filesystem at any time, so we don't read anything twice, we just probe to find out which one is on the disk.

> We could also encode the compression format in our file format if we need to.
We could, but if we go with the extension, we don't need that.

> So if we go with compression today, what is our path to reduce RAM usage?
My aim was to mainly address the disk pressure problem we had recently. Compressing in-memory needs some more work as it would touch the current circular buffer model we are using. Should I also go for that, in this PR?

smira: > So only one of a.log or a.log.zst would exist on the filesystem at any time, so we don't read anything twice, we just probe to find out which one is on the disk.

what if we have both files? now I'm even more confused :exploding_head: 

> My aim was to mainly address the disk pressure problem we had recently. Compressing in-memory needs some more work as it would touch the current circular buffer model we are using. Should I also go for that, in this PR?

No, I'm not asking for that of course, but I want us to draw a plan what it might be if we go with that. So that our current implementation at least fits the future direction. E.g. if the circular buffer is already compressed, we don't need to compress on writes, and we don't need to decompress on reads. But certainly we can't compress it the way we do today as well.

internal/pkg/siderolink/loghandler.go
```diff
@@ -68,24 +72,35 @@ func (h *LogHandler) Start(ctx context.Context) error {
...
+		timer       *time.Timer
+		timerCh     <-chan time.Time
+		storagePath string
+	)
 
-		defer ticker.Stop()
+	storage, ok := h.Cache.Storage.Get()
+	if ok {
+		timer = time.NewTimer(nextFlushDuration())
+		defer timer.Stop()
```
Unix4ever, line 94: you might also need to drain the timer when you stop it:
https://pkg.go.dev/time#Timer.Stop

Unix4ever, line 94: You might need to drain the timer channel after you stop it:
https://pkg.go.dev/time#Timer.Stop

I guess it should be done, when you stop the timer without reading from the channel, which might happen if the context is canceled.

internal/pkg/siderolink/logstorage/logstorage.go
```diff
@@ -0,0 +1,302 @@
...
+	}
+
+	return false, nil
+}
+
+// Load loads logs of the machine with the given id, if exists, into the given writer.
+func (l *Storage) Load(id string, writer io.Writer) error {
+	compressed := l.Compress
+
+	bufferDataOptional, hashHexActual, err := l.readLogFile(id, l.Compress)
```
Unix4ever, line 82: Can we detect file compression by the file extension instead of trying to read it twice?

Unix4ever: ah, so the ID doesn't have the extension and we probe the files here

Unix4ever: Though I think something like this might make the code more readable:

```go
compressed, err := l.probeLogFile(id)
if err != nil {
  return err
}

bufferDataOptional, hashHexActual, err := l.readLogFile(id, compressed)
```

utkuozdemir:
Closed for now - doing a different rework after the discussion with @smira

Issue #41, opened 2024-02-06
Make `MachineStatusSnapshot` use `MachineStatus` to get the current state of the machine


Unix4ever:
## Problem

`MachineStatus` is the current state of the machine. Updates to `MachineStatus` resource are sent by Talos as events and delivered over SideroLink to Omni.

If the machine can't send events, it might skip sending events, and an update might be lost.

Also we can watch `MachineStatus` resource over Talos API, but this requires the API to be up. Talos API is down e.g. when initial install is performed.

## Solution

Do not rely on the machine events, but pull the current state from the `runtime.MachineStatus` resource.

PR #186, opened 2024-04-29
feat: improve reliability of machine status snapshots

utkuozdemir:
Use the Talos resource API as well as the siderolink event sink to determine the status of a machine.

Follow the agreed decision tree of:
- if the update came over the same channel as before, use it
- if the update came over a different channel than before, and the timestamp is newer than the previous update, use it
- otherwise, drop it

Closes siderolabs/omni#41.


client/api/omni/specs/omni.proto
```diff
@@ -158,6 +158,11 @@ message MachineStatusSpec {
     string config = 1;
   }
 
+  message TalosMachineStatus {
+    machine.MachineStatusEvent status = 1;
```
utkuozdemir, line 162: This type actually differs from the Talos's `MachineStatus` resource, despite being functionally the same, but I chose to reuse this type for simplicity, as it is already imported in this proto file.

Unix4ever (approved):

PR #199, opened 2024-04-30
feat: automatically pick up extensions when upgrading Talos

Unix4ever:
1 rule is enabled currently: when the node was ever running 1.5.x it will always have `bnx` and `intel` extensions installed.

Fixes: https://github.com/siderolabs/omni/issues/178

### DIFF
--- a/internal/backend/runtime/omni/controllers/omni/schematic_configuration.go
+++ b/internal/backend/runtime/omni/controllers/omni/schematic_configuration.go
@@ -8,7 +8,9 @@ package omni
 import (
 	"context"
 	"errors"
+	"slices"
 
+	"github.com/blang/semver"
 	"github.com/cosi-project/runtime/pkg/controller"
 	"github.com/cosi-project/runtime/pkg/controller/generic/qtransform"
 	"github.com/cosi-project/runtime/pkg/resource"
@@ -89,14 +91,10 @@ func (helper *schematicConfigurationHelper) reconcile(
 		return err
 	}
 
-	var extensionsList []string
-
 	if extensions != nil {
 		if err = updateFinalizers(ctx, r, extensions); err != nil {
 			return err
 		}
-
-		extensionsList = extensions.TypedSpec().Value.Extensions
 	}
 
 	clusterName, ok := clusterMachine.Metadata().Labels().Get(omni.LabelCluster)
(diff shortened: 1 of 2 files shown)


internal/backend/runtime/omni/controllers/omni/schematic_configuration.go
```diff
@@ -217,3 +216,78 @@ func updateFinalizers(ctx context.Context, r controller.ReaderWriter, extensions
...
+	if machineStatus.TypedSpec().Value.InitialTalosVersion == "" {
+		return nil, errors.New("machine initial version is not set")
+	}
+
+	initialVersion, err := semver.ParseTolerant(machineStatus.TypedSpec().Value.InitialTalosVersion)
+	if err != nil {
+		return nil, err
+	}
+
+	currentVersion, err := semver.ParseTolerant(cluster.TypedSpec().Value.TalosVersion)
```
smira, line 273: it's minor right now, but we should do `currentVersion.Pre = nil`, otherwise `v1.6.0-alpha.0 >= v1.6.0` is not true

internal/backend/runtime/omni/controllers/omni/schematic_configuration.go
```diff
@@ -217,3 +216,78 @@ func updateFinalizers(ctx context.Context, r controller.ReaderWriter, extensions
...
+
+	me.extensionsList = append(me.extensionsList, detected...)
+
+	if me.machineExtensions != nil {
+		for _, e := range me.machineExtensions.TypedSpec().Value.Extensions {
+			if slices.Contains(me.extensionsList, e) {
+				continue
+			}
+
+			me.extensionsList = append(me.extensionsList, e)
```
smira, line 250: we probably want it general sort list of extensions, so that we normalize schematics like `a, b, c` and `b, a, c`

Unix4ever: I had the same idea initially, but then I remember you had the discussion with Utku that we shouldn't sort them.
As one extension can have an overlap with another one and we should allow users to change the order.

smira: I meant that on Image Factory level we probably can't as it changes the output, but our extensions shouldn't depend on the order they are added, so we can sort them to reduce the schematic bloat.

utkuozdemir (approved):

PR #205, opened 2024-05-03
feat: implement the page that shows list of extensions per node

Unix4ever:
Also show the current status of each extension: `installed`, `removing` or `installing`.
`UpdateExtensions` modal now can be used to update the machine extensions.

Additional changes:
- bump Vue to 3.3 and make `Watch.vue` typed.
- fix an issue with `MachineExtensions` controller that wasn't triggering reconcile after removing extensions config for a machine.
- drop `ResourceTyped<T>` and replace it with `Resource<T, S>` everywhere.

The UI previews:
image
extensions-list

Fixes: #201


internal/backend/runtime/omni/controllers/omni/machine_extensions.go
```diff
@@ -83,14 +84,25 @@ func (ctrl *MachineExtensionsController) MapInput(ctx context.Context, _ *zap.Lo
 
 	switch ptr.Type() {
 	case omni.MachineExtensionsType:
-		root, ok := res.Metadata().Labels().Get(labelRootConfiguration)
+		clusterName, ok := res.Metadata().Labels().Get(omni.LabelCluster)
```
Unix4ever, line 87: The bug I discovered:
- `ExtensionsConfiguration` is defined for a machine and generates one `MachineExtensions` for it.
- another `ExtensionsConfiguration` is defined on the cluster level.
- `ExtensionsConfiguration` for a machine is dropped, and removes it's `MachineExtensions` resource.
- Nothing triggers reconcile for `ExtensionsConfiguration` on the cluster level.

Make `MachineExtensions` destroy trigger reconcile for the whole cluster so that other `ExtensionsConfigurations` could regenerate their own `MachineExtensions`.

smira (approved):

PR #210, opened 2024-05-04
feat: allow setting some url params for api endpoint

Simon-Boyer:
# feat: Accept apiendpoint url params

## Description

This lets the operator define url params for the api endpoint. For example https://<endpoint>/?grpc_tunnel=true. Instead of only appending the jointoken, we are parsing the url and adding it using Query.Set.

## Why

When configuring an ingress in front of omni, it is easier to use the grpc tunnel instead of opening an UDP port. But there is currently no way of setting grpc_tunnel to true by default. This change has a minimal effect on how the endpoint is handled while allowing an operator to set default param values.

### DIFF
--- a/internal/pkg/config/config.go
+++ b/internal/pkg/config/config.go
@@ -48,6 +48,7 @@ type Params struct {
 	SiderolinkWireguardBindAddress       string `yaml:"siderolinkWireguardBindAddress"`
 	SiderolinkWireguardAdvertisedAddress string `yaml:"siderolinkWireguardAdvertisedAddress"`
 	SiderolinkDisableLastEndpoint        bool   `yaml:"siderolinkDisableLastEndpoint"`
+	SiderolinkUseGRPCTunnel              bool   `yaml:"siderolinkUseGRPCTunnel"`
 
 	EventSinkPort    int                `yaml:"eventSinkPort"`
 	SideroLinkAPIURL string             `yaml:"siderolinkAPIURL"`
--- a/cmd/omni/main.go
+++ b/cmd/omni/main.go
@@ -292,6 +292,7 @@ func init() {
 		config.Config.SiderolinkWireguardAdvertisedAddress,
 		"advertised wireguard address which is passed down to the nodes.")
 	rootCmd.Flags().StringVar(&config.Config.SiderolinkWireguardBindAddress, "siderolink-wireguard-bind-addr", config.Config.SiderolinkWireguardBindAddress, "Siderolink wireguard bind address.")
+	rootCmd.Flags().BoolVar(&config.Config.SiderolinkUseGRPCTunnel, "siderolink-use-grpc-tunnel", false, "use gRPC tunnel to wrap wireguard traffic instead of UDP")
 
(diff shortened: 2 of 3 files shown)


Unix4ever:
I would rather suggest to add another flag to the command flags.
See `cmd/omni/main.go`.
Can add the new flag there and read it into `config.go` module.

If it's set then enable HTTP tunnel mode in the `manager.go`

Unix4ever (approved):

Simon-Boyer:
@Unix4ever shouldnt it be "siderolink-use-grpc-tunnel" instead of "siderolink-use-http-tunnel"? We might be able to use grpc over https but the name of the functionnality is grpc_tunnel, would be more consistent. Other than that looks good, wasnt able to work on this in past 2 days so thank you!

Unix4ever:
yeah, sure, makes sense :)
No problem. It's a small change to make

internal/pkg/config/config.go
```diff
@@ -48,6 +48,7 @@ type Params struct {
 	SiderolinkWireguardBindAddress       string `yaml:"siderolinkWireguardBindAddress"`
 	SiderolinkWireguardAdvertisedAddress string `yaml:"siderolinkWireguardAdvertisedAddress"`
 	SiderolinkDisableLastEndpoint        bool   `yaml:"siderolinkDisableLastEndpoint"`
+	SiderolinkUseGRPCTunnel              bool   `yaml:"siderolinkUseHTTPTunnel`
```
utkuozdemir, line 51: since we are at it (consistent naming), maybe also
```suggestion
	SiderolinkUseGRPCTunnel              bool   `yaml:"siderolinkUseGRPCTunnel`
```
?

Simon-Boyer: Git is really playing some tricks on me right ow haha. Locally i have this, it says its pushed, but on github it's not there. I'll try a force push i guess

Simon-Boyer: that worked, that was a bit weird haha

Unix4ever: sorry, it was me force pushing :)
I've added your changes too now. Thanks

internal/pkg/config/config.go
```diff
@@ -48,6 +48,7 @@ type Params struct {
 	SiderolinkWireguardBindAddress       string `yaml:"siderolinkWireguardBindAddress"`
 	SiderolinkWireguardAdvertisedAddress string `yaml:"siderolinkWireguardAdvertisedAddress"`
 	SiderolinkDisableLastEndpoint        bool   `yaml:"siderolinkDisableLastEndpoint"`
+	SiderolinkUseGRPCTunnel              bool   `yaml:"siderolinkUseGRPCTunnel`
```
Simon-Boyer, line 51: @Unix4ever your forced pushed removed this mistake i fixed before. Can i just push another commit or do you want to change that yourself? Just trying to make sure the commit doesnt get erased again haha

```suggestion
	SiderolinkUseGRPCTunnel              bool   `yaml:"siderolinkUseGRPCTunnel"`
```

Unix4ever: yeah, good catch. Fixed

PR #216, opened 2024-05-07
feat: implement the UI for configuring extensions during cluster create

Unix4ever:
Fixes: https://github.com/siderolabs/omni/issues/207

The UI allows selecting installed extensions before adding machines to a cluster during cluster creation or scaling.


Unix4ever:
New UI:

image
image

utkuozdemir (approved):
Maybe we can enable hello-world-service in the e2e tests on cluster create?

Unix4ever:
> Maybe we can enable hello-world-service in the e2e tests on cluster create?

Well, hello-world-service is already installed there. But we can add qemu-guest-agent there

utkuozdemir:
> > Maybe we can enable hello-world-service in the e2e tests on cluster create?
> 
> Well, hello-world-service is already installed there. But we can add qemu-guest-agent there

ah, right, I mean any small, lightweight extension to get this screen tested.

Unix4ever:
Tested both the modal window and the fact that extension is installed.

Issue #12, opened 2024-02-21
UI: unify Users and Backup Storage on the same page


Unix4ever:
Replace `Users` and `Backups Storage` link with only `Settings` in the sidebar.
And on the Setting page it should have 2 tabs on top, similar to what we have on the NodeOverview page with `Users`, `Backup Storage` options.

PR #217, opened 2024-05-08
feat: unify admin settings under `Settings` page

Unix4ever:
Use two tabs on that page: `Users`, `Backups`.

Fixes: #12 

image

### DIFF
--- a/internal/e2e-tests/main_test.go
+++ b/internal/e2e-tests/main_test.go
@@ -337,13 +337,17 @@ func (s *E2ESuite) assertClusterCreation() {
 
 		s.click(page, `button:has-text("Save")`)
 
-		s.click(page, `button#extensions-CP`)
-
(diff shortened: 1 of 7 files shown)


internal/e2e-tests/main_test.go
```diff
@@ -337,13 +337,17 @@ func (s *E2ESuite) assertClusterCreation() {
...
 		s.click(page, `button:has-text("Save")`)
 
+		err = page.Locator("button#extensions-CP").WaitFor()
+
+		s.Require().NoError(err)
+
 		s.click(page, `button#extensions-CP`)
 
-		err = page.Locator(`span:has-text("qemu-guest-agent")`).WaitFor()
+		err = page.Locator(`span:has-text("usb-modem-drivers")`).WaitFor()
```
Unix4ever, line 346: changed extension name to avoid clashing with the integration tests, as I think we enable `qemu-guest-agent` there.

utkuozdemir (approved):

Issue #27, opened 2023-12-21
UI for configuring max parallelism for workers update/scale down


utkuozdemir:
Followup on siderolabs/omni-archive#1446.

PR #229, opened 2024-05-13
feat: implement the UI for adjusting machine sets update strategies

Unix4ever:
Allow setting update and delete strategies along with max parallelism.

Fixes: https://github.com/siderolabs/omni/issues/27


Unix4ever:
https://github.com/siderolabs/omni/assets/830532/e408c9c7-91da-40f0-98e8-10f48c4bc69d

rothgar:
You may want to validate updates to control plane does not cause etcd instability (eg if 3 nodes only allow 1 node rolling). 

For worker nodes a percentage is sometimes better than hard coded rolling because if the cluster scales up (eg 300+ nodes) 5 nodes would still take a long time. IIRC EKS we would allow both settings (hard coded or percentage) and favored the larger one.

Unix4ever:
We don't allow changing this config for control planes, so that's already validated :)
It's always Rolling and 1 node at a time.

smira:
Should the `Unset` be something like "All at the same time" ? idk...

smira (approved):

PR #238, opened 2024-05-15
fix: do not try to audit machine which no longer has `MachineStatus`

Unix4ever:
It won't work for such machines, so treat their etcd member ID as orphaned.

This fix mitigates the following issue:

- The cluster has 3 controlplanes running.
- Controlplane `A` is added and joined etcd, now the cluster has 4
  members.
- Controlplane `A` is removed ungracefully by removing it's
  `Link`.
- Before `MachineSetEtcdAudit` controller removes controlplane `A`
  member controlplane `B` is added and removed ungracefully by removing it's
  `Link`.

Before the fix, after these operations the cluster would get stuck in
scaling down state due to controllers deadlock:

- `MachineSetStatus` controller waits for the etcd member `A` to be
  removed.
- `MachineSetEtcdAudit` controller tries to reach controlplane `B` which
  will never be available because it's `MachineStatus` no longer exist.


client/api/omni/specs/omni.proto
```diff
@@ -446,6 +446,9 @@ message ClusterMachineStatusSpec {
   string management_address = 6;
 
   ConfigApplyStatus config_apply_status = 7;
+
+  // IsRemoved means that the machine removed from Omni.
+  bool is_removed = 8;
```
Unix4ever, line 451: Add an extra field to not pull MachineStatus in the MachineSetEtcdAuditController

smira (approved):

Issue #24, opened 2024-01-31
UI: auto-complete for labels in the list view


smira:
Auto-complete on label key, ignoring the label prefix/domain.

PR #252, opened 2024-05-17
feat: implement labels completion for clusters and machines

Unix4ever:
New virtual resource `LabelsCompletion` can now be used to query all labels.
And it works for any resources in the default namespace. To get the completions for any resource it is enough to pass the full resource type as the completions ID.

Labels input is reusable and can be enabled for any views we need. Both mouse and keyboard control work for the completion selection. Labels now appear as the same boxes as they are displayed in the machines and cluster lists.

Label completion search looks both in keys and values until you add `:` to the line. Then it looks for keys that contain whatever is before `:` with values that contain whatever is after `:`.

Fixes: https://github.com/siderolabs/omni/issues/24


Unix4ever:
https://github.com/siderolabs/omni/assets/830532/cb8ad2eb-aec6-4098-9435-0aaf1da69e63

smira:
Looks cool... I wonder if we shouldn't apply the label to the search (showing "no matches") until it's fully built?

Unix4ever:
> Looks cool... I wonder if we shouldn't apply the label to the search (showing "no matches") until it's fully built?

Well, the thing is that you can both add the label and search by text.
So when it shows `no matches` it means it couldn't find text like that.

Unix4ever:
I don't want to drop search by text as it's still useful if you need to fine a node with some specific uuid for example.

smira (approved):

Unix4ever:
I think I _can_ do it smarter, but I think that might be even more confusing. So I think to leave it as is.

Issue #148, opened 2024-04-15
show warning when running `omnictl` of a different version than the server


Unix4ever:
It shouldn't hard fail, but it should at least display a warn

PR #254, opened 2024-05-20
feat: warn about using `omnictl` version different from the backend

Unix4ever:
Fixes: https://github.com/siderolabs/omni/issues/148

### DIFF
--- a/client/pkg/omnictl/internal/access/client.go
+++ b/client/pkg/omnictl/internal/access/client.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"os"
 
+	"github.com/blang/semver"
 	"github.com/cosi-project/runtime/pkg/resource"
 	"github.com/cosi-project/runtime/pkg/safe"
 	"github.com/cosi-project/runtime/pkg/state"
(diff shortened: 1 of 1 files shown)


client/pkg/omnictl/internal/access/client.go
```diff
@@ -159,3 +162,19 @@ If you want to enable the version validation and disable this warning, set githu
...
+		return
+	}
+
+	clientVersion, err := semver.ParseTolerant(version.Tag)
+	if err != nil {
+		return
+	}
+
+	if clientVersion.Major != backendVersion.Major || clientVersion.Minor != backendVersion.Minor {
+		fmt.Printf("[WARN] omnictl major or minor version differs from the backend version: %q vs %q.\n", clientVersion.String(), backendVersion.String())
```
smira, line 178:
```suggestion
		fmt.Fprintf(os.Stderr, "[WARN] omnictl major or minor version differs from the backend version: %q vs %q.\n", clientVersion.String(), backendVersion.String())
```

otherwise it would go into the output of e.g. `omnictl get`

smira (approved):

client/pkg/omnictl/internal/access/client.go
TimJones, line 178: I feel a simpler message is clearer. Maybe something like:
```suggestion
		fmt.Printf("[WARN] omnictl version differs from the backend version: %q vs %q.\n", clientVersion.String(), backendVersion.String())
```

TimJones (approved):

PR #264, opened 2024-05-21
chore: update Talos to 1.7.2 in the tests

Unix4ever:
Bump machinery to 1.7.2.
Enable partial machine config tests.


smira:
@Unix4ever should we update the defaults?

Unix4ever:
> @Unix4ever should we update the defaults?

Yep, sure

smira (approved):
gentle reminder about bumping default Talos/Kubernetes version

utkuozdemir (approved):

Makefile
```diff
@@ -20,9 +20,9 @@ REGISTRY_AND_USERNAME ?= $(REGISTRY)/$(USERNAME)
 PROTOBUF_GRPC_GATEWAY_TS_VERSION ?= 1.2.1
 TESTPKGS ?= ./...
 NODE_BUILD_ARGS ?=
-PROTOBUF_GO_VERSION ?= 1.34.1
+PROTOBUF_GO_VERSION ?= 1.33.0
 GRPC_GO_VERSION ?= 1.3.0
-GRPC_GATEWAY_VERSION ?= 2.20.0
+GRPC_GATEWAY_VERSION ?= 2.19.1
```
utkuozdemir, line 25: These going down is expected I guess?

utkuozdemir (approved):

Makefile
Unix4ever: Yeah. We might need to look into gRPC proxy to check if it supports that properly.
With these versions enabled all Omni API stops working.

PR #276, opened 2024-05-23
fix: use proper name for fetching existing extension configuration

Unix4ever:
The bug was leading to fetching `schematic-schematic-<machine-uuid>`, which is the incorrect name and it was always trying to create the resources instead of the update.

### DIFF
--- a/frontend/src/views/omni/Modals/UpdateExtensions.vue
+++ b/frontend/src/views/omni/Modals/UpdateExtensions.vue
@@ -161,7 +161,7 @@ const updateExtensionsConfig = async () => {
 
   try {
     const existing: Resource<ExtensionsConfigurationSpec> = await ResourceService.Get({
-      id: `schematic-${extensionsConfiguration.metadata.id}`,
+      id: extensionsConfiguration.metadata.id,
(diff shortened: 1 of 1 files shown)


utkuozdemir (approved):
I just fixed it in my secure boot branch as well :)

smira (approved):

frezbo (approved):
i guess we can't easily test this :thinking:

Unix4ever:
> i guess we can't easily test this 🤔

I will think about that, we'll need to update the extensions list and do it through the UI.

frontend/src/views/omni/Modals/UpdateExtensions.vue
```diff
@@ -161,7 +161,7 @@ const updateExtensionsConfig = async () => {
 
   try {
     const existing: Resource<ExtensionsConfigurationSpec> = await ResourceService.Get({
-      id: `schematic-${extensionsConfiguration.metadata.id}`,
+      id: `${extensionsConfiguration.metadata.id}`,
```
utkuozdemir, line 164: small nit
```suggestion
      id: extensionsConfiguration.metadata.id,
```

PR #277, opened 2024-05-23
feat: add secure boot support

### LINKED ISSUE #253: support secureboot enabled nodes
- Read secureboot info from the node resources
- Add `-secureboot` to the installer url when upgrading/installing Talos.
- We might need to do special handling for preserving kernel params when the node runs in the secureboot mode.

utkuozdemir:
Correctly handle the retrieval and updates of schematics when a Talos node has secure boot enabled.

When secure boot is enabled, we now
- Use the correct installer image, `installer-secureboot` instead of `installer`
- Preserve the kernel args in the schematic on install/upgrade instead of stripping them away.

For non-secureboot, we keep everything as-is, to avoid triggering an upgrade of existing nodes.

Closes #253.


internal/backend/runtime/omni/controllers/omni/internal/task/machine/poll.go
```diff
@@ -58,6 +60,12 @@ var resourcePollers = map[string]machinePollFunction{
 var machinePollers = map[string]machinePollFunction{
 	"version": pollVersion,
 	"disks":   pollDisks,
+
+	// we do not use a resource poller for the secure boot status, as we want to mark
+	// secure boot explicitly to disabled (contrary to leaving it nil) if the feature is not available (i.e., older Talos versions).
+	//
+	// resourcePollers skip polling the resource if it is not defined on the Talos API.
+	"secureBootStatus": pollSecureBootStatus,
```
Unix4ever, line 68: I guess it should also do it only once for the machine :thinking: 
As soon as we read it, no need to continue polling as it can't be changed.

internal/backend/runtime/omni/controllers/omni/machine_status.go
```diff
@@ -431,14 +435,24 @@ func (ctrl *MachineStatusController) handleNotification(ctx context.Context, r c
...
 			}
 
 			if spec.Schematic.Invalid {
 				spec.Schematic.InitialSchematic = ""
-			} else if spec.Schematic.InitialSchematic == "" {
-				spec.Schematic.InitialSchematic = spec.Schematic.Id
+				spec.Schematic.InitialFullId = ""
+			} else {
+				if spec.Schematic.InitialSchematic == "" {
+					spec.Schematic.InitialSchematic = spec.Schematic.Id
```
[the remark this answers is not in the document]
utkuozdemir: makes sense!

internal/backend/runtime/omni/controllers/omni/internal/task/machine/poll.go
utkuozdemir: as discussed, it will only poll when the collect task crashes, but not periodically. so it is good enough for now.

internal/backend/runtime/omni/controllers/omni/machine_status.go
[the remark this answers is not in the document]
utkuozdemir: Done, removed `initial_full_id` altogether.

hack/test/integration.sh
```diff
@@ -133,45 +133,86 @@ mkdir -p "${PARTIAL_CONFIG_DIR}"
...
+    \
+    --name test-3 \
+    --skip-injecting-config \
     --with-init-node \
-    --with-uuid-hostnames
+    --cidr=172.22.0.0/24 \
+    --no-masquerade-cidrs=172.20.0.0/24,172.21.0.0/24 \
+    --with-tpm2 \
+    --iso-path="https://factory.talos.dev/image/${SECURE_BOOT_SCHEMATIC_ID}/v${TALOS_VERSION}/metal-amd64-secureboot.iso" \
+    --encrypt-state \
```
Unix4ever, line 213: I don't think `--encrypt-state` and `--encrypt-ephemeral` do anything if you add `--skip-injecting-config`.
It generates the config with the encryption enabled, but it doesn't apply it.

Unix4ever: So I think we should rather use an Omni config patch to enable that.

utkuozdemir: I think we can just skip it then.

frezbo: > I don't think --encrypt-state and --encrypt-ephemeral do anything if you add --skip-injecting-config.
It generates the config with the encryption enabled, but it doesn't apply it.

yeh, unless `--with-apply-config` is used

utkuozdemir: removed

Unix4ever (approved):

utkuozdemir:
🎉

utkuozdemir:
As a note, the PR was passing the whole e2e suite except a test flake and a Talos issue that was resolved in main but not yet released (due 1.7.3).

PR #293, opened 2024-05-27
enable ALPN for machine API

### LINKED ISSUE #292: [feature] Support ALPN negotiation for machine api endpoint
### Problem Description

I was trying to use GKE ingress for my Omni deployment, but encountered a problem: load balancer needs HTTP/2 backends to be configured with SSL and uses ALPN negotiation for connection. I suggest enabling this capability for machine API listener.

### Solution

_No response_

### Alternative Solutions

_No response_

### Notes

_No response_

kjubybot:
Closes #292

### DIFF
--- a/internal/pkg/siderolink/manager.go
+++ b/internal/pkg/siderolink/manager.go
@@ -219,6 +219,7 @@ func (p *Params) NewListener() (net.Listener, error) {
 
 	return tls.Listen("tcp", p.APIEndpoint, &tls.Config{
 		Certificates: []tls.Certificate{cert},
+		NextProtos:   []string{"h2"},
 	})
 }
 


internal/pkg/siderolink/manager.go
```diff
@@ -219,6 +219,7 @@ func (p *Params) NewListener() (net.Listener, error) {
 
 	return tls.Listen("tcp", p.APIEndpoint, &tls.Config{
 		Certificates: []tls.Certificate{cert},
+		NextProtos:   []string{"h2"},
```
smira, line 222: this is not a machine API? not sure what kind of machine API you're looking into?

this is gRPC API so, it's always using HTTP/2

kjubybot: I refer to the machine API as in [here](https://github.com/siderolabs/omni/blob/0bd2a420e85901ba2e4d4137823bafc391896020/internal/pkg/config/config.go#L42-L44). It uses HTTP/2, yes, but does not provide ALPN negotiation, required by GKE ingress.

smira (approved):
should be fine I guess

smira:
> Um, what's up with the workflows?

They won't run until someone re-signs and pushes your PR back. That's fine.

PR #300, opened 2024-05-28
fix: use last read time as timestamp for `MachineStatus` resource

Unix4ever:
When handling events it should detect that the Talos `MachineStatus` in Omni's `MachineStatus` is stale and ignore the updates of the resource.


internal/backend/runtime/omni/controllers/omni/internal/task/machine/poll.go
```diff
@@ -317,7 +318,7 @@ func pollTalosMachineStatus(ctx context.Context, c *client.Client, info *Info) e
 		}
 
 		return &specs.MachineStatusSpec_TalosMachineStatus{
-			UpdatedAt: timestamppb.New(r.Metadata().Updated()),
+			UpdatedAt: timestamppb.New(time.Now()),
```
smira, line 321: would it cause endless `MachineStatus` updates?

Unix4ever: Shouldn't, as we poll only when the watch marks it as dirty. 
But we still will update all resources on startup.

We can probably store the version in the resource instead. And then compare it when we update `MachineStatus`.

Unix4ever: Added version check. Now it won't do any updates if the resource version isn't increased, that should guarantee that it won't update too frequently.

internal/backend/runtime/omni/controllers/omni/machine_status.go
```diff
@@ -417,7 +417,8 @@ func (ctrl *MachineStatusController) handleNotification(ctx context.Context, r c
 			spec.PlatformMetadata = event.PlatformMetadata
 		}
 
-		if event.TalosMachineStatus != nil {
+		if event.TalosMachineStatus != nil && spec.TalosMachineStatus != nil &&
```
smira, line 420: hm... if spec.TalosMachineStatus is nil (initially), how it would be set for the first time?

Unix4ever: oh, good catch

internal/backend/runtime/omni/controllers/omni/machine_status.go
```diff
@@ -417,7 +417,8 @@ func (ctrl *MachineStatusController) handleNotification(ctx context.Context, r c
 			spec.PlatformMetadata = event.PlatformMetadata
 		}
 
-		if event.TalosMachineStatus != nil {
+		if spec.TalosMachineStatus != nil && (event.TalosMachineStatus == nil ||
```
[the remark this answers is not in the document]
Unix4ever: :facepalm:

internal/backend/runtime/omni/controllers/omni/internal/task/machine/poll.go
```diff
@@ -314,6 +314,7 @@ func pollTalosMachineStatus(ctx context.Context, c *client.Client, info *Info) e
 		}
 
 		return &specs.MachineStatusSpec_TalosMachineStatus{
+			Version:   r.Metadata().Version().Value(),
 			UpdatedAt: timestamppb.New(r.Metadata().Updated()),
```
smira, line 318: so this is Talos timestamp (I guess?)

internal/backend/runtime/omni/controllers/omni/machine_status.go
```diff
@@ -417,7 +417,8 @@ func (ctrl *MachineStatusController) handleNotification(ctx context.Context, r c
 			spec.PlatformMetadata = event.PlatformMetadata
 		}
 
-		if event.TalosMachineStatus != nil {
+		if event.TalosMachineStatus != nil && (spec.TalosMachineStatus == nil ||
+			spec.TalosMachineStatus.Version < event.TalosMachineStatus.Version) {
```
smira, line 421: here we compare versions, but on reboot they reset to '1', so not sure how it would succeed until it reaches the version it was before?

internal/pkg/machinestatus/handler.go
```diff
@@ -109,7 +109,7 @@ func (handler *Handler) handleMachineStatusResourceUpdate(ctx context.Context, m
 	}
 
 	if err := handler.handleMachineStatusEvent(ctx, talosMachineStatus.Status, machineStatus.Metadata().ID(),
-		handler.Clock.Now(), false); err != nil {
+		talosMachineStatus.UpdatedAt.AsTime(), false); err != nil {
```
smira, line 112: here we compare Talos timestamps with Omni?

sorry, I'm probably lost

internal/backend/runtime/omni/controllers/omni/internal/task/machine/poll.go
Unix4ever: yeah, I've messed it up. Sorry I'm still fighting with my network setup... Will submit the fix. Now for sure.

internal/backend/runtime/omni/controllers/omni/machine_status.go
smira: I think my question here is still valid... machine was running, version of `TalosMachineStatus` went through e.g. 1..2..3.. ... 128, then it reboots and version starts counting from 1 once again.

Unix4ever: yeah. Then I don't think we have a way to limit the updates frequency.
Unless we rely on the updated at of the `MachineStatus`, which is also unreliable.

Unix4ever: unless we also count reboot count of the node and take that into equation, but that's too complicated...

PR #311, opened 2024-05-31
fix: implement the controller for handling machine status snapshot

Unix4ever:
Make the controller run tasks that can collect machine status from each machine.
Instead of changing the `MachineStatusSnapshot` directly in the siderolink events handler pass these events to the controller throught the channel, so that all events are handled in the same place.

If either event comes from siderolink or if task runner gets the machine status it updates the `MachineStatusSnapshot` resource.


internal/backend/runtime/omni/controllers/omni/internal/task/snapshot/snapshot.go
```diff
@@ -0,0 +1,214 @@
...
+					return err
+				}
+
+				snapshot.TypedSpec().Value.MachineStatus = ev
+
+				if !spec.sendInfo(ctx, snapshot, notifyCh) {
+					return nil
+				}
+			}
+		}
```
Unix4ever, line 133:
```suggestion
		case event = <-watchCh:
		}
		
		switch event.Type {
			case state.Errored:
				return fmt.Errorf("error watching COSI resource: %w", event.Error)
			case state.Bootstrapped, state.Destroyed:
				// ignore
			case state.Created, state.Updated:
				snapshot := omni.NewMachineStatusSnapshot(resources.DefaultNamespace, spec.MachineID)

				machineStatusSpec := event.Resource.Spec().(*runtime.MachineStatusSpec) //nolint:forcetypeassert,errcheck

				ev, err := convertStatus(machineStatusSpec)
				if err != nil {
					return err
				}

				snapshot.TypedSpec().Value.MachineStatus = ev

				if !spec.sendInfo(ctx, snapshot, notifyCh) {
					return nil
				}
			}
		}
```

utkuozdemir (approved):

Issue #219, opened 2024-05-08
UI: report the reason for 'Not Ready' machine


smira:
Report the reason for the `Machine` to be !ready.

Read the `MachineStatus` reason for Not Ready.

PR #316, opened 2024-06-03
feat: display more data on the node overview page

Unix4ever:
Now the page has an additional panel with:

- machine stage from the `ClusterMachineStatus`.
- last active time from `MachineStatusLink`.
- unmet conditions from Talos `MachineStatus`.
- CPU and memory information from the Omni `MachineStatus`.

Additionally display labels and allow editing them in the separate panel.

Also add missing service phases to the service events UI.

Fixes: https://github.com/siderolabs/omni/issues/219


Unix4ever:
image
image

internal/backend/runtime/cosi/cosi.go
```diff
@@ -169,7 +169,9 @@ func WatchLegacy(ctx context.Context, st state.State, md resource.Metadata, out
 					return fmt.Errorf("failed to encode resource to send back to the client %w", err)
 				}
 
-				items = append(items[:index], items[index+1:]...)
+				if items != nil {
+					items = append(items[:index], items[index+1:]...)
```
DmitriyMV, line 173:
```suggestion
					items = slices.Delete(items, index, index+1)
```

DmitriyMV (approved):
Small nit.

utkuozdemir (approved):

utkuozdemir:
Just an idea, you can consider showing the secure boot status on the overview page as well. I put it on the machines screen, but here also could be nice, with some color maybe.

Unix4ever:
> Just an idea, you can consider showing the secure boot status on the overview page as well. I put it on the machines screen, but here also could be nice, with some color maybe.

Sure, not a problem.
I'm going to drop `Hardware` title though to make it occupy the same size

Unix4ever:
image

utkuozdemir:
> image

beautiful

PR #314, opened 2024-06-03
chore: bump deps

DmitriyMV:
- run rekres and fix nolint directives
- bump deps (keep gen to 0.4.8 for now) for server, client and tests

PR #318, opened 2024-06-03
chore: replace grpc.Dial* with grpc.NewClient

DmitriyMV:
That should silence `staticcheck` linter.


DmitriyMV:
Continues #314

internal/backend/grpc/router/router.go
```diff
@@ -75,7 +75,12 @@ func NewRouter(
 	authEnabled bool,
 	verifier grpc.UnaryServerInterceptor,
 ) (*Router, error) {
-	omniConn, err := grpc.Dial(transport.Address, //nolint:staticcheck
+	address := transport.Address
+	if address == "grpc-conn" {
+		address = "passthrough:whatever"
```
Unix4ever, line 80: Is that the fix 🙃

DmitriyMV: Yup

smira (approved):

Unix4ever (approved):

PR #321, opened 2024-06-04
fix: fix config generation failure due to secure boot status

utkuozdemir:
A change in the `SecureBootStatus` in the `MachineStatus` did not cause a new config to be generated. This was caused by the `ClusterMachineConfigController` not taking the `MachineStatus` into the calculation of the hash from the inputs (`UpdateInputsVersions`) to determine if it should generate a new config or not.

Modify the existing `MachineConfigGenOptions` resource to hold the required information to build the install image. Instead of using `MachineStatus+ClusterMachineTalosVersion` to build the install image URL, use only `MachineConfigGenOptions` resource.

Additionally, add a migration to pre-populate the new fields in the `MachineConfigGenOptions` resources, so that they will not be modified on the next Omni upgrade. Also update the input versions of the `ClusterMachineConfig` to not contain `ClusterMachineTalosVersion` anymore, so that we won't trigger a config re-apply on the next Omni upgrade.


internal/backend/runtime/omni/controllers/omni/cluster_machine_config.go
```diff
@@ -450,8 +456,8 @@ func buildInstallImage(clusterMachineTalosVersion *omni.ClusterMachineTalosVersi
 	schematicID := schematicConfig.Id
 
 	secureBootStatus := machineStatus.TypedSpec().Value.SecureBootStatus
-	if secureBootStatus == nil {
-		return "", xerrors.NewTaggedf[qtransform.SkipReconcileTag]("secure boot status for machine %q is not yet set", machineStatus.Metadata().ID())
+	if secureBootStatus == nil { // should never happen - must have been handled before entering this function
+		return "", fmt.Errorf("machine %q has no secure boot status set", machineStatus.Metadata().ID())
```
Unix4ever, line 460: do we need this check then?

Unix4ever (approved):
LGTM.

But the fact that the machines running Talos 1.7.2 have no secure boot status information also looks like a bug.

Unix4ever:
And maybe we should add a unit-test in the controller that will verify that there's no bug anymore

internal/backend/runtime/omni/controllers/omni/cluster_machine_config.go
utkuozdemir: we don't really need it, but isn't it still better than accidentally switching from secureboot to non-secureboot if we ever use this function from a 3rd place in the future?

utkuozdemir:
> LGTM.
> 
> But the fact that the machines running Talos 1.7.2 have no secure boot status information also looks like a bug.

They actually have the status set correctly on the machinestatus - the generation error we see is obsolete/outdated.

Unix4ever:
>They actually have the status set correctly on the machinestatus - the generation error we see is obsolete/outdated

But why it's Unknown? Should it be disabled instead?

utkuozdemir:
> > They actually have the status set correctly on the machinestatus - the generation error we see is obsolete/outdated
> 
> But why it's Unknown? Should it be disabled instead?
> 
> I think there's some bug in the JS code.

Makes sense.

So, the MachineStatus seems completely right:
```
    securebootstatus:
        enabled: false
```

But the `ClusterMachineConfig` is not aware of it.

I think I just found another issue here - here, for the obsolete error to remain, the change in `MachineStatus` (the change that sets the `securebootstatus` to be not-nil and `enabled: false`) should not trigger another generation of config. I think it also happened here, because before generating a new config, we calculate the input versions, and if there is no change in the versions, we just skip it. And while doing it, it seems we don't take `MachineStatus` into consideration.

internal/backend/runtime/omni/controllers/omni/cluster_machine_config.go
Unix4ever: This function is internal for this controller.
I don't think we should ever call it from the other places.

utkuozdemir: We already call it from two places - while generating config from `ClusterMachineConfigController` and while upgrading a node from `ClusterMachineConfigStatusController`.

Unix4ever: ahh, that's `buildInstallImage`. I got confused.

Unix4ever: ignore me then :)) I thought it's part of reconcileMachineConfig

Unix4ever:
> Question is, if I add it to the input versions calculation, will it trigger too many config generations?

I guess it will, but every update with the same state of the secure boot will be no-op.
I would actually prefer to get rid of the `MachineStatus` in this controller.
It would be better to split secure boot and schematic parts to a separate resource.

utkuozdemir:
> > Question is, if I add it to the input versions calculation, will it trigger too many config generations?
> 
> I guess it will, but every update with the same state of the secure boot will be no-op. I would actually prefer to get rid of the `MachineStatus` in this controller. It would be better to split secure boot and schematic parts to a separate resource.

Yep, I got a similar idea now - we copy over the secure boot status into `ClusterMachineTalosVersion`, just like we do with the schamatic ID. And here we read it from that instead.

utkuozdemir:
@Unix4ever I ended up using `ClusterMachineTalosVersion` for now, as it already has a working flow. Avoided adding a new resource into the mix atm which potentially cause some other parts to break.

I think we can pair and do a bigger rework later to simplify the whole machine customization flow.

Unix4ever:
We might need to do migration here to avoid triggering the config apply.

And if we rework this later we'll have to do another migration.

As both changes are no-op but change the inputs version, the config will get regenerated in the controllers if we don't do anything.

utkuozdemir:
> We might need to do migration here to avoid triggering the config apply.
> 
> And if we rework this later we'll have to do another migration.
> 
> As both changes are no-op but change the inputs version, the config will get regenerated in the controllers if we don't do anything.

Done as discussed, including the migration. Please have another look so we don't miss anything.

internal/backend/runtime/omni/controllers/omni/talos_upgrade_status.go
```diff
@@ -448,13 +453,16 @@ func updateMachine(ctx context.Context, r controller.ReaderWriter, logger *zap.L
...
 
 	helpers.CopyAllLabels(machine, res)
 
-	res.TypedSpec().Value.TalosVersion = talosVersion
-	res.TypedSpec().Value.SchematicId = schematicID
+	res.TypedSpec().Value.TalosVersion = schematicConfig.TypedSpec().Value.TalosVersion
+	res.TypedSpec().Value.SchematicId = schematicConfig.TypedSpec().Value.SchematicId
+	res.TypedSpec().Value.SchematicInitialized = schematicConfig.TypedSpec().Value.SchematicInitialized
+	res.TypedSpec().Value.SchematicInvalid = schematicConfig.TypedSpec().Value.SchematicInvalid
+	res.TypedSpec().Value.SecureBootStatus = schematicConfig.TypedSpec().Value.SecureBootStatus
```
Unix4ever, line 465: Not sure about this forwarding.
`TalosUpgradeStatusController` still has `MachineStatus` as the input, so maybe we can read it from there directly instead of adding the same fields to 2 types of resources?

utkuozdemir: ah cool, that makes sense, also makes things easier. will do.

utkuozdemir: done

utkuozdemir:
@Unix4ever ended up adding it into `MachineConfigGenOptions` which simplified things a lot.

internal/backend/runtime/omni/controllers/omni/cluster_machine_config.go
```diff
@@ -218,37 +200,30 @@ func reconcileClusterMachineConfig(
...
 	// skip if the machine schematic information is not yet detected
-	if machineStatus.TypedSpec().Value.Schematic == nil {
+	if !installImage.SchematicInitialized {
 		logger.Error("machine schematic is not set, skip reconcile")
 
 		return xerrors.NewTagged[qtransform.SkipReconcileTag](errors.New("machine schematic is not set detected"))
 	}
 
+	if installImage.SecureBootStatus == nil {
+		logger.Error("secure boot status is not set, skip reconcile")
```
Unix4ever, line 218:
```suggestion
		logger.Error("secure boot status is not detected, skip reconcile")
```

internal/backend/runtime/omni/migration/helpers.go
```diff
@@ -30,7 +30,7 @@ func reconcileConfigInputs(ctx context.Context, st state.State, item *omni.Clust
 		return err
 	}
 
-	// update input versions on the cluster machine config to avoid it's reconciliation
+	// update input versions on the cluster machine config to avoid its reconciliation
```
Unix4ever, line 33: As you dropped `ClusterMachineTalosVersion` from the config inputs you also need to drop it here.

PR #325, opened 2024-06-05
feat: add Cache-Control header for head requests

### LINKED ISSUE #324: [feature] Add cache-control header to head requests
### Problem Description

Hey! I faced with issue using Cloudflare as a proxy for Omni. Omni frontend uses this workflow to download the image from the backend:
1. Check/trigger the creation of the image via a `HEAD` request to the `/image/xxx` route
2. Download the image via `GET` request

The problem is here, that Cloudflare caches all requests, and they [convert](https://developers.cloudflare.com/cache/concepts/cache-behavior/#interaction-of-head-requests-with-cache) `HEAD` request to `GET` request by default, which broke the authentication because [payload](https://github.com/siderolabs/go-api-signature/blob/main/pkg/message/http.go#L129) contains the method too.

### Solution

Cloudflare (and others I believe) support no caching behavior based on standard [headers in the request](https://developers.cloudflare.com/cache/concepts/default-cache-behavior/), so we can just add the header `Cache-Control: no-store` to our HEAD request to bypass caching in CF.

### Alternative Solutions

Currently, I bypass the caching by using Cloudflare caching rules with exceptions for all `/image` URIs, but it isn't very flexible.

### Notes

I will prepare PR to resolve this issue

selfuryon:
Added `Cache-Control: no-store` header to all `HEAD` requests in `/image` path.

Close #324

### DIFF
--- a/frontend/src/views/omni/Modals/DownloadInstallationMedia.vue
+++ b/frontend/src/views/omni/Modals/DownloadInstallationMedia.vue
@@ -320,7 +320,7 @@ const download = async () => {
 
     phase.value = Phase.Generating;
 
-    await doRequest(url, { signal: controller.signal, method: "HEAD" });
+    await doRequest(url, { signal: controller.signal, method: "HEAD", headers: { "Cache-Control": "no-store" } });
 
     phase.value = Phase.Loading;
 


frontend/src/methods/key.ts
```diff
@@ -262,6 +262,10 @@ const registerInterceptors = () => {
         } else if (url.indexOf("/image/") == 0) {
           config.headers.set(TimestampHeaderKey, ts);
 
+          if (config.method == "HEAD") {
+            config.headers.set(`Cache-Control`, "no-store");
+          }
```
Unix4ever, line 267: This works, but I'd rather suggest to add this header here:

https://github.com/siderolabs/omni/blob/main/frontend/src/views/omni/Modals/DownloadInstallationMedia.vue#L323

Unix4ever (commented):
Thanks for the PR

frontend/src/methods/key.ts
selfuryon: Do you mean to add a third option argument with headers and pass it from this place?

Unix4ever: Remove this `if`.
And change the line I posted to:
```
    await doRequest(url, { signal: controller.signal, method: "HEAD", headers: { "Cache-Control": "no-store" } });
```

will do the trick

Unix4ever: So yeah, I guess that what you mean :)

selfuryon: Done! Didn't know that I can do that, sorry, I'm not a TS developer :D

Unix4ever (approved):

frontend/src/methods/key.ts
Unix4ever: None of us is a TS developer here :D

Looks like my suggestion wasn't correct. But I'll fix that.

Issue #294, opened 2024-05-27
experiment: replace `node.js` & `npm` with Bun


smira:
https://bun.sh/

PR #326, opened 2024-06-05
chore: migrate to Vite and Bun to build the frontend

Unix4ever:
Dropped some dependencies which no longer worked:

- monaco-editor-vue3 - replaced with a tiny wrapper.
- polyfill for streaming classes - replaced with vite plugin for polyfills.
- buffer - replaced base64 encoding with calls from the `fetch.pb.ts`.

Dropped the code that sets up monaco in `main.ts` in favor of Vite monaco plugin that does that for you.

Updated `monaco-yaml` as the old version didn't work with Vite.

Switched from `Jest` to `bun:test`.

Use TypeScript in the tailwind config, as it imports `colors.ts` now. Had to use TypeScript for `colors` as Vite can't import CommonJS modules.

Add eslint linter to the `lint` dependencies, enable a bit more linters and fix all detected errors.

Fixes: https://github.com/siderolabs/omni/issues/294


smira (approved):
I personally love the -34k lines

Do we need to have a matching kres update?

Unix4ever:
> I personally love the -34k lines
> 
> Do we need to have a matching kres update?

yep, will create the PR for it now

utkuozdemir (approved):

Unix4ever:
I'll cut 0.38.0-alpha.0 with this change and roll it out to sandbox.
We need to properly test that no UI features were broken after this change, as I've basically recreated vue project from scratch.

PR #330, opened 2024-06-06
fix: pass through the `talosctl -n` args if they cannot be resolved

utkuozdemir:
We were not correctly checking if the nodes passed via `talosctl --nodes` were resolved before replacing the target in the GRPC metadata with the resolved ones.

We were handling a single node in the metadata correctly, but not doing the same handing for multiple nodes. This PR fixes that.

### DIFF
--- a/internal/backend/grpc/router/resolve.go
+++ b/internal/backend/grpc/router/resolve.go
@@ -9,10 +9,14 @@ import (
 	"strings"
 
 	"github.com/siderolabs/gen/xslices"
-	"github.com/siderolabs/go-api-signature/pkg/message"
 	"google.golang.org/grpc/metadata"
 
 	"github.com/siderolabs/omni/internal/backend/dns"
+)
+
+const (
+	nodeHeaderKey  = "node"
+	nodesHeaderKey = "nodes"
 )
 
 // NodeResolver resolves a given cluster and a node name to an IP address.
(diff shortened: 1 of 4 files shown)


internal/backend/grpc/router/resolve.go
```diff
@@ -25,6 +25,20 @@ type resolvedNodeInfo struct {
 	nodes []dns.Info
 }
 
+func (i resolvedNodeInfo) nodesOK() bool {
```
smira, line 28: I think this is wrong fix, as it goes "all or nothing", so if a mix of resolved/unresolved is treated as "unresolved"

I feel much easier fix is to make backend resolver return same record if not found (or wrap it to return same entry if not found)

utkuozdemir: Reworked it a bit, pls have another look.

internal/backend/grpc/router/talos_backend.go
```diff
@@ -107,17 +107,22 @@ func (l *TalosBackend) GetConnection(ctx context.Context, fullMethodName string)
...
 	// overwrite the node headers with the resolved ones
+
 	resolved := resolveNodes(l.nodeResolver, md)
-	if resolved.node.Address != "" {
-		md = md.Copy()
 
-		setHeaderData(ctx, md, "node", resolved.node.Address)
-	} else if len(resolved.nodes) > 0 {
+	if len(resolved) > 0 {
+		headerKey := "node"
```
smira, line 114: idk if this is going to be a problem, but it seems to rewrite `node`/`nodes` based on result count?

I don't quite know all the details, but would it be easier to write something like:

```go
if node, ok := md.Get("node"); ok {
    md.Set("node", resolveNodes(node))
}

if nodes, ok := md.Get("nodes"); ok {
    md.Set("node", resolveNodes(nodes)...)
}
```

Unix4ever:
Can we get some unit tests implemented for it?

internal/backend/grpc/router/talos_backend.go
utkuozdemir: That wouldn't work exactly because of the code structure, but I did something similar to that.

utkuozdemir:
> Can we get some unit tests implemented for it?
> 
> I see we have some tests, but could be cool to verify that we no longer have the bug we're trying to fix

Agreed, done.

Unix4ever (approved):

Issue #111, opened 2023-10-09
[feature] support kubelogin grant-type=authcode-keyboard in omni


nberlee:
### Problem Description

Our team utilizes an SSH stepping stone or bastion host to access our Kubernetes clusters. This approach ensures segmented access and enhanced security as it eliminates the possibility to store manifests with potential secrets on employee laptops. However, our SSH stepping stone doesn’t support Xorg/Wayland, hence making it impossible to login into Keycloak using browsers, also textbrowsers like lynx lack JavaScript support.

We came across a potential solution in the kubelogin docs that outlines the use of grant-type=authcode-keyboard. This feature facilitates an authorization code flow login utilizing a keyboard, bypassing the need for a graphical interface on the same host as the terminal.

```sh
$ kubectl get nodes
Please visit the following URL in your browser: https://<interal-url-to-omni>/oidc/authorize?access_type=offline&client_id=native&code_challenge=<redacted>&code_challenge_method=S256&nonce=<redacted>&redirect_uri=urn%3Aietf%3Awg%3Aoauth%3A2.0%3Aoob&response_type=code&scope=cluster%3<cluster>+openid&state=<redacted>
Enter code:
```

However, upon attempting this method, we encountered an error indicating that the requested redirect_uri is not configured in the client settings. The error message: `The requested redirect_uri is missing in the client configuration. If you have any questions, you may contact the administrator of the application.` Our initial assumption was that this issue originated from Keycloak, but further investigation suggested that Omni might be the source of this message. We did attempt to add urn:ietf:wg:oauth:2.0:oob to the redirect_uri within Keycloak for the Omni client, but the issue persisted.

### Solution

Support authcode-keyboard mode of kubectl oidc-login (kubelogin)

### Alternative Solutions

Any way to login with only a terminal (authorization code flow login utilizing a keyboard, bypassing the need for a graphical interface)

rothgar:
@nberlee can you review this configuration in my PR and see if that works for you? https://github.com/siderolabs/omni-docs/pull/77/files

rothgar:
I thought about port conflicts for shared/jump hosts. Thanks for sharing your script on working around that issue. It looks like you run the tunnel in the background, do you ever clean it up when someone disconnects?

I'm assuming support for `authcode-keyboard` would still be preferred for your environment.

nberlee:
[...]

PR #333, opened 2024-06-06
feat: support authcode login in `kubectl oidc-login`

Unix4ever:
Omni now allows using `--grant-type=authcode-keyboard` option in `oidc-login`.
That allows to log in without forwarding `:8000` port from the remote machine. Instead kubelogin will present a link and ask for the one time code.

Clicking the link will open Omni web page with one time code.

Fixes: https://github.com/siderolabs/omni/issues/111


Unix4ever:
The full flow looks like this

https://github.com/siderolabs/omni/assets/830532/079ae6f7-3902-4e17-9896-565ce4f32a85

client/pkg/omnictl/kubeconfig.go
```diff
@@ -35,13 +35,22 @@ var kubeconfigCmdFlags struct {
...
 	serviceAccount       bool
 }
 
+var allGrantType = strings.Join([]string{
+	"auto",
+	"authcode",
+	"authcode-keyboard",
+	"password",
+	"device-code",
+}, "|")
```
utkuozdemir, line 52:
```suggestion
var allGrantTypes = strings.Join([]string{
	"auto",
	"authcode",
	"authcode-keyboard",
	"password",
	"device-code",
}, "|")
```

not all of these are working, right?

Unix4ever: Probably not, I haven't tried

Unix4ever: But the kubeconfig is valid with that, so I don't know

utkuozdemir: Should we validate against only the ones we know to be working? To avoid generating broken kubeconfigs?

Unix4ever: I guess that makes sense

utkuozdemir (approved):
🆒

client/pkg/omnictl/kubeconfig.go
Unix4ever: updated. Left only `auto`, `authcode` and `authcode-keyboard`. Other 2 don't work

internal/backend/grpc/oidc.go
```diff
@@ -49,11 +51,50 @@ func (s *oidcServer) Authenticate(ctx context.Context, req *oidc.AuthenticateReq
...
+func encodeToString(max int) (string, error) {
+	b := make([]byte, max)
+
+	n, err := io.ReadAtLeast(rand.Reader, b, max)
+	if n != max {
+		return "", err
+	}
+
+	for i := 0; i < len(b); i++ {
+		b[i] = table[int(b[i])%len(table)]
```
smira, line 94: this is not a exactly a good way, as %10 is biased towards some numbers more when the input is a byte (0..255). 

the easiest might be to have 16 characters (any power of 2 would work).

Unix4ever: added letters

PR #335, opened 2024-06-06
test: fix flaky test in `ClusterMachineConfigStatus` unit tests

utkuozdemir:
The test `TestSchematicChanges` was flaky, as it was doing the following:

1. Trigger one or more `factory.talos.dev` build image generations & talos node upgrades by modifying resources
2. Capture these upgrade calls
3. Assert the build image in these calls (to be one or more with, always the expected image)
4. Clear captured upgrade calls
5. Trigger one or more `ghcr.io` build image generations & talos node upgrades by modifying resources
6. Capture these upgrade calls
7. Assert the build image in these new calls (to be one or more with, always the expected image)

The problem was, between step 4 and 7, an old `factory.talos.dev` build image could sometimes still be captured due to the nature of controllers asynchronously reconciling resources.

Replace this logic by not clearing the upgrade requests at all, instead, simply ignoring the previous image's preceding captured requests in step 7.

### DIFF
--- a/internal/backend/runtime/omni/controllers/omni/cluster_machine_config_status_test.go
+++ b/internal/backend/runtime/omni/controllers/omni/cluster_machine_config_status_test.go
@@ -16,6 +16,7 @@ import (
 	"github.com/cosi-project/runtime/pkg/resource/rtestutils"
 	"github.com/cosi-project/runtime/pkg/safe"
 	"github.com/google/uuid"
+	"github.com/siderolabs/gen/xslices"
 	"github.com/siderolabs/go-retry/retry"
 	"github.com/siderolabs/talos/pkg/machinery/api/machine"
 	"github.com/siderolabs/talos/pkg/machinery/resources/runtime"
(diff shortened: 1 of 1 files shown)


internal/backend/runtime/omni/controllers/omni/cluster_machine_config_status_test.go
```diff
@@ -530,24 +531,23 @@ func (suite *ClusterMachineConfigStatusSuite) TestSchematicChanges() {
 		suite.Require().NoError(err)
 	}
 
+	expectedFactoryImage := "factory.talos.dev/installer/bbbb:v1.3.0"
```
frezbo, line 534: what's `bbbb`, or maybe call unexpected image?

utkuozdemir: `bbbb` is just the schematic ID we use for testing. It is used at the previous step. We can use any string here, we just test with it.

Here we set it as the schematic ID: https://github.com/siderolabs/omni/blob/main/internal/backend/runtime/omni/controllers/omni/cluster_machine_config_status_test.go#L524

And here we assert that it is in the install URL: https://github.com/siderolabs/omni/blob/main/internal/backend/runtime/omni/controllers/omni/cluster_machine_config_status_test.go#L539

The fix I did is not related to it. My fix is: "if we still receive an Talos upgrade request with the previous install image after updating the resources to trigger a new one, do not fail, as it can happen".

Unix4ever (approved):

PR #336, opened 2024-06-07
fix: change Transport.Address field to Transport.Address method

DmitriyMV:
With new gRPC (both gateway and modules) it uses `grpc.NewClient` call to create clients. It no longer support custom addresses without a `passthrough:` prefix. Previous fix didn't account for that in some places, so this one changes the structure of `Transport` to always return address in proper form for external users.


utkuozdemir (approved):

Unix4ever (approved):

smira (approved):

DmitriyMV:
All tests [passed](https://github.com/siderolabs/omni/actions/runs/9417959412/job/25944300229?pr=336) rebasing on main and merging. Removing `integration` tags.

PR #346, opened 2024-06-10
fix: proper time adjustment to fix flaky TestEtcdManualBackup

DmitriyMV:
I have run it with `–count=100` and it no longer fails.

The https://github.com/golang/go/issues/67434 can't come soon enough, so we can drop all fake clocks.


utkuozdemir (approved):

smira (approved):
https://github.com/jonboulle/clockwork

This packages has interesting `BlockUntil` helpers which allow to wait for the clock consumers to get blocked

DmitriyMV:
> https://github.com/jonboulle/clockwork
> This packages has interesting BlockUntil helpers which allow to wait for the clock consumers to get blocked

Which we actually use :D. Then problem is that it doesn't account for goroutines blocking on chan read or mutexes or whatever.  https://github.com/golang/go/issues/67434  actually does, so when you will ask it for world to stop, it will actually wait for the world to stop.

PR #350, opened 2024-06-10
feat: use circular buffer's new persistence option for machine logs

utkuozdemir:
Switch from the old machine log storage approach to using the new persistence feature in the go-circular: https://github.com/siderolabs/go-circular/pull/4

Implement a migration code to migrate from old format to the new format when a machine log buffer is initialized.

Move some hardcoded log buffer settings into config (CLI args).

Use a default jitter of 0.1 (10+-1 minutes).

Rework the CLI arg names to be consistent.

Closes siderolabs/omni#118.


internal/pkg/config/config.go
```diff
@@ -219,10 +226,15 @@ var (
...
 		},
 		LogServerPort: 8092,
-		LogStorage: LogStorageParams{
-			Enabled:     true,
-			Path:        "_out/logs",
-			FlushPeriod: 10 * time.Minute,
+		MachineLogConfig: MachineLogConfigParams{
+			BufferInitialCapacity: 16384,
+			BufferMaxCapacity:     131072,
+			BufferSafetyGap:       2048,
```
smira, line 232: probably you can lower this, as with compressed chunks it's mostly unused?

smira (approved):

Unix4ever (approved):

Issue #33, opened 2023-11-08
provide break the glass Talos and Kubernetes API workload access


smira:
Allow Admin users to download at any time:

* raw `taloscnfig` with `os:operator` role (we can populate endpoints with machine IPs)
* raw `kubeconfig` with `system:masters` role (we can't provide an endpoint? use first cp machine IP?)

Endpoints/LBs/network access are the responsibility of the user.

If the `talosconfig`/`kubeconfig` leaks, it's the responsibility of the user.

This doesn't give more access to Talos/Kubernetes API when compared to access via Omni as a proxy.

This feature might be enabled per instance (?).

smira:
Follow-up work: #220

PR #360, opened 2024-06-11
feat: provide a way to get admin `talosconfig` and `kubeconfig`

Unix4ever:
Fixes: https://github.com/siderolabs/omni/issues/33

It is now possible to get full access `kubeconfig` and `talosconfig` (operator role), if the Omni instance has `enable-break-glass-configs` flag enabled.

They can be downloaded using cli commands:

`omnictl kubeconfig --break-glass --cluster <name>`
`omnictl talosconfig --break-glass --cluster <name>`

After you download the config the cluster will be marked with `omni.sidero.dev/tainted` annotation to keep in mind that this cluster has weaker security and might need to get secrets rotation in the future.


internal/backend/grpc/management.go
```diff
@@ -291,6 +345,47 @@ func (s *managementServer) adminTalosconfig(ctx context.Context) (*management.Ta
...
+	if err != nil {
+		return nil, err
+	}
+
+	data, err := kubernetesRuntime.AdminKubeconfig(ctx, clusterName)
+	if err != nil {
+		return nil, err
+	}
+
+	if err = s.markClusterAsTainted(ctx, clusterName); err != nil {
```
smira, line 380: Cluster is a user-owned resource, so they could remove that taint? Should we use another resource?

internal/backend/runtime/omni/omni.go
```diff
@@ -496,6 +499,40 @@ func (r *Runtime) AdminTalosconfig(ctx context.Context, clusterName string) ([]b
 	return omni.NewTalosClientConfig(talosConfig, endpoints...).Bytes()
 }
 
+// AdminTalosconfig returns the os:operator talosconfig for the cluster with the given name.
+func (r *Runtime) AdminTalosconfig(ctx context.Context, clusterName string) ([]byte, error) {
```
smira, line 503: Should it be OperatorTalosconfig? As it's not fully admin

utkuozdemir:
Maybe only the `Admin` should be able to do that, as it bypasses the user management system. Any specific reason to pick `Operator`?

client/api/omni/management/management.proto
```diff
@@ -91,6 +93,7 @@ message KubeconfigRequest {
   string service_account_user = 3;
   repeated string service_account_groups = 4;
   string grant_type = 5;
+  bool admin = 6;
```
utkuozdemir, line 96: Does this mean "break-the-glass service account"? Isn't the feature only for emergency use cases? Using it in automation would be abuse of the feature, no?

client/pkg/omnictl/talosconfig.go
```diff
@@ -140,9 +142,10 @@ func init() {
 	talosconfigCmd.Flags().StringVarP(&talosconfigCmdFlags.cluster, "cluster", "c", "", "cluster to use")
 	talosconfigCmd.Flags().BoolVarP(&talosconfigCmdFlags.force, "force", "f", false, "force overwrite of talosconfig if already present")
 	talosconfigCmd.Flags().BoolVarP(&talosconfigCmdFlags.merge, "merge", "m", true, "merge with existing talosconfig")
+	talosconfigCmd.Flags().BoolVar(&talosconfigCmdFlags.getAdminConfig, "admin", false, "get admin talosconfig (if enabled)")
```
utkuozdemir, line 145: Naming question: should we call it simply "admin kubeconfig"? It kinda implies that it is part of a regular workflow and not something reserved for break-glass scenarios.

internal/backend/grpc/management.go
utkuozdemir: Applying a cluster template would remove it as well.

client/api/omni/management/management.proto
Unix4ever: We can rename it, not sure about the new name though

internal/backend/grpc/management.go
utkuozdemir: I think `ClusterStatus` is the right place, but why copy? Just write on it?

client/api/omni/management/management.proto
utkuozdemir: no idea :)

Unix4ever:
> Maybe only the Admin should be able to do that, as it bypasses the user management system. Any specific reason to pick Operator?

Only Admin can do that, I've updated that check in the API.

We pick Operator to give the same access level as Omni would provide, except you can talk to the nodes directly.

client/api/omni/management/management.proto
Unix4ever: Can call it break-glass

internal/backend/grpc/management.go
Unix4ever: It's owned by the controller, didn't want to interfere that

utkuozdemir: Oh, you are right. Copying would be kind of ok, my only concern would be, when I sync my cluster template next time, it would report as changed, when actually nothing should be changed.

Would adding another resource be an overkill? Is there any other existing ownerless resource we can use?

client/api/omni/management/management.proto
utkuozdemir: I'm not super sure if we should change it, open for more opinions. @smira wdyt?

internal/backend/grpc/management.go
Unix4ever: Well, not a big deal. Can add a resource too. But it will be possible to destroy this resource too unless we set an owner

Unix4ever: Ah wait nvm, we can block write access there.

client/api/omni/management/management.proto
smira: `break_glass` sounds better to me, and probably we should do same for the CLI? 

as it's not 'admin', it's same level of access as via Omni

andrewrynhard:
The only thing I would add would be that “tainted” is a pretty broad term and as a user I would want to know why it is “tainted”. I can’t think of a better descriptive name at the moment. Especially with the meaning of tainted in Kubernetes being different than what we mean here.

It also feels like there is some overlap in what we mean by admin here. We have admin roles but this kubeconfig/talosconfig is different from those roles. I think “emergency” better describes the intent of this feature and separates things clearly.

Unix4ever:
> The only thing I would add would be that “tainted” is a pretty broad term and as a user I would want to know why it is “tainted”. I can’t think of a better descriptive name at the moment. Especially with the meaning of tainted in Kubernetes being different than what we mean here.
> 
> It also feels like there is some overlap in what we mean by admin here. We have admin roles but this kubeconfig/talosconfig is different from those roles. I think “emergency” better describes the intent of this feature and separates things clearly.

Tainted in the sense that we can no longer guarantee secrets protection. You grabbed admin config, if you break your cluster or if you forget to rotate the secrets and it leaks it's your fault.

client/api/omni/management/management.proto
Unix4ever: updated to `break_glass` everywhere

Unix4ever:
The name of the resource isn't that important, it's more for our internal usage right now.
And if we decide to show the UI, we can give a broad explanation there.

utkuozdemir (approved):

Issue #17, opened 2024-02-07
🐛 dns name for an exposed service takes long time to be created


smira:
## Problem

If we should workload proxy service entry before it's ready (fully provisioned), and the user clicks on it, negative DNS response is cached on the user side (for ~3 minutes).

## Solution 1: New Naming Scheme

`*.proxy-$CELL.omni.siderolabs.io` -> DNS entry, TLS cert

`p-g3a4ana-demo.proxy-us.omni.siderolabs.io`

### Solution 1a 

Use nginx ingresses (one per workload service).

### Solution 1b

Implement a mini-router which will pparse `p-<workload>-<user>` and forward the request to a matching instance.

## Solution 2: Backwards Flow (only show service when it's ready)

Add a backwards flow from the CRD/omni-controller to report when the exposed service entry is ready (both DNS and Ingress are ready), and report it via sidecar to Omni.

If the workload service is not read yet, show it as 'Pending' in the UI.

## Decision

Go with solution 1a.

PR #364, opened 2024-06-12
feat: use the new domain scheme for exposed services

utkuozdemir:
Rework the workload service proxying feature to support the new domain format in addition to the existing one.

Old format:
```
p-g3a4ana-demo.omni.siderolabs.io
```

New format:
```
g3a4ana-demo.proxy-us.omni.siderolabs.io
```

The old format required a new DNS records to be added for each new workload service, causing issues with the resolution on clients. The new format addresses it by leveraging wildcard records.

Additionally, build the full exposed service URL on the backend and make it a field on `ExposedService` resource, so they can be accessed using `omnictl get exposedservice`.

Part of siderolabs/omni#17.


internal/backend/runtime/omni/controllers/omni/kubernetes_status.go
```diff
@@ -240,6 +257,29 @@ func (ctrl *KubernetesStatusController) updateExposedServices(ctx context.Contex
...
+	restParts := strings.SplitN(rest, ".", 2)
+	if len(restParts) != 2 {
+		return "", fmt.Errorf("invalid advertised API URL: %s", ctrl.advertisedAPIURL)
+	}
+
+	instanceName := restParts[0]
+	rest = restParts[1]
+
+	// example: p-g3a4ana-demo.proxy-us.omni.siderolabs.io
+	url := protocol + "//" + workloadproxy.HostPrefix + "-" + alias + "-" + instanceName + "." + ctrl.workloadProxySubdomain + "." + rest
```
Unix4ever, line 278: Do we still use `p-` to route the requests?

utkuozdemir: Actually, good point, we don't need it anymore. The instance name is still good to ensure there are no collisions, but `p-` is not needed really. Should we just drop it?

Unix4ever: Could be cool to provide a way to also make alias static, to get even prettier URL :)

Unix4ever: But I guess we should definitely drop `p-` prefix

utkuozdemir: We had something like that as the initial idea back then, but there was a problem with potential collisions.

If it's ok, I'd leave that for another iteration to make sure we get it right if we do it again. There are some other topics we might want to address in a next iteration.

I'd say let's address the DNS propagation issue in this one. The changes we would add later would still use the new subdomain introduced here anyway.

But I'm gonna drop the `p-` - it's definitely not needed.

utkuozdemir: dropped `p-`, also updated the omni-controller.

internal/backend/workloadproxy/handler.go
```diff
@@ -218,3 +245,16 @@ func getMainDomain(url *url.URL) string {
 
 	return url.Host
 }
+
+// getWorkloadProxyDomain returns the full domain used by the workload proxy as the parent domain.
+//
+// Example: proxy-us.omni.siderolabs.io.
+func getWorkloadProxyDomain(subdomain string, mainDomain string) string {
+	parts := strings.SplitN(mainDomain, ".", 2)
```
DmitriyMV, line 253:
```suggestion
	left, right, ok := strings.Cut(mainDomain, ".")
```

frontend/src/api/resources.ts
```diff
@@ -21,13 +21,13 @@ export const authBearerHeaderPrefix = "Bearer ";
 export const SignatureVersionV1 = "siderov1";
 export const samlSessionHeader = "saml-session";
 export const DefaultKubernetesVersion = "1.30.1";
+export const workloadProxyHostPrefix = "p";
```
DmitriyMV, line 24: Do we need this const?

utkuozdemir: ah, I removed the `tsgen:` hint but forgot to re-generate afterwards.

Unix4ever (approved):

PR #369, opened 2024-06-12
fix: properly delete the item from the cached items slice

Unix4ever:
Should fix nil reference errors.

### DIFF
--- a/internal/backend/runtime/cosi/cosi.go
+++ b/internal/backend/runtime/cosi/cosi.go
@@ -171,7 +171,7 @@ func WatchLegacy(ctx context.Context, st state.State, md resource.Metadata, out 
 				}
 
 				if items != nil {
-					slices.Delete(items, index, index+1) //nolint:govet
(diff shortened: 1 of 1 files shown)


andrewrynhard (approved):

internal/backend/runtime/cosi/cosi.go
```diff
@@ -171,7 +171,7 @@ func WatchLegacy(ctx context.Context, st state.State, md resource.Metadata, out
 				}
 
 				if items != nil {
-					slices.Delete(items, index, index+1) //nolint:govet
+					items = slices.Delete(items, index, index+1)
```
utkuozdemir, line 174: Omg, there should be a linter catching this

utkuozdemir (approved):

internal/backend/runtime/cosi/cosi.go
Unix4ever: there's but I've disabled it 🫠

Issue #31, opened 2023-11-29
Support config patches on the machine class level


Unix4ever:
Priority wise:

- Cluster
-> MachineClass <-
- MachineSet
- ClusterMachine
- Machine

We introduce a new controller that copies config patch to the machine set.

garymm:
Would be useful for applying NVidia GPU support patches.

PR #374, opened 2024-06-13
feat: support machine patches on the machine class level

Unix4ever:
Now the `ConfigPatch` resources created with
`omni.sidero.dev/machine-class: <name>` label will be assigned to each machine which matches the picked machine class.

It will have higher priority than Machine and Cluster patches but will be overwritten by the ClusterMachine and MachineSet level patches.

Fixes: https://github.com/siderolabs/omni/issues/31


Unix4ever:
Screenshot 2024-06-13 at 18 32 57
Screenshot 2024-06-13 at 18 33 35

internal/backend/runtime/omni/controllers/omni/machine_class_patch.go
```diff
@@ -0,0 +1,236 @@
...
+			state.WithLabelQuery(resource.LabelEqual(omni.LabelMachineClass, ptr.ID())))
+		if err != nil {
+			return nil, err
+		}
+
+		return xslices.Map(patches.Items, func(r resource.Resource) resource.Pointer {
+			return r.Metadata()
+		}), nil
+	case omni.MachineStatusType:
+		machineStatus, err := safe.ReaderGetByID[*omni.MachineStatus](ctx, r, ptr.ID())
```
smira, line 97: my only concern is that it's computationally intensive - how are we going to scale with that?

on any MachineStatus change we basically try to re-apply all machine class config patches

Unix4ever: Yeah, I tried to limit it only by the machines which are added to a cluster, but still.

Unix4ever: We can do an optimization I guess.
Can extract `MachineStatus` labels into a separate resource, this will lower the count of wake ups of the controller.
And this resource can also be used in the `MachineSetNode` controller that allocates the machines.

smira:
We had a long discussion with Artem and identified two potential directions:

* optimize MachineClass by extracting MachineStatus labels to a separate resource to match on (reduce wakeups to only label changes)
* instead of full ConfigPatch propagation from the Machine Class on any change, copy MachineClass patches to the Machine as it joins the MachineSet from the MachineClass (less surpises on config patches changes, less computational complexity); also potentially provide a way to force patch sync

PR #375, opened 2024-06-13
chore: migrate to oidc v3

DmitriyMV:
Update to latest oidc implementation.


internal/backend/oidc/internal/storage/keys/storage.go
```diff
@@ -84,85 +69,79 @@ func (s *Storage) GetKeySet(context.Context) (*jose.JSONWebKeySet, error) {
...
+		return fmt.Errorf("failure to generate the key: %w", err)
+	}
 
-		keyID := uuid.NewString()
+	keyID := uuid.NewString()
 
-		if err = s.storeKey(ctx, keyID, privateKey); err != nil {
-			return fmt.Errorf("failure to store the key: %w", err)
-		}
+	ctx = actor.MarkContextAsInternalActor(ctx)
```
DmitriyMV, line 123: @Unix4ever @utkuozdemir I'm not sure about this part, but we never pass this context anywhere else. Without this the state below complains about "no access is permitted".

utkuozdemir: It is required for `storeKey` to work.

In the old (current) code, if you trace the ctx passed to `storeKey` all the way to where it originated from (yo go into the library and back out), you'll see that is it the main context marked as the internal actor already.

Probably something changed in the library where they don't propagate it all the way here anymore.

DmitriyMV: The OIDC library no longer starts its own gorotuines. Which means I had to redo logic a bit - now public and sign requests will check for expiration instead of relying on other gorotuine to set the required field in the Storage struct.

smira:
We need to test `kubectl` operations thoroughly, as I think we don't have any auto-tests for it.

internal/backend/oidc/internal/storage/keys/storage.go
DmitriyMV: @utkuozdemir @Unix4ever restored old logic where key rotation is independent of key requests.

DmitriyMV:
@smira 

```sh
~ > omnictl kubeconfig --cluster myk8scluster
Could not authenticate: key expired
Attempting to open URL: https://999.999.999.999/omni/authenticate?flow=cli&public-key-id=323e15e1e33474cfc9e91b67955c9fdbad819518
Please visit this page to authenticate: https://999.999.999.999/omni/authenticate?flow=cli&public-key-id=323e15e1e33474cfc9e91b67955c9fdbad819518
Public key 323e15e1e33474cfc9e91b67955c9fdbad819518 is now registered for user dmitry.matrenichev@siderolabs.com
PGP key saved to /home/dmitry/.local/share/omni/keys/local-https-dmitry.matrenichev@siderolabs.com.pgp

~ > omnictl kubeconfig --cluster myk8scluster --grant-type=authcode-keyboard
auth "default-myk8scluster-dmitry.matrenichev@siderolabs.com" already exists [(r)ename/(o)verwrite]: o
cluster "default-myk8scluster" already exists [(r)ename/(o)verwrite]: o

~ > kubectl run alpine --image alpine -it -- ash
Please visit the following URL in your browser: https://999.999.999.999/oidc/authorize?access_type=offline&client_id=native&code_challenge=kAFCwg2EQ1SLigPaW1V_fPEy9fGSfj6-prEfJ-jDKac&code_challenge_method=S256&nonce=8Zf6T9WVAotrIZk2MZfn9yWWviR4kZ1McyNNr31T-38&redirect_uri=urn%3Aietf%3Awg%3Aoauth%3A2.0%3Aoob&response_type=code&scope=cluster%3Amyk8scluster+openid&state=_wBno7DwbN9GfDRZd_zfFR3zq97avtcebtJ4vC_kZhs
Enter code: 12df1a
Warning: would violate PodSecurity "restricted:latest": allowPrivilegeEscalation != false (container "alpine" must set securityContext.allowPrivilegeEscalation=false), unrestricted capabilities (container "alpine" must set securityContext.capabilities.drop=["ALL"]), runAsNonRoot != true (pod or container "alpine" must set securityContext.runAsNonRoot=true), seccompProfile (pod or container "alpine" must set securityContext.seccompProfile.type to "RuntimeDefault" or "Localhost")
If you don't see a command prompt, try pressing enter.
/ #  ping 8.8.8.8
PING 8.8.8.8 (8.8.8.8): 56 data bytes
64 bytes from 8.8.8.8: seq=0 ttl=116 time=1.188 ms
64 bytes from 8.8.8.8: seq=1 ttl=116 time=0.750 ms
64 bytes from 8.8.8.8: seq=2 ttl=116 time=1.194 ms
64 bytes from 8.8.8.8: seq=3 ttl=116 time=1.165 ms
^C
--- 8.8.8.8 ping statistics ---
4 packets transmitted, 4 packets received, 0% packet loss
round-trip min/avg/max = 0.750/1.074/1.194 ms
/ #
```

What else I can test?

DmitriyMV:
> We need to test kubectl operations thoroughly, as I think we don't have any auto-tests for it.

Done. With both `omnictl kubeconfig --cluster <cluster_name>` and `omnictl kubeconfig --cluster <cluster_name> --grant-type=authcode-keyboard`. Found bugs along the way.

Unix4ever (approved):

utkuozdemir (approved):

DmitriyMV:
[e2e-short](https://github.com/siderolabs/omni/actions/runs/9602829777/job/26485329949?pr=375#logs) passes, so I also remove label here before rebasing.

PR #376, opened 2024-06-14
fix: skip invalid machines in `TalosUpgradeStatusController`

Unix4ever:
Do not exit the loop, but just ignore such machines. Otherwise we don't create `ClusterMachineTalosVersion` resources for all new machines in the machine set, which blocks
`ClusterMachineConfigController` and the machines are not configured.


utkuozdemir (approved):
Good find

DmitriyMV (approved):

smira (approved):

Issue #373, opened 2024-06-13
protect from panics in goroutines


smira:
Inspired by #369, follow-up for #372.

There are two major sources of goroutines launched by Omni which might not have panic protection: 

1. `errgroup.Go`
2. `go func()`

I think we need to ensure all these runs have a panic handling mechanism.

I propose to add the following helper function:

```go
func recoverPanic(f func() error, logger zap.Logger) (err error) {
   defer func() {
       if p := recover(); p != nil {
             ....
             err =  
             logger.
       }
     }()

     return f()
```

And the introduce sub-versions of this function - the one which accepts `func()` and returns no error (for `go func()`) case.

And then ensure that every invocation of a goroutine is wrapped accordingly

smira:
let's also make sure it's unit-tested :)

PR #389, opened 2024-06-18
feat: handle all goroutine panics gracefully

utkuozdemir:
Convert goroutine panics to errors or error logs.

Disallow usage of `golang.org/x/sync/errgroup` package in the backend by `depguard` linter. This linter configuration depends on: https://github.com/siderolabs/kres/pull/417

Rekres the project to include the feature (also bump Go to 1.22.4), but revert `PROTOBUF_GO_VERSION` and `GRPC_GATEWAY_VERSION` manually to not break the frontend.

Disallowing the named `go` statement was not possible at the moment using existing linters, raised an issue in `forbidigo` for it: https://github.com/ashanbrown/forbidigo/issues/47

Closes https://github.com/siderolabs/omni/issues/373.


DmitriyMV:
Hmm, I think we can do better:

1. Create our own `errgroup.Group` implementation which does what we want.
2. Forbid the usage of `errgroup` using https://golangci-lint.run/usage/linters/#depguard
3. Forbid the usage of `go` statement using https://golangci-lint.run/usage/linters/#forbidigo

utkuozdemir:
> Hmm, I think we can do better:
> 
> 1. Create our own `errgroup.Group` implementation which does what we want.
> 2. Forbid the usage of `errgroup` using [golangci-lint.run/usage/linters/#depguard](https://golangci-lint.run/usage/linters/#depguard)
> 3. Forbid the usage of `go` statement using [golangci-lint.run/usage/linters/#forbidigo](https://golangci-lint.run/usage/linters/#forbidigo)

ah this makes perfect sense, thanks.

utkuozdemir:
> Hmm, I think we can do better:
> 
> 1. Create our own `errgroup.Group` implementation which does what we want.
> 2. Forbid the usage of `errgroup` using [golangci-lint.run/usage/linters/#depguard](https://golangci-lint.run/usage/linters/#depguard)
> 3. Forbid the usage of `go` statement using [golangci-lint.run/usage/linters/#forbidigo](https://golangci-lint.run/usage/linters/#forbidigo)

done something along the lines of your suggestion and our discussion.

utkuozdemir:
> > Rekres the project to include the feature (also bump Go to 1.22.4), but revert PROTOBUF_GO_VERSION and GRPC_GATEWAY_VERSION manually to not break the frontend.
> 
> Does it still break the frontend, I thought it was fixed?

Actually, everything seemed to be working when I ran locally with newer versions and generated sources, but I saw some 500s for the watch requests in the browser developer console, so to be on the safe side I reverted manually anyway. So, not 100% sure. @DmitriyMV?

Unix4ever (approved):

DmitriyMV (approved):
I think we should proceed and look into those 500s if they become a problem.

utkuozdemir:
> I think we should proceed and look into those 500s if they become a problem.

Ok, but I'll still merge this PR as-is, and make rekres a separate PR to not block this with any potential issues.

PR #405, opened 2024-06-21
feature: added HelmChart

kevintijssen:


kevintijssen:
It this something that you guys want, do I need to do extra or should I close this PR? @TimJones @utkuozdemir

utkuozdemir:
> It this something that you guys want, do I need to do extra or should I close this PR? @TimJones @utkuozdemir

Hi Kevin, thank you very much for your effort.

We at the moment have multiple PRs for a Helm chart and have something internal as well. We are not sure yet which direction we want to go with the Helm chart and with which one.

So let's keep this on hold for now - when we can allocate some time to the Helm chart topic, we will get back to you on that.

Mershab99:
Amazing work @kevintijssen ! Love how you added Cert Manager integrations, etc. 

Just one note, as wireguard needs to know the IP of the LoadBalancer, how would we be able to know what service LoadBalancer's ExternalIP will be when we apply the helm chart? Perhaps we should create the service as part of the setup and use that svc IP in reference. Just an idea but it seems like the last piece missing to tie this all together!

smira:
Superseded by #152

PR #409, opened 2024-06-21
release(v0.38.1): prepare release

Unix4ever:
This is the official v0.38.1 release.


.conform.yaml
```diff
@@ -12,7 +12,7 @@ policies:
           gitHubOrganization: siderolabs
       spellcheck:
         locale: US
-      maximumOfOneCommit: true
```
utkuozdemir, line 15: this expected?

utkuozdemir (approved):

.conform.yaml
Unix4ever: yep, that's to simplify further backports if any

Issue #396, opened 2024-06-20
Unhealthy workload exposer nodes are not excluded from the load balancer


Unix4ever:
Omni should run health checks for all nodes and do not route request to the nodes which are unhealthy.
Otherwise one down node brings down all exposed services.

PR #410, opened 2024-06-24
feat: use tcp loadbalancer for exposed services

utkuozdemir:
Improve the exposed service reliability by using a TCP loadbalancer between the nodes exposing the service.

Rework the exposed service proxy registry to be a COSI controller instead to simplify the logic, improve reliability and testability.

Closes siderolabs/omni#396.


internal/backend/runtime/omni/controllers/omni/cluster_workload_proxy_status.go
```diff
@@ -0,0 +1,141 @@
...
+
+				return nil
+			},
+		},
+		qtransform.WithExtraMappedInput(
+			mappers.MapByClusterLabel[*omni.ExposedService, *omni.Cluster](),
+		),
+		qtransform.WithExtraMappedInput(
+			mappers.MapByClusterLabel[*omni.ClusterMachineStatus, *omni.Cluster](),
+		),
```
Unix4ever, line 137: I see you don't have tests for exposed service resource and cluster machine status deletion.
As you don't add finalizers to these two inputs the test might fail.

You can use `helpers.HandleInputs` for that, it both reads the resources and adds finalizers to them.

Unix4ever: Actually maybe `HandleInputs` won't fit here, but in any case there should be finalizer added to each input, which should be removed if either this additional input is tearing down or if cluster is tearing down.

internal/backend/workloadproxy/reconciler.go
```diff
@@ -0,0 +1,208 @@
...
+	}
+
+	return nil
+}
+
+// hostPortForAlias returns a unique IP:port for the given cluster and alias.
+//
+// The value is arbitrary, as it uses in-memory transport, never reached via the network.
+func (registry *Reconciler) hostPortForAlias(clusterID resource.ID, alias string) string {
+	return fmt.Sprintf("%s_%s:4242", clusterID, alias)
```
Unix4ever, line 207: What stops us from running lb on the exposed service port directly?
Why do we need memu layer?

utkuozdemir: I thought about that, but then we would need to allocate a distinct port or a loopback address for each and every exposed svc. I initially thought something like:
```
exposed svc 1 -> 127.0.0.1:40001
exposed svc 2 -> 127.0.0.1:40002
```
or
```
exposed svc 1 -> 127.0.40.1:40000
exposed svc 2 -> 127.0.40.2:40000
```

and so on.

But I didn't like that solution much, as when Omni receives the request for the exposed service, the data is already in-memory. It just needs to hit one of the healthy upstreams. It does not need to leave the process at all, just to come back to it to be forwarded to the upstream. So this felt like a more efficient and neater solution to me.

internal/backend/runtime/omni/controllers/omni/cluster_workload_proxy_status.go
utkuozdemir: Ah, good catch, I added a test that deletes a clustermachinestatus and an exposedservice, and you were right. I modified the controller code to add and remove finalizers to those resources, and now it seems to be working fine. Please have another look.

Unix4ever: your fix looks good except that it doesn't remove finalizers when the cluster enters tearing down phase.
But that probably works now as cluster resource is always removed after we clean up clustermachinestatuses and exposedservice resources for it.

Unix4ever (approved):

internal/backend/workloadproxy/reconciler.go
```diff
@@ -0,0 +1,209 @@
...
+		aliasToCluster:           map[string]resource.ID{},
+		connProvider:             provider,
+		logger:                   logger,
+		logLevel:                 logLevel,
+	}
+}
+
+// Reconcile reconciles the workload proxy load balancers for a cluster.
+func (registry *Reconciler) Reconcile(cluster resource.ID, aliasToUpstreamAddresses map[string][]string) error {
+	registry.logger.Log(registry.logLevel, "reconcile LBs", zap.String("cluster", cluster), zap.Any("aliasToUpstreamAddresses", aliasToUpstreamAddresses))
```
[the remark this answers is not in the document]
utkuozdemir: ah right. ensure and delete functions already log in themselves as the first thing, so I just removed the `zap.Any` from this (which was actually the copilot :)) and now I think it should be fine.

internal/backend/workloadproxy/reconciler.go
```diff
@@ -0,0 +1,209 @@
...
+	}
+
+	return nil
+}
+
+// ensureLB ensures that a load balancer exists and started for the given cluster and alias, targeting the given upstream addresses.
+func (registry *Reconciler) ensureLB(cluster resource.ID, alias string, upstreamAddresses []string) error {
+	registry.logger.Log(registry.logLevel, "ensure LB", zap.String("cluster", cluster), zap.String("alias", alias), zap.Strings("upstreamAddresses", upstreamAddresses))
+
+	lbSts := registry.clusterToAliasToLBStatus[cluster][alias]
```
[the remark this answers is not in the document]
utkuozdemir: done

internal/backend/workloadproxy/reconciler.go
```diff
@@ -0,0 +1,209 @@
...
+	}
+
+	// ensure new LBs
+	for alias, upstreamAddresses := range aliasToUpstreamAddresses {
+		if err := registry.ensureLB(cluster, alias, upstreamAddresses); err != nil {
+			errs = multierror.Append(errs, fmt.Errorf("failed to register load balancer: %w", err))
+		}
+	}
+
+	return nil
```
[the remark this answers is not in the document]
utkuozdemir: 🤦 yep, certainly :)

internal/backend/runtime/omni/controllers/omni/cluster_workload_proxy_status.go
utkuozdemir: Just to be on the safe side, I added the same finalizer removal logic into `FinalizerRemovalExtraOutputFunc` as well.

Issue #284, opened 2024-05-24
[feature] support --insecure talosctl commands over siderolink


rsmitty:
### Problem Description

We should figure out how we might allow this, because take a machine in maintenance mode. There's no good way to find, for example, the bus path of a disk I want to install to (or use as extraDisks) if I want to be explicit about the chosen disk with a config patch.

Issue #236, opened 2024-05-15
[feature] provide a mechanism for reinstalling talos at a given version from maintenance mode


rsmitty:
### Problem Description

This issue is mostly a stub for some much longer discussion. But we should brainstorm how we might offer the ability for a reinstall of talos at any given version, upgrade or downgrade, when the box is in maintenance mode. I'd mostly like to scope it there, as I don't think it makes sense to discuss in the context of the machine being part of a cluster. But if the machine has been de-allocated from a cluster, we should have some way to go back to an older version of Talos and thus be able to easily reuse the node in question without the need for a manual reset or pxe boot, etc.

Unix4ever:
Provide a page for downgrading the machines in maintenance mode:

- on Machines page display the button to change Talos version.
- we shouldn't display it for the machines running from ISO and PXE.
- detect some obvious cases when downgrade is not possible, like:
  - do not allow downgrading nodes with overlays from 1.7 to 1.6.

steverfrancis:
This shouldn't be something a user has to think about, so we don't want to give a page that makes them downgrade.

Ideally we just want to support a pool of Talos machines, of any version supported by Omni. (So 1.4 to 1.8 currently)

You can add any machine, running any version of Talos, to any cluster running any version of Talos,  or specify a cluster to be created with any version of Talos.

Omni will automatically downgrade or upgrade the Talos that is installed to the machine, so that it can join the cluster.
So if the machine is running Talos off ISO, and has not been installed to disk, then it needs to install to disk, then do the upgrade/downgrade as necessary to join the cluster.

Unix4ever:
> Omni will automatically downgrade or upgrade the Talos that is installed to the machine, so that it can join the cluster.
So if the machine is running Talos off ISO, and has not been installed to disk, then it needs to install to disk, then do the upgrade/downgrade as necessary to join the cluster.

I don't think installing ISO on _some_ disk to do an upgrade is a viable solution: you won't be able to change the disk afterwards and we can't pick the disk automatically.

But the biggest blocker here is that you have to apply the machine config to even start the upgrade and if the config has new fields which are not supported in the older Talos version, the machine will never boot.

Maintenance upgrade, ISO upgrade story will be addressed in Talos 1.8.
I can think about how we can address that in Omni a bit more, but that's more of a Talos API issue, we should fix it there first

PR #411, opened 2024-06-25
feat: support insecure access to the nodes running in maintenance

Unix4ever:
Any insecure `talosctl` commands now work with Omni per-instance `talosconfig`.
User should have at least `Operator` Omni role to be able to use the insecure `talosctl` mode.

DNS resolver was updated to react on the `MachineStatus` resource creation, not only the `ClusterMachineConfigStatus` resource. That makes the DNS record for UUID appear as soon as machine joins Omni, not when the machine gets allocated into a cluster.

Machines list now has maintenance Talos version update button. The UI will issue `talosctl upgrade` when another Talos version is picked.

`MachineStatus` controller was updated a bit: version poller wasn't marked as dirty after maintenance upgrades. Now we mark it as dirty every time we get Talos `MachineStatus` resource update.

Also fixed UI issues here and there.

Fixes: #284 
Fixes: #236


frontend/src/views/omni/ItemLabels/ItemLabels.vue
```diff
@@ -90,7 +87,15 @@ const labels = computed((): Array<Label> => {
   }
 
   labelsArray.sort((a, b) => {
-    return getLabelOrder(a) - getLabelOrder(b);
+    if (getLabelOrder(a) === getLabelOrder(b)) {
```
Unix4ever, line 90: Label order was broken.

frontend/src/views/omni/Overview/Overview.vue
```diff
@@ -278,7 +278,7 @@ url: "${talosLoggingKernel}"
 }
 
 .overview-copy-icon {
-  @apply cursor-pointer w-5 text-primary-P2 p-0.5 rounded;
+  @apply cursor-pointer w-5 text-primary-P2 p-0.5 rounded w-5 h-5;
```
Unix4ever, line 281: fixes icons gigantomania

Unix4ever:
https://github.com/siderolabs/omni/assets/830532/a6915a9f-aab6-495d-8a92-eb2e33b7e16d

frontend/src/methods/key.ts
```diff
@@ -24,7 +24,6 @@ import {
   workloadProxyPublicKeyIdCookie,
   workloadProxyPublicKeyIdSignatureBase64Cookie,
 } from '@/api/resources';
-// import { Buffer } from "buffer";
```
utkuozdemir, line 27: maybe remove?

Unix4ever: it was removed 🤔

utkuozdemir: oh sorry, I'm blind, it's red :D

internal/backend/grpc/router/router.go
```diff
@@ -154,16 +156,22 @@ func (r *Router) Director(ctx context.Context, fullMethodName string) (proxy.Mod
...
 	clusterName := getClusterName(md)
 
-	if backend, ok := r.talosBackends.Get(clusterName); ok {
+	id := clusterName
+
+	if clusterName == "" {
+		id = getNodeID(md)
+	}
+
+	if backend, ok := r.talosBackends.Get(id); ok {
```
Unix4ever, line 168: use struct as a key

Unix4ever: or add prefix to `cluster` and `machine` backends if we stick to strings.

Unix4ever:
verify insecure access in the integration tests

PR #426, opened 2024-06-29
fix: add resource caches for missing resource types

smira:
Some are recently added resources, some are resources we kept back on the initial cache set due to the concerns on the memory usage. Let's try to cache more and use all of them to see if it works better.


Unix4ever (approved):

smira:
Added even more resource types, and disabled `govulncheck` manually

PR #428, opened 2024-06-30
fix: add resource caches for missing resource types

Unix4ever:
Some are recently added resources, some are resources we kept back on the initial cache set due to the concerns on the memory usage. Let's try to cache more and use all of them to see if it works better.

Signed-off-by: Andrey Smirnov <andrey.smirnov@siderolabs.com>
(cherry picked from commit d60e819bc8a074d25f9ccaf01e53540dd4b09b6f)


smira:
I'll get to the computer today and double check the list, the backport should be good to go though

smira (approved):

PR #435, opened 2024-07-01
fix: do not re-create peer on the remote addr change

Unix4ever:
And exclude port from the saved address.

Additionally fix Talos backends cache to not to
react on the `MachineType` `Create` and `Update` events.

### DIFF
--- a/internal/backend/runtime/omni/controllers/omni/cluster_machine_status.go
+++ b/internal/backend/runtime/omni/controllers/omni/cluster_machine_status.go
@@ -188,8 +188,7 @@ func NewClusterMachineStatusController() *ClusterMachineStatusController {
 				case machineapi.MachineStatusEvent_RUNNING:
 					cmsVal.Stage = specs.ClusterMachineStatusSpec_RUNNING
 				case machineapi.MachineStatusEvent_MAINTENANCE:
-					// shouldn't happen as machine config should be submitted
-					cmsVal.Stage = specs.ClusterMachineStatusSpec_UNKNOWN
+					cmsVal.Stage = specs.ClusterMachineStatusSpec_CONFIGURING
 				}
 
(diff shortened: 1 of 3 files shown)


internal/backend/runtime/omni/controllers/omni/cluster_machine_status.go
```diff
@@ -188,8 +188,7 @@ func NewClusterMachineStatusController() *ClusterMachineStatusController {
 				case machineapi.MachineStatusEvent_RUNNING:
 					cmsVal.Stage = specs.ClusterMachineStatusSpec_RUNNING
 				case machineapi.MachineStatusEvent_MAINTENANCE:
-					// shouldn't happen as machine config should be submitted
-					cmsVal.Stage = specs.ClusterMachineStatusSpec_UNKNOWN
+					cmsVal.Stage = specs.ClusterMachineStatusSpec_CONFIGURING
```
Unix4ever, line 191: Now the machines won't be in Unknown state when they're added to the cluster and are still in maintenance.
I've noticed that Talos might have a slight delay before it reports `BOOTING` stage.

internal/pkg/siderolink/manager.go
```diff
@@ -751,6 +755,16 @@ func (manager *Manager) Provision(ctx context.Context, req *pb.ProvisionRequest)
 		dirty = true
 	}
 
+	if spec.RemoteAddr != host {
+		if _, err = safe.StateUpdateWithConflicts(ctx, manager.state, link.Metadata(), func(r *siderolink.Link) error {
+			r.TypedSpec().Value.RemoteAddr = host
```
smira, line 760: dumb question - why do even collect that?

Unix4ever: That's a long story, but it might help in case when you have a machine joining your account and Omni can't get access to it. And you can't even tell which node is that.

smira (approved):

Issue #75, opened 2024-03-15
Omni Support bundle download button in the UI


Unix4ever:
Same as `omnictl support` but in the UI.
Can be placed to the cluster overview page.
When clicked should open the dialog with 2 buttons: download and cancel.

If download is clicked it should start loading the bundle showing the progress bars for each bundle generation step.
Errors should be displayed as the list in the end of the download.
When download completes the server will send the final bundle as zip archive bytes.
The UI should simulate file download same way as we do it in the kubeconfig download flow.

PR #436, opened 2024-07-01
feat: add support bundle download button to cluster overview

utkuozdemir:
Allow downloading support bundles from the UI.
Part of siderolabs/omni#75.


utkuozdemir:
@Unix4ever not sure if this will be useful at all, but handing this over to you as it is beyond my FE/design skills.

Unix4ever:
Polished the UI, fixed the bug related to the download not being canceled after you close the modal.

Final UI:

https://github.com/user-attachments/assets/1f601a5b-d924-41b2-8593-d6b1b5a948ec

smira (approved):

frezbo:
> Polished the UI, fixed the bug related to the download not being canceled after you close the modal.
> 
> Final UI:
> Screen.Recording.2024-07-24.at.19.52.00.mov

Cool, can we use nodenames instead of siderolink addresses?

Unix4ever:
> Cool, can we use nodenames instead of siderolink addresses?

Well, that's actually node internal IP, it's just how Talos emulator shows them.
It's how Talos support bundle collector works, so we might need to map it in Omni somehow.

PR #438, opened 2024-07-02
fix: use proper permissions for cluster taint resource

Unix4ever:
This should fix `omnictl kubeconfig/talosconfig --break-glass` command.

### DIFF
--- a/internal/backend/grpc/management.go
+++ b/internal/backend/grpc/management.go
@@ -274,6 +274,8 @@ func (s *managementServer) ValidateConfig(ctx context.Context, request *manageme
 }
 
 func (s *managementServer) markClusterAsTainted(ctx context.Context, name string) error {
+	ctx = actor.MarkContextAsInternalActor(ctx)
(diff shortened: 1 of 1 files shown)


internal/backend/runtime/omni/state_access.go
```diff
@@ -482,6 +482,14 @@ func filterAccessByType(access state.Access) error {
 		}
 
 		return status.Error(codes.PermissionDenied, "only read, update and delete access is permitted")
+	case omni.ClusterTaintType:
+		// Allow read, update and delete
+		// Update access is required for siderolink by rtestutils.Destroy[*siderolink.Link] call on integration tests
```
smira, line 487: comment is outdated?

smira (approved):

internal/backend/runtime/omni/state_access.go
Unix4ever: yeah

PR #443, opened 2024-07-03
fix: get rid of config patches for the maintenance configs

Unix4ever:
It's unreliable as the config patches may be created with some delay after the machine is added to Omni and if it gets immediatelly allocated into a cluster it will lose Omni connection params.

Always bake in the connection params in the `cluster_machine_config` controller instead.


cmd/integration-test/root.go
```diff
@@ -31,7 +31,7 @@ var rootCmd = &cobra.Command{
 	RunE: func(*cobra.Command, []string) error {
 		return withContext(func(ctx context.Context) error {
 			// hacky hack
-			os.Args = append(os.Args[0:1], "-test.v")
+			os.Args = append(os.Args[0:1], "-test.v", "-test.parallel", "4")
```
Unix4ever, line 34: an attempt to improve CI tests stability

utkuozdemir (approved):

PR #445, opened 2024-07-04
fix: provide cached access to the state via Omni API

smira:
This affects Get/List operations, but not Watch and Create/Update/Destroy.


smira:
Based on https://github.com/cosi-project/runtime/pull/470

utkuozdemir (approved):

PR #449, opened 2024-07-08
feat: add cloud provider resources and role

utkuozdemir:
Add the new resources for the cloud provider feature: `CloudConfigs`, `MachineRequests` and `MachineRequestStatuses`.

Add a new role, `CloudProvider` with access to the resources a cloud provider plugin requires.

Introduce the concept of "cloud provider service accounts" which are a special type of service accounts in the format `cloud-provider:<id>`. They must have the `CloudProvider` role and their id is matched against the label `omni.sidero.dev/cloud-provider-id` label on the `MachineRequest*` type resources.


internal/backend/runtime/omni/cloudprovider/state.go
```diff
@@ -0,0 +1,353 @@
...
+					}
+				}
+
+				channel.SendWithContext(ctx, eventCh, event)
+			}
+		}
+	}, st.logger)
+
+	return innerEventCh
+}
```
utkuozdemir, line 296: @smira I am not sure if I got this filtering logic right, pls have a look :)

internal/backend/grpc/serviceaccount.go
```diff
@@ -10,22 +10,317 @@ import (
...
+		if !isSa {
+			return nil, fmt.Errorf("unexpected service account ID %q", identity.Metadata().ID())
+		}
+
+		name := sa.NameWithPrefix()
+
+		serviceAccounts = append(serviceAccounts, &management.ListServiceAccountsResponse_ServiceAccount{
+			Name:          name,
+			PgpPublicKeys: publicKeys,
+			Role:          user.TypedSpec().Value.GetRole(),
```
Unix4ever, line 255: maybe lets add IsCloudProvider flag to this response

internal/backend/runtime/omni/state_access.go
```diff
@@ -484,6 +495,8 @@ func filterAccessByType(access state.Access) error {
 
 		return status.Error(codes.PermissionDenied, "only read, update and delete access is permitted")
 	case
+		cloud.MachineRequestType,       // read-only for all except for CloudProvider role (checked in filterAccess)
```
Unix4ever, line 498: I guess this comment is valid for the machine request status resource, but it should be possible to manage MachineRequests from Omni, isn't it?
Also, do we apply access policy rules for the MachineRequests?

utkuozdemir: Omni will not be subject to this, as it will be an internal actor - will simply bypass this. (here I assume that `MachineRequest` is not a user-managed resource, therefore users will only be able to read it)

utkuozdemir: About AccessPolicies - didn't do anything related to them, as we don't have the Cluster<>MachineRequest relation yet.

I could actually already add that logic based on a (future) cluster label. But I am not sure if they would carry cluster label - maybe they'll only carry machine class label, don't know yet.

internal/backend/grpc/serviceaccount.go
utkuozdemir: I can, but I think it would actually be redundant to this `Role` field, as only cloud providers can have `CloudProvider` role (it is enforced by validations). So the client can simply check for that instead, no?

Unix4ever: ah, right.

internal/backend/runtime/omni/state_access.go
Unix4ever: gotcha :heavy_check_mark:

Unix4ever (approved):

Issue #37, opened 2023-05-12
Omni audit logs


steverfrancis:
Omni needs to have an audit log accessible in the UI for any action that changes configurations, whether by omnictl or the UI.

I'd suggest different labels for actions affecting Omni (user additions/deletions/role changes etc), Clusters and Machines.
For machines, things like user initiated reboots, or deletions, as well as changes to patches and machine configs, need to be logged.

## Update 2024-08-12 Implementation status and tasks

```[tasklist]
## Remaining steps
- [ ] https://github.com/siderolabs/omni/issues/545
- [ ] https://github.com/siderolabs/omni/issues/546
- [ ] https://github.com/siderolabs/omni/issues/578
- [ ] https://github.com/siderolabs/omni/issues/586
- [ ] https://github.com/siderolabs/omni/issues/593
- [x] Deploy to production
```

```[tasklist]
## Future ideas
- [ ] Allow user to specify the retention policy
- [ ] Allow user to speciby the `begin` and `end` dates for getting logger
- [ ] Add support for MachineClass resources.
- [ ] Resource deletion should also print associated labels.
- [ ] Log when user downloads `talosconfig`.
- [ ] Log when user downloads `kubeconfig`.
- [ ] Maybe drop `GET` requests from logging k8s access.
```

DmitriyMV:
@smira also added `TestLogFile_CurrentDayConcurrent` test to be sure.

PR #461, opened 2024-07-15
chore: rework auth.* keys, add `ctxstore` package

DmitriyMV:
Using so-called phantom types we can use the types themselves as keys directly without loosing performance. You no longer need to remember which type was attached to the thing you passed in context and can look up all fields access directly.

Part of #37

### DIFF
--- a/internal/pkg/ctxstore/ctxstore.go
+++ b/internal/pkg/ctxstore/ctxstore.go
@@ -1,0 +1,35 @@
+// Copyright (c) 2024 Sidero Labs, Inc.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+
+// Package ctxstore provides a way to store values in the context with the key based on the type.
+package ctxstore
+
+import "context"
+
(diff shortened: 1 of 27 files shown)


internal/pkg/ctxstore/ctxstore.go
```diff
@@ -0,0 +1,35 @@
...
+// phantomKey represent key based on type. The cool thing about this empty struct,
+// is that two instances of phantomKey with the different type are different, while
+// two instances of phantomKey with the same type are the same. This is useful for
+// creating a unique key for each type. It also helps to avoid collision with other
+// keys in the context.
+//
+// It also does not allocate when used as a key in the context (aka converted to any).
+// Same goes for int and struct containing single int field with value below 256.
+// Same goes for bool and struct containing single bool field.
+type phantomKey[T any] struct{}
```
utkuozdemir, line 20: 🆒

smira (approved):

utkuozdemir (approved):

Issue #270, opened 2024-05-22
automatic machine provisioning


smira:
## Provisioning

1. Omni creates a `MachineRequest` to provision a `Machine` in some cloud with some settings (region, size, ...)
2. Omni Cloud Provider (a plugin to Omni) picks up the `MachineRequest`, and provisions the machine in the cloud.
3. Omni Cloud Provider creates `MachineRequestStatus` describing the future machine which will join Omni (e.g. it has the UUID of the Machine that is going to be created/going to join Omni).
4. As a provisioned `Machine` joins Omni, provider matches it with the request and labels/annotates it as requested (for example, label indicates that this Machine should be part of a `MachineClass` and it will join the `MachineSet`).

## De-provisioning

1. Machine is removed in the cloud --> Omni Cloud Provider picks it up and starts removing `Machine` from Omni.
2. MachineRequest is removed -> Cloud Provider removes the `Machine`, removes machine in the cloud, cleans up everything.
3. `Machine` is removed from Omni (e.g. faulty hardware) -> Omni Cloud Provider cleans it in the cloud.

smira:
## Omni Cloud Provider Deployment

1. Might run with Omni (customer cloud access keys are given to Omni)
2. Might run outside of Omni (auths to Omni using a service account, access keys are within external deployment)

smira:
# Omni Cloud Provider 

1. Connects to Omni API (some specific role, like 'Cloud Provider').
2. Fetches cloud provider config resource from Omni (or uses config supplied via the environment).
3. Watches `MachineRequest` resources with specific label (targeting this provider ID).
4. If this is a new `MachineRequest` that needs to be satisfied:
   * puts a finalizer on the `MachineRequest`
   * creates a `MachineRequestStatus` in not ready state, uses that to store internal progress status
   * starts provisioning a machine in the cloud, it might update `MachineRequestStatus` with some data to restart operations on failure
   * once the machine is ready (booting), update `MachineRequestStatus` to be ready and put the UUID of the machine which is going to join Omni (or labels the machine in a unique way); Omni should be able to match a joining `Machine` back to the `MachineRequest`
5. If this is a tearing down `MachineRequest`:
  * destroy the infrastructure
  * destroy the `MachineRequestStatus` resource
  * remove the finalizer on `MachineRequest`

smira:
## `MachineRequest` Resource

```yaml
metadata:
   id: machine-xyz
spec:
```

smira:
[...]

PR #468, opened 2024-07-18
feat: compute machineclass machine requirement (pressure)

utkuozdemir:
Introduce the resource `MachinesetRequiredMachines` containing the information about how many more machines are requested from a machine set using a machine class.

Introduce the resource `MachineClassStatus`, containing the information of the total amount of required additional machines (i.e., machine pressure) on a machine set.

The pressure is only created if a machine set is using a machine class with static allocation type (not unlimited).

Related to siderolabs/omni#270.


internal/backend/runtime/omni/controllers/omni/machine_set_node.go
```diff
@@ -175,6 +188,28 @@ func (ctrl *MachineSetNodeController) Run(ctx context.Context, r controller.Runt
 	}
 }
 
+// handleMachinePress.
```
smira, line 191: the comment looks a bit odd

internal/backend/runtime/omni/controllers/omni/machine_set_node.go
```diff
@@ -184,12 +219,14 @@ func (ctrl *MachineSetNodeController) reconcileMachineSet(
...
 	visited map[resource.ID]struct{},
 	logger *zap.Logger,
-) error {
-	var err error
-
+) (requiredAdditionalMachines int, err error) {
 	spec := machineSet.TypedSpec()
 	if spec.Value.MachineClass == nil || machineSet.Metadata().Phase() == resource.PhaseTearingDown {
-		return nil
+		if err = r.Destroy(ctx, omni.NewMachineSetRequiredMachines(resources.DefaultNamespace, machineSet.Metadata().ID()).Metadata()); err != nil && !state.IsNotFoundError(err) {
```
smira, line 225: Artem knows better, idk if we need some extra finalizers to handle this?

smira (approved):

internal/backend/runtime/omni/controllers/omni/machine_set_node.go
Unix4ever: This controller is of the old type, so I guess we don't need finalizers here.

But it may be necessary to handle `Teardown` + `Destroy` flow if `MachineSetRequiredMachines` resource gets some finalizers.

internal/backend/runtime/omni/controllers/omni/machine_set_node.go
```diff
@@ -175,6 +188,28 @@ func (ctrl *MachineSetNodeController) Run(ctx context.Context, r controller.Runt
...
+
+				res.Metadata().Labels().Set(omni.LabelMachineClassName, machineSet.TypedSpec().Value.MachineClass.Name)
+
+				res.TypedSpec().Value.RequiredAdditionalMachines = uint32(numRequired)
+
+				return nil
+			})
+	}
+
+	if err := r.Destroy(ctx, omni.NewMachineSetRequiredMachines(resources.DefaultNamespace, machineSet.Metadata().ID()).Metadata()); err != nil && !state.IsNotFoundError(err) {
```
Unix4ever, line 206: As we destroy machine set required machines if the number == 0 we need to add the finalizer on resource in the MachineClassStatus controller.
Then we'll need to run the Teardown + Destroy flow.

I think maybe we can avoid all that if we just set the number of required machines to 0 in the resource and do not remove it.

internal/backend/runtime/omni/controllers/omni/machine_class_status_test.go
```diff
@@ -0,0 +1,92 @@
...
+
+		return nil
+	})
+	suite.Require().NoError(err)
+
+	assertResource(&suite.OmniSuite, mcs.Metadata(), func(r *omni.MachineClassStatus, assertion *assert.Assertions) {
+		assertion.Equal(uint32(1), r.TypedSpec().Value.RequiredAdditionalMachines)
+	})
+
+	rtestutils.Destroy[*omni.MachineSetRequiredMachines](suite.ctx, suite.T(), suite.state, []string{msrm1.Metadata().ID()})
```
Unix4ever, line 77: I'd rather suggest to destroy `omni.MachineSetRequiredMachines` with non 0 value.
It will expose the issues with the finalizers you have.

internal/backend/runtime/omni/controllers/omni/machine_set_node.go
utkuozdemir: 👍  I applied the change you suggested in this controller.

But ofc, still, `MachineSetController` (the `cleanup` controller) removes them when the corresponding `MachineSet` is removed.

internal/backend/runtime/omni/controllers/omni/machine_class_status_test.go
utkuozdemir: But this already destroys the one with non-zero value (with value `1`), and then asserts that the total number is reduced by one on the `MachineClassStatus`. (pls note that I applied your other suggestion with not removing. But this was passing just fine even before that)

I don't see how it is a problem, as the moment `MachineSetRequiredMachines` is being destroyed, it gets an update first (because of the `Teardown` call), this controller wakes up and recalculates for the matching machine set. And since it skips the `TearingDown` ones while doing that, it computes the correct number.

internal/backend/runtime/omni/controllers/omni/machine_set_node.go
utkuozdemir: Removed `Destroy` from this controller as Artem's suggestion, so no more applicable.

internal/backend/runtime/omni/controllers/omni/machine_class_status_test.go
Unix4ever: Yeah, I see now. There are actually 3 resources. All good then

Unix4ever (approved):

PR #472, opened 2024-07-19
fix: make image pre pull failure block the kubernetes update

Unix4ever:
If the wrong version of the kubernetes components is used it won't bring down any pods as the upgrade will stop on the pre-pull phase.

### DIFF
--- a/internal/backend/runtime/omni/controllers/omni/internal/task/image/pulltask.go
+++ b/internal/backend/runtime/omni/controllers/omni/internal/task/image/pulltask.go
@@ -77,25 +77,31 @@ func (p PullTaskSpec) RunTask(ctx context.Context, _ *zap.Logger, pullStatusCh P
 		currentNode = imageList.Node
 
 		for _, img := range imageList.Images {
+			currentError = nil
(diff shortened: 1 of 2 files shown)


internal/backend/runtime/omni/controllers/omni/internal/task/image/pulltask.go
```diff
@@ -77,25 +77,31 @@ func (p PullTaskSpec) RunTask(ctx context.Context, _ *zap.Logger, pullStatusCh P
...
-					return errs
-				}
+
+			if !nodeImageInfoMap.shouldPull(imageList.Node, img) {
+				currentNum++
+
+				continue
+			}
+
+			currentError = p.pullImage(ctx, clusterID, imageList.Node, img)
```
smira, line 89: I wonder - does it retry on failure? or it simply fails and dies?

Unix4ever: retries, it stops only if you return a nil error

smira (approved):

PR #473, opened 2024-07-22
chore: add rotating log for audit data

DmitriyMV:
Adds rotating audit log writer. Also minor improvements.

For #37


internal/backend/server.go
```diff
@@ -352,13 +352,15 @@ func (s *Server) buildServerOptions() ([]grpc.ServerOption, error) {
...
 
-	unaryAuthInterceptors, streamAuthInterceptors, err := s.getAuthInterceptors()
+	authIntereceptors, err := s.getAuthInterceptors()
 	if err != nil {
 		return nil, err
 	}
 
-	unaryInterceptors = append(unaryInterceptors, unaryAuthInterceptors...)
-	streamInterceptors = append(streamInterceptors, streamAuthInterceptors...)
+	for _, authIntereceptor := range authIntereceptors {
```
Unix4ever, line 360:
```suggestion
	for _, authInterceptor := range authInterceptors {
```

internal/backend/runtime/omni/audit/log_file.go
```diff
@@ -0,0 +1,87 @@
...
+func NewLogFile(dir string) *LogFile {
+	return &LogFile{dir: dir}
+}
+
+// Write writes data to the log file, creating new one on demand.
+func (l *LogFile) Write(data any) error {
+	return l.writeAt(data, time.Now())
+}
+
+func (l *LogFile) writeAt(data any, at time.Time) error {
```
smira, line 40: I would rename this to something else to avoid confusion with `io.WriterAt`

internal/backend/runtime/omni/audit/log_file.go
```diff
@@ -0,0 +1,87 @@
...
+// the last write date. Otherwise, it opens a new file.
+func (l *LogFile) openFile(at time.Time) (*os.File, error) {
+	l.mu.Lock()
+	defer l.mu.Unlock()
+
+	// TODO(Dmitriy): decide what to do with when we write entry for the previous day
+	// into the current day (happens on day's border). One solution is to have two decriptors opened:
+	// one for the current day and one for the previous day. We can close the previous day's descriptor
+	// when we open the 3rd day's descriptor. This way we will have at most two descriptors opened.
+	if l.f != nil && ordered.MakeTriple(at.Date()).Compare(ordered.MakeTriple(l.lastWrite.Date())) <= 0 {
```
smira, line 67: idk, but it feels it might be easier to run a goroutine which "flips" the file at midnight, and all writes go to a single currently opn file, so they are at least chronologically ordered?

internal/backend/runtime/omni/audit/log_file.go
```diff
@@ -0,0 +1,87 @@
...
+	return l.writeAt(data, time.Now())
+}
+
+func (l *LogFile) writeAt(data any, at time.Time) error {
+	f, err := l.openFile(at)
+	if err != nil {
+		return err
+	}
+
+	if err := json.NewEncoder(f).Encode(data); err != nil {
```
smira, line 46: this won't work well - we use direct file (without buffering), and json encoder would issue tons of small writes directly to the OS.

plus this won't be atomic, as append mode is only guaranteed for a single write (records will be interleaved)

in this case I would say it's easier to to write to the `bytes.Buffer` and then write bytes as a single write to the file.

internal/backend/server.go
DmitriyMV: Thx!

DmitriyMV:
So on further thinking I reworked `writeAt` (now `dumpAt` because of the similarity to `spew.Dump` and other pretty printers/dumpers) to take the lock right after we are finished with encoding. I looked at other solutions (such as creating locked wrapper around the file and implement synced `WriteTo`), but overall I don't think it worth it. I also moved the time assertion inside the locked part, so now all writers are fully ordered.

internal/backend/runtime/omni/audit/log_file.go
DmitriyMV: Yeah, I agree.

internal/backend/runtime/omni/audit/log_file.go
DmitriyMV: I reworked this entirely, more explanation below.

internal/backend/runtime/omni/audit/log_file.go
DmitriyMV: Renamed to `dumpAt` and `Write` -> `Dump`.

DmitriyMV:
@smira I also added `TestLogFile_CurrentDayConcurrent` test to be sure.

internal/backend/runtime/omni/audit/log_file.go
```diff
@@ -0,0 +1,107 @@
...
+	if l.f != nil {
+		err := l.f.Close()
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	logPath := filepath.Join(l.dir, at.Format("2006-01-02")) + ".jsonlog"
+
+	f, err := os.OpenFile(logPath, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0o644)
```
smira, line 99: not exactly clear how this could would recover from a failure.

I would say `.Close()` error should be ignored (as there's no recovery path, and it doesn't actually fail).

While we should zero `l.f` to make the code jump into creating a new file.

Creating new file might fail vs. closing it

internal/backend/runtime/omni/audit/log_file.go
```diff
@@ -0,0 +1,107 @@
...
+	if at.IsZero() {
+		at = time.Now()
+	}
+
+	f, err := l.openFile(at)
+	if err != nil {
+		return err
+	}
+
+	_, err = io.Copy(f, b)
```
smira, line 73: if we just `Write(b.Bytes())` that would be atomic, we don't need to lock the writes.

DmitriyMV: Yeah, but that will break `lastWrite` field access, and since `*os.File.Write` uses mutex internally I don't think it makes a lot of sense to unlock and then atomically write `lastWrite` field.

internal/backend/runtime/omni/audit/log_file.go
DmitriyMV: > not exactly clear how this could would recover from a failure.

I'm not sure it can without alerting the user. We can create a temporary ring buffer if we failed to create a file, but then what? Try again at the next write (and the next after that and so on) and write all stuff which we stored if we succeed?

My current idea is to write metrics/logs if the error occured.

> While we should zero l.f to make the code jump into creating a new file.

Yeah, agreed.

internal/backend/runtime/omni/audit/log_file.go
DmitriyMV: Replaced `io.Copy` with `b.WriteTo(f)`

internal/backend/runtime/omni/audit/log_file.go
smira: if we failed to create the file, the audit log will have missing entries, but once the error condition is clear, it should continue to log audit entries

smira (approved):

Unix4ever (approved):

PR #478, opened 2024-07-22
fix: properly update `ClusterMachineIdentity` resource

Unix4ever:
The code had the bug which was leading to not removing the old labels from the resource and not resetting the etcd member.


frezbo (approved):
no tests?

smira (approved):

Unix4ever:
> no tests?

It's a bit PITA to implement as this controller needs to talk to Talos API.

PR #497, opened 2024-07-29
test: run prometheus in tests and check metrics after talemu tests

Unix4ever:
Make the tests fail if the metrics do not meet the expected thresholds.

### DIFF
--- a/hack/compose/docker-compose.yml
+++ b/hack/compose/docker-compose.yml
@@ -102,6 +102,16 @@ services:
       MINIO_SECRET_KEY: secret123
     command: server /minio-server/export
 
+  prometheus:
+    network_mode: host
+    image: prom/prometheus
+    depends_on:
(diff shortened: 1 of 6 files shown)


hack/compose/docker-compose.yml
```diff
@@ -102,6 +102,16 @@ services:
...
 
+  prometheus:
+    network_mode: host
+    image: prom/prometheus
+    depends_on:
+      - omni
+    ports:
+      - "9090:9090"
+    volumes:
+      - ./prometheus/prometheus.yml:/etc/prometheus/prometheus.yml
```
frezbo, line 113: file not included in PR?

Unix4ever: yep and that's not the only file I forgot :) Added both

frezbo (approved):

PR #520, opened 2024-08-07
add a stringArray example for extensions

brantgurga:
This addresses ticket siderolabs/omni-docs#91


Unix4ever:
Thanks for the contribution!
You probably need to do `git commit --amend -s`, and change the commit message to be `feat: add a stringArray example for extensions`.
After you re-push it, I'll sign the commit and merge it.

PR #533, opened 2024-08-08
fix: workload service reconciler

smira:
Most of the code is simplifying/refactoring, but there are few fixes:

* increase LB upstream healthcheck interval to 1 minute
* pass a logger to the LB (as otherwise it creates its own)
* shutdown the LB by waiting for it to shutdown
* close the LB even when it fails to start to avoid leaking health check goroutines

Additionally, add an integration test for workload proxying.


internal/backend/workloadproxy/reconciler.go
```diff
@@ -25,8 +26,27 @@ import (
 )
 
 type lbStatus struct {
-	lb                *loadbalancer.TCP
-	upstreamAddresses []string
+	lb *loadbalancer.TCP
+}
+
+func (lbSts *lbStatus) Close() error {
+	var lb *loadbalancer.TCP
```
frezbo, line 33: do we need a nil check for `lbStatus`>?

frezbo (approved):

internal/backend/workloadproxy/reconciler.go
smira: I think it can't be nil (as we check before we call this func)

internal/backend/workloadproxy/reconciler.go
```diff
@@ -94,60 +110,50 @@ func (registry *Reconciler) Reconcile(cluster resource.ID, aliasToUpstreamAddres
...
+func (registry *Reconciler) ensureLB(cluster resource.ID, alias string, upstreamAddresses []string) error {
 	registry.logger.Log(registry.logLevel, "ensure LB", zap.String("cluster", cluster), zap.String("alias", alias), zap.Strings("upstreamAddresses", upstreamAddresses))
 
-	var existingUpstreamAddresses []string
+	hostPort := registry.hostPortForAlias(cluster, alias)
+	lbSts := registry.clusterToAliasToLBStatus[cluster][alias]
+
+	if lbSts == nil { // no LB yet, create and start it
+		tcpLB := &loadbalancer.TCP{
+			Logger: registry.logger.WithOptions(zap.IncreaseLevel(zapcore.ErrorLevel)),
```
smira, line 121: should create this one and re-use for all loadbalancers

smira:
Utku was kind enough to get this tested once he's back.

utkuozdemir (approved):

internal/backend/workloadproxy/reconciler.go
utkuozdemir: done

cmd/integration-test/pkg/tests/workload_proxy.go
```diff
@@ -0,0 +1,290 @@
...
+func AssertWorkloadProxy(testCtx context.Context, apiClient *client.Client, clusterName string) TestFunc {
+	return func(t *testing.T) {
+		ctx, cancel := context.WithTimeout(testCtx, 150*time.Second)
+		t.Cleanup(cancel)
+
+		st := apiClient.Omni().State()
+		kubeClient := getKubernetesClient(ctx, t, apiClient.Management(), clusterName)
+		port := 12345
+		label := "test-nginx"
+		icon := "H4sICKNuvmYAA1NpZGVyb19tYXJrX29uZWNvbG9yX1JHQi5zdmcAfVRLaxtBDP4ry/aqkUead+oYUl/bU6CHXsqycR7gPIhNnFD63/tpvYSmpMF4RszI0vfQeLl7uuqeb7d3u9P+er9/OFksDocDHwLfP14t1Hu/QEbfPd1sDl/un0973/lO" +
```
smira, line 47: not a big deal, but we could probably embed the icon itself via `go:embed` and base64/gzip it in the tests

cmd/integration-test/pkg/tests/workload_proxy.go
```diff
@@ -0,0 +1,290 @@
...
+				return http.ErrUseLastResponse // disable follow redirects
+			},
+			Timeout: 5 * time.Second,
+		}
+
+		t.Logf("do a GET request to the exposed service URL, expect a redirect to authentication page")
+
+		doRequestAssertResponseWithRetries(ctx, t, httpClient, svcURL, http.StatusSeeOther, "", "")
+
+		t.Logf("do the same request but with proper cookies set, expect a successful response from nginx")
```
smira, line 109:
```suggestion
		t.Logf("do the same request but with proper cookies set, expect a successful response from Omni")
```

?

smira (commented):
:cool: :+1:  (I can't self-approve)

PR #536, opened 2024-08-09
feat: implement `MachineRequestSets` and support links cleanup flow

Unix4ever:
The `MachineRequestSet` creates a number of `MachineRequests`. The machine request count is defined in the `MachineRequestSet` along with schematic id, Talos version, provider data and provider id to use.

Also introduce `MachineRequestCleanup` controller that can initiate the machine set node removal if the machine is added to a cluster and after the cluster machine is gone it should tear down the link.

`MachineSetNodeController` now also reacts on the `MachineRequestStatus` teardown: it will remove the related `MachineSetNode`.

Fixes: https://github.com/siderolabs/omni/issues/501


client/api/omni/specs/omni.proto
```diff
@@ -1103,3 +1103,17 @@ message KubernetesNodeAuditResultSpec {
...
 }
+
+// MachineRequestSetSpec creates the number of machine requests.
+// Can be either connected to a machine class or be manual.
+message MachineRequestSetSpec {
+  string provider_id = 1;
+  int32 machine_count = 2;
+
+  string talos_version = 3;
+  string schematic_id = 4;
```
Unix4ever, line 1114: I'm not sure about using raw schematic here though.
Maybe we should use the list of extensions, kernel args and meta args here and in the `MachineRequest` resource.

Unix4ever:
I will add the integration tests in the followup PR. Need to update talemu first.

client/pkg/omni/resources/omni/machine_request_set.go
```diff
@@ -0,0 +1,62 @@
...
+		protobuf.NewResourceSpec(&specs.MachineRequestSetSpec{}),
+	)
+}
+
+// MachineRequestSetType is the type of MachineRequestSet resource.
+//
+// tsgen:MachineRequestSetType
+const MachineRequestSetType = resource.Type("MachineRequestSets.omni.sidero.dev")
+
+// MachineRequestSet resource describes connected Talos server (bare-metal, VM, RPi, etc).
```
utkuozdemir, line 30: copy-paste artifact i believe

client/pkg/omni/resources/omni/machine_request_set.go
```diff
@@ -0,0 +1,62 @@
...
+				JSONPath: "{.providerid}",
+			},
+			{
+				Name:     "MachineCount",
+				JSONPath: "{.machinecount}",
+			},
+			{
+				Name:     "MachineClass",
+				JSONPath: "{.machineclass}",
+			},
```
utkuozdemir, line 59: this is no more a field, right?

client/pkg/omni/resources/omni/machine_request_set.go
Unix4ever: yep

client/pkg/omni/resources/omni/machine_request_set.go
Unix4ever: yep

utkuozdemir (approved):

Issue #527, opened 2024-08-08
extract ExposedService creation from the KubernetesStatus controller


Unix4ever:
Dump services status into `KubernetesStatus` then consume that in another controller that will create `ExposedService` resources.
This should make the code more testable.

utkuozdemir:
Not moved into a separate controller but into a separate reconciler type with its unit tests, which cleaned up the `KubernetesStatusController` quite a bit: https://github.com/siderolabs/omni/blob/main/internal/backend/runtime/omni/controllers/omni/internal/exposedservice/reconciler.go

PR #553, opened 2024-08-14
fix: properly cleanup tearing down exposed services

utkuozdemir:
The controller was missing to remove some `ExposedService`s, because the code path triggered by the `DestroyReady` phase input did not reach any cleanup code.

Rework the `ExposedService` cleanup code slightly to also remove the `DestroyReady` resources.


smira (approved):

smira:
We should do #527 I guess to really fix this... and do a QTransform one-to-many

utkuozdemir:
yep, we should still do that, let this be an immediate fix for now.

PR #568, opened 2024-08-19
chore: add periodic cleanup of old log files

### LINKED ISSUE #546: Add support for audit log files rotation

DmitriyMV:
Delete files older than 30 days. Also bump minimal Go version to 1.23 and fix lint issues.

Closes #546


cmd/integration-test/pkg/tests/stats.go
```diff
@@ -91,7 +92,7 @@ func AssertStatsLimits(testCtx context.Context) TestFunc {
 					}
 
 					if agg.hadErrors {
-						return retry.ExpectedErrorf(agg.String())
+						return retry.ExpectedError(errors.New(agg.String()))
```
Unix4ever, line 95: Why do we need this change? Linter?

Unix4ever (approved):

cmd/integration-test/pkg/tests/stats.go
DmitriyMV: Yup

Issue #516, opened 2024-08-06
keep machine status link resources for machine statuses in tearing down phase


Unix4ever:
Currently it's being destroyed as soon as machine status enters tearing down.
But that leads to machine not being displayed in the list of connected machines, which is very confusing.

We can make it still display the machines but show it as grayed out or something.

PR #570, opened 2024-08-19
fix: display machines in tearing down state

Unix4ever:
We display `MachineStatusLink` on the `Machines` page, which was removed right after `MachineStatus` enters tearing down.
But the `MachineStatus` can have finalizers pending for a long time, so the user gets false impression that the machine was already removed, while it's not.

Make `MachineStatusLink` remain until the corresponding `MachineStatus` exist.
Add the flag to the spec to allow displaying that the machine is in tearing down phase in the UI.

Show machines in tearing down phase with 50% opacity and disable any operations with such machine.

Fixes: https://github.com/siderolabs/omni/issues/516


Unix4ever:
Screenshot 2024-08-19 at 21 58 53

utkuozdemir (approved):

Issue #505, opened 2024-08-02
[bug] Talos update shows `Installing Extensions` when reverting an upgrade


Unix4ever:
Should show reverting upgrade instead.

PR #571, opened 2024-08-20
fix: use a separate phase for the extensions installation

Unix4ever:
Fixes: https://github.com/siderolabs/omni/issues/505

This should make the UI more reliable.


internal/backend/runtime/omni/controllers/omni/talos_upgrade_status.go
```diff
@@ -403,6 +403,10 @@ func reconcileTalosUpdateStatus(ctx context.Context, r controller.ReaderWriter,
 	totalMachines := clusterMachines.Len()
 	pendingMachines := totalMachines - len(machinesToUpdate) + 1
 
+	if schematicUpdates && !versionMismatch {
```
smira, line 406: should be in the same `if` block as below (or better, `switch`) statement?

otherwise this value will be overwritten with the `if` block below

Unix4ever: actually it can never be overwritten as if `schematicUpdates && !versionMismatch` is true it will always be true for `versionMismatch || schematicUpdates`
I can move this `if` inside block:
```
if versionMismatch || schematicUpdates {
```

That'll be an equivalent version

Unix4ever: Or the switch that looks like that:

```
	switch {
	case versionMismatch || schematicUpdates:
		upgradeStatus.TypedSpec().Value.Status = fmt.Sprintf("updating machines %d/%d", pendingMachines, totalMachines)
	case schematicUpdates && !versionMismatch:
		upgradeStatus.TypedSpec().Value.Status = fmt.Sprintf("updating machines %d/%d", pendingMachines, totalMachines)
		upgradeStatus.TypedSpec().Value.Phase = specs.TalosUpgradeStatusSpec_InstallingExtensions
	default:
		upgradeStatus.TypedSpec().Value.Phase = specs.TalosUpgradeStatusSpec_Reverting
		upgradeStatus.TypedSpec().Value.Status = fmt.Sprintf("reverting update %d/%d", pendingMachines, totalMachines)
	}
```

smira (approved):
ah, got it, I was confused

utkuozdemir (approved):

Issue #557, opened 2024-08-15
generate new BUSL license end date every time we do release


Unix4ever:
Make release.sh generate new BUSL license end date every time `commit` is called.
Should update `kres` to support that.

smira:
Note: this should happen only on `main`

PR #577, opened 2024-08-21
chore: rekres to get BUSL license change date updated on releases

utkuozdemir:
Bring in https://github.com/siderolabs/kres/pull/436 from kres.

Closes siderolabs/omni#557.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -7,7 +7,7 @@ Licensed Work:        Omni
                       The Licensed Work is (c) 2024 Sidero Labs, Inc.
 Additional Use Grant: None
 
-Change Date:          2028-08-16
+Change Date:          2028-08-22
 
 Change License:       Mozilla Public License, version 2.0
 
--- a/.kres.yaml
+++ b/.kres.yaml
@@ -509,7 +509,7 @@ spec:
         Licensor: Sidero Labs, Inc.
         LicensedWork: Omni
         Copyright: (c) 2024 Sidero Labs, Inc.
-        ChangeDate: "2028-02-22"
+        ChangeDate: "2028-08-22"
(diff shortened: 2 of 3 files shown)


DmitriyMV (approved):

LICENSE
```diff
@@ -7,7 +7,7 @@ Licensed Work:        Omni
                       The Licensed Work is (c) 2024 Sidero Labs, Inc.
 Additional Use Grant: None
 
-Change Date:          2028-08-16
+Change Date:          2028-02-22
```
smira, line 10: why did it go back in time?

utkuozdemir: Oh, this is kres's doing, but I don't know why. Let me have a look.

utkuozdemir: It is because it takes the `ChangeDate` template param as-is from `.kres.yaml` when rekres-ed. I updated it. Maybe we should make kres smarter about it somehow, if you have a good idea. For example, we could leave out `ChangeDate` to be a param, and always set it to be `today+4y` or something like that.

smira: the whole idea was to bump `Change Date` to be `now` + N years each time you do a release on `main`.

smira: otherwise this is basically no-op

utkuozdemir: it still works, but only when you rekres, it takes it back unless you go and update `.kres.yaml`. that is the problem. when we don't rekres, it works as expected.

smira: then it should actually update `.kres.yaml`? idk

but the goal I stated above, the implementation can be anything

utkuozdemir: yes, that would work, just did that here: https://github.com/siderolabs/kres/pull/437

PR #581, opened 2024-08-23
chore: enable 'github.com/planetscale/vtprotobuf' encoding

DmitriyMV:
Use it for both client and server gRPC.

### DIFF
--- a/client/pkg/client/encoding.go
+++ b/client/pkg/client/encoding.go
@@ -1,0 +1,54 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+package client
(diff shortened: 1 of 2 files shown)


smira (approved):
we also use gRPC to etcd, so I hope it doesn't break things here

client/pkg/client/encoding.go
```diff
@@ -0,0 +1,54 @@
...
+}
+
+func (vtprotoCodec) Unmarshal(data []byte, v any) error {
+	switch v := v.(type) {
+	case vtprotoMessage:
+		return v.UnmarshalVT(data)
+	case proto.Message:
+		return proto.Unmarshal(data, v)
+	case protoadapt.MessageV1:
+		return proto.Unmarshal(data, protoadapt.MessageV2Of(v))
```
DmitriyMV, line 41: This is what allows us to support legacy protobuf format.

Issue #508, opened 2024-08-05
[bug] Omni doesn't reload certs


rsmitty:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When running a self hosted Omni, we're using certbot to handle certificate generation and renewal. Omni then uses these certs. However, even though certbot renews the certs as expected, Omni doesn't reload the certs upon this renewal. Leads to an error in browser and kube proxy not functioning at all.

### Expected Behavior

Omni sees cert changes and reloads as necessary.

### Steps To Reproduce

1. Gen certs with certbot
2. Use these certs with a basic docker run for the omni container
3. Wait for cert renewal
4. Omni shows expired cert errors.

smira:
The easiest solution is to copy from `etcd` - they load the certs each time `tls.Config` is used.

PR #587, opened 2024-08-26
feat: reload TLS certs without restart

Unix4ever:
Omni now watches gRPC/HTTP TLS cert files for changes and loads them without restart.

Fixes: https://github.com/siderolabs/omni/issues/508


internal/backend/server.go
```diff
@@ -847,21 +849,123 @@ func setRealIPRequest(req *http.Request) *http.Request {
...
+
+	if err = w.Add(c.keyFile); err != nil {
+		return fmt.Errorf("error adding watch for file %s: %w", c.keyFile, err)
 	}
 
-	return s.server.ListenAndServe()
+	for {
+		select {
+		case e := <-w.Events:
+			if !e.Has(fsnotify.Write) && !e.Has(fsnotify.Create) {
```
smira, line 912: my only concern - does it work with Kubernetes `Secretes` being mounted? 

how do these get updated? do they replace/rename the files?

internal/backend/server.go
```diff
@@ -847,21 +849,123 @@ func setRealIPRequest(req *http.Request) *http.Request {
...
+				break
+			}
+
+			if err = c.load(); err != nil {
+				logger.Error("failed to load certs", zap.Error(err))
```
smira, line 917: should we keep watching here? as it might be incomplete file that we see right now (e.g. file being overwritten), but it will be "good" soon?

internal/backend/server.go
Unix4ever: That's a good question. I can probably copy that code from the etcd codebase.

internal/backend/server.go
```diff
@@ -847,21 +854,142 @@ func setRealIPRequest(req *http.Request) *http.Request {
...
+func (c *certData) load() error {
+	cert, err := tls.LoadX509KeyPair(c.certFile, c.keyFile)
+
+	if err == nil {
+		c.loaded = true
+	}
+
+	if err != nil {
+		return err
+	}
```
utkuozdemir, line 878: This is a bit weird, shouldn't we just check for the error first like usual, and below simply set loaded=true?

internal/backend/server.go
```diff
@@ -847,21 +854,142 @@ func setRealIPRequest(req *http.Request) *http.Request {
...
 	keyFile  string
+	mu       sync.Mutex
+	loaded   bool
+}
+
+func (c *certData) load() error {
+	cert, err := tls.LoadX509KeyPair(c.certFile, c.keyFile)
+
+	if err == nil {
+		c.loaded = true
```
utkuozdemir, line 873: Shouldn't the mutex protect this variable as well?

utkuozdemir (approved):

Issue #519, opened 2024-08-07
map nvidia extension names


Unix4ever:
Do the similar thing in Omni:

https://github.com/siderolabs/image-factory/pull/136

Related to: https://github.com/siderolabs/talos/issues/9086

PR #589, opened 2024-08-27
chore: rewrite renamed extension names on Talos version updates

utkuozdemir:
If an extension is renamed between minor Talos versions, rewrite its name to the newer one on upgrades and to older one on downgrades.

Handle the rename of `xe-guest-utilities` to `xen-guest-agent` happened on `1.6->1.7`, which was previously not handled correctly.

Additionally, closes siderolabs/omni#519.

### DIFF
--- a/internal/backend/runtime/omni/controllers/omni/schematic_configuration.go
+++ b/internal/backend/runtime/omni/controllers/omni/schematic_configuration.go
@@ -26,6 +26,7 @@ import (
 	"github.com/siderolabs/omni/client/pkg/omni/resources"
 	"github.com/siderolabs/omni/client/pkg/omni/resources/omni"
 	"github.com/siderolabs/omni/client/pkg/omni/resources/siderolink"
+	"github.com/siderolabs/omni/internal/backend/extensions"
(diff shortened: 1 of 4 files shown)


internal/backend/runtime/omni/controllers/omni/schematic_configuration.go
```diff
@@ -377,8 +385,8 @@ func detectExtensions(initialVersion, currentVersion semver.Version) ([]string,
 	if initialVersion.Major == 1 && initialVersion.Minor == 5 && currentVersion.GTE(semver.MustParse("1.6.0")) {
 		// install firmware
 		return []string{
-			"siderolabs/bnx2-bnx2x",
-			"siderolabs/intel-ice-firmware",
+			extensions.OfficialPrefix + "bnx2-bnx2x",
```
frezbo, line 388: i guess we can move this also to `extensions.MapByVersion`, follow somothing similar to quirks in talos, so all these customization is handled in one package

Unix4ever (approved):

smira (approved):

frezbo (approved):

PR #591, opened 2024-08-28
chore: allow users with admin role to download audit log from UI

### LINKED ISSUE #586: Add audit log to UI (stream RPC)

DmitriyMV:
Currently, there is no way to write to file using stream-like API, so we have to concatenate the full log in memory, before we save it to the file system. In practice, if this proves to be problematic we can always redo the logic using standard http file download.

Closes #586


frontend/src/views/omni/Overview/Overview.vue
```diff
@@ -170,6 +173,12 @@ const hasRoleNone = computed(() => {
   return !role || role === RoleNone;
 });
 
+const hasRoleAdmin = computed(() => {
+  const role = currentUser.value?.spec?.role
```
Unix4ever, line 177: I'd suggest to create a separate flag like `canDownloadAuditLog` in the currentUser resource.

hack/release.toml
```diff
@@ -7,3 +7,19 @@ match_deps = "^github.com/((talos-systems|siderolabs)/[a-zA-Z0-9-]+)$"
 
 pre_release = true
 previous = "v0.40.0"
+
+[notes]
+
+  [notes.extensions]
```
Unix4ever, line 13: :heart:

DmitriyMV: Yeah, sorry, updated to proper name.

frontend/src/views/omni/Overview/Overview.vue
DmitriyMV: Done!

Unix4ever (approved):

PR #602, opened 2024-09-02
feat: implement compression of config fields on resources

utkuozdemir:
Add compression support.

I did a manual comparison using 100 clusters with the whole ArgoCD manifest set as an `inlineManifest` config patch on each of them.
Before the compression, unmarshaling of these four resources occupied 477MB combined in memory, amounting to 52% of the total in-use memory (1.13GB).
After the compression, the number reduced to 34MB total, less than 10% of the total in-use memory (390MB).

Difference in the size of the compressed data when using a dictionary vs not (small inputs give much better results, the ideal dictionary size seems to be 64K according to manual comparison of different sizes):
```
    compression_bench_test.go:76: small input no dict: 2.9 KiB
    compression_bench_test.go:77: small input with dict: 1.3 KiB
    compression_bench_test.go:78: large input no dict: 69 KiB
    compression_bench_test.go:79: large input with dict: 63 KiB
```

The benchmarks:
```
goos: linux
goarch: amd64
pkg: github.com/siderolabs/omni/client/pkg/compression
cpu: AMD Ryzen 9 3900X 12-Core Processor
BenchmarkAccess/small_no-compression_no-pool_no-dict-24         	 4574419	       256.1 ns/op	      24 B/op	       1 allocs/op
BenchmarkAccess/large_no-compression_no-pool_no-dict-24         	 4557902	       257.9 ns/op	      24 B/op	       1 allocs/op
BenchmarkAccess/small_with-compression_no-pool_no-dict-24       	  118248	      9722 ns/op	    5462 B/op	       3 allocs/op
BenchmarkAccess/large_with-compression_no-pool_no-dict-24       	    1627	    725780 ns/op	 1665912 B/op	       3 allocs/op
BenchmarkAccess/small_with-compression_with-pool_no-dict-24     	  144502	      8062 ns/op	      77 B/op	       2 allocs/op
BenchmarkAccess/large_with-compression_with-pool_no-dict-24     	    2132	    545663 ns/op	    3072 B/op	       2 allocs/op
BenchmarkAccess/small_with-compression_with-pool_with-dict-24   	  155839	      7402 ns/op	      77 B/op	       2 allocs/op
BenchmarkAccess/large_with-compression_with-pool_with-dict-24   	    1420	    754848 ns/op	    4508 B/op	       2 allocs/op
```


client/api/omni/specs/compression.go
```diff
@@ -0,0 +1,128 @@
...
+
+var jsonMarshaler = protojson.MarshalOptions{
+	UseProtoNames:  true,
+	UseEnumNumbers: true,
+}
+
+func doCompress(r io.Reader) ([]byte, error) {
+	var buf bytes.Buffer
+
+	writer, err := zstd.NewWriter(&buf)
```
smira, line 29: I believe there was a way to re-use the `ztsd` instance to re-use buffers on compression/decompression

I think it'd cool to have a pool of slices one way or another, but it'd be tricky to track the lifecycle of these...

internal/backend/runtime/omni/controllers/omni/internal/machineset/operations.go
```diff
@@ -205,16 +206,19 @@ func (d *Destroy) Apply(ctx context.Context, r controller.ReaderWriter, logger *
 	return nil
 }
 
-func setPatches(clusterMachineConfigPatches *omni.ClusterMachineConfigPatches, patches []*omni.ConfigPatch) {
+func setPatches(clusterMachineConfigPatches *omni.ClusterMachineConfigPatches, patches []*omni.ConfigPatch, compressConfigData bool) error {
 	patchesRaw := make([]string, 0, len(patches))
 
 	for _, patch := range patches {
-		data := patch.TypedSpec().Value.Data
+		data, err := patch.TypedSpec().Value.GetUncompressedData()
```
smira, line 213: be :cool: and don't recompress patches if they are already compressed

client/api/omni/specs/compression.go
utkuozdemir: addressed both

internal/backend/runtime/omni/controllers/omni/internal/machineset/operations.go
utkuozdemir: done

client/api/omni/specs/cluster_machine_config_patches.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+
+	if x.CompressedPatches == nil {
+		buffers := make([]Buffer, 0, len(x.Patches))
+
+		for _, patch := range x.Patches {
+			buffers = append(buffers, newNoOpBuffer([]byte(patch)))
+		}
+
+		return buffers, nil
```
Unix4ever, line 89: `xslices.Map` can make it much shorter:

```suggestion
		return xslices.Map(x.Patches, func(patch string) []Buffer {
			return newNoOpBuffer([]byte(patch))
		})
```

client/api/omni/specs/cluster_machine_config_patches.go
```diff
@@ -0,0 +1,166 @@
...
+		patches := make([]string, 0, len(data))
+
+		for _, patch := range data {
+			patches = append(patches, string(patch))
+		}
+
+		x.Patches = patches
+		x.CompressedPatches = nil
+
+		return nil
```
Unix4ever, line 126:
```suggestion
		x.Patches = xslices.Map(data, func(patch []byte) string { return string(patch) }
		x.CompressedPatches = nil

		return nil
```

client/api/omni/specs/cluster_machine_config_patches.go
utkuozdemir: I kinda avoided to bring that package into the `specs` package (as it is public), but now that I checked, it was already used on `ClusterMachineConfigPatchesSpec.MarshalYAML` (I just moved this function, didn't change).

So I guess it makes sense to use it here as well now.

client/api/omni/specs/cluster_machine_config_patches.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+
+	compressedPatches := make([][]byte, 0, len(data))
+
+	for _, patch := range data {
+		compressedPatches = append(compressedPatches, doCompress(patch, config))
+	}
+
+	x.CompressedPatches = compressedPatches
```
Unix4ever, line 135:
```suggestion
	x.CompressedPatches = xslices.Map(data, func(patch []byte) []byte { doCompress(patch, config) }
```

client/api/omni/specs/cluster_machine_config_patches.go
```diff
@@ -0,0 +1,166 @@
...
+		return nil, err
+	}
+
+	patches := make([]string, 0, len(buffers))
+
+	for _, buffer := range buffers {
+		patches = append(patches, string(buffer.Data()))
+	}
+
+	return patches, nil
```
Unix4ever, line 154:
```suggestion
	return xslices.Map(buffers, func(patch Buffer) string { string(buffer.Data()) }, nil
```

client/pkg/template/internal/models/patch.go
```diff
@@ -157,7 +157,10 @@ func (patch *Patch) Translate(prefix string, weight int, labels ...pair.Pair[str
 
 	patch.Descriptors.Apply(patchResource)
 
-	patchResource.TypedSpec().Value.Data = string(raw)
+	// we do not know if the Omni server enables patch compression, so we always compress the data
```
Unix4ever, line 160: I'd say this config should go to `root.go` `InitConfig` call.

hack/zstd-dict/data/argocd-manifests.yaml
Unix4ever, line 1: Do we really need this file?

hack/zstd-dict/main.go
```diff
@@ -0,0 +1,240 @@
+// Copyright (c) 2024 Sidero Labs, Inc.
```
Unix4ever, line 1: I would be nice to have some README on how this intended to be ran and where.
I Kubernetes? Or locally?

I guess the goal is to generate `config.zdict` right?

utkuozdemir: Yes, I'll write it down to clarify. This code does two things
- as you said, generate a config.zdict. But then you are responsible of putting it to its place.
- it also writes all the input files of the dictionary to the `files/` dir there, so `zstd` command line tool can be used to generate the dictionary instead (gives better result than the go code).

I'll write them down nicely in README.

internal/backend/grpc/resource.go
```diff
@@ -362,8 +363,13 @@ func CreateResource(resource *resources.Resource) (cosiresource.Resource, error)
...
 	}
 
+	unmarshalValue, err := initUnmarshalResourceSpec(obj)
+	if err != nil {
+		return nil, err
+	}
+
 	if resource.Spec != "" {
-		err = json.Unmarshal([]byte(resource.Spec), obj.Spec())
+		err = json.Unmarshal([]byte(resource.Spec), unmarshalValue)
```
Unix4ever, line 372: This approach worked for me in the infra provider:

```
		if r, ok := res.Spec().(interface {
			UnmarshalJSON(bytes []byte) error
		}); ok {
			if err = r.UnmarshalJSON([]byte(resource.Spec)); err != nil {
				return err
			}
		}
```

utkuozdemir: hmm, this is cool, I'll try that.

internal/backend/runtime/omni/controllers/omni/cluster_machine_config_test.go
```diff
@@ -157,24 +172,37 @@ func (suite *ClusterMachineConfigSuite) TestGenerationError() {
...
         bridge: invalidValueType`)
 
-			return nil
+			return config.TypedSpec().Value.SetUncompressedPatches(patches)
 		},
 	)
 
 	suite.Require().NoError(err)
 
+	time.Sleep(2 * time.Second)
```
Unix4ever, line 190: intended?

internal/backend/runtime/omni/controllers/omni/cluster_machine_encryption.go
```diff
@@ -27,7 +27,9 @@ import (
 var encryptionConfigPatchData string
 
 // ClusterMachineEncryptionController manages disk encryption configuration for the cluster.
-type ClusterMachineEncryptionController struct{}
+type ClusterMachineEncryptionController struct {
+	CompressConfigData bool
```
Unix4ever, line 31: is it used?

internal/backend/runtime/omni/controllers/omni/cluster_machine_config_test.go
utkuozdemir: nope, sorry

internal/backend/runtime/omni/controllers/omni/cluster_machine_encryption.go
utkuozdemir: nope, seems I left some traces from my previous approach where I explicitly passed this to the accessors. will clean all of them up.

hack/zstd-dict/main.go
utkuozdemir: done

utkuozdemir: Also removed the argo manifests - now the dict tool downloads the chart and does `helm template` on it itself.

internal/backend/grpc/resource.go
utkuozdemir: Couldn't make that work, as the type of `Spec()` here is:
`*protobuf.ResourceSpec[specs.ConfigPatchSpec,*specs.ConfigPatchSpec]`

This type implements `UnmarshalJSON`, but in the `protobuf` package, on `ResourceSpec`. And it simply ignores my overridden `UnmarshalJSON` function on the `ConfigPatchSpec`, as it uses `protojson`'s unmarshal function.

So what I need here is to call the `.Value` on the spec to get the actual `ConfigPatchSpec` object, and call its `UnmarshalJSON` function.

Am I missing something? Is there a neat way of doing it?

Unix4ever: Oh I see now. You have `UnmarshalJSON` defined on the wrapped spec.

I think the right way would be to customize COSI runtime code to call `UnmarshalJSON` on the wrapped spec if it implements it.
I'd really prefer to avoid the code with lots of reflection here.

Unix4ever: The change in the COSI runtime should be simple:
```go
// UnmarshalJSON implements json.Unmarshaler.
func (spec *ResourceSpec[T, S]) UnmarshalJSON(bytes []byte) error {
	spec.Value = new(T)

	if unmarshaler, ok := spec.(json.Unmarshaler); ok {
		return unmarshaler.UnmarshalJSON(bytes)
	}

	opts := protojson.UnmarshalOptions{}

	return opts.Unmarshal(bytes, spec.Value)
}
```

Unix4ever (approved):

PR #604, opened 2024-09-02
feat: support passing extra data through the siderolink join token

Unix4ever:
This extra data is used in the infra provider to add the annotation to the `siderolink.Link` as early as possible.
Then the `Machine` controller is changed to skip the `Links` that have annotation `omni.sidero.dev/infra-provider` and do not have the label `omni.sidero.dev/machine-request`.
This change makes not consistent `Links` to be ignore by the system, until the are fully populated.

Also changed the infra provider iterface to take siderolink connection params as string instead of the resource.

Fixes: https://github.com/siderolabs/omni/issues/603


client/pkg/jointoken/jointoken.go
```diff
@@ -0,0 +1,54 @@
...
+	"strings"
+)
+
+// EncodedPrefix is used for the tokens encoded in base64.
+const EncodedPrefix = "b64:"
+
+// JoinToken is the siderolink join token.
+// Custom type adds methods for encoding/decoding extra data from the token.
+type JoinToken struct {
+	ExtraData map[string]string `json:"extra_data"`
```
smira, line 20: I wonder if we should instead of just passing extra data completely unverified, use HMAC with the password to simply "sign" the data? This way the join token can't modified without knowing the password, and it is still verifiable/

Unix4ever: yep, makes sense

internal/pkg/siderolink/manager.go
```diff
@@ -684,6 +697,12 @@ func (manager *Manager) getLink(ctx context.Context, req *pb.ProvisionRequest, i
 
 		link := siderolink.NewLink(siderolink.Namespace, id, spec)
 
+		if token.ExtraData != nil {
```
utkuozdemir, line 700: this nil check shouldn't be necessary

client/pkg/omni/resources/siderolink/connection_params.go
```diff
@@ -93,3 +98,50 @@ func APIURL(cfg *ConnectionParams, useTunnel bool) (string, error) {
 
 	return url.String(), nil
 }
+
+// GetConnectionArgsForProvider composes connection args for the specific provider.
+func GetConnectionArgsForProvider(connectionParams *ConnectionParams, providerID string) (string, error) {
```
utkuozdemir, line 103: a small unit test for this can be a good idea

client/pkg/jointoken/jointoken.go
```diff
@@ -0,0 +1,115 @@
...
+
+	if _, err = mac.Write(data); err != nil {
+		return nil, err
+	}
+
+	return mac.Sum(nil), nil
+}
+
+// V1Prefix is used for the tokens encoded using the first token version.
+const V1Prefix = "v1:"
```
utkuozdemir, line 35: does this need to be exported?

client/pkg/infra/provision/context.go
```diff
@@ -0,0 +1,194 @@
...
+}
+
+// SetMachineInfraID in the machine request status.
+func (context *Context[T]) SetMachineInfraID(value string) {
+	context.MachineRequestStatus.Metadata().Labels().Set(omni.LabelMachineInfraID, value)
+}
+
+// GenerateSchematicID generate the final schematic out of the machine request.
+// This method also calls the image factory and uploads the schematic there.
+func (context *Context[T]) GenerateSchematicID(ctx context.Context, opts ...SchematicOption) (string, error) {
```
utkuozdemir, line 126: being able to pass a logger here could be helpful by `.WithLogger` or something, and maybe we could log the generated schematic yaml

client/pkg/infra/imagefactory/client.go
```diff
@@ -0,0 +1,57 @@
...
+		return "", fmt.Errorf("failed to create schematic: %w", err)
+	}
+
+	return schematicID, nil
+}
+
+// NewClient creates new Omni based image factory client.
+func NewClient(options ClientOptions) (*Client, error) {
+	if options.FactoryEndpoint == "" {
+		options.FactoryEndpoint = "https://factory.talos.dev"
```
utkuozdemir, line 46: can we use the constant `ImageFactoryBaseURL` here?

client/pkg/infra/controllers/provision.go
```diff
@@ -14,35 +15,39 @@ import (
...
+	"github.com/siderolabs/omni/client/api/omni/specs"
 	infrares "github.com/siderolabs/omni/client/pkg/infra/internal/resources"
 	"github.com/siderolabs/omni/client/pkg/infra/provision"
 	"github.com/siderolabs/omni/client/pkg/omni/resources"
 	"github.com/siderolabs/omni/client/pkg/omni/resources/infra"
-	"github.com/siderolabs/omni/client/pkg/omni/resources/omni"
 	"github.com/siderolabs/omni/client/pkg/omni/resources/siderolink"
 )
 
+const currentStepAnnotation = "omni.sidero.infra.dev/step"
```
utkuozdemir, line 28: maybe `"infra." + SystemLabelPrefix + "step"` or something like that, for the sake of consistency in the used domain for the labels?

internal/pkg/siderolink/manager.go
Unix4ever: this map can be nil if the token is plain, so I don't think we can't remove this check

utkuozdemir: I mean we just read data from map, so nil wouldn't panic.

client/pkg/jointoken/jointoken.go
Unix4ever: we can make it private for now

client/pkg/infra/provision/context.go
Unix4ever: makes sense

internal/pkg/siderolink/manager.go
Unix4ever: I keep thinking that reading from a nil map will panic :D

utkuozdemir: yep, tricky to remember :)

utkuozdemir (approved):

PR #623, opened 2024-09-11
fix: better handle machine move between machine sets

Unix4ever:
It looks like `qtransform` controller skips reconciliation if it gets `ConflictErrors`:
https://github.com/cosi-project/runtime/blob/8911486fcad50b151399e4652c7c62f2e037e7dd/pkg/controller/generic/qtransform/qtransform.go#L256

Cluster machine creation operation can return conflict errors, and the controller should actually fail, not skip the reconciliation, so we mask this error in the `reconcileMachines` func.


internal/backend/runtime/omni/controllers/omni/machine_set_status.go
```diff
@@ -195,6 +196,12 @@ func (handler *machineSetStatusHandler) reconcileRunning(ctx context.Context, r
 
 	requeue, err := handler.reconcileMachines(ctx, r, logger, rc)
 	if err != nil {
+		// qtransform controller ignores PhaseConflict errors
+		// mask the error and make the controller fail explicitly instead
+		if state.IsPhaseConflictError(err) || state.IsConflictError(err) {
+			return fmt.Errorf("operations failed %s", err.Error())
```
utkuozdemir, line 202: intended, or?
```suggestion
			return fmt.Errorf("operations failed: %w", err)
```

utkuozdemir: nvm, didn't pay a attention to the comment above.

Unix4ever: intended, see the comment above

smira: I think we kept ignoring `PhaseConflictError` e.g. when you try update something while it is being torn down, that was I guess for primary output only, but since that time we got extra outputs as well

so idk how we can best fix that on the COSI side - enhance the error to track  which resource it was?

(I guess here it's an extra output)

Unix4ever: Yeah, adding the resource ref to the error can work I guess.

Unix4ever: https://github.com/cosi-project/runtime/pull/492

utkuozdemir (approved):

Issue #388, opened 2024-06-18
Allow for GRPC Tunnel when creating install media


rsmitty:
We should provide a tick box or something to allow users to easily enable the GRPC tunnel.

PR #627, opened 2024-09-12
feat: allow specifying grpc tunnel option explicitly for install media

utkuozdemir:
By default, generated Talos installation media uses `grpc_tunnel` for SideroLink based on the Omni instance configuration, namely via `--siderolink-use-grpc-tunnel` flag.

Allow overriding this setting in `omnictl download` and in Download Installation Media screen on the web.

On the Download Installation Media screen, the default value of the checkbox is based on the instance default.

Closes siderolabs/omni#388.

## Preview

---


client/pkg/omnictl/download.go
```diff
@@ -58,6 +63,9 @@ func init() {
 	downloadCmd.Flags().StringArrayVar(&downloadCmdFlags.labels, "initial-labels", nil, "Bake initial labels into the generated installation media")
 	downloadCmd.Flags().StringArrayVar(&downloadCmdFlags.extraKernelArgs, "extra-kernel-args", nil, "Add extra kernel args to the generated installation media")
 	downloadCmd.Flags().StringArrayVar(&downloadCmdFlags.extensions, "extensions", nil, "Generate installation media with extensions pre-installed")
+	downloadCmd.Flags().StringVar(&downloadCmdFlags.siderolinkGRPCTunnelMode, "siderolink-grpc-tunnel-mode",
```
Unix4ever, line 66: Maybe just add a bool flag?
I understand the idea behind AUTO mode, but maybe we can make it work like that:
- no flag set: auto mode.
- flag set to false: override mode to false.
- flag set to true: override mode to true.

Something like pointer to bool could work here, but I'm not sure if flags lib supports that.

Unix4ever: Not having to remember constants for different modes is better I think.

utkuozdemir: did the change to work exactly that way

Unix4ever (approved):

PR #629, opened 2024-09-16
chore: bump default versions: Talos `1.7.6`, Kubernetes `1.30.5`

utkuozdemir:
Bump them to the latest and greatest versions.


frezbo (approved):

TimJones (approved):
Feels weird that we can't pull those go constants into the typescript frontend :(

smira (approved):

smira:
> Feels weird that we can't pull those go constants into the typescript frontend :(

this TypeScript file is generated from Go actually

utkuozdemir:
> Feels weird that we can't pull those go constants into the typescript frontend :(

Actually the `tsgen:*` instructions above them do exactly that, but statically - we have to run `make generate` after editing the constants.

We could easily change it to fetch dynamically at runtime and/or pick the latest one automatically.

smira:
Should also update https://github.com/siderolabs/omni/blob/c4a4151d7a5d9030ec82fab434932f3d002e59cf/hack/test/integration.sh#L23

Unix4ever (approved):

Issue #595, opened 2024-08-28
machine set node controller should support the mode which allocates machines from a specific machine request set


Unix4ever:
This mode should be mutually exclusive with machine class mode and manual mode.
If the machine set has some machines allocated it shouldn't be possible to switch modes.

PR #630, opened 2024-09-16
feat: support attaching machine sets to a machine request sets

Unix4ever:
With that it becomes possible to get the machines from the machine request sets instead of the machine classes.

This opens the way for automated machine provisioning. Fixes: https://github.com/siderolabs/omni/issues/595

### DIFF
--- a/client/pkg/omni/resources/omni/machine_set.go
+++ b/client/pkg/omni/resources/omni/machine_set.go
@@ -73,3 +73,16 @@ func (MachineSetExtension) ResourceDefinition() meta.ResourceDefinitionSpec {
 		PrintColumns:     []meta.PrintColumn{},
 	}
 }
+
+// GetMachineAllocation from the machine set resource.
+func GetMachineAllocation(res *MachineSet) *specs.MachineSetSpec_MachineAllocation {
+	if res.TypedSpec().Value.MachineClass != nil { //nolint:staticcheck
+		return res.TypedSpec().Value.MachineClass //nolint:staticcheck
+	}
+
+	if res.TypedSpec().Value.MachineAllocation != nil {
+		return res.TypedSpec().Value.MachineAllocation
+	}
+
+	return nil
+}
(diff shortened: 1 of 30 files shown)


client/pkg/omni/resources/omni/machine_set.go
```diff
@@ -73,3 +73,16 @@ func (MachineSetExtension) ResourceDefinition() meta.ResourceDefinitionSpec {
 		PrintColumns:     []meta.PrintColumn{},
 	}
 }
+
+// GetMachineAllocation from the machine set resource.
+func GetMachineAllocation(res *MachineSet) *specs.MachineSetSpec_MachineAllocation {
+	if res.TypedSpec().Value.MachineClass != nil {
+		return res.TypedSpec().Value.MachineClass
```
utkuozdemir, line 80: Should we mark this and the other field as `Deprecated:` and tell that the right way of getting allocation is using this function instead? The same way I did in compressed resources.

utkuozdemir: Also, I'm thinking - what if we reduced this into a single field instead, and add another boolean or enum to indicate whether it's a class or request set? Or you think this is a clearer way?

Unix4ever: > Should we mark this and the other field as Deprecated: and tell that the right way of getting allocation is using this function instead? The same way I did in compressed resources.

Well, this method is internal to our code base. And we don't deprecate `MachineClass` usage.

> Also, I'm thinking - what if we reduced this into a single field instead, and add another boolean or enum to indicate whether it's a class or request set? Or you think this is a clearer way?

I had the same thoughts. But maybe later I'll add more fields to the `MachineRequestSet` allocation time. I'm thinking to put the config for machine request set scaling.

Unix4ever: > Also, I'm thinking - what if we reduced this into a single field instead, and add another boolean or enum to indicate whether it's a class or request set? Or you think this is a clearer way?

Actually after stitching more pieces together I'm starting to lean towards the flag in the single allocation config field.

utkuozdemir: as you wish, you know the internals better here. I was just pointing out in case you haven't though of that option.

utkuozdemir (approved):

Issue #54, opened 2024-03-05
Integrate with Talos Diagnostics


andrewrynhard:
### Problem Description

Original problem:

> There is the possibility that a nodes' IP address is in the pod and/or service subnet(s). This results in a broken cluster.

Talos implements a set of checks, and results of these checks are populated as resources (see https://github.com/siderolabs/talos/issues/8682). The remaining piece is the glue to show diasnostics in Omni.

(This is a Talos 1.8 feature).

### Solution

Original solution:

> Warn and even refuse to create a cluster if a machine has an IP address in the pod and/or service subnet(s)

smira:
Should be implemented via https://github.com/siderolabs/talos/issues/8682

PR #641, opened 2024-09-18
feat: integrate with Talos diagnostics

utkuozdemir:
Gather the diagnostic information from Talos machines into `MachineStatus` resource. Aggregate them into a single `ClusterDiagnostics` resource per cluster.

Display this information on the cluster overview and node overview pages.

Closes siderolabs/omni#54.

# Preview

---

---

---

---


smira (commented):
:cool:

Unix4ever:
Reusing notification for this works, but I'd rather implement a dedicated UI element for this.

utkuozdemir:
> Reusing notification for this works, but I'd rather implement a dedicated UI element for this.

You mean moving it into its own component? And are we talking about the big one on the node overview page?

Unix4ever:
> You mean moving it into its own component? And are we talking about the big one on the node overview page?
If you mean that, I just changed it to be that way.

So for cluster overview I think it would be better to show the warning on the right sidebar. Maybe in the cluster details.
Or another option is to not show it at all, having this yellow warning sign near the node might be enough already.

On the node overview page, I think it might be better to put it right above the labels and below the 3 cards we have.
And use the same style for it: Title separated by the line and the same background.

Have a title: `Node Diagnostic Warnings`.

Maybe something like this, but stretched to the full width:
image

Not sure if we can have the level of the alerts there though

utkuozdemir:
> > You mean moving it into its own component? And are we talking about the big one on the node overview page?
> > If you mean that, I just changed it to be that way.
> 
> So for cluster overview I think it would be better to show the warning on the right sidebar. Maybe in the cluster details. Or another option is to not show it at all, having this yellow warning sign near the node might be enough already.
> 
> On the node overview page, I think it might be better to put it right above the labels and below the 3 cards we have. And use the same style for it: Title separated by the line and the same background.
> 
> Have a title: `Node Diagnostic Warnings`.
> 
> Maybe something like this, but stretched to the full width: image
> 
> Not sure if we can have the level of the alerts there though
> 
> Or instead of `view >` button we can use the same element we have for Machines for example. Arrow that can expand it.

Got it, will rework that.
The reason I showed it on top of Cluster Overview is because if cluster have many nodes, the warning not be visible at all, as we hide them after a certain threshold.

Unix4ever:
> The reason I showed it on top of Cluster Overview is because if cluster have many nodes, the warning not be visible at all, as we hide them after a certain threshold.

Ah I see. That's a good point.

But then I'd suggest to show it on the right sidebar.

smira:
One thing I forgot to ask - do we have a link to the docs for each warning?

https://github.com/danrue/talos/blob/d498f647cd9dfcd575f51005c9b78c2c1c7b51ca/pkg/machinery/resources/runtime/diagnostic.go#L32-L36

utkuozdemir:
> One thing I forgot to ask - do we have a link to the docs for each warning?
> 
> 
> 
> https://github.com/siderolabs/talos/blob/0ec75463eecebfb543a64b0c859ba0b2477e406f/pkg/machinery/resources/runtime/diagnostic.go#L32-L36
> 
> 

Oh nope, missed that one. Let me add that.

@Unix4ever do you think simply making the warning texts in node overview hyperlinks is a good idea? Or would you do it differently?

utkuozdemir:
> One thing I forgot to ask - do we have a link to the docs for each warning?
> 
> [siderolabs/talos@`0ec7546`/pkg/machinery/resources/runtime/diagnostic.go#L32-L36](https://github.com/siderolabs/talos/blob/0ec75463eecebfb543a64b0c859ba0b2477e406f/pkg/machinery/resources/runtime/diagnostic.go#L32-L36)

did something like this (opens in a new tab):

gif

Unix4ever:
Probably makes sense to show underline when you hover the message to indicate that the title is clickable

utkuozdemir:
> Some color change might be necessary on hover, maybe we can use:
> 
> ```css
> @apply transition-colors hover:brightness-125;
> ```

Thanks, done, looks like 
hover

Unix4ever:
I actually see a small UI issue on the gif. I don't know why it does that, but the bulletpoint size becomes smaller when there's lots of content on the line

frontend/src/views/cluster/Nodes/NodeDiagnosticWarnings.vue
```diff
@@ -0,0 +1,49 @@
...
+    <t-list-item v-for="diagnostic in diagnostics" :key="diagnostic.id" disable-border-on-expand>
+      <a class="diagnostic-item" :href="'https://talos.dev/diagnostic/' + diagnostic.id" target="_blank">
+        <t-icon icon="warning" class="w-4 h-4 text-yellow-400 mr-2"/>
+        <div class="text-yellow-400">{{ diagnostic.message }}</div>
+      </a>
+      <template #details>
+        <div class="diagnostic-sublist">
+          <div class="diagnostic-subitem" v-for="(detail, index) in diagnostic.details" :key="index">
+            <t-icon icon="dot" class="w-4 h-4 mr-2"/>
+            <span>{{ detail }}</span>
```
Unix4ever, line 18: probably adding `class="flex-1"` here will fix the icon changing size bug

Unix4ever (approved):

frontend/src/views/cluster/Nodes/NodeDiagnosticWarnings.vue
utkuozdemir: yep, done that, fixed the issue, thanks

utkuozdemir:
> I actually see a small UI issue on the gif. I don't know why it does that, but the bulletpoint size becomes smaller when there's lots of content on the line

yep, noticed that after posting the gif, was about to ask you that but you were quicker :)

PR #643, opened 2024-09-20
fix: get rid of the exceptions in the ui

Unix4ever:
Use proper var names and fields in the UI for cluster machine list and machine set config.

### DIFF
--- a/frontend/src/views/cluster/ClusterMachines/ClusterMachine.vue
+++ b/frontend/src/views/cluster/ClusterMachines/ClusterMachine.vue
@@ -23,7 +23,7 @@ included in the LICENSE file.
           </div>
         </div>
         <div class="flex gap-2 items-center"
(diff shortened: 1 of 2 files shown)


frontend/src/views/cluster/ClusterMachines/ClusterMachine.vue
```diff
@@ -23,7 +23,7 @@ included in the LICENSE file.
           </div>
         </div>
         <div class="flex gap-2 items-center"
-          v-if="machineSpec?.spec?.machine_allocation == undefined && machine?.metadata?.labels?.[LabelWorkerRole] !== undefined">
+          v-if="machineSet?.machine_allocation == undefined && machine?.metadata?.labels?.[LabelWorkerRole] !== undefined">
```
utkuozdemir, line 26: hehe I fixed this in my pr as well :)

smira (approved):

utkuozdemir (approved):

PR #651, opened 2024-09-26
feat: support auto provisioned machine classes

Unix4ever:
The `MachineClass` now can operate in two different modes:

- manual, which selects the machines using the label selectors.
- auto provisioned, which can automatically create and delete machines when they are requested by the machine sets which are using the machine class.

`MachineRequest` and `MachineRequestSet` now have the provider data field which is validated by the JSON schema which is uploaded to Omni by the infrastructure providers.

On the UI side introduce the new components for creating the machine classes which auto provision machines.

Fixes: https://github.com/siderolabs/omni/issues/644


internal/backend/runtime/omni/state_validation.go
```diff
@@ -421,8 +424,83 @@ func machineSetValidationOptions(st state.State, etcdBackupStoreFactory store.Fa
...
+			if err != nil {
+				return fmt.Errorf("failed to get provider: %w", err)
+			}
+
+			if err = validateSchema(autoProvision, providerStatus); err != nil {
+				return err
+			}
+		}
+
+		_, err := labels.ParseSelectors(res.TypedSpec().Value.MatchLabels)
```
smira, line 489: do we need to check for `nil` here ? or it doesn't care?

internal/backend/runtime/omni/state_validation.go
```diff
@@ -421,8 +424,83 @@ func machineSetValidationOptions(st state.State, etcdBackupStoreFactory store.Fa
...
+			if err := validateTalosVersion(ctx, st, "", autoProvision.TalosVersion); err != nil {
+				return err
+			}
+
+			providerStatus, err := safe.ReaderGetByID[*infra.ProviderStatus](ctx, st, autoProvision.ProviderId)
+			if err != nil {
+				return fmt.Errorf("failed to get provider: %w", err)
+			}
+
+			if err = validateSchema(autoProvision, providerStatus); err != nil {
```
smira, line 484: I wonder what happens if the schema changes, while the `autoProvision` block is not valid with new schema?

should we add a check to make sure we validate the schema only if it changes?

smira (approved):
looks really :cool:

internal/backend/runtime/omni/state_validation.go
Unix4ever: well, maybe we can check that it shouldn't be empty

internal/backend/runtime/omni/state_validation.go
Unix4ever: I think that schema should be changed in the backward compatible way.
In any case we can check if the data if it changes.

cmd/integration-test/pkg/tests/auth.go
```diff
@@ -1013,6 +1013,9 @@ func AssertResourceAuthz(rootCtx context.Context, rootCli *client.Client, client
 			{
 				resource: omni.NewBackupData(uuid.New().String()),
 			},
+			{
```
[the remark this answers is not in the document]
Unix4ever: for the controller, yes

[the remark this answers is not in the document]
Unix4ever: maybe I can create a test that uses talemu too

PR #658, opened 2024-10-01
fix: retry fetching Talos extensions for prod versions after 1.8.0

Unix4ever:
If Talos release gets published to the registry before the Image Factory extensions list is populated Omni might cache the empty extensions list.

There's no way to reliably watch extensions being published, so just retry fetching them every 5 minutes. But do it only for new Talos versions and ignore pre-releases.

### DIFF
--- a/internal/backend/runtime/omni/controllers/omni/talos_extensions.go
+++ b/internal/backend/runtime/omni/controllers/omni/talos_extensions.go
@@ -12,6 +12,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/blang/semver/v4"
 	"github.com/cosi-project/runtime/pkg/controller"
 	"github.com/cosi-project/runtime/pkg/controller/generic/qtransform"
 	"github.com/siderolabs/gen/xerrors"
@@ -22,6 +23,8 @@ import (
 	"github.com/siderolabs/omni/client/pkg/omni/resources/omni"
 	"github.com/siderolabs/omni/internal/backend/imagefactory"
 )
+
+var retryExtensionsFetchMinVersion = semver.MustParse("1.8.0")
 
 // TalosExtensionsController creates omni.TalosExtensions for each omni.TalosVersion.
 //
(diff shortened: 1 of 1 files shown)


internal/backend/runtime/omni/controllers/omni/talos_extensions.go
```diff
@@ -52,7 +55,16 @@ func NewTalosExtensionsController(imageFactoryClient *imagefactory.Client) *Talo
 
 				// skip fetching Talos extensions for a version which isn't registered in the image factory
 				if !hasVersion {
-					return xerrors.NewTaggedf[qtransform.SkipReconcileTag]("version %q is not registered in the image factory", version.Metadata().ID())
+					v, err := semver.ParseTolerant(version.Metadata().ID())
```
smira, line 58: I would rather fix this in the Image Factory if that is actually the case. I don't think Image Factory reports empty result, but should report an error actually

Unix4ever: But how would that help? There are no errors, but it's rather the version is not yet in the image factory.

Unix4ever: But the `Versions` call won't fail, no?
The problem is that the versions list doesn't have the new version of Talos, so the controller skips it.

smira: What should happen is that `Versions` would report v1.8.0 at some point, but fetching extensions would fail, and after some time once the release is finalized, the extensions will appear, but they should never be empty.

Unix4ever: It will, but the old version of controller will skip reconciliation and won't trigger an update until the instance gets restarted.
It won't react to the version list updates.

smira: I'm confused then... it should just fail and retry to fetch extensions later?

Probably I misunderstand the issue

smira: we shouldn't have a TalosVersion which is not in the Image Factory in the first place I guess..

Unix4ever: I guess this happens: 1.8.0 gets added to the list of versions in Omni, but it's not in the list of versions returned by the image Factory. Omni skips reconciliation and never comes back to check it again

smira: should we base Omni version on Image Factory versions? as if Image Factory doesn't report a version, anyways all operations would fail

Unix4ever: Well, yeah, that's possible.

Issue #666, opened 2024-10-03
[feature] add manifest properties on Windows binaries


brantgurga:
### Problem Description

Similar to https://github.com/siderolabs/talos/issues/9429 the Windows binary for omnictl is currently not filled with properties. This is mostly informational though I do know Windows Installer looks at the file versions if it were used to distribute binaries. The file version field seems like the most useful and informative field to fill in. Those properties are currently blank.
![image](https://github.com/user-attachments/assets/46749ec2-890c-4ad7-95dd-941ff9a01021)

### Solution

Something like go-winres (https://github.com/tc-hib/go-winres) looks like it can be used with the Go ecosystem to fill in that manifest information.

### Alternative Solutions

Investigation into tooling to support it also had led me to using windres from something like mingw to compile the manifest into a resource, but that's bringing in tooling outside the Go ecosystem so seems like a less favorable approach.

brantgurga:
I got this proof of concept working as illustrated. The executable has an icon which I left as the default one go-winres produced and some basic executable properties are filled in though they likely need reviewed and tweaked. Build pipeline integration likely needs tweaked too. I'll make a branch/pull request with what I did.
![image](https://github.com/user-attachments/assets/2288f9e4-241c-44a3-b3a2-44dce09da928)

brantgurga:
still applicable

brantgurga:
Still applicable.

brantgurga:
Still applicable.

PR #667, opened 2024-10-06
POC Windows Resources implementation

brantgurga:
POC draft for https://github.com/siderolabs/omni/issues/666

I'm sure there's tweaks from what I did on this.


brantgurga:
I got a branch of the kres repository and figured out the basic place to output the steps. Need to figure out if I need the --mount-type=cache arguments to RUN which were in my manually edited Dockerfile. Also need to use a tagged version of go-winres most likely instead of @latest and right now it puts the lines for all the targets which I think is okay since it creates objects that only get compiled in for Windows, but it's probably better to just output the lines when windows is the target

brantgurga:
The key file in kres seems to be internal\projects\golang\build.go.
Run make image-kres to build a local image. Copy the output local image and use in running kres on the omni folder.
MSYS_NO_PATHCONV=1 docker run --rm --net=host --user 197609:197609 -v /C/Users/brant/repositories/omni:/src -w /src -e GITHUB_TOKEN ghcr.io/brant/kres:34e72ac-dirty

it's smart enough to skip whatever GITHUB interaction without a GITHUB_TOKEN set. The MSYS_NO_PATHCONV=1 is needed since I'm currently doing this from MSYS but docker is not MSYS so it doesn't need to do path conversion to run and actually fails if run in the msys shell without that set.

utkuozdemir:
Thanks @brantgurga for the work.

We have decided that if we decide to tackle this, we would do it in https://github.com/siderolabs/kres, so all our other projects would also benefit from it. Besides, it is not a priority at the moment.

PR #671, opened 2024-10-07
fix: filter requests in the infra provision controller

Unix4ever:
Make each controller process only resources labeled with it's provider ID.
Allow overriding gRPC tunnel options for the machine classes/request sets.
Expose join configs to the infra providers.

Also publish Omni integration tests as the part of releases.


client/pkg/omni/resources/siderolink/connection_params_test.go
```diff
@@ -5,10 +5,14 @@
...
 
 import (
+	"bytes"
+	"errors"
+	"io"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"gopkg.in/yaml.v2"
```
smira, line 15: should be `.v3` (we shouldn't use .v2 anywhere)

client/pkg/omni/resources/siderolink/data/joinconfig.yaml
```diff
@@ -0,0 +1,13 @@
...
+apiUrl: '{{ .APIURL }}'
+---
+apiVersion: v1alpha1
+kind: EventSinkConfig
+endpoint: '{{ .EventSinkEndpoint }}'
+---
+apiVersion: v1alpha1
+kind: KmsgLogConfig
+name: omni-kmsg
+url: '{{ .LogEndpoint }}'
```
smira, line 13: this also works, but you can generate it programmatically as well to avoid drift and issues with YAML escaping

e.g. https://github.com/siderolabs/talos/blob/519a48302e771fd9b331913166d55c50fff4961a/pkg/machinery/config/types/siderolink/siderolink_test.go#L39-L40

if you combine them into a config Container and then encode the whole, you should always get a valid machine config

smira (approved):

client/pkg/omni/resources/siderolink/connection_params_test.go
Unix4ever: oh, yeah, that's automatic import, will fix

PR #675, opened 2024-10-08
fix: get proper IP from peer metadata

### LINKED ISSUE #668: Incorrect IP in audit logs
Omni sees incorrect IP in `peer.address` gRPC metadata. Which is set using `x-forwarded-for` and `x-real-ip` and should contain proper IP, but for some reason it's not.

DmitriyMV:
Currently, if gateway meets existing X-Forwarded-For header, it will append peer address that it sees to the existing value using comma. Our IP extraction function didn't account for that, and so it failed to parse IP and it used the original `peer.address` which set deep below in the gRPC middleware.

This commit ensures that we try to split the string value using `,`.

Closes #668

### DIFF
--- a/internal/pkg/grpcutil/logger.go
+++ b/internal/pkg/grpcutil/logger.go
@@ -8,7 +8,10 @@ package grpcutil
 import (
 	"context"
 	"fmt"
+	"iter"
 	"net/netip"
+	"strings"
+	"unicode/utf8"
 
 	"github.com/cosi-project/runtime/api/v1alpha1"
 	grpc_zap "github.com/grpc-ecosystem/go-grpc-middleware/logging/zap"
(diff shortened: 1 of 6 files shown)


internal/pkg/grpcutil/logger.go
```diff
@@ -131,15 +134,52 @@ func setRealIPAddress(ctx context.Context) {
 				continue
 			}
 
-			addr, err := netip.ParseAddr(peer[0])
-			if err != nil {
-				continue
+			for str := range splitString(peer[0], ",") {
```
Unix4ever, line 137: Why can't we just use this?
```suggestion
			for str := strings.Split(peer[0], ",") {
```

DmitriyMV: Just an optimization. String split returns a slice, which we don't need. We can delete this code, once Go 1.24 lands https://go-review.googlesource.com/c/go/+/587095

Unix4ever: lets put a TODO comment here then

DmitriyMV:
@TimJones we can't stick to `X-Real-IP` because it's not what grpc-gateway uses [internally](https://github.com/grpc-ecosystem/grpc-gateway/blob/8c8df189b24baa2133236c0f4fc4532c9a298e90/runtime/context.go#L198). Aka we will be forced to pass it manually in our headers. We can move from `X-Real-IP` to `X-Forwarded-For` tho.

@Unix4ever does this sounds good?

internal/pkg/grpcutil/logger.go
DmitriyMV: Okay, will do that.

Unix4ever (approved):

Issue #662, opened 2024-10-02
Auto provision machines for each machine set instead for the whole machine class


Unix4ever:
Rewrite the Provision Controller to create MachineRequestSets for each MachineSet and use the Talos version and extensions configuration from the Cluster/MachineSet.
This will help to get rid of the issues when the machine class is created with the Talos version different from the Cluster version.

Add MachineSet/Cluster label to the MachineRequest and MachineRequestStatus resources.
This way it will be possible to fetch all pending machine requests in the MachineSets.

PR #680, opened 2024-10-10
feat: display pending machine requests in the machine sets

Unix4ever:
The provision controller was changed: now it creates individual machine request sets for each machine set which is using the machine class with auto provision.
That allows to map machine requests to each machine set.

Talos version and initial extensions was dropped from the autoprovisioned machine classes as Talos version is now read from the Cluster to which the machine request is going to be added. Extensions support is dropped as there's no way to reliably set them without a Talos version.

Machine request set pressure resource was removed as the controller now reads the required machine count directly from the machine sets.

Infra provider controllers were updated to report human readable status.

Introduced the new resource type `ClusterMachineRequestStatus` that is used in the UI to show pending machine requests.

Closes: https://github.com/siderolabs/omni/issues/662


internal/backend/runtime/omni/controllers/omni/cluster_machine_request_status.go
```diff
@@ -0,0 +1,138 @@
...
+		qtransform.Settings[*infra.MachineRequest, *omni.ClusterMachineRequestStatus]{
+			Name: clusterMachineRequestStatusControllerName,
+			MapMetadataFunc: func(machineRequest *infra.MachineRequest) *omni.ClusterMachineRequestStatus {
+				return omni.NewClusterMachineRequestStatus(resources.DefaultNamespace, machineRequest.Metadata().ID())
+			},
+			UnmapMetadataFunc: func(clusterMachineStatus *omni.ClusterMachineRequestStatus) *infra.MachineRequest {
+				return infra.NewMachineRequest(clusterMachineStatus.Metadata().ID())
+			},
+			TransformFunc: func(ctx context.Context, r controller.Reader, _ *zap.Logger,
+				machineRequest *infra.MachineRequest, clusterMachineStatus *omni.ClusterMachineRequestStatus,
```
Unix4ever, line 44:
```suggestion
				machineRequest *infra.MachineRequest, clusterMachineRequestStatus *omni.ClusterMachineRequestStatus,
```

utkuozdemir (approved):

DmitriyMV (approved):

PR #681, opened 2024-10-10
chore: move from Codec to CodecV2

DmitriyMV:
This should be protocol backward compatible change (and breaking change in the source code) which should speed-up the process of gRPC data encoding / decoding. Also unify our CodecV2 usage.

For https://github.com/siderolabs/talos/issues/9404


Unix4ever (approved):

DmitriyMV:
> It looks like something doesn't play well with the emulator...

This should be fixed in https://github.com/siderolabs/crypto/pull/34 but lets see...

PR #696, opened 2024-10-22
fix: fetch Talos version from the image factory

Unix4ever:
The fix will help to avoid issues with Omni having Talos versions without extensions, as Talos versions appear in ghcr.io registry before the image factory picks them up.

Fixes: https://github.com/siderolabs/omni/issues/661

### DIFF
--- a/internal/backend/runtime/omni/controllers/omni/versions.go
+++ b/internal/backend/runtime/omni/controllers/omni/versions.go
@@ -15,7 +15,9 @@ import (
 	"github.com/blang/semver"
 	"github.com/cosi-project/runtime/pkg/controller"
 	"github.com/cosi-project/runtime/pkg/safe"
+	"github.com/cosi-project/runtime/pkg/state"
 	"github.com/siderolabs/gen/xslices"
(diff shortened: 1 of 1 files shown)


internal/backend/runtime/omni/controllers/omni/versions.go
```diff
@@ -113,20 +114,26 @@ var allowedPreVersionStrings = map[string]struct{}{
...
+	return registry.UpgradeCandidates(ctx, source)
+}
+
+func (ctrl *VersionsController) fetchTalosVersions(ctx context.Context, factoryURL string) ([]string, error) {
+	client, err := client.New(factoryURL)
 	if err != nil {
 		return nil, err
 	}
 
+	return client.Versions(ctx)
```
frezbo, line 127: do we also need to fetch this periodically or that happens already?

frezbo (approved):

internal/backend/runtime/omni/controllers/omni/versions.go
Unix4ever: that happens already, we just change the source here

PR #701, opened 2024-10-23
chore: do not audit log `GET` requests to k8s

DmitriyMV:
Since we do not audit log read-like requests, I think it makes sense to do the same here.


utkuozdemir (approved):

smira:
I don't think we should log the request body, as it might be HUGE, and I don't think that level of details should go to audit logs. The fact that a user X accessed Kubernetes API at time Y for cluster Z is enough. 

API server has its own set of audit logs which provides additional (configurable) details.

smira:
If we don't log the request, we can drop half of the complexity there

DmitriyMV:
@smira 

> If we don't log the request, we can drop half of the complexity there

Done.

smira (approved):

smira:
@Unix4ever we should get it backported I guess

Issue #419, opened 2024-06-26
Run talosctl reset for machines removed from Omni


Unix4ever:
Destroy Talos on the node if the API is reachable.

`Remove Machine` should try calling `talosctl reset` without options for the removed machine before destroying the link.

PR #703, opened 2024-10-24
feat: reset machine when it's removed from Omni

Unix4ever:
The reset removes Talos from the machine completely. Fixes: https://github.com/siderolabs/omni/issues/419

### DIFF
--- a/internal/backend/runtime/omni/controllers/omni/machine_teardown.go
+++ b/internal/backend/runtime/omni/controllers/omni/machine_teardown.go
@@ -1,0 +1,194 @@
+// Copyright (c) 2024 Sidero Labs, Inc.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+
+package omni
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/cosi-project/runtime/pkg/controller"
+	"github.com/cosi-project/runtime/pkg/controller/generic"
+	"github.com/cosi-project/runtime/pkg/resource"
+	"github.com/cosi-project/runtime/pkg/safe"
+	"github.com/cosi-project/runtime/pkg/state"
+	"github.com/siderolabs/gen/optional"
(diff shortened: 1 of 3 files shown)


DmitriyMV (approved):

internal/backend/runtime/omni/controllers/omni/machine_teardown.go
```diff
@@ -0,0 +1,194 @@
...
+				Kind:      controller.InputQMapped,
+			},
+		},
+		Concurrency: optional.Some[uint](4),
+	}
+}
+
+// MapInput implements controller.QController interface.
+func (ctrl *MachineTeardownController) MapInput(ctx context.Context, _ *zap.Logger, r controller.QRuntime, ptr resource.Pointer) ([]resource.Pointer, error) {
+	if ptr.Type() == omni.TalosConfigType {
```
smira, line 66: this is a bit confusing, if it's a one-shot operation, why do we want to reconcile on TalosConfig change?

internal/backend/runtime/omni/controllers/omni/machine_teardown.go
```diff
@@ -0,0 +1,194 @@
...
+				opts,
+				client.WithTLSConfig(insecureTLSConfig),
+				client.WithEndpoints(address),
+			)...)
+	}
+
+	talosConfig, err := safe.ReaderGet[*omni.TalosConfig](ctx, r, omni.NewTalosConfig(resources.DefaultNamespace, clusterName).Metadata())
+	if err != nil {
+		if state.IsNotFoundError(err) {
+			return nil, xerrors.NewTaggedf[qtransform.SkipReconcileTag]("cluster '%s' talosconfig not found: %w", clusterName, err)
```
smira, line 173: and if it hasn't been found... should the machine just be removed? if there's no talosconfig, there should be no Talos on it? or it's in maintenance mode?

Issue #620, opened 2024-09-10
[feature] Automatically resolve node part of a cluster


rothgar:
### Problem Description

When querying nodes in an omni instance it would be nice to be able to globally query it (assuming the name is unique) no matter what cluster it is in.

```
talosctl services -n my_node
```
Should work even if the node is part of a cluster and I don't specify the cluster at the CLI.

Currently I get an error that is unclear why it happens
```
rpc error: code = Unavailable desc = last connection error: connection error: desc = "transport: Error while dialing: dial tcp: lookup mini on 192.168.128.10:53: no such host"
```
but if I add `--cluster my_cluster` it works

### Solution

If there is only one machine with a specific name return that machine with `talosctl` commands.

### Alternative Solutions

We could alternatively make the error message more clear that the user needs to specify a `--cluster` for querying nodes.

smira:
> We could alternatively make the error message more clear that the user needs to specify a --cluster for querying nodes.

I think we should do this asap, as it's clearly a very confusing behavior right now.

PR #704, opened 2024-10-24
feat: automatically resolve cluster in `talosctl` calls

Unix4ever:
`talosctl --cluster` flag is now optional, Omni will automatically resolve the cluster if the machine is a part of one.

Fixes: https://github.com/siderolabs/omni/issues/620

### DIFF
--- a/internal/backend/dns/service.go
+++ b/internal/backend/dns/service.go
@@ -10,6 +10,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"slices"
 	"sync"
 
 	"github.com/cosi-project/runtime/pkg/resource"
@@ -34,6 +35,8 @@ type Info struct {
 
 	address            string
 	managementEndpoint string
+
+	Ambiguous bool
 }
 
 // NewInfo exports unexported.
@@ -55,6 +58,34 @@ func (i Info) GetAddress() string {
 	return i.managementEndpoint
 }
 
+type resolverMap map[string][]resource.ID
+
+func (m resolverMap) get(key string) []resource.ID {
+	return m[key]
+}
+
+func (m resolverMap) add(key string, id resource.ID) {
+	if slices.Index(m[key], id) != -1 {
+		return
+	}
+
+	m[key] = append(m[key], id)
+}
+
+func (m resolverMap) remove(key string, id resource.ID) {
+	ids, ok := m[key]
+	if !ok {
+		return
+	}
+
+	ids = slices.DeleteFunc(ids, func(item string) bool { return item == id })
+	if len(ids) == 0 {
+		delete(m, key)
+	} else {
+		m[key] = ids
+	}
+}
+
 // Service is the DNS service.
 type Service struct {
 	omniState state.State
@@ -62,6 +93,8 @@ type Service struct {
 
 	recordToMachineID map[record]resource.ID
 	machineIDToInfo   map[resource.ID]Info
+	addressToID       resolverMap
+	nodenameToID      resolverMap
 
 	lock sync.Mutex
 }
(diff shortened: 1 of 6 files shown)


internal/backend/grpc/router/router.go
```diff
@@ -173,10 +173,28 @@ func (r *Router) Director(ctx context.Context, fullMethodName string) (proxy.Mod
 func (r *Router) getTalosBackend(ctx context.Context, md metadata.MD) ([]proxy.Backend, error) {
 	clusterName := getClusterName(md)
 
+	nodeID := getNodeID(md)
+
 	id := fmt.Sprintf("cluster-%s", clusterName)
 
 	if clusterName == "" {
-		id = fmt.Sprintf("machine-%s", getNodeID(md))
+		machineSetNode, err := safe.ReaderGetByID[*omni.MachineSetNode](actor.MarkContextAsInternalActor(ctx), r.cosiState, nodeID)
```
smira, line 181: should we add an integration test for all the ways we have right now?

internal/backend/grpc/router/router.go
```diff
@@ -173,10 +173,28 @@ func (r *Router) Director(ctx context.Context, fullMethodName string) (proxy.Mod
 func (r *Router) getTalosBackend(ctx context.Context, md metadata.MD) ([]proxy.Backend, error) {
 	clusterName := getClusterName(md)
 
+	nodeID := getNodeID(md)
```
smira, line 176: it's not clear, does it work only with machine UUIDs?

internal/backend/grpc/router/router.go
smira: we already had one, it just might need to be updated

internal/backend/grpc/router/router.go
```diff
@@ -173,10 +173,28 @@ func (r *Router) Director(ctx context.Context, fullMethodName string) (proxy.Mod
...
+			var ok bool
+
+			clusterName, ok = machineSetNode.Metadata().Labels().Get(omni.LabelCluster)
+			if !ok {
+				return nil, fmt.Errorf("failed to resolve machine cluster relation: no cluster name in the machine set node %q", id)
+			}
+
+			id = fmt.Sprintf("cluster-%s", clusterName)
+		} else {
+			id = fmt.Sprintf("machine-%s", nodeID)
```
DmitriyMV, line 196:
```suggestion
			id ="machine-" + nodeID
```

Minor nit: perhaps this is clearer?

internal/backend/grpc/router/router.go
Unix4ever: that's a great question. I'm afraid that with the hostname the resolution might be ambiguous.
I think I can make it work: we can show the error if the resolution gives several clusters.

internal/backend/grpc/router/router.go
Unix4ever: yeah, will do that

internal/backend/dns/service.go
```diff
@@ -55,13 +58,41 @@ func (i Info) GetAddress() string {
...
+func (m resolverMap) remove(key string, id resource.ID) {
+	ids, ok := m[key]
+	if !ok {
+		return
+	}
+
+	ids = slices.DeleteFunc(ids, func(item string) bool { return item == id })
+	if len(ids) == 0 {
+		delete(m, key)
+	}
```
DmitriyMV, line 84:
```suggestion
	if len(ids) == 0 {
		delete(m, key)
	} else {
		m[key] = ids	
	}
```

DmitriyMV (approved):

internal/backend/dns/service.go
Unix4ever: good catch!

PR #707, opened 2024-10-25
feat: add resources and controllers for bare metal infra provider

utkuozdemir:
Introduce the concept of "static" infra providers, e.g., bare-metal infra provider, which manage a static set of machines contrary to the "regular" infra providers.

Add the following resources:

- `infra.Machine`: similar to `MachineRequest`, lives in the `infra-provider` namespace, serving as the input of the owning static provider. It is created in the `MachineController` if there is a SideroLink connection with the static provider ID. Regular flow of `Machine` creation is blocked, until this `infra.Machine` is accepted.

- `infra.MachineStatus`: similar to `MachineRequestStatus`, lives in the `infra-provider` namespace, serving as the output of the owning static provider. Its lifecycle must be bound to the corresponding `infra.Machine`.

- `infra.MachineState`: a resource that is supposed to be shared by Omni and bare-metal provider bi-directionally - they both can read from and write to it. It is currently used to mark the machine as installed when we observe an installation (through `SequenceEvent`s in the event sink), and to mark it as non-installed after we wipe it in the provider.

- `omni.InfraMachineConfig`: a user-managed resource to mark the `infra.Machine`s as accepted or set their desired power state. The acceptance information is then propagated to the `infra.Machine` resource. A machine which was already accepted cannot be unaccepted (checked by a validation), and this resource can only be removed when the `siderolink.Link` for the matching infra machine is removed.


internal/backend/runtime/omni/controllers/omni/infra_machine.go
```diff
@@ -0,0 +1,132 @@
...
+					infraMachine.TypedSpec().Value.TalosVersion = clusterMachineConfigStatus.TypedSpec().Value.TalosVersion
+					infraMachine.TypedSpec().Value.Allocated = true
+
+					if clusterMachineConfigStatus.TypedSpec().Value.LastConfigError == "" { // we observed that a config was successfully applied at least once - assume installed
+						infraMachine.TypedSpec().Value.Installed = true // todo: probably not reliable
+					}
+				} else {
+					infraMachine.TypedSpec().Value.TalosVersion = ""
+					infraMachine.TypedSpec().Value.Allocated = false
+					infraMachine.TypedSpec().Value.Installed = false // todo: probably not reliable, how do we actually know if it's installed or not?
```
utkuozdemir, line 77: To the reviewer: I need some good ideas here. I basically want to reliably detect if Talos is installed on the disk or not. This affects the wipe decision on the provider.

client/pkg/omni/resources/infra/machine_status.go
```diff
@@ -0,0 +1,58 @@
...
+}
+
+const (
+	// InfraMachineStatusType is the type of MachineStatus resource.
+	//
+	// tsgen:InfraMachineStatusType
+	InfraMachineStatusType = resource.Type("InfraMachineStatuses.omni.sidero.dev")
+)
+
+// MachineStatus resource describes a machine request.
```
Unix4ever, line 32: Probably the description is off here.

client/pkg/omni/resources/infra/machine.go
```diff
@@ -0,0 +1,74 @@
...
+}
+
+const (
+	// InfraMachineType is the type of Machine resource.
+	//
+	// tsgen:InfraMachineType
+	InfraMachineType = resource.Type("InfraMachines.omni.sidero.dev")
+)
+
+// Machine resource describes a machine request.
```
Unix4ever, line 32: Probably the description is off here.

client/api/omni/specs/omni.proto
```diff
@@ -1251,3 +1251,15 @@ message ClusterMachineRequestStatusSpec {
...
+message InfraMachineConfigSpec {
+  enum MachinePowerState {
+    POWER_STATE_DEFAULT = 0;
+    POWER_STATE_OFF = 1;
+    POWER_STATE_ON = 2;
+  }
+
+  MachinePowerState power_state = 1;
+  bool accepted = 2;
+}
```
Unix4ever, line 1265: Looks like a duplicate of `InfraMachineStatusSpec`, maybe we can use the same spec for both resources?

internal/backend/runtime/omni/controllers/omni/infra_machine.go
```diff
@@ -0,0 +1,132 @@
...
+					return err
+				}
+
+				if clusterMachineConfigStatus != nil {
+					infraMachine.TypedSpec().Value.TalosVersion = clusterMachineConfigStatus.TypedSpec().Value.TalosVersion
+					infraMachine.TypedSpec().Value.Allocated = true
+
+					if clusterMachineConfigStatus.TypedSpec().Value.LastConfigError == "" { // we observed that a config was successfully applied at least once - assume installed
+						infraMachine.TypedSpec().Value.Installed = true // todo: probably not reliable
+					}
```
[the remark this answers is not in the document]
utkuozdemir: I think you mean `!=`, right?

internal/backend/runtime/omni/controllers/omni/machine.go
```diff
@@ -41,19 +45,43 @@ func NewMachineController() *MachineController {
...
+								"the link is created by the infra provider, but doesn't have the machine request label yet",
+							)
+						}
+
+						machineRequestSetID, ok = link.Metadata().Labels().Get(omni.LabelMachineRequestSet)
+						if !ok {
+							return xerrors.NewTaggedf[qtransform.SkipReconcileTag](
+								"the link is created by the infra provider, but doesn't have the machine request label yet",
+							)
+						}
```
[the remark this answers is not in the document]
utkuozdemir: done, refactored this controller a bit.

client/api/omni/specs/omni.proto
utkuozdemir: There are some differences: 
- the zero value in this one (`Config`) is `DEFAULT`, while in the `Status` spec it is `UNKNOWN`, and logically they represent different things.
- in the Config, there's the `accepted` field, in Status, there's `readyToUse` field, again, semantically different.

I can imagine them diverging even more soon, so I'd rather keep them this way for now, if that's not an issue.

internal/backend/runtime/omni/controllers/omni/machine.go
```diff
@@ -38,57 +42,116 @@ func NewMachineController() *MachineController {
...
+	if err != nil {
+		return err
+	}
+
+	if _, isStaticProvider := providerStatus.Metadata().Labels().Get(omni.LabelIsStaticInfraProvider); isStaticProvider { // static infra provider flow, e.g., bare-metal
+		return h.handleStaticInfraProvider(ctx, r, machine)
+	}
+
+	// provisioning (regular) infra provider flow, e.g., kubevirt
+	return h.handleProvisioningInfraProvider(ctx, r, link, machine)
```
utkuozdemir, line 106: Other than this branching here, the logic in this file is totally unchanged - just reworked for readability.

internal/backend/runtime/omni/controllers/omni/machine_status.go
```diff
@@ -249,7 +249,7 @@ func (ctrl *MachineStatusController) reconcileRunning(ctx context.Context, r con
 
 		spec.Connected = connected
 
-		helpers.CopyLabels(machine, m, omni.LabelMachineRequest, omni.LabelMachineRequestSet, omni.LabelNoManualAllocation)
+		helpers.CopyLabels(machine, m, omni.LabelMachineRequest, omni.LabelMachineRequestSet, omni.LabelNoManualAllocation, omni.LabelIsManagedByStaticInfraProvider)
```
utkuozdemir, line 252: Propagate this label  from Machine to MachineStatus. We then check this label on `MachineStatus` in `ClusterMachineConfigStatusController` to find out if the machine is being prepared by the static infra provider (still not ready to be used, e.g., being wiped), and if it is, skip applying config to it.

internal/backend/runtime/omni/controllers/omni/machine.go
```diff
@@ -38,57 +42,116 @@ func NewMachineController() *MachineController {
...
+		)
+	}
+
+	return nil
+}
+
+// handleProvisioningInfraProvider checks if the machine managed by a provisioning infra provider has the machine request and machine request set labels set.
+//
+// It then matches the machine request set owner to the machine provision controller, and if it matches, sets the no manual allocation label on the machine.
+func (h *machineControllerHelper) handleProvisioningInfraProvider(ctx context.Context, r controller.Reader, link *siderolink.Link, machine *omni.Machine) error {
```
utkuozdemir, line 132: better naming suggestions are welcome - my thought process here was: we have 2 distinct types of infra providers: "**static**", e.g., bare-metal, and "**provisioning**", e.g. kubevirt, aws etc.

internal/pkg/machinestatus/handler.go
utkuozdemir, line 1: For this to not be confusing: I simply this from `machinestatus.Handler` to be `machineevent.Handler`, as it now also handles `SequenceEvent`s. GitHub couldn't figure out that it was moved and also modified afterwards.

internal/pkg/machineevent/handler.go
utkuozdemir, line 1: See: https://github.com/siderolabs/omni/pull/707/files#r1857434405

internal/pkg/machineevent/handler.go
```diff
@@ -0,0 +1,180 @@
...
+		return fmt.Errorf("no machines found for address %s", ip)
+	}
+
+	machineID := machines.Get(0).Metadata().ID()
+
+	switch event := event.Payload.(type) {
+	case *machineapi.MachineStatusEvent:
+		return handler.handleMachineStatusEvent(ctx, event, machineID)
+	case *machineapi.SequenceEvent:
+		return handler.handleSequenceEvent(ctx, event, machineID)
```
utkuozdemir, line 76: Added this, otherwise no changes.

client/api/omni/specs/infra.proto
```diff
@@ -29,6 +29,38 @@ message MachineRequestStatusSpec {
   string status = 4;
 }
 
+message InfraMachineSpec {
+  enum MachinePowerState {
+    POWER_STATE_OFF = 0;
```
smira, line 34: should it also be `unset` if the user doesn't want anything? (i.e. follow the defaults?)

utkuozdemir: User's preference is here: https://github.com/siderolabs/omni/pull/707/files/d6063fcdb22fd49f859ba01bcae61a91b5f39343#diff-64ee15284d47f2291bfba2e406d10cb1aaf16ffc27e1af89008a114b58304c4eR1260-R1266

This is the final verdict, taking the user's preference and platform default (later to be defined in a resource). Not sure if it makes sense to add `unset` here.

smira: ah, ok, I was confused `preferred_power_state` name

go.mod
```diff
@@ -3,6 +3,9 @@ module github.com/siderolabs/omni
 go 1.23.3
 
 replace (
+	// todo: remove when merged
+	github.com/siderolabs/omni-infra-provider-bare-metal v0.0.0-20241014092637-c3e892b87244 => github.com/utkuozdemir/sidero-omni-infra-provider-bare-metal v0.0.0-20241021213554-cc93e594a0d2
```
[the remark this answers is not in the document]
utkuozdemir: That's a good point, and actually it was unused. The initial idea and the reason was to import the GRPC server API types that the provider serves, so that Omni could send commands to the machines (e.g. Reboot) over the reverse GRPC tunnel, `Omni->Provider->Machine`.

I initially did that, but then removed to keep the scope smaller in this PR.

So I now removed this import, and also decided to remove the whole reverse tunnel logic. I'll keep that in a separate Draft PR for now, as I'm having second thoughts on whether it is necessary at all. It would only be used for reboot for the time being, and imo it is perfectly doable over the resources at hand as well (similar to wipeID, something like rebootID).

So I'll discuss this with @Unix4ever when he is back :)

Issue #705, opened 2024-10-24
create a controller for handling machine request set which is created for managed control planes


Unix4ever:
Implement the controller that will create machine request sets for the clusters which have managed control planes enabled.
Each cluster will have it's own machine request set with 3 requests.
Add `omni.sidero.dev/managed` label to the `Cluster`.

Open questions:
- should we propagate the label to the `MachineSet`?
- should the `MachineSet` be created by the controller?

PR #723, opened 2024-11-04
feat: implement managed control planes

Unix4ever:
Cluster template now has
```
managed:
  enable: true|false
```
configuration in the control planes model.
Creating managed control planes will be possible only if the Omni account is configured with the provider to use for managed control planes, otherwise resource validation won't allow creating the machine set with such configuration.

Also drop `MachineRequestSet` allocation mode from the machine set, remove source there, as it doesn't make much sense: for managed control planes it should always use the machine request set name with the same name as the machine set.

Config patches for managed mode are going to be added in the follow-up PRs.

Fixes: https://github.com/siderolabs/omni/issues/705


client/pkg/template/internal/models/controlplane.go
```diff
@@ -19,6 +20,12 @@ const KindControlPlane = "ControlPlane"
 // ControlPlane describes Cluster controlplane nodes.
 type ControlPlane struct {
 	MachineSet `yaml:",inline"`
+	Managed    Managed `yaml:"managed,omitempty"`
+}
+
+// Managed describes managed control planes mode.
+type Managed struct {
+	Enable bool `yaml:"enable"`
```
smira, line 28: I don't quite see how managed controlplane can be in a non-managed cluster ?

Unix4ever: probably we should name it differently: should indicate that the machines are allocated from our infra.

internal/pkg/config/config.go
```diff
@@ -110,6 +112,13 @@ type InitialServiceAccount struct {
 	Enabled  bool
 }
 
+// ManagedControlPlanes specifies the configuration for the provider used in managed control planes.
+type ManagedControlPlanes struct {
+	ProviderID   string `yaml:"providerID" json:"providerID"`
+	ProviderData string `yaml:"providerData" json:"providerData"`
+	Enable       bool   `yaml:"enable" json:"enable"`
```
Unix4ever, line 119:
```suggestion
	ProviderID   string `yaml:"providerID"`
	ProviderData string `yaml:"providerData"`
	Enable       bool   `yaml:"enable"`
```

cmd/omni/cmd/cmd.go
```diff
@@ -593,5 +593,26 @@ var initOnce = sync.OnceValue(func() *cobra.Command {
 		"the lifetime duration of the initial service account key",
 	)
 
+	rootCmd.Flags().BoolVar(
+		&config.Config.ManagedControlPlanes.Enable,
+		"managed-control-planes-enable",
```
Unix4ever, line 598:
```suggestion
		"experimental-managed-control-planes-enable",
```

Unix4ever: Hide the flag.
And show the big warning on Omni startup.

utkuozdemir:
not stale

Unix4ever:
Closing this for now. Will get back to it when we return to the managed control planes idea.

Issue #728, opened 2024-11-05
support setting machine patches in the infra provider


Unix4ever:
Provide a way to set machine patches in the infrastructure provider.
Infrastructure provider can create patches in the `infra-provider` namespace, to which it has access to.
Then some controller can copy them to the main namespace.

The patch requests must be labeled with the infra provider id and system patch labels.
Validation should block creating patches without the infra provider label or if the id doesn't match the provider id which tries to create the patch.

The controller should verify that the machine UUID specified in the patch really belongs to the infra provider and only after that create the patch.

PR #729, opened 2024-11-05
feat: support creating config patches in the infrastructure providers

Unix4ever:
Patches can be created for a single machine in the machine provision flow: the provider can call `CreateConfigPatch` method at any point. This will create a `ConfigPatchRequest` resource which will be turned into a `ConfigPatch` after the `MachineRequestStatus` UUID gets populated.

Fixes: https://github.com/siderolabs/omni/issues/728

### DIFF
--- a/cmd/integration-test/pkg/tests/stats.go
+++ b/cmd/integration-test/pkg/tests/stats.go
@@ -35,7 +35,7 @@ func AssertStatsLimits(testCtx context.Context) TestFunc {
 			{
 				name:  "resource CRUD",
 				query: `sum(omni_resource_operations_total{operation=~"create|update", type!="MachineStatusLinks.omni.sidero.dev"})`,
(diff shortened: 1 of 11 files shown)


cmd/integration-test/pkg/tests/stats.go
```diff
@@ -45,7 +45,7 @@ func AssertStatsLimits(testCtx context.Context) TestFunc {
 			{
 				name:  "controller wakeups",
 				query: `sum(omni_runtime_controller_wakeups{controller!="MachineStatusLinkController"})`,
-				check: func(assert *assert.Assertions, value float64) { assert.Less(value, float64(10000)) },
+				check: func(assert *assert.Assertions, value float64) { assert.Less(value, float64(11000)) },
```
Unix4ever, line 48: I've checked the stats, it looks like we've reached `10022` gradually, before it was `9690`.
So I've bumped the limits here.

smira (approved):

PR #733, opened 2024-11-12
fix: properly map config patch requests in the infra provision ctrl

Unix4ever:
The bug was causing missing reconciliations.

### DIFF
--- a/client/pkg/infra/controllers/provision.go
+++ b/client/pkg/infra/controllers/provision.go
@@ -106,16 +106,41 @@ func (ctrl *ProvisionController[T]) Settings() controller.QSettings {
 }
 
 // MapInput implements controller.QController interface.
-func (ctrl *ProvisionController[T]) MapInput(_ context.Context, _ *zap.Logger,
-	_ controller.QRuntime, ptr resource.Pointer,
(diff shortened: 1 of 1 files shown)


client/pkg/infra/controllers/provision.go
```diff
@@ -106,15 +106,34 @@ func (ctrl *ProvisionController[T]) Settings() controller.QSettings {
...
-func (ctrl *ProvisionController[T]) MapInput(_ context.Context, _ *zap.Logger,
-	_ controller.QRuntime, ptr resource.Pointer,
+func (ctrl *ProvisionController[T]) MapInput(ctx context.Context, _ *zap.Logger,
+	r controller.QRuntime, ptr resource.Pointer,
 ) ([]resource.Pointer, error) {
 	if ptr.Type() == siderolink.ConnectionParamsType {
 		return nil, nil
 	}
 
+	id := ptr.ID()
```
smira, line 116: should it be more explicit about types of the inputs here?

as it's not clear what happens in the flow if the type is ConfigPatchRequestType?

might be `switch ptr.Type() {` ?

Unix4ever: ok, let's change it to switch :)

smira (approved):

Issue #711, opened 2024-10-27
Omni : problem to use an OnPrem image-factory


flpajany:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

I have installed an image-factory (v0.5.0) in a docker running on my machine where omni OnPrem is running too (app01-lvn-re.phys.rece).
Here is the command I execute :

```
docker run \
-d \
--net=host \
--cap-add=NET_ADMIN \
--device /dev/net/tun \
--name omni \
--restart unless-stopped \
-v /root/omni/etcd:/_out/etcd \
-v /root/omni/tls.crt:/tls.crt \
-v /root/omni/tls.key:/tls.key \
-v /root/omni/omni.asc:/omni.asc \
-v /root/omni/descriptor.xml:/saml-descriptor \
-v /root/omni/certs:/etc/ssl/certs \
hub.valentine.sfr.com/proxy-ghcr-io/siderolabs/omni:v0.42.3 \
--account-id=$(cat /root/omni/omni-account-uuid) \
--name=onprem-omni \
--enable-break-glass-configs \
--private-key-source=file:///omni.asc \
--event-sink-port=8091 \
--cert=/tls.crt \
--key=/tls.key \
--machine-api-cert=/tls.crt \
--machine-api-key=/tls.key \
--bind-addr=0.0.0.0:443 \
--machine-api-bind-addr=0.0.0.0:8090 \
--k8s-proxy-bind-addr=0.0.0.0:8100 \
--advertised-api-url=https://omni-test.pack.valentine.sfr.com/ \
--siderolink-api-advertised-url=https://omni-test.pack.valentine.sfr.com:8090/ \
--siderolink-wireguard-advertised-addr=10.144.18.178:50180 \
--advertised-kubernetes-proxy-url=https://omni-test.pack.valentine.sfr.com:8100/ \
--auth-saml-enabled=true \
--talos-installer-registry=app01-lvn-re.phys.rece:5005/siderolabs/installer \
--image-factory-pxe-address=https://factory-talos-test.pack.valentine.sfr.com/ \
--image-factory-address=https://factory-talos-test.pack.valentine.sfr.com/ \
--auth-saml-metadata=/saml-descriptor

```
But when I launch an upgrade for a machine, in the logs, I found this line :

```
[...]

PR #744, opened 2024-11-21
fix: use the custom image factory host in backend and frontend

utkuozdemir:
Contains two fixes:

- In the frontend, `factory.talos.dev` was hardcoded for the maintenance upgrade calls. Move the image factory URL into the `FeaturesConfig` resource, and switch to reading the host from this resource on the frontend.

- In the backend, we were initializing the `imageFactoryHost` in an `init` block, and because of init, it was running before the user override was actually written into the config by Cobra. Fix it by initializing the `imageFactoryHost` lazily when it is used for the first time.

Fixes siderolabs/omni#711.


internal/backend/runtime/omni/controllers/omni/cluster_machine_config.go
```diff
@@ -432,16 +433,14 @@ func stripTalosAPIAccessOSAdminRole(cfg config.Provider) (config.Provider, error
 	return container.New(updatedDocs...)
 }
 
-var imageFactoryHost string
-
-func init() {
+var getImageFactoryHost = sync.OnceValues(func() (string, error) {
```
utkuozdemir, line 436: Backend fix

frontend/src/methods/machine.ts
```diff
@@ -151,8 +152,12 @@ const copyUserLabels = (src: Resource, dst: Resource) => {
 }
 
 export const updateTalosMaintenance = async (machine: string, talosVersion: string, schematic?: string) => {
+  const imageFactoryBaseURL = await getImageFactoryBaseURL();
+
+  const host = new URL(imageFactoryBaseURL).host;
+
   const image = schematic ?
-    `factory.talos.dev/installer/${schematic}:v${talosVersion}` :
+    `${host}/installer/${schematic}:v${talosVersion}` :
```
utkuozdemir, line 160: Frontend fix

frontend/src/methods/machine.ts
```diff
@@ -151,8 +152,12 @@ const copyUserLabels = (src: Resource, dst: Resource) => {
...
     `ghcr.io/siderolabs/installer:v${talosVersion}`;
```
frezbo, line 161: i guess this is also hardcoded in a way :sweat_smile:

frezbo (approved):

smira (approved):

frontend/src/methods/machine.ts
utkuozdemir: Yep, but probably much less of a problem than the other. Can live with it for now I think.

Issue #660, opened 2024-10-02
Bare Metal Infrastructure Provide Phase 1


smira:

utkuozdemir:
Closing, the remaining work is tracked in #757.

PR #746, opened 2024-11-26
feat: add GRPC reverse tunnel server

utkuozdemir:
Allow Omni API clients to receive GRPC calls in the reverse direction, `Omni->Client`. Might be used by the infra providers, e.g., bare-metal infra provider.

Possibly related to #660.


go.mod
```diff
@@ -3,6 +3,9 @@ module github.com/siderolabs/omni
 go 1.23.3
 
 replace (
+	// todo: remove when merged
+	github.com/siderolabs/omni-infra-provider-bare-metal v0.0.0-20241014092637-c3e892b87244 => github.com/utkuozdemir/sidero-omni-infra-provider-bare-metal v0.0.0-20241021213554-cc93e594a0d2
```
smira, line 7: I think if we need an API, it should be common for all providers, and should either live in Omni or in a separate repo.

my 2 cents

utkuozdemir: Yes you are right, I'll anyway keep this as draft until we make a decision.

PR #753, opened 2024-12-03
feat: add Turing RK1 SoM to SBCs dropdown

nberlee:
This has hardcoded "turingrk1" strings as discussed in https://github.com/siderolabs/talos/pull/9869

### DIFF
--- a/internal/backend/runtime/omni/controllers/omni/installation_media.go
+++ b/internal/backend/runtime/omni/controllers/omni/installation_media.go
@@ -343,6 +343,14 @@ var installationMedia = []installationMediaSpec{
 		SBC:          true,
 		ContentType:  "application/x-xz",
 	},
+	{
+		Name:         "Turing RK1",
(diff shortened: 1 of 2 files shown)


Unix4ever:
Do we support this board for Talos 1.7 and 1.8?
Asking that because we might need to introduce a new check, which will disable the SBC for the old versions of Talos.

nberlee:
No, its it needs kernel 6.12. So starting from 1.9

Unix4ever:
I will add a couple changes to your PR to support limiting the min supported Talos version for some configurations.

Unix4ever (approved):

Unix4ever:
Ok, updated. Now Turing RK1 will only be available for Talos 1.9.x+.

Screenshot 2024-12-10 at 22 45 15

nberlee:
Ty @Unix4ever

Unix4ever:
Thank you for the PR :)

Issue #197, opened 2024-04-30
[feature] User management sync between `omnictl` and Omni's web UI


ArcherSeven:
### Problem Description

Users permissions in the Web UI seem largely divorced from their permissions in the CLI, both in practice, and in how they are displayed.

### Solution

"Role" should mean the same thing in the UI and in the CLI, and a users full permissions in both contexts should be visible from either context, and ideally editable from either context.

### Alternative Solutions

The very hard to understand situation that currently exists is I guess the alternative.

### Notes

Assigning RBAC to users / groups / roles today is very convoluted and not the simplest to test, especially with SAML enabled.  It would be extremely nice to help mitigate that by at least having terms mean the same thing in the CLI and in the UI, and be able to trivially view them in both places.

https://github.com/siderolabs/omni-feedback/issues/30

Unix4ever:
Implement user management methods in the CLI:
- user add
- user remove
- user list

PR #763, opened 2024-12-09
feat: implement CLI commands for user management

Unix4ever:
Add the UI for showing and editing service accounts.

Added CLI commands:
```
omnictl user list
omnictl user create <email> --role <role>
omnictl user delete <email>
omnictl user set-role --role <role>
```

Fixes: https://github.com/siderolabs/omni/issues/197


Unix4ever:
Screenshot 2024-12-09 at 18 44 03
Screenshot 2024-12-09 at 18 44 17

client/pkg/omnictl/user/set_role.go
```diff
@@ -0,0 +1,63 @@
...
+}
+
+func setUserRole(email string) func(ctx context.Context, client *client.Client) error {
+	return func(ctx context.Context, client *client.Client) error {
+		identity, err := safe.ReaderGetByID[*auth.Identity](ctx, client.Omni().State(), email)
+		if err != nil {
+			return err
+		}
+
+		_, err = safe.StateUpdateWithConflicts(ctx, client.Omni().State(),
```
smira, line 42: the commands seem to be untested (?)

even adding a simple integration test to create/delete/update a user would ensure the commands are not broken in the future

PR #778, opened 2024-12-11
release(v0.44.1): prepare release

Unix4ever:
This is the official v0.44.1 release.


smira:
Do we backport fixes skipping PR process?

frezbo (approved):
shouldn't the `gendata/tag` should also be updated?

smira (approved):

Unix4ever:
> shouldn't the gendata/tag should also be updated?

Thanks, good catch! I updated it but forgot to do `git add`

frezbo (approved):

Issue #797, opened 2024-12-20
Do not preserve `metal-agent` extension when installing Talos to disk


utkuozdemir:
Problem:
- Provider boots Talos into agent mode
- User accepts machine on Omni
- Omni creates a `MachineStatus` with the list of extensions and schematic ID in it, including all firmware extensions and `metal-agent` extension
- User adds the machine to a cluster
- Provider PXE boots Talos
- Before config is pushed to install Talos to disk, Omni 
  - either misses to update set of extensions in `MachineStatus`
  - or something else happens (?)
- Talos gets installed to the disk with the `metal-agent`
- It is a broken Talos installation

We have a temporary fix for it, but need to find the right solution.

utkuozdemir:
The current fix addresses the issue atm. If we face another issue related to that in the future, we can open a new ticket and see what we can do.

PR #799, opened 2024-12-20
fix: do not preserve extensions on Talos agent mode

utkuozdemir:
If Talos is in agent mode, we should avoid preserving the list of extensions for the actual Talos installation, as the `metal-agent` extension breaks the Talos installation.

Related to/workaround for https://github.com/siderolabs/omni/issues/797

### DIFF
--- a/internal/backend/runtime/omni/controllers/omni/internal/talos/schematic.go
+++ b/internal/backend/runtime/omni/controllers/omni/internal/talos/schematic.go
@@ -11,6 +11,7 @@ import (
 	"strings"
 
 	"github.com/cosi-project/runtime/pkg/safe"
+	"github.com/siderolabs/go-pointer"
 	"github.com/siderolabs/image-factory/pkg/constants"
 	"github.com/siderolabs/image-factory/pkg/schematic"
 	"github.com/siderolabs/talos/pkg/machinery/client"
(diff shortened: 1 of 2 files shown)


Unix4ever (commented):
I'd suggest to add some tests. We have unit tests for the extensions discovery, it should be a good place to add it.

internal/backend/runtime/omni/controllers/omni/internal/talos/schematic.go
```diff
@@ -83,6 +91,18 @@ func GetSchematicInfo(ctx context.Context, c *client.Client, defaultKernelArgs [
 		return SchematicInfo{}, err
 	}
 
+	if inAgentMode {
+		id, idErr := pointer.To(schematic.Schematic{}).ID()
+		if idErr != nil {
+			return SchematicInfo{}, fmt.Errorf("failed to calculate extensions schematic ID: %w", idErr)
+		}
+
+		return SchematicInfo{
```
frezbo, line 100: can't we just return an empty schematic here?

utkuozdemir: We need to return the ID of an empty schematic, which we do here. Not an empty string.

internal/backend/runtime/omni/controllers/omni/machine_status_test.go
```diff
@@ -408,11 +397,50 @@ func (suite *MachineStatusSuite) TestMachineSchematic() {
...
+			id := "test-" + tt.name
+
+			machine := omni.NewMachine(resources.DefaultNamespace, id)
+			spec := machine.TypedSpec().Value
+
+			spec.Connected = true
+			spec.ManagementAddress = suite.socketConnectionString
+
+			suite.Require().NoError(suite.state.Create(suite.ctx, machine))
+
```
utkuozdemir, line 443: Moved this here to create a different machine ID each time - otherwise `MachineStatus` preserves the initial schematic, and my new test was giving different results when it was run in isolation vs when it was run together with other tests.

utkuozdemir:
> I'd suggest to add some tests. We have unit tests for the extensions discovery, it should be a good place to add it.

done

Unix4ever (approved):

Issue #774, opened 2024-12-09
UI: display powered off machines as `powered off`, not `disconnected`, allow adding them to clusters


utkuozdemir:
When a machine is disconnected, it cannot be added to a cluster on the UI. It needs to be addressed as well.

PR #800, opened 2024-12-20
feat: properly handle powered off machines in the UI and machine classes

Unix4ever:
Introduce the new label `omni.sidero.dev/ready-to-use`, which is now used in the UI when creating/scaling a cluster and in the machine set scaling controller instead of `omni.sidero.dev/connected`.

The machine status controller was changed to add the ready to use label when the machine is created by the bare metal infra provider. Then the controller looks on the state of the `InfraMachineStatus` resource and sets `omni.sidero.dev/ready-to-use` label when the `ReadyToUse` flag is true in the `InfraMachineStatus`.

If the machine is managed by the bare metal infra provider and is not a part of a cluster we no longer
set `omni.sidero.dev/disconnected` label. It is used only in the UI. We still remove `connected` label when the machine gets disconnected from Omni.

Update the `MachineStatus` resource to also include power state of the machine. Then display it in the UI for the machines list.

Introduce `POWERING_ON` stage in the cluster machine status. Although it doesn't cover the power on stage fully. It will show powering on only for a short period until the bare metal provider issues the power on. We can improve that in the follow-up PR when we start tracking powering on stage more precisely.

The new UI:
Screenshot 2024-12-20 at 22 12 22

Fixes: https://github.com/siderolabs/omni/issues/774


utkuozdemir:
🆒

Unix4ever (commented):
/m

Issue #750, opened 2024-12-02
Omni erroneously removes a live etcd member


smira:
https://github.com/siderolabs/omni/issues/745

justyns:
FWIW, I also ran into this issue when upgrading from Talos 1.8.1 to 1.8.4.  I didn't do anything special, just updated the cluster template and ran `omnictl cluster template sync`.  The first node upgraded without issue, the second control plane node got stuck in an etcd restart loop with errors like:

```
100.64.0.77: {"level":"info","ts":"2024-12-24T22:14:15.262169Z","caller":"etcdserver/server.go:864","msg":"starting etcd server","local-member-id":"66de3babc0abe198","local-server-version":"3.5.17","cluster-id":"78fa62a055012012","cluster-version":"3.5"}                           
100.64.0.77: {"level":"info","ts":"2024-12-24T22:14:15.262330Z","caller":"etcdserver/server.go:773","msg":"starting initial election tick advance","election-ticks":5}                                                                                                                   
100.64.0.77: {"level":"warn","ts":"2024-12-24T22:14:15.262417Z","caller":"etcdserver/server.go:1154","msg":"server error","error":"the member has been permanently removed from the cluster"}                                                                                            
100.64.0.77: {"level":"warn","ts":"2024-12-24T22:14:15.262494Z","caller":"etcdserver/server.go:1155","msg":"data-dir used by this member must be removed"}                                                                                                                               
```

To "fix" it, I selected the "Reset installation and reboot into maintenance mode" for the 100.64.0.77 node.  It eventually came back up.

I made a support bundle with omnictl before resetting the node.  Would it be helpful to send somewhere?

Unix4ever:
Did you have any extensions installed on the nodes?

Unix4ever:
I've identified the root cause. The fix is landed in 0.45.1.

PR #804, opened 2024-12-25
fix: never remove etcd members which ID is discovered at least once

Unix4ever:
The logic of the etcd audit got outdated with the more recent Talos versions. `apid` now runs in the states where it wasn't available before, so the check for the etcd member might lead to the false-positives.
Also reorder the `auditMember` check sequence to be more correct.

Fixes: https://github.com/siderolabs/omni/issues/750


smira (approved):

smira:
Do we need to run `e2e-upgrades` as well?

utkuozdemir (approved):
🆒

Will it be compatible wit both older and newer Talos versions?

Unix4ever:
> Will it be compatible wit both older and newer Talos versions?

Yep, it should be.

PR #809, opened 2025-01-03
fix: fix immutability checks in infra provider state

utkuozdemir:
Found a couple of bugs in the infra provider state:

- For some reason, sometimes a resource update is detected where nothing other than the `.metadata.version` was changing. In these cases, the validation blocked the update. Fix this by ignoring the version field. Since we ignore the version field, we now compare the specs as well as the metadata.

- Infra providers were able to modify `infra.Machine` resources when they shouldn't have. The fix above also addresses that.

### DIFF
--- a/internal/backend/runtime/omni/infraprovider/state.go
+++ b/internal/backend/runtime/omni/infraprovider/state.go
@@ -142,13 +142,7 @@ func (st *State) Update(ctx context.Context, newResource resource.Resource, opts
 
 	switch newResource.Metadata().Type() {
 	case infra.MachineRequestType, infra.InfraMachineType:
-		oldMd := oldResource.Metadata().Copy()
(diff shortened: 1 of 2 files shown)


internal/backend/runtime/omni/infraprovider/state.go
```diff
@@ -163,6 +157,44 @@ func (st *State) Update(ctx context.Context, newResource resource.Resource, opts
 	return status.Errorf(codes.NotFound, "not found")
 }
 
+func (st *State) resourcesAreEqual(res1, res2 resource.Resource) bool {
```
Unix4ever, line 160: You can also take a look at how equality is checked in the COSI internals maybe

Unix4ever (approved):

PR #810, opened 2025-01-03
fix: wait for infra machine info to be collected before powering off

### LINKED ISSUE #798: The machine gets powered off before Omni is able to collect it's status
We should let Omni collect the `MachineStatus` before powering off the machine.
Otherwise it ends up in the state when it can't be allocated into a cluster, because it's Talos version is not known.

Maybe the controller can be changed to wait until the machine status is populated and then switch the desired state to powered off.

utkuozdemir:
There were cases when an infra machine was accepted, it was powered off by the infra provider too quickly, before its specs were populated by the `MachineStatus` poller. This caused additional issues as some other resources like `SchematicConfiguration` were also never created, blocking cluster creation.

Address this by setting the preferred power state of the infra machine to ON until its status is populated (we check this by checking the secure boot status field).

Additionally, clean up `InfraMachineConfig` resources (user-managed) when a machine is deleted.

Closes #798.

### DIFF
--- a/internal/backend/runtime/omni/controllers/omni/infra_machine.go
+++ b/internal/backend/runtime/omni/controllers/omni/infra_machine.go
@@ -65,6 +65,9 @@ func NewInfraMachineController() *InfraMachineController {
 			qtransform.MapperSameID[*omni.ClusterMachine, *siderolink.Link](),
 		),
 		qtransform.WithExtraMappedInput(
+			qtransform.MapperSameID[*omni.MachineStatus, *siderolink.Link](),
+		),
+		qtransform.WithExtraMappedInput(
 			func(ctx context.Context, _ *zap.Logger, runtime controller.QRuntime, res *infra.ProviderStatus) ([]resource.Pointer, error) {
 				linkList, err := safe.ReaderListAll[*siderolink.Link](ctx, runtime, state.WithLabelQuery(resource.LabelEqual(omni.LabelInfraProviderID, res.Metadata().ID())))
 				if err != nil {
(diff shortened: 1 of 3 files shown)


internal/backend/runtime/omni/controllers/omni/infra_machine.go
```diff
@@ -108,7 +116,9 @@ func (h *infraMachineControllerHelper) transformExtraOutput(ctx context.Context,
 		return xerrors.NewTaggedf[qtransform.SkipReconcileTag]("the link is not created by a static infra provider")
 	}
 
-	if err = h.applyInfraMachineConfig(infraMachine, config); err != nil {
+	machineInfoCollected := machineStatus != nil && machineStatus.TypedSpec().Value.SecureBootStatus != nil
```
frezbo, line 119: I don't quite get this, not all machines would boot with secureboot, so why check this?

utkuozdemir: It's just an indicator that machine info was collected. This struct has an `Enabled` field that is false if we collected the machine info and found out that it's not a secure boot machine. So, if this struct is not nil, we know that we collected the info.

I could have checked one of the other fields instead and it would have worked, but I wanted specifically to make sure this check would pass here in this controller: https://github.com/siderolabs/omni/blob/main/internal/backend/runtime/omni/controllers/omni/schematic_configuration.go#L166-L169

So I went with the exact same field.

frezbo: feels weird to check for that status instead of something like hardware info, maybe some historical reason.

frezbo (approved):

smira (approved):

PR #816, opened 2025-01-08
feat: sync infra machine labels onto the machine status

utkuozdemir:
Copy the labels set on infra.MachineStatus to omni.MachineStatus resources.

Introduce a special prefix for the labels set by infra providers, so that they can keep track and synchronize the labels as needed.

Related to/depends on https://github.com/siderolabs/omni-infra-provider-bare-metal/pull/16.


utkuozdemir:
@Unix4ever do you think we need a makeup for this in the UI?

image

Issue #175, opened 2024-04-25
[feature] Secure boot images in web UI


jfroy:
### Problem Description

There does not appear to be a way to download secure boot images.

Unix4ever:
Actually you can use CLI, like that:

```
omnictl download --secureboot iso
```

jfroy:
> Actually you can use CLI, like that:
> 
> 
> 
> ```
> 
> omnictl download --secureboot iso
> 
> ```

Ah great. I hadn't tried the CLI before.

Unix4ever:
@smira do you think we're ready to enable secure boot checkbox in the UI?

smira:
> @smira do you think we're ready to enable secure boot checkbox in the UI?

yes, only for bare-metal for now, and make sure you link to the docs (https://www.talos.dev/v1.7/talos-guides/install/bare-metal-platforms/secureboot/), and this is Talos 1.5+ only.

CappyT:
Is there any update on this? Also, once the iso from the command line is used, do upgrade from UI respect the secureboot status?

Unix4ever:
> Is there any update on this? Also, once the iso from the command line is used, do upgrades from UI respect the secureboot status?

We should enable it, will prioritize it.

The upgrades should handle the secureboot.

PR #825, opened 2025-01-09
fix: enable secure boot checkbox in the UI

Unix4ever:
Fixes: https://github.com/siderolabs/omni/issues/175

### DIFF
--- a/frontend/src/views/omni/Modals/DownloadInstallationMedia.vue
+++ b/frontend/src/views/omni/Modals/DownloadInstallationMedia.vue
@@ -60,18 +60,11 @@ included in the LICENSE file.
 
       <t-input v-model="kernelArguments"/>
 
-      <!-- TODO(image-factory): enable this after further testing and making sure that it works -->
-      <template v-if="false">
-        <h3 class="text-sm text-naturals-N14">
-          Secure Boot
-        </h3>
-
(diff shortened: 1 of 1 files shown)


frontend/src/views/omni/Modals/DownloadInstallationMedia.vue
```diff
@@ -60,18 +60,11 @@ included in the LICENSE file.
 
       <t-input v-model="kernelArguments"/>
 
-      <!-- TODO(image-factory): enable this after further testing and making sure that it works -->
```
smira, line 63: I don't know if Omni uses this already, but Image Factory exports a list of supported platforms/images and secure boot support as well: https://github.com/siderolabs/image-factory/blob/main/pkg/metadata/platforms.go

smira: We could move it back to Talos machinery as well

smira (approved):

frontend/src/views/omni/Modals/DownloadInstallationMedia.vue
Unix4ever: Omni doesn't use it yet. I guess it'll require a bit of a research.

PR #826, opened 2025-01-10
feat: add mgmt endpoint & cli command for infra machine reboot

utkuozdemir:
Introduce a new mgmt endpoint to reboot an infra machine and wait for the reboot request to be processed by the respective infra provider. It does the heavy lifting of modifying/waiting for the resources.

Additionally:
- In the `InfraMachineController`, to verify that the machine info was collected before powering it off, check the `TalosVersion` on the `MachineStatus` instead of `SecureBootStatus`. It is a more suitable field to check this.
- Rename the infra provider label prefix format from being `omni.sidero.dev/infra-provider[provider-id]/key=value` to be `omni.sidero.dev/infra-provider/provider-id/key=value`.


internal/backend/grpc/management.go
```diff
@@ -654,6 +656,81 @@ func (s *managementServer) ReadAuditLog(req *management.ReadAuditLogRequest, srv
...
 }
 
+func (s *managementServer) RebootMachine(ctx context.Context, req *management.RebootMachineRequest) (*management.RebootMachineResponse, error) {
+	if _, err := s.authCheckGRPC(ctx, auth.WithRole(role.Admin)); err != nil {
+		return nil, err
+	}
+
+	ctx = actor.MarkContextAsInternalActor(ctx)
+
+	if _, err := safe.StateGetByID[*infra.Machine](ctx, s.omniState, req.MachineId); err != nil {
```
Unix4ever, line 666: Maybe a bit crazy idea, but we could make reboot work seamlessly for all kind of machines.
If the infra machine is not found we can fall back to the talos API call.
What do you think?

utkuozdemir: Sure, we can do that. I'm just concerned with the fact that this endpoint does not block or report any status (like stream the phases back or something), so not sure how much value it actually brings.

Unix4ever: Does it block in case of `omnictl power reboot`?

utkuozdemir: nope, none. It blocks until reboot is issued, not until machine comes back.

Unix4ever: so it will be more or less consistent I guess

Unix4ever: another crazy idea is to intercept `talosctl reboot` command and make it go through the infra provider 🙃

internal/backend/runtime/omni/controllers/omni/infra_machine_test.go
```diff
@@ -55,7 +55,7 @@ func (suite *InfraMachineControllerSuite) TestReconcile() {
 	})
 
 	machineStatus := omni.NewMachineStatus(resources.DefaultNamespace, "machine-1")
-	machineStatus.TypedSpec().Value.SecureBootStatus = &specs.SecureBootStatus{}
+	machineStatus.TypedSpec().Value.TalosVersion = "v1.9.1"
```
DmitriyMV, line 58: Any particular reason for `v1.9.1`?

internal/backend/grpc/management.go
```diff
@@ -654,6 +656,81 @@ func (s *managementServer) ReadAuditLog(req *management.ReadAuditLogRequest, srv
...
+	eventCh := make(chan safe.WrappedStateEvent[*infra.MachineStatus])
+
+	if err = safe.StateWatch[*infra.MachineStatus](ctx, s.omniState, infra.NewMachineStatus(req.MachineId).Metadata(), eventCh); err != nil {
+		return nil, err
+	}
+
+	var (
+		machineStatus *infra.MachineStatus
+		event         safe.WrappedStateEvent[*infra.MachineStatus]
+	)
```
DmitriyMV, line 708: No reason do declare those outside of `for { ... }` since they are not used outside. Actually `if machineStatus, err = event.Resource(); err != nil {` can be split and there will be no need to declare `machineEvent`.

internal/backend/grpc/management.go
```diff
@@ -654,6 +656,81 @@ func (s *managementServer) ReadAuditLog(req *management.ReadAuditLogRequest, srv
...
+	newRebootID := uuid.NewString()
+
+	machineConfig, err := safe.StateGetByID[*omnires.InfraMachineConfig](ctx, s.omniState, req.MachineId)
+	if err != nil && !state.IsNotFoundError(err) {
+		return nil, err
+	}
+
+	if machineConfig == nil {
+		machineConfig = omnires.NewInfraMachineConfig(resources.DefaultNamespace, req.MachineId)
+
```
DmitriyMV, line 683: Empty line?

Unix4ever:
I'd propose to close this PR. It was pending for a long time and it looks like no one really asked for that so far.

utkuozdemir:
> I'd propose to close this PR. It was pending for a long time and it looks like no one really asked for that so far.

Yep, definitely. Doing it now.

PR #827, opened 2025-01-10
fix: change the look of the infra provider labels

Unix4ever:
Fixes: https://github.com/siderolabs/omni/issues/822

### DIFF
--- a/frontend/src/methods/labels.ts
+++ b/frontend/src/methods/labels.ts
@@ -3,6 +3,8 @@
 // Use of this software is governed by the Business Source License
 // included in the LICENSE file.
 
+import { InfraProviderLabelPrefix, SystemLabelPrefix } from "@/api/resources";
+
 export const parseLabels = (...labels: string[]): Record<string, string> => {
     const labelsMap: Record<string, string> = {};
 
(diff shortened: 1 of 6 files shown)


Unix4ever:
Screenshot 2025-01-10 at 22 38 03

frontend/src/methods/labels.ts
```diff
@@ -75,3 +78,36 @@ export const selectors = (labels: Label[]) => {
...
+
+export const getLabelFromID = (key: string, value: string): Label => {
+  const isUser = key.indexOf(SystemLabelPrefix) !== 0;
+
+  let strippedKey = key.replace(new RegExp(`^${SystemLabelPrefix}`), "");
+  let icon: string | undefined;
+  let color = getLabelColor(strippedKey);
+  let description = labelDescriptions[strippedKey];
+
+  if (strippedKey.indexOf("infra-provider") === 0) {
```
utkuozdemir, line 94: No big deal but should we instead have this as constant on backend and use `tsgen:` and use the var both on backend and frontend?

utkuozdemir (approved):
Looks neat!

frontend/src/methods/labels.ts
Unix4ever: Good point!

frezbo (approved):

PR #829, opened 2025-01-13
fix: do not allow using static infra providers in the machine classes

Unix4ever:
Auto-provision mode should be disabled for them.

This change has two parts:

1. Filter the static providers in the UI.
2. Block creating machine classes which reference infra provider with the `is-static-infra-provider` label set.


utkuozdemir (approved):
🆒

PR #834, opened 2025-01-14
fix: properly reset `MachineStatus` hostname for deallocated machines

Unix4ever:
Need to treat `nil` hostname as no hostname and wipe the current value in the `MachineStatus`.

### DIFF
--- a/internal/backend/runtime/omni/controllers/omni/internal/task/machine/poll.go
+++ b/internal/backend/runtime/omni/controllers/omni/internal/task/machine/poll.go
@@ -99,6 +99,9 @@ func pollVersion(ctx context.Context, c *client.Client, info *Info) error {
 }
 
 func pollHostname(ctx context.Context, c *client.Client, info *Info) error {
+	info.Hostname = pointer.To("")
+	info.Domainname = pointer.To("")
+
 	return forEachResource(
 		ctx,
 		c,


internal/backend/runtime/omni/controllers/omni/machine_status.go
```diff
@@ -417,10 +417,14 @@ func (ctrl *MachineStatusController) handleNotification(ctx context.Context, r c
 
 		if event.Hostname != nil {
 			spec.Network.Hostname = *event.Hostname
+		} else {
+			spec.Network.Hostname = ""
```
smira, line 421: this I believe is not exactly great change, as by the design some parts of the event might be `nil` if they're missing.

The proper way probably is to treat missing `HostnameStatus` as a way to set `event.Hostname` to `&("")`

Unix4ever: Yeah. I guess I have an idea

utkuozdemir (approved):

internal/backend/runtime/omni/controllers/omni/internal/task/machine/poll.go
```diff
@@ -99,6 +99,9 @@ func pollVersion(ctx context.Context, c *client.Client, info *Info) error {
 }
 
 func pollHostname(ctx context.Context, c *client.Client, info *Info) error {
+	info.Hostname = pointer.To("")
```
smira, line 102: this works, but just fyi: the way Talos works, on boot, there might be a window when hostname is not known, and it might flicker a bit between `""` and whatever known hostname.

not a big deal though

smira (approved):

internal/backend/runtime/omni/controllers/omni/internal/task/machine/poll.go
Unix4ever: I think if the machine reboots it's machine status gets updated anyways.

PR #838, opened 2025-01-15
feat: introduce resource for infra provider health checks

utkuozdemir:
Add the ephemeral resource `InfraProviderHealthStatus` for the providers to report their health.

Refactor the infra provider state to define resource characteristics/configuration in a single place.

Add a new ephemeral underlying state to the infra provider state for it to be able to store ephemeral resources.

Improve/do fixes on resource operation checks in infra provider state.

Introduce the new `ProviderHealthStatusController` for the infra providers to report their health status.

This controller is used by the common infra provider library, but can be used standalone as well.

Part of https://github.com/siderolabs/omni/pull/838.


internal/backend/runtime/omni/state_access.go
```diff
@@ -504,12 +502,6 @@ func filterAccessByType(access state.Access) error {
 
 		return status.Error(codes.PermissionDenied, "only read, update and delete access is permitted")
 	case
-		infra.MachineRequestType,       // read-only for all except for InfraProvider role (checked in filterAccess)
-		infra.MachineRequestStatusType, // read-only for all except for InfraProvider role (checked in filterAccess)
-		infra.InfraMachineType,         // read-only for all except for InfraProvider role (checked in filterAccess)
-		infra.InfraMachineStateType,    // read-only for all except for InfraProvider role (checked in filterAccess)
-		infra.InfraMachineStatusType,   // read-only for all except for InfraProvider role (checked in filterAccess)
-		infra.InfraProviderStatusType,  // read-only for all except for InfraProvider role (checked in filterAccess)
```
utkuozdemir, line 512: This was a bug - the code never reached here for these types.

internal/backend/runtime/omni/infraprovider/state.go
```diff
@@ -389,3 +426,54 @@ func (st *State) checkNamespaceAndType(ns resource.Namespace, infraProviderID st
...
+	case infra.InfraProviderHealthStatusType:
+		return resourceConfig{
+			ephemeral: true,
+			checkID:   true,
+		}, true
+	case infra.ConfigPatchRequestType:
+		return resourceConfig{}, true
+	default:
+		return resourceConfig{}, false
+	}
```
utkuozdemir, line 478: This is now the central place where we define all configurations for the infra provider resources.

internal/backend/runtime/omni/state.go
```diff
@@ -90,7 +90,8 @@ func newNamespacedState(params *config.Params, primaryStorageCoreState state.Cor
 		return nil, nil, fmt.Errorf("failed to create etcd backup store: %w", err)
 	}
 
-	infraProviderState := infraprovider.NewState(primaryStorageCoreState, logger.With(logging.Component("infraprovider_state")))
+	infraProviderEphemeralState := inmem.NewStateWithOptions(inmem.WithHistoryGap(20))(resources.InfraProviderNamespace)
```
Unix4ever, line 93: I might understand the reasoning to use the same namespace for the ephemeral resources here, but it will be really hard to understand what's going on.
We route the resource to the ephemeral storage by it's type in the infra provider state.
But my proposal was to to use `resources.EphemeralNamespace` for the `InfraProviderHealthStatus` and just allow to write to that namespace.

internal/backend/runtime/omni/infraprovider/state.go
Unix4ever: Overall it looks good to me, but I don't like that for the normal resources it's easy to guess: ephemeral namespace -- goes to the ephemeral state, default -- goes to the etcd.
Here we use the same namespace.

I would prefer to use the same schema everywhere.

internal/backend/runtime/omni/state.go
Unix4ever: The current schema with the new inmem state can be still used, but I'd propose to use the ephemeral namespace.

I think namespaced state can be used to achieve that. You put that one on top, then ephemeral and etcd states underneath.

utkuozdemir: Do you mean the infra provider state should be the "top-level" state now, and it should only catch/modify the infra provider related resources, and pass everything else through without touching them?

It doesn't feel right to me, because imo, the infra provider state should only get the resources it is interested in, not more.

If the issue with current approach is "looking at the resource, it is not obvious that it is ephemeral" (because it's namespace does not reflect that), then actually an easier solution would be to create a new ns called `infra-provider-ephemeral`. I also considered that as an option.

utkuozdemir: One solution I also considered:
- introduce a new state wrapper called `RootState` let's say
- that is responsible of routing resources to the correct underlying state - not the "namespaced state" anymore
- it observes `namespace+type` to route the resources to the right underlying state
- the "infra provider state" is aware of both persistent and ephemeral states (passed via constructor)
- it does its intercepting logic and sends it to one of the underlying states

sth like:
```
root state
├── etcd state
├── ephemeral state
└── infra provider state
    ├── etcd state
    └── ephemeral state
```

Unix4ever: > Do you mean the infra provider state should be the "top-level" state now, and it should only catch/modify the infra provider related resources, and pass everything else through without touching them?

Not top level. It can be like it is right now.

I think I get the problem with using `ephemeral` namespace name.
Yeah, we can `infra-provider-ephemeral` namespace name and then make the infra provider handle them both, then next will be the namespaced state and then etcd and the ephemeral state.

utkuozdemir: alright then I guess I'll go with `infra-provider-ephemeral`.

utkuozdemir: Updated the PR to do that.

utkuozdemir:
> LGTM, I only think that we will need another resource on top of this one and the infra provider status to present a merged version of them, which will be queried by the UI.

Ok, I'm going to merge this, I think we can do the unified request later when we do the UI part.

PR #843, opened 2025-01-16
feat: implement power states as machine stage events

utkuozdemir:
Introduce a new component to watch infra.MachineStatus resources to produce complementary "synthetic" machine events for "powering on" and "powered off" stages.

Change the logic in `MachineStatusSnapshot` and `ClusterMachineStatus` controllers to take these changes into account.

This is required to display the correct status for the machines that are powered on/off by the infra providers.


internal/backend/powerstage/watcher.go
```diff
@@ -0,0 +1,92 @@
...
+
+		if err := event.Error(); err != nil {
+			return fmt.Errorf("failed to watch machine status: %w", err)
+		}
+
+		resource, err := event.Resource()
+		if err != nil {
+			return fmt.Errorf("failed to read resource from the event: %w", err)
+		}
+
```
Unix4ever, line 66: I think we might need to ignore `Destroyed` events here.

internal/backend/runtime/omni/controllers/omni/machine_status_snapshot.go
```diff
@@ -32,17 +32,19 @@ type MachineStatusSnapshotController struct {
 	runner       *task.Runner[snapshot.InfoChan, snapshot.CollectTaskSpec]
 	notifyCh     chan *omni.MachineStatusSnapshot
 	siderolinkCh <-chan *omni.MachineStatusSnapshot
+	powerStageCh <-chan *omni.MachineStatusSnapshot
```
Unix4ever, line 35: I wonder if we could reuse `siderolinkCh` here instead of creating a new channel.

utkuozdemir: I just thought this was cleaner, as they don't come from SideroLink. I think it is more obvious this way that there are 2 different sources. Wdyt?

Unix4ever: I guess we can keep the two channels

internal/backend/powerstage/watcher.go
utkuozdemir: good point, done.

internal/backend/runtime/omni/controllers/omni/machine_status_snapshot.go
utkuozdemir: yes I think it's ok, `select` works with multiple channels beautifully :)

Unix4ever (approved):

PR #848, opened 2025-01-20
fix: detect infra provider internal resources correctly

utkuozdemir:
Fix the regression where the resources in namespace `infra-provider:<provider-id>` were not correctly captured and managed by the infra provider state.


utkuozdemir:
false alarm, it was working correctly.

PR #856, opened 2025-01-21
test: extract static infra provider machine accept tests into hooks

Unix4ever:
Change the integration tests structure to enable running
any tests against the static infra providers: extract machine
acceptance/cleanup code into the pre/post hooks.
Introduce the flags to disable pre-installed extensions checks.

Signed-off-by: Artem Chernyshev <artem.chernyshev@talos-systems.com>

### DIFF
--- a/cmd/integration-test/pkg/root.go
+++ b/cmd/integration-test/pkg/root.go
@@ -46,9 +46,10 @@ var rootCmd = &cobra.Command{
 			testOptions := tests.Options{
 				RunTestPattern: rootCmdFlags.runTestPattern,
 
-				ExpectedMachines: rootCmdFlags.expectedMachines,
-				CleanupLinks:     rootCmdFlags.cleanupLinks,
-				RunStatsCheck:    rootCmdFlags.runStatsCheck,
+				ExpectedMachines:            rootCmdFlags.expectedMachines,
+				CleanupLinks:                rootCmdFlags.cleanupLinks,
+				RunStatsCheck:               rootCmdFlags.runStatsCheck,
+				SkipExtensionsCheckOnCreate: rootCmdFlags.skipExtensionsCheckOnCreate,
 
 				MachineOptions:           rootCmdFlags.machineOptions,
 				AnotherTalosVersion:      rootCmdFlags.anotherTalosVersion,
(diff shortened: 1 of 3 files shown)


utkuozdemir:
Good one, I'll have a look tomorrow latest.

smira:
Can we get a more thorough explanation about "ready to be used"? what is the logic behind it? why applying a machine config updates makes it "not ready to be used"?

Unix4ever:
> Can we get a more thorough explanation about "ready to be used"? what is the logic behind it? why applying a machine config updates makes it "not ready to be used"?

I guess it's related to the fact machine being not-clean. But in terms of config apply I guess it only makes sense only before we apply the config for the first time.
Another way to fix it would be to keep `ReadyToUse` as true after the machine is allocated. But I'm not sure what else will break if we do that.

smira:
> I guess it's related to the fact machine being not-clean. But in terms of config apply I guess it only makes sense only before we apply the config for the first time.
> Another way to fix it would be to keep `ReadyToUse` as true after the machine is allocated. But I'm not sure what else will break if we do that.

My concern is that a machine should stay ready to use as long as it's allocated? Machine going no ready to use sounds like a real bug there? Otherwise we seem to be fixing consequences, not the real problem?

utkuozdemir:
> > I guess it's related to the fact machine being not-clean. But in terms of config apply I guess it only makes sense only before we apply the config for the first time.
> > Another way to fix it would be to keep `ReadyToUse` as true after the machine is allocated. But I'm not sure what else will break if we do that.
> 
> My concern is that a machine should stay ready to use as long as it's allocated? Machine going no ready to use sounds like a real bug there? Otherwise we seem to be fixing consequences, not the real problem?

You are right, that field's meaning is under-defined, and right now it is marked as ReadyToUse when it is ready (does not need wipe for example) AND not allocated. So it kinda means "ready to be allocated".

We can just change it to stay ReadyToUse after allocation, or rename/break down that field so multiple fields to make it more clear.

smira:
Ok, but what kind of problem does it solve specifically in this controller then? It should control allocation, and this controller only works on allocated machines? no?

utkuozdemir:
> Ok, but what kind of problem does it solve specifically in this controller then? It should control allocation, and this controller only works on allocated machines? no?

We just discussed it with @Unix4ever , and that flag is supposed to only represent dirtiness status (if there's a pending wipe or not). But atm it represents that + being unallocated (means, it kinda represented "ok you can add this machine to a cluster now").

This was wrong and unneeded logic, and we will fix it today by making it stay `true` even when it is allocated/installed etc.

The main change in this PR will not be needed anymore after that.

utkuozdemir (approved):

Issue #855, opened 2025-01-21
[feature] Report machine usage to Stripe


andrewrynhard:
### Problem Description

In order to track machine usage for more accurate billing we need a way to report near real-time machine usage to Stripe.

The idea is to create a new controller that watches for a resource sufficient enough to represent when a machine is created or destroyed. We should not immediately update Stripe. There needs to be some sort of "smoothing" algorithm to help in the cases when there is a lot of churn.

smira:
Omni already reports machine metrics, smoothing I guess can be applied as a Prometheus query, and now it can be fed to another service which does reporting.

andrewrynhard:
> Omni already reports machine metrics, smoothing I guess can be applied as a Prometheus query, and now it can be fed to another service which does reporting.

Yes. Spoke with Artem and we landed on the same thing. The controller watches the metrics and reports that to Stripe if the count is stable for 10 minutes.

Unix4ever:
Is this done @andrewrynhard

PR #857, opened 2025-01-21
feat: add Stripe machine reporting

andrewrynhard:
Adds a controller for reporting machine count to Stripe.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -4,7 +4,7 @@ Parameters
 
 Licensor:             Sidero Labs, Inc.
 Licensed Work:        Omni
-                      The Licensed Work is (c) 2025 Sidero Labs, Inc.
+                      The Licensed Work is (c) 2024 Sidero Labs, Inc.
 Additional Use Grant: None
 
 Change Date:          2029-01-17
--- a/internal/backend/runtime/omni/controllers/omni/reporting.go
+++ b/internal/backend/runtime/omni/controllers/omni/reporting.go
@@ -1,0 +1,170 @@
+// Copyright (c) 2024 Sidero Labs, Inc.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+
+package omni
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/cenkalti/backoff/v4"
+	"github.com/cosi-project/runtime/pkg/controller"
+	"github.com/cosi-project/runtime/pkg/controller/generic"
+	"github.com/cosi-project/runtime/pkg/safe"
+	"github.com/cosi-project/runtime/pkg/state"
+	"github.com/stripe/stripe-go/v74"
+	"github.com/stripe/stripe-go/v74/subscriptionitem"
+	"go.uber.org/zap"
+
+	"github.com/siderolabs/omni/client/pkg/omni/resources"
+	"github.com/siderolabs/omni/client/pkg/omni/resources/omni"
+)
+
+// DefaultDebounceDuration defines the default time to wait after the last event before processing updates.
+var DefaultDebounceDuration = 10 * time.Minute
+
+// StripeMetricsReporterControllerName is the name of the StripeMetricsReporterController.
+const StripeMetricsReporterControllerName = "StripeMetricsReporterController"
+
+// StripeMetricsReporterController reports machine metrics to Stripe.
+type StripeMetricsReporterController struct {
+	generic.NamedController
+
+	stripeAPIKey             string
+	stripeSubscriptionItemID string
+	debounceDuration         time.Duration
+}
+
+// StripeMetricsReporterControllerOptions defines the options for StripeMetricsReporterController.
+type StripeMetricsReporterControllerOptions struct {
+	DebounceDuration time.Duration
+}
+
+// StripeMetricsReporterControllerOption defines the option type for StripeMetricsReporterController.
+type StripeMetricsReporterControllerOption func(*StripeMetricsReporterControllerOptions)
+
+// WithDebounceDuration sets the debounce duration for StripeMetricsReporterController.
+func WithDebounceDuration(duration time.Duration) StripeMetricsReporterControllerOption {
+	return func(opts *StripeMetricsReporterControllerOptions) {
+		opts.DebounceDuration = duration
+	}
+}
+
+// NewStripeMetricsReporterController initializes StripeMetricsReporterController.
+func NewStripeMetricsReporterController(stripeAPIKey, stripeSubscriptionItemID string, opts ...StripeMetricsReporterControllerOption) *StripeMetricsReporterController {
+	options := &StripeMetricsReporterControllerOptions{
+		DebounceDuration: DefaultDebounceDuration,
+	}
+
+	for _, opt := range opts {
+		opt(options)
+	}
+
+	return &StripeMetricsReporterController{
+		NamedController: generic.NamedController{
+			ControllerName: StripeMetricsReporterControllerName,
+		},
+		stripeAPIKey:             stripeAPIKey,
+		stripeSubscriptionItemID: stripeSubscriptionItemID,
+		debounceDuration:         options.DebounceDuration,
+	}
+}
+
+// Inputs implements controller.QController interface.
+func (ctrl *StripeMetricsReporterController) Inputs() []controller.Input {
+	return []controller.Input{
+		{
+			Namespace: resources.EphemeralNamespace,
+			Type:      omni.MachineStatusMetricsType,
+			Kind:      controller.InputWeak,
+		},
+	}
+}
+
+// Outputs implements controller.Controller interface.
+func (ctrl *StripeMetricsReporterController) Outputs() []controller.Output {
+	return nil
+}
+
+// Run implements controller.Controller interface.
+func (ctrl *StripeMetricsReporterController) Run(ctx context.Context, r controller.Runtime, log *zap.Logger) error {
+	stripe.Key = ctrl.stripeAPIKey
+
+	var (
+		pendingCount uint32
+		timerCh      <-chan time.Time
+	)
+
+	processPending := func() {
+		count := pendingCount
+		pendingCount = 0
+
+		err := updateStripeSubscriptionItemQuantity(ctrl.stripeSubscriptionItemID, count, log)
+		if err != nil {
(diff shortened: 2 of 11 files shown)


andrewrynhard:
Closes #855

LICENSE
```diff
@@ -4,7 +4,7 @@ Parameters
 
 Licensor:             Sidero Labs, Inc.
 Licensed Work:        Omni
-                      The Licensed Work is (c) 2025 Sidero Labs, Inc.
```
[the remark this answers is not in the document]
utkuozdemir: Looked into it, and it seems this is because it is not dynamic, comes from `.kres.yaml` here: https://github.com/siderolabs/omni/blob/main/.kres.yaml#L545

Feel free to update it and rekres one more time.

internal/backend/runtime/omni/controllers/reporting/reporting.go
```diff
@@ -0,0 +1,204 @@
...
+	subscriptionItemID := os.Getenv("STRIPE_SUBSCRIPTION_ITEM_ID")
+	if subscriptionItemID == "" {
+		return fmt.Errorf("STRIPE_SUBSCRIPTION_ITEM_ID environment variable not set")
+	}
+
+	// Channel to send count changes to the debouncer.
+	countChan := make(chan uint32, 100) // Buffered to handle bursts.
+
+	// Start the debouncer goroutine.
+	go debouncer(ctx, countChan, subscriptionItemID)
```
utkuozdemir, line 84: We use `panichandler.Go(...)` instead of naked `go` statements.

internal/backend/runtime/omni/controllers/reporting/reporting.go
```diff
@@ -0,0 +1,204 @@
...
+	stripe.Key = os.Getenv("STRIPE_API_KEY")
+	if stripe.Key == "" {
+		return fmt.Errorf("STRIPE_API_KEY environment variable not set")
+	}
+
+	// Retrieve Subscription Item ID from environment variables or configuration.
+	subscriptionItemID := os.Getenv("STRIPE_SUBSCRIPTION_ITEM_ID")
+	if subscriptionItemID == "" {
+		return fmt.Errorf("STRIPE_SUBSCRIPTION_ITEM_ID environment variable not set")
+	}
```
utkuozdemir, line 78: Should we instead do this outside of the controller and crash the whole instance if these are not set? Or this controller crash-looping but functional Omni is more preferable?

internal/backend/runtime/omni/controllers/reporting/reporting.go
```diff
@@ -0,0 +1,204 @@
...
+			if err != nil {
+				if state.IsNotFoundError(err) {
+					continue
+				}
+
+				return err
+			}
+
+			count := metrics.TypedSpec().Value.RegisteredMachinesCount
+			countChan <- count
```
utkuozdemir, line 101: Maybe protect this with `select` from blocking by selecting on `ctx.Done()` and this channel send?

internal/backend/runtime/omni/controllers/reporting/reporting.go
utkuozdemir, line 1: No big deal but this could instead be a `qcontroller` or even `qtransform` controller outputting a dummy (e.g., ephemeral, never changing) resource. Can simplify the logic. But this is also fine I think.

internal/backend/runtime/omni/controllers/reporting/reporting.go
andrewrynhard: I could see an argument for either way. I want to lean towards functional Omni but that has it's downsides. I will let others chime in.

internal/backend/runtime/omni/controllers/reporting/reporting.go
```diff
@@ -0,0 +1,204 @@
+// Copyright (c) 2024 Sidero Labs, Inc.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+
+package reporting
```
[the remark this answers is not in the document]
Unix4ever: we do, but mainly because we reuse common parts there. If we don't then I don't see a reason not to have a separate folder.

internal/backend/runtime/omni/controllers/reporting/reporting.go
andrewrynhard: I'll add a little context as to why I made these environment variables too: With Kubernetes we can at least reference these from a secret.

Unix4ever: I'd suggest to pass stripe configs to the `NewStripeMetricsReporterController` methods as the arguments.
It will be more obvious that the controller needs some params.

internal/backend/runtime/omni/controllers/reporting/reporting.go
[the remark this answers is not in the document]
andrewrynhard: I will move into `omni`.

internal/backend/runtime/omni/controllers/reporting/reporting.go
```diff
@@ -0,0 +1,214 @@
...
+
+			// Start a new timer.
+			timer = time.NewTimer(DebounceDuration)
+
+		case <-func() <-chan time.Time {
+			if timer != nil {
+				return timer.C
+			}
+			// Return a never-blocking channel if timer is nil.
+			return make(chan time.Time)
```
smira, line 173: reading from `nil` channel blocks forever, so usually it's way easier to write this as `var timerCh <- chan time.Time` and make it non-nil when timer is running

internal/backend/runtime/omni/controllers/reporting/reporting.go
```diff
@@ -0,0 +1,214 @@
...
+	panichandler.Go(func() {
+		debouncer(ctx, countChan, subscriptionItemID)
+	}, nil)
+
+	for {
+		select {
+		case <-ctx.Done():
+			return nil
+		case <-r.EventCh():
+			metrics, err := safe.ReaderGetByID[*omni.MachineStatusMetrics](ctx, r, omni.MachineStatusMetricsID)
```
smira, line 94: this can be un-indented one level by closing `select` before this line

internal/backend/runtime/omni/controllers/reporting/reporting.go
```diff
@@ -0,0 +1,214 @@
...
+				// Successfully sent to the channel.
+			default:
+				log.Printf("Dropping count update: %d", count)
+			}
+		}
+	}
+}
+
+// debouncer accumulates count changes and updates Stripe after a debounce duration.
+func debouncer(ctx context.Context, countChan <-chan uint32, subscriptionItemID string) {
```
smira, line 117: I don't think I understand why `debouncer` is a separate goroutine, it hurts error reporting, but I think it can be perfectly merged into the main controller loop reducing complexity

internal/backend/runtime/omni/controllers/reporting/reporting.go
```diff
@@ -0,0 +1,214 @@
...
+	var (
+		pendingCount uint32 = 0
+		timer        *time.Timer
+	)
+
+	// Define a function to process the pending count.
+	processPending := func() {
+		if pendingCount == 0 {
```
smira, line 125: what if the number of machines went e.g. 1->0, we will never report it?

internal/backend/runtime/omni/controllers/omni/reporting.go
```diff
@@ -0,0 +1,158 @@
...
+
+		if timer != nil {
+			if !timer.Stop() {
+				<-timer.C
+			}
+		}
+
+		timer = time.NewTimer(DebounceDuration)
+
+		timerCh = timer.C
```
utkuozdemir, line 126:
```suggestion
		timerCh = timer.C
```

internal/backend/runtime/omni/controllers/omni/reporting.go
```diff
@@ -0,0 +1,158 @@
...
+	}
```
utkuozdemir, line 127: As a common practice, I suggest putting `r.ResetRestartBackoff()` as the last statement in the `for` loop of events.

utkuozdemir (approved):
LGTM overall with a couple of remarks 👍

Unix4ever:
I'd still recommend to add the unit test for the `reporting` controller to test it in isolation.
It can submit the data to either a mock running an http server, or we can make it accept stripe client interface and mock the methods in the tests.

andrewrynhard:
> I'd still recommend to add the unit test for the `reporting` controller to test it in isolation. It can submit the data to either a mock running an http server, or we can make it accept stripe client interface and mock the methods in the tests.

Sorry. I missed this recommendation. I can look into it.

internal/backend/runtime/omni/controllers/omni/reporting.go
```diff
@@ -0,0 +1,170 @@
...
+		r.ResetRestartBackoff()
+	}
+}
+
+// updateStripeSubscriptionItemQuantity adjusts the quantity of a subscription.
+// Implements retry logic with exponential backoff for transient failures.
+func updateStripeSubscriptionItemQuantity(subscriptionItemID string, count uint32) error {
+	operation := func() error {
+		newQuantity := count
+		log.Printf("Updating subscription item %s quantity to %d", subscriptionItemID, newQuantity)
```
Unix4ever, line 147: Should use `zap.Logger`. The logger is passed into the `Run` method above.

Unix4ever (approved):

Unix4ever:
TBH I'm still a bit torn if that's the right way to put it inside Omni, as it's more SaaS specific.
I don't quite understand how it can help us to enforce machine limits for on-prem. As they will have to pass our stripe creds then, which feels a bit weird.

In any case we can extract exactly this code into a separate service at any moment, if we decide to.

andrewrynhard:
> TBH I'm still a bit torn if that's the right way to put it inside Omni, as it's more SaaS specific. I don't quite understand how it can help us to enforce machine limits for on-prem. As they will have to pass our stripe creds then, which feels a bit weird.
> 
> In any case we can extract exactly this code into a separate service at any moment, if we decide to.

For now it is geared towards Stripe. It could be made more generic in the future. As you said, we can always pull this out. It fixes some important things for now and it isn't too intrusive.

PR #863, opened 2025-01-22
chore: migrate uncompressed resources data

### LINKED ISSUE #853: DB migration to re-compress resources containing machine configuration (and patches)
Only new config patches are compressed right now.

The migration should be added to the list of the Omni migrations that will:
- read all `ConfigPatch`, `ClusterMachineConfig`, `ClusterMachineConfigPatches` and `RedactedClusterMachineConfig`resources.
- for each resource check if it's using uncompressed field (`Patches` for `ClusterMachineConfigPatches` and `Data` for the others).
- call `GetUncompressedData`, then do `SetUncompressedData` to write the data back in the compressed mode and write the resource back.

DmitriyMV:
Abusing generics to compress uncompressed data in `*omni.ConfigPatch`, `*omni.ClusterMachineConfig`, `*omni.RedactedClusterMachineConfig` and `*omni.ClusterMachineConfigPatches`.

Fixes #853

### DIFF
--- a/internal/backend/runtime/omni/migration/migrations.go
+++ b/internal/backend/runtime/omni/migration/migrations.go
@@ -11,8 +11,10 @@ import (
 	"slices"
 	"strings"
 
+	"github.com/cosi-project/runtime/pkg/controller/generic"
 	"github.com/cosi-project/runtime/pkg/resource"
 	"github.com/cosi-project/runtime/pkg/resource/kvutils"
+	"github.com/cosi-project/runtime/pkg/resource/protobuf"
 	"github.com/cosi-project/runtime/pkg/safe"
 	"github.com/cosi-project/runtime/pkg/state"
 	"github.com/google/uuid"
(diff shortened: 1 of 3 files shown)


smira (approved):
I guess we should have a test for it

utkuozdemir (approved):
I wouldn't do it using generics for the sake of readability, but no big deal :)
I also suggest adding some tests. After that, ✅

DmitriyMV:
> I wouldn't do it using generics for the sake of readability, but no big deal :)

I wanted to preserve type safety so it was either duplicating code for different types or use generics. I tried to simplify the end result as much as I could

internal/backend/runtime/omni/migration/migrations.go
```diff
@@ -1333,3 +1335,95 @@ func removeMaintenanceConfigPatchFinalizers(ctx context.Context, st state.State,
 		return st.RemoveFinalizer(ctx, item.Metadata(), "MaintenanceConfigPatchController")
 	})
 }
+
+func compressMachineAndPatches(ctx context.Context, st state.State, l *zap.Logger) error {
```
Unix4ever, line 1339:
```suggestion
func compressMachineConfigsAndPatches(ctx context.Context, st state.State, l *zap.Logger) error {
```

DmitriyMV:
@smira @Unix4ever should be good to go

Unix4ever:
Now the clusterInfo fails 🤦

Unix4ever (approved):

Issue #858, opened 2025-01-21
[feature] Add Omni URL to add service account create screen


rothgar:
### Problem Description

If users create a service account via the CLI they are shown output for the service account key and omni endpoint so they can easily consume that information in their config. In the web UI they are only shown the key.

### Solution

We should add a copyable field in the web UI so they can easily copy the correct URL for the omni instance to be used in their configuration.

This is helpful in product integrations such as Portainer.

### Alternative Solutions

We could add a note about the URL so they make sure they use the correct endpoint which is different than the API endpoint shown on the home screen.

PR #873, opened 2025-01-23
feat: display `OMNI_ENDPOINT` in the service account creation UI

Unix4ever:
Fixes: https://github.com/siderolabs/omni/issues/858

### DIFF
--- a/.conform.yaml
+++ b/.conform.yaml
@@ -50,7 +50,7 @@ policies:
         - .pb.ts
         - babel.config.js
       header: |
-        // Copyright (c) 2024 Sidero Labs, Inc.
+        // Copyright (c) 2025 Sidero Labs, Inc.
         //
         // Use of this software is governed by the Business Source License
         // included in the LICENSE file.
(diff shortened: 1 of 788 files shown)


Unix4ever:
Screenshot 2025-01-23 at 21 32 57

utkuozdemir:
> Screenshot 2025-01-23 at 21 32 57

does it also have the copy button on top right on hover?

Unix4ever:
> does it also have the copy button on top right on hover?

Yep

smira:
Should we show a way to set/export these variables as ready-to-use script?

Unix4ever:
I've tried it but it looked a bit weird.
Nothing stops us from doing that of course

Unix4ever:
Ok, changed it to show the same stuff as we do in CLI.
Also updated all file headers to have `(c) 2025`

Unix4ever:
Screenshot 2025-01-23 at 22 17 47

utkuozdemir:
> Ok, changed it to show the same stuff as we do in CLI. Also updated all file headers to have `(c) 2025`

cool, btw, in the CLI output I believe we don't have the `export` command, but it's a good idea to add it, more handy to copy/paste that way.

utkuozdemir (approved):
This PR breaks my Chrome :) but LGTM

PR #885, opened 2025-01-29
fix: do not use memconn for tcpproxy load balancer

DmitriyMV:
Our `tcpproxy` module makes too many assumptions about `net.Conn` being `*net.TCPConn` so I don't think it's correct to use memconn there.


smira (commented):
this might be tricky due to local port exhaustion, this needs to be thoroughly assessed

Unix4ever (approved):

utkuozdemir:
> this might be tricky due to local port exhaustion, this needs to be thoroughly assessed

Yes that was one of the reasons for going with memconn, along with performance.

utkuozdemir:
I think this might break workload proxying completely, no?

smira:
See #886

Issue #888, opened 2025-01-30
[bug] Bug: Upgrade Requests are Sent to the Wrong Machine


utkuozdemir:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

When upgrading Talos using **Omni (UI) or `talosctl`**, upgrade requests are sent to the wrong machine.  

### Expected Behavior

Request hits the correct machine.

### Steps To Reproduce

1. Upgrade **machine1** (e.g., 1.7.7 → 1.8.2).  
2. Try upgrading **machine2** → request is sent to **machine1** instead.  
3. After ~30 min, upgrading **machine2** works, but **machine1** can no longer be upgraded.  

### Anything else?

- Happens across different versions.  
- Machines are in **maintenance mode** (not clustered).  
- Upgrading a **third machine** while **machine1** is rebooting results in errors:  
  - `i/o timeout` (expected due to reboot).  
  - `"cluster XXXX-XXXX doesn't exist"` (invalid UUID).

More info: https://sidero.zendesk.com/agent/tickets/419

PR #890, opened 2025-01-30
fix: fix node resolution cache for nodes in maintenance mode

utkuozdemir:
There was a problem with the node resolution (a.k.a. DNS) cache of the nodes. When a machine is in maintenance mode, there is a corresponding `MachineStatus` resource for it, but there isn't any `ClusterMachineIdentity`. Both of these types trigger updates in the node resolution cache. When a machine was never part of a cluster, the only source is `MachineStatus`, and the cache updates on it did not populate the machine ID in the cache. This caused the GRPC router to pick the wrong destination.

Furthermore, we did not remove the cluster and node name information from the cache when a machine was removed from a cluster. This caused the cache to contain obsolete cluster information, causing Talos GRPC proxy to not proxy the requests correctly after a machine was removed from a cluster.

Closes https://github.com/siderolabs/omni/issues/888.

### DIFF
--- a/internal/backend/dns/service.go
+++ b/internal/backend/dns/service.go
@@ -281,6 +281,7 @@ func (d *Service) updateEntryByMachineStatus(res *omni.MachineStatus) {
 
 	info := d.machineIDToInfo[res.Metadata().ID()]
 
+	info.ID = res.Metadata().ID()
 	info.TalosVersion = version
 	info.managementEndpoint = res.TypedSpec().Value.ManagementAddress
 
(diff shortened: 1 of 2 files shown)


Unix4ever (commented):
Maybe we should also test the transition unallocated -> allocated (unless we do that)

Unix4ever (approved):

utkuozdemir:
> Maybe we should also test the transition unallocated -> allocated (unless we do that)

Sure, will do.

Issue #870, opened 2025-01-23
allow user-provided bmc credentials in bare-metal infra provider


utkuozdemir:
So we can skip the credential configuring over the agent via IPMI (go full Redfish).

PR #910, opened 2025-02-06
feat: allow manual bmc configuration for bare metal machines

utkuozdemir:
Introduce the resources and controllers to configure the BMC address and credentials manually:

- Add the user-managed `omni.InfraMachineBMCConfig` resource which can be read/modified only by admins.
- Add the controller-managed `infra.BMCConfig` resource which lives in the `infra-provider` namespace and holds the matching provider ID label to satisfy the data access checks for the infra providers (can only be read by the matching infra provider and admins).

Omni part of https://github.com/siderolabs/omni/issues/870.


Unix4ever (approved):

internal/backend/runtime/omni/infraprovider/state_test.go
```diff
@@ -473,6 +526,15 @@ func prepareInfraProviderServiceAccount(ctx context.Context) context.Context {
 	return ctx
 }
 
+func prepareRole(ctx context.Context, role role.Role) context.Context {
```
DmitriyMV, line 529: I think it makes sense to unify with `prepareInfraProviderServiceAccount` so we don't create duplicate functions.

utkuozdemir: 👍 done

DmitriyMV (approved):

Issue #840, opened 2025-01-15
Put the machine into the limbo state when it join Omni for the first time


Unix4ever:
Change the process of the machine join. The process will look like this:

0. Store the unique token in the link resources.
1. Talos does the request to the siderolink API.
2. Omni takes the decision what to do with the machine:
	1. if the machine has no unique token, put the machine into the limbo state unconditionally.
	2. If the machine has unique token and it matches with what is stored in the link, let the machine join.
	3. if the machine has the unique token and it doesn't match, reject the machine.
4. For the machines in the limbo state:
	1. check the policy for the machine join (can depend on the token type) and let the machine in or leave it in the limbo state.
	2. user can accept the machine which is in the limbo state, Omni will let it fully join.
	3. if there's a link with non-empty unique token and the same UUID, Omni should generate a unique UUID for the machine in the limbo state.
5. For each machine in the limbo state which was accepted, generate a unique token.
6. For each machine with the unique token, create a link.

Migration should be handled: if the link exists and there's no unique token, let the machine in without putting it into a limbo state.

We can use the Wireguard address of the connected machine as the ID for the machines in the limbo state.

smira:
The unique machine token flow is supported since Talos (>=1.5.0, need to double check version). 

If the Omni instance has machines before that version, it should support "insecure" join for machines which report versions below that.

We probably need a flag to Omni to support joining SideroLink with old versions, and if that flag is not enabled, secure join flow is enforced.

PR #924, opened 2025-02-11
feat: implement secure node join flow

Unix4ever:
Fixes: https://github.com/siderolabs/omni/issues/840

This PR changes the Talos machine join flow drastically:

- newly joined machine is first put into a limbo state where Omni creates a temporary Wireguard connection to it.
- the controller picks up and tries to write a unique machine token to the newly joined machine, in the mean time it also resolves UUID conflicts automatically and writes UUID override to the META partition.
- the machine re-joins Omni, now with the unique token.
- the unique token is saved in the `siderolink.Link` resource and any subsequent join checks that `siderolink.Link` has matching unique token.

Siderolink manager was refactored, as it was a huge monolithic poorly testable chunk, it was split to:
- LinkStatus controller, which creates/removes wireguard peers.
- PendingMachineStatus controller, which ensures all joined machines have unique node tokens.
- Provision handler, which implements gRPC server and has all logic related to the machine acceptance now.
- PeersPool, which is used by LinkStatus controllers. It deduplicates peers creation, reuses them when possible.

Additionally updated siderolink loghandler to not accept logger connection for the machines which do not have corresponding log buffers.

Nodes which do not support secure flow are still able to join by default.
Secure join flow can be forced by setting `--disable-legacy-join-tokens` flag.


client/api/omni/specs/siderolink.proto
```diff
@@ -28,6 +28,17 @@ message SiderolinkSpec {
...
+  // the machine is accepted in Omni.
+  // Only for Talos >= 1.6.
+  string node_unique_token = 9;
+}
+
+// LinkStatusSpec is created when the link peer event was submitted.
+message LinkStatusSpec {
+  string node_subnet = 1;
+  string node_public_key = 2;
+  string virtual_addrport = 3;
```
utkuozdemir, line 41: Why virtual?

client/pkg/omni/resources/omni/labels.go
```diff
@@ -184,6 +184,9 @@ const (
 	// MachineStatusLabelInstalled means that Talos is installed on the machine.
 	// tsgen:MachineStatusLabelInstalled
 	MachineStatusLabelInstalled = SystemLabelPrefix + "installed"
+
+	// MachineUUID is set on the PendingMachine resource.
+	MachineUUID = "machine-uuid"
```
utkuozdemir, line 189: No system label prefix?

client/pkg/omni/resources/siderolink/pending_machine.go
```diff
@@ -0,0 +1,47 @@
...
+	"github.com/cosi-project/runtime/pkg/resource/meta"
+	"github.com/cosi-project/runtime/pkg/resource/protobuf"
+	"github.com/cosi-project/runtime/pkg/resource/typed"
+
+	"github.com/siderolabs/omni/client/api/omni/specs"
+	"github.com/siderolabs/omni/client/pkg/omni/resources"
+)
+
+// NewPendingMachine creates a new machine being in the limbo state.
+func NewPendingMachine(ns, id string, spec *specs.SiderolinkSpec) *PendingMachine {
```
utkuozdemir, line 18: Do we need `ns` as argument? If it always needs to be Ephemeral, we can remove it from the constructor.
Also wondering, why do we have spec as an argument unlike other resources on `SideroLink` and this one?

client/pkg/omni/resources/siderolink/pending_machine_status.go
```diff
@@ -0,0 +1,47 @@
...
+	"github.com/cosi-project/runtime/pkg/resource/meta"
+	"github.com/cosi-project/runtime/pkg/resource/protobuf"
+	"github.com/cosi-project/runtime/pkg/resource/typed"
+
+	"github.com/siderolabs/omni/client/api/omni/specs"
+	"github.com/siderolabs/omni/client/pkg/omni/resources"
+)
+
+// NewPendingMachineStatus creates a new machine being in the limbo state.
+func NewPendingMachineStatus(ns, id string) *PendingMachineStatus {
```
utkuozdemir, line 18: Should we remove `ns`?

cmd/omni/cmd/cmd.go
```diff
@@ -607,5 +607,14 @@ var initOnce = sync.OnceValue(func() *cobra.Command {
 		"enable Stripe machine usage reporting",
 	)
 
+	rootCmd.Flags().BoolVar(
+		&config.Config.DisableLegacyJoinTokens,
+		"disable-legacy-join-tokens",
+		config.Config.DisableLegacyJoinTokens,
+		`disables joining the machines running Talos below 1.6
+		by forcing strict policy that uses join token only for the initial machine join
+		then uses randomly generated secret stored in the machine META partition`,
```
utkuozdemir, line 616: no big deal, just I hope the line breaks do not mess up how it renders on `--help`.

internal/backend/runtime/omni/controllers/omni/link_status.go
```diff
@@ -0,0 +1,114 @@
...
+	"go.uber.org/zap"
+
+	"github.com/siderolabs/omni/client/api/omni/specs"
+	siderolinkres "github.com/siderolabs/omni/client/pkg/omni/resources/siderolink"
+	"github.com/siderolabs/omni/internal/backend/runtime/omni/controllers/helpers"
+	"github.com/siderolabs/omni/internal/pkg/siderolink"
+)
+
+// LinkStatusControllerName is the name of the LinkStatusController.
+const LinkStatusControllerName = "LinkStatusController"
```
utkuozdemir, line 26: Maybe make this a function like
```suggestion
func LinkStatusControllerName(typ resource.Type) string { return fmt.Sprintf("LinkStatusController[%s]", typ) }
```

Just to ensure that if we later refer to this variable from somewhere else to set it as finalizer or something, we don't introduce a bug.

internal/backend/runtime/omni/controllers/omni/link_status.go
```diff
@@ -0,0 +1,114 @@
...
+		qtransform.WithOutputKind(controller.OutputShared),
+		qtransform.WithConcurrency(32),
+	)
+}
+
+type linkStatusHandler[T res] struct {
+	peers *siderolink.PeersPool
+}
+
+func (handler *linkStatusHandler[T]) reconcileRunning(ctx context.Context, _ controller.Reader, _ *zap.Logger, link T, linkStatus *siderolinkres.LinkStatus) error {
```
utkuozdemir, line 74: Do we need the `handleInputs` thing here?

internal/backend/runtime/omni/controllers/omni/link_status.go
```diff
@@ -0,0 +1,114 @@
...
+func (handler *linkStatusHandler[T]) reconcileTearingDown(ctx context.Context, _ controller.Reader, _ *zap.Logger, link T) error {
+	return handler.peers.Remove(ctx, siderolink.GetPeerID(link.TypedSpec().Value), link.Metadata())
+}
+
+type linkSpec interface {
+	GetVirtualAddrport() string
+	GetNodePublicKey() string
+}
+
+func needsPeerUpdate(oldSpec, newSpec linkSpec) bool {
```
utkuozdemir, line 104: Make this a method on `linkStatusHandler` to reduce its scope?

internal/backend/runtime/omni/controllers/omni/pending_machine_status.go
```diff
@@ -0,0 +1,237 @@
...
+				defer c.Close() //nolint:errcheck
+
+				ctx, cancel := context.WithTimeout(ctx, time.Second*10)
+				defer cancel()
+
+				if err = handleUUIDConflicts(ctx, r, c, logger, pendingMachine, pendingMachineStatus); err != nil {
+					return err
+				}
+
+				return generateUniqueNodeToken(ctx, c, logger, pendingMachineStatus)
```
utkuozdemir, line 69: make a helper and move this 2 functions to be methods on it?

internal/backend/runtime/omni/controllers/omni/pending_machine_status.go
```diff
@@ -0,0 +1,237 @@
...
+			},
+		},
+		qtransform.WithExtraMappedInput(qtransform.MapperSameID[*siderolink.LinkStatus, *siderolink.PendingMachine]()),
```
utkuozdemir, line 72: Do we need `handleInputs` on this?

internal/backend/runtime/omni/controllers/omni/pending_machine_status.go
```diff
@@ -0,0 +1,237 @@
...
+	if configured {
+		return nil
+	}
+
+	machineUUID, ok := pendingMachine.Metadata().Labels().Get(omni.MachineUUID)
+	if !ok {
+		return fmt.Errorf("machine UUID is not set on the pending machine")
+	}
+
+	conflict, err := detectUUIDCOnflict(ctx, r, pendingMachine, machineUUID)
```
utkuozdemir, line 123: typo
```suggestion
	conflict, err := detectUUIDConflict(ctx, r, pendingMachine, machineUUID)
```

internal/pkg/logreceiver/logreceiver.go
```diff
@@ -74,6 +75,19 @@ func (srv *Server) Serve() error {
...
+				"ignoring log handler connection attempt from the remote address: machine is not accepted yet",
+				zap.String("addr", remoteAddr.String()),
+			)
+
+			if err = conn.Close(); err != nil {
+				return err
+			}
+
+			continue
+		}
```
utkuozdemir, line 89: This would take care of unaccepted bare metal machines spamming the logs with errors, right?

client/api/omni/specs/siderolink.proto
Unix4ever: that's for the wireguard over gRPC. We keep it there to keep track of it being updated.

client/pkg/omni/resources/omni/labels.go
Unix4ever: It was annotation before... Will update

internal/pkg/logreceiver/logreceiver.go
Unix4ever: yep

internal/backend/runtime/omni/controllers/omni/pending_machine_status.go
Unix4ever: I think we don't care about the link status removal here. Only for it to appear, so I think we can ignore that

cmd/omni/cmd/cmd.go
Unix4ever: yeah, it does. Thanks, fixed

internal/backend/runtime/omni/controllers/omni/pending_machine_status.go
```diff
@@ -0,0 +1,237 @@
...
+	}
+
+	return link != nil && link.TypedSpec().Value.NodeUniqueToken != "", nil
+}
+
+func getClient(
+	ctx context.Context,
+	r controller.Reader,
+	pendingMachine *siderolink.PendingMachine,
+) (*client.Client, error) {
```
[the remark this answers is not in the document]
Unix4ever: I've extracted that and reused that in the place where I copy-pasted it from. Other places are slightly different.

internal/pkg/siderolink/provision.go
```diff
@@ -0,0 +1,419 @@
...
+	_, err := h.state.Teardown(ctx, pendingMachine.Metadata())
+	if err != nil {
+		if state.IsNotFoundError(err) {
+			return nil
+		}
+
+		return nil
+	}
+
+	_, err = h.state.WatchFor(ctx, pendingMachine.Metadata(), state.WithFinalizerEmpty())
```
Unix4ever, line 249: Drop the watch for, run teardown and destroy for the ready machines, then on the next loop run teardown again and destroy ready ones.

internal/pkg/siderolink/provision.go
```diff
@@ -0,0 +1,419 @@
...
+// ProvisionHandler is the gRPC service that handles provision responses coming from the Talos nodes.
+type ProvisionHandler struct {
+	pb.UnimplementedProvisionServiceServer
+	logger                  *zap.Logger
+	state                   state.State
+	disableLegacyJoinTokens bool
+}
+
+func (h *ProvisionHandler) runCleanup(ctx context.Context) error {
+	timer := time.NewTicker(time.Second * 30)
```
Unix4ever, line 85:
```suggestion
	ticker := time.NewTicker(time.Second * 30)
```

internal/backend/runtime/omni/controllers/omni/pending_machine_status.go
```diff
@@ -0,0 +1,212 @@
...
+	"github.com/siderolabs/omni/client/pkg/omni/resources/omni"
+	"github.com/siderolabs/omni/client/pkg/omni/resources/siderolink"
+	"github.com/siderolabs/omni/internal/backend/runtime/omni/controllers/helpers"
+)
+
+// PendingMachineStatusController is a controller that writes the meta key to the pending machine.
+type PendingMachineStatusController = qtransform.QController[*siderolink.PendingMachine, *siderolink.PendingMachineStatus]
+
+// NewPendingMachineStatusController initializes PendingMachineStatusController.
+func NewPendingMachineStatusController() *PendingMachineStatusController {
```
Unix4ever, line 34: Write tests for this controller.

internal/backend/runtime/omni/controllers/omni/pending_machine_status.go
```diff
@@ -0,0 +1,212 @@
...
+		return false, err
+	}
+
+	if _, conflictFound := conflicts.Find(func(r *siderolink.PendingMachine) bool {
+		otherUUID, found := r.Metadata().Labels().Get(omni.MachineUUID)
+		if !found {
+			return false
+		}
+
+		return r.Metadata().ID() != pendingMachine.Metadata().ID() && machineUUID == otherUUID
```
Unix4ever, line 176:
```suggestion
		return r.Metadata().ID() != pendingMachine.Metadata().ID()
```

utkuozdemir (approved):

PR #926, opened 2025-02-12
Changes redirection logic so that the iframe doesn't navigate away fr…

Tricho340:
…om the parent application


Unix4ever:
Did you try this change out, does it fix your issue?

utkuozdemir:
Hi @Tricho340,

We will reach out to you on this topic, as we want to better understand your workflow and how we can enable it in another way. I'm gonna close this now, as this change is not how it needs to be solved, as the redirect on the main windows was intentional in the first place.

cc @rothgar.

Issue #886, opened 2025-01-29
workload proxy refactoring


smira:
## Ignoring Healthchecks

1. HTTP request from the user arrives to `<uniqueID>.proxy.omni`, `uniqueID` can be mapped to `cluster`, `workload service`.
2. We validate the request, and we can convert `cluster,service` pair to a list of `[IP:port, ...]`
3.  We use Go's http ReverseProxy, we rewrite the URL to be `http://<cluster>.<service>/<original URL>`.
4. In Reverse Proxy HTTP transport, we override `Dial` function to "resolve" `<cluster>.<service>` to one of the `[IP:port]` pairs.

The benefit: http.Client can re-use outgoing HTTP connections, including HTTP/1.1 keep-alive if the host part of the URL is the same.

We can control size of the HTTP transport idle connection pool, etc.

## With Healthchecks

Running healthchecks has a cost: it will do `Dial` on all configured endpoints with a configured interval, which has both CPU & network resource usage.

We can run healthchecks on demand - if some `cluster-service` pair is being used, we start healthchecking, if it is idle for some time, we can shut down healthchecks. We can bootstrap initial healthcheck state with machine connection status.

> Note: don't run healthchecks for single-node clusters.

In the flow above, on step (4), when we `Dial`, we do `upstreams.Pick()` to get a random healthy machine.

> Note: we can enable healthcheck to use latency tiers.

DmitriyMV:
- [ ] Simplify map->map to cluster. We have an cluster list with list of upstreams where for each alias only port information differs. That means that we can simplify map of upstreams from two levels to one. Create a separate map of aliases for port info. 
- [ ] Before set upstream list we "down" those who are definitely down. 
- [ ] Use latency checker from standard prometheus exporter. GotConn and WriteReq are important. GotResponse maybe interesting for their workload. Count can be used for idleConnection because in-flight requests and in-flight connections.
- [ ] Set metrics for idleConnections and inused connections: either using prometheus or using Go internals.
- [ ] Put those to Grafana.

PR #929, opened 2025-02-13
fix: enhance workloadproxy to be lazy and internally consistent  

DmitriyMV:
- [x] Ensure `Reconciler` is internally consistent on all variations of `Reconcile` call (including parallel). Track aliases and clusters
side by side.
- [x] Add tests for the above.
- [x] Replace HealthCheck logic with the actual tcp probing.
- [x] Replace probing port on alias removal. That is - if we lost an alias to the probing port, find a new one and use it.
- [x] Expose metrics. Specifically for `connectionLatency`, `requestStartLatency`, `responseStartLatency`, `inFlightRequests` and `workingProbes`. Register those in prometheus.
- [x] Add happy path test for the http.Handler.

For #886


DmitriyMV:
Pending:
- [ ] Unit tests using `httptest`
- [ ] Proper backend tiers logic.

internal/backend/workloadproxy/reconciler.go
```diff
@@ -6,83 +6,68 @@
...
 	}
 
-	return &Reconciler{
-		clusterToAliasToLBStatus: map[resource.ID]map[string]*lbStatus{},
-		aliasToCluster:           map[string]resource.ID{},
-		connProvider:             provider,
-		logger:                   logger,
-		lbLogger:                 logger.WithOptions(zap.IncreaseLevel(zapcore.ErrorLevel)),
-		logLevel:                 logLevel,
+	trans := &http.Transport{
```
smira, line 47: let's use https://github.com/hashicorp/go-cleanhttp instead of hand-rolling it

internal/backend/workloadproxy/reconciler.go
```diff
@@ -6,83 +6,68 @@
...
+		Proxy:                 http.ProxyFromEnvironment,
+		ForceAttemptHTTP2:     true,
+		MaxIdleConns:          100,
```
smira, line 50: we need to consider idle connections in a specific way, as the host would be service name

internal/backend/workloadproxy/reconciler.go
```diff
@@ -182,49 +156,50 @@ func (registry *Reconciler) GetProxy(alias string) (http.Handler, resource.ID, e
...
+	return &httputil.ReverseProxy{
+		Director:  func(req *http.Request) { req.URL.Host = hostPortForAlias(clusterID, alias) },
+		Transport: registry.transport,
+	}, clusterID, nil
+}
 
-	return proxy, clusterID, nil
+// hostPortForAlias returns a unique IP:port for the given cluster and alias.
+func hostPortForAlias(clusterID resource.ID, alias string) string {
+	return fmt.Sprintf("%s_%s:4242", clusterID, alias)
```
smira, line 172: do we have some kind of guarantee that `_` can't be in the `clusterID` ?

internal/backend/workloadproxy/reconciler.go
smira: there's also stuff like MaxConnsPerHost

internal/backend/workloadproxy/reconciler.go
DmitriyMV: That's a good question. Maybe pick some other (perhaps unicode) symbol? But there is no guarantee that it will not be picked in the future either.There is one in `ClusterDiscoveryConfig.Validate` and one in `Config.Validate` checks in Talos, but both only ensure that ClusterID is not empty. As for Omni, I'm currently try to dig usages.

smira: use `net.JoinHostPort`

internal/backend/workloadproxy/reconciler.go
```diff
@@ -182,49 +156,50 @@ func (registry *Reconciler) GetProxy(alias string) (http.Handler, resource.ID, e
...
-// hostPortForAlias returns a unique IP:port for the given cluster and alias.
-//
-// The value is arbitrary, as it uses in-memory transport, never reached via the network.
-func (registry *Reconciler) hostPortForAlias(clusterID resource.ID, alias string) string {
-	return fmt.Sprintf("%s_%s:4242", clusterID, alias)
+type backend struct {
+	Addr string
 }
+
+func (b *backend) HealthCheck(context.Context) (upstream.Tier, error) { return 0, nil }
```
smira, line 205: that should be a Dial check actually ?

internal/backend/workloadproxy/reconciler.go
```diff
@@ -118,58 +103,47 @@ func (registry *Reconciler) Reconcile(cluster resource.ID, aliasToUpstreamAddres
...
-			// we still need to close the loadbalancer, so that the health checks goroutines get terminated
-			if closeErr := tcpLB.Close(); closeErr != nil {
-				return errors.Join(startErr, fmt.Errorf("failed to close LB: %w", closeErr))
-			}
+	return nil
+}
 
-			return startErr
-		}
+func (registry *Reconciler) removeLB(cluster resource.ID, alias string) {
```
smira, line 132: we should call this on some "idle" timeout - that is if we don't have DialContext for the cluster for some timeout, we can shut it down

DmitriyMV:
I think it's better if we split this PR into two, because this is getting too big. Remaining parts:

- [ ] Find a way to test the dialer logic. Pass the cache as DI? Or add probes factory?
- [ ] Maybe use [latency checker from standard prometheus exporter](https://pkg.go.dev/github.com/prometheus/client_golang/prometheus/promhttp)? Since we already have metrics for connections exposed I dunno it's warrented. GotConn and WriteReq, WroteRequest and InFlight timings may look interesting.
- [ ] Ensure that active probes are actually deleted on time expiration.
- [ ] On setting upstream.List[T] ensure that "down" clusters are added with the negative score so we don't see those on dial attempt. Looks like it's already the case, since `NewClusterWorkloadProxyStatusController` only passes `healthyTargetHosts` to `Reconcile`.

Makefile
```diff
@@ -26,7 +26,7 @@ GRPC_GATEWAY_VERSION ?= 2.25.1
 VTPROTOBUF_VERSION ?= 0.6.0
 GOIMPORTS_VERSION ?= 0.29.0
 DEEPCOPY_VERSION ?= v0.5.6
-GOLANGCILINT_VERSION ?= v1.64.4
```
smira, line 29: is it some manual change bypassing kres?

internal/backend/workloadproxy/healthcheck.go
```diff
@@ -0,0 +1,79 @@
...
+
+	return calcTier(err, elapsed)
+}
+
+func (r *remote) healthCheck(ctx context.Context) error {
+	d := probeDialer()
+
+	c, err := d.DialContext(ctx, "tcp", r.AddrPort)
+	if err != nil {
+		r.Logger.Warn("healthcheck failed", zap.String("address", r.AddrPort), zap.Error(err))
```
Unix4ever, line 40: I'm not sure we should log that. Might be too verbose.

internal/backend/workloadproxy/reconciler.go
```diff
@@ -6,225 +6,655 @@
...
+			Help: "Number of in-use connections in the workload proxy.",
+		}),
+	}
+
+	setTransportDialer(rec)
+
+	waitForEvicts := rec.setEvictionFunc()
+	stopped := make(chan struct{})
+
+	go func(u *ttlcache.Cache[resource.ID, remoteList]) { defer close(stopped); u.Start() }(rec.activeUpstreams)
```
Unix4ever, line 90: Should probably use panichandler wrapper?

internal/backend/workloadproxy/reconciler.go
```diff
@@ -6,225 +6,655 @@
...
-	clusterToAliasToLBStatus map[resource.ID]map[string]*lbStatus
-	aliasToCluster           map[string]resource.ID
-
-	connProvider *memconn.Provider
-	logger       *zap.Logger
-	lbLogger     *zap.Logger
-	logLevel     zapcore.Level
-	mu           sync.Mutex
+// ReconcileData is the data structure used to reconcile the load balancer for a specific cluster.
+type ReconcileData struct {
```
smira, line 187: I would appreciate if these small pieces would be split into separate files/internal packages and some tests added for them to simply understand what is going on better, as this file contains at least 5 or 6 different things

Makefile
DmitriyMV: Yes, golangci lint had a weird bug on some code, but looks like it no longer reproduces on the last version of code. Anyway I created a bump for kres https://github.com/siderolabs/kres/pull/496

internal/backend/workloadproxy/reconciler.go
DmitriyMV: Okay, done! Added tests.

internal/backend/workloadproxy/healthcheck.go
DmitriyMV: Moved down to Debug level.

internal/backend/workloadproxy/reconcile_utils.go
```diff
@@ -0,0 +1,259 @@
...
+// ReplaceCluster replaces the cluster data in the aliasToCluster. If the ReconcileData is nil or doesn't contain any
+// aliases, the cluster will be removed from the structure.
+func (a *aliasToCluster) ReplaceCluster(clusterID resource.ID, rd *ReconcileData, onAliasChange func(alias) port) {
+	if rd == nil || len(rd.AliasPort) == 0 {
+		a.slc.Delete(func(data *clusterData) bool { return clusterID == data.clusterID })
+
+		return
+	}
+
+	aliases := xmaps.Map(rd.AliasPort, func(k string, v string) (alias, port) { return alias(k), port(v) })
```
Unix4ever, line 34:
```suggestion
	aliases := xmaps.Map(rd.AliasPort, func(k, v string) (alias, port) { return alias(k), port(v) })
```

go.mod
```diff
@@ -177,6 +176,7 @@ require (
 	github.com/hashicorp/go-sockaddr v1.0.7 // indirect
 	github.com/hashicorp/hcl v1.0.1-vault-7 // indirect
 	github.com/inconshreveable/mousetrap v1.1.0 // indirect
+	github.com/jellydator/ttlcache/v3 v3.3.0 // indirect
```
smira, line 179: should it be direct dependency?

internal/backend/workloadproxy/reconciler.go
```diff
@@ -6,225 +6,374 @@
...
+// active probing.
+type Description struct {
+	Alias     string
+	ClusterID string
+	Port      string
+	InUsePort string
+	Upstream  []string
+}
+
+func setTransportDialer(rec *Reconciler) {
```
smira, line 247: dumb question, but why this is not `func (rec *Reconciler) setTransportDialer()`? it seems that is actually builds a transport, so it might be `func buildTransport(rec *Reconciler) http.RountTripper` ?

internal/backend/workloadproxy/reconciler.go
```diff
@@ -6,225 +6,374 @@
...
+			}
 		}
 	}
+}
 
-	delete(aliasToLB, alias)
-	delete(registry.aliasToCluster, alias)
+// setEvictionFunc sets the eviction function for the activeProbes cache. It will shut down the active load balancer
+// probe.
+func (rec *Reconciler) setEvictionFunc() func() {
```
smira, line 325: would inlining this into the call site make it more readable? 

it seems it can be reduced to `res.activeProbes.OnEviction(func() { ... })` ?

internal/backend/workloadproxy/reconciler.go
```diff
@@ -6,225 +6,374 @@
...
-		logger:                   logger,
-		lbLogger:                 logger.WithOptions(zap.IncreaseLevel(zapcore.ErrorLevel)),
-		logLevel:                 logLevel,
+	rec := &Reconciler{
+		logger:   logger,
+		logLevel: logLevel,
+		activeProbes: ttlcache.New[resource.ID, remoteList](
+			ttlcache.WithTTL[resource.ID, remoteList](5 * time.Minute),
+		),
+		idleConnections: prometheus.NewGauge(prometheus.GaugeOpts{
```
smira, line 73: I think for debugging/verification it'd be cool to also have a simple Gauge metric which returns `len(activeProbes)` to make sure we shut them down properly on inactivity.

internal/backend/workloadproxy/reconciler.go
```diff
@@ -6,225 +6,374 @@
...
+			GotConn: func(info httptrace.GotConnInfo) {
+				if info.WasIdle {
+					rec.idleConnections.Dec()
+				}
+
+				rec.inUseConnections.Inc()
+			},
+			PutIdleConn: func(err error) {
+				if err == nil {
+					rec.idleConnections.Inc()
```
smira, line 370: I'm curious if this is accurate in case idle connection is closed on timeout - there will be no httptrace here?

go.mod
DmitriyMV: Thanks, fixed!

internal/backend/workloadproxy/reconciler.go
DmitriyMV: Reworked the whole object creation process.

internal/backend/workloadproxy/reconciler.go
DmitriyMV: Done!

internal/backend/workloadproxy/reconciler.go
DmitriyMV: Yeah, I agree that looks like a problem. I tried to fix this in the new version - instead of idleConnections we have total connections and finalizer attached to them. The difference between `inUse` and `total` should tell us how many are idle (more or less).

DmitriyMV:
@Unix4ever reworked as you proposed.

internal/backend/workloadproxy/reconcile_utils.go
```diff
@@ -0,0 +1,283 @@
...
+}
+
+type elem[E any] interface {
+	comparable
+	Equal(E) bool
+}
+
+// actionMap is a data structure that maps keys to values.
+type actionMap[K comparable, V any] struct {
+	m map[K]V
```
Unix4ever, line 230: I don't think it should be called `actionMap` anymore. As there are no actions here.

Unix4ever: And it can be actually replaced by a regular map

Unix4ever (approved):

Unix4ever:
> @Unix4ever reworked as you proposed.

Thanks! It looks better without a bunch of callbacks

internal/backend/workloadproxy/reconcile_utils.go
DmitriyMV: 👍

internal/backend/runtime/omni/controllers/omni/cluster_workload_proxy_status_test.go
```diff
@@ -218,11 +227,50 @@ func (m *mockWorkloadProxyReconciler) Reconcile(cluster resource.ID, aliasToUpst
 		m.data = map[resource.ID]map[string][]string{}
 	}
 
-	m.data[cluster] = aliasToUpstreamAddresses
+	newCluster := true
+
+	for als := range rd.AliasPort {
+		for clusterID, alsToHost := range m.data {
+			if _, ok := alsToHost[als]; ok && clusterID != cluster {
+				panic(fmt.Errorf("alias %q already exists and used by cluster %q", als, clusterID))
```
smira, line 235: I don't quite see panic here doing anything good for real?

internal/backend/runtime/omni/controllers/omni/cluster_workload_proxy_status_test.go
```diff
@@ -218,11 +227,50 @@ func (m *mockWorkloadProxyReconciler) Reconcile(cluster resource.ID, aliasToUpst
...
+			}
+
+			if clusterID == cluster {
+				m.t.Logf("replacing data for cluster %q :: %v :: %v", cluster, rd.Hosts, rd.AliasPort)
```
smira, line 239: do we need all these logs for every change every time reconcile is called (it will be called excessively)

internal/backend/workloadproxy/reconcile_utils.go
```diff
@@ -0,0 +1,231 @@
...
+	"github.com/cosi-project/runtime/pkg/resource"
+	xmaps "github.com/siderolabs/gen/maps"
+	"github.com/siderolabs/gen/pair"
+)
+
+// aliasToCluster is a data structure that maps aliases to clusters and ports. It doesn't keep tarck of active probes,
+// but it keeps track of the in-use port for the said probe for each cluster.
+type aliasToCluster struct {
+	aliasToPort map[alias]pair.Pair[*clusterData, port]
+	slc         uniqueSlice[*clusterData]
```
smira, line 22: can we do something more simple here? 

this is so many layers of abstraction, that imho it makes it virtually impossible to understand and read.

my 2 cents

internal/backend/runtime/omni/controllers/omni/cluster_workload_proxy_status_test.go
Unix4ever: I guess it's fine for the test code 🤔

DmitriyMV: This is test package, so it doesn't mater that much?

internal/backend/workloadproxy/reconcile_utils.go
DmitriyMV: This is actually my Nth attempt of refactoring this. I don't have any better solutions here, since we need a fast access for both cluster and aliases. You could store every cluster inside the slice and lookup aliases using linear search and accessing aliases inside clusterData, but that will look even worse and will add O(N) complexity to all alias lookups.

I didn't want to leave only `aliasToPort` map, because that way, when you need to update/remove cluster you have to walk entire map.

internal/backend/runtime/omni/controllers/omni/cluster_workload_proxy_status_test.go
DmitriyMV: This should't happen actually, so it's just a check in tests to ensure no bugs.

DmitriyMV:
Yay. I managed to test it locally, but had to write custom cookie generator.

```
curl --cookie 'publicKeyId=c5d7c58c7f92e11b16c3aaabf7ac398ab6dc28fb' --cookie 'publicKeyIdSignatureBase64=wrsEABYKAG0Fgme/ZeMJkPesOYq23Cj7RRQAAAAAABwAIHNhbHRAbm90YXRpb25zLm9wZW5wZ3Bqcy5vcmcSPvNPEnAezhLyus2x2iNeMgLvWA4j5FAe28qAjAwcLRYhBMXXxYx/kuEbFsOqq/esOYq23Cj7AACXlgD/YMjAg6D+gApwaL/yMYwbFI52CMeAxv+XWXpCSzn2/FMBAIsvDSySjBKKJ+7XxoeJZH+Lvoj/bgFFAsfNK2/4RaIO' -v -H "Host: 85fx58-147.proxy-us.75.xx.xx" 'https://147.75.xx.xx'
*   Trying 147.75.xx.xx:443...
* Connected to 147.75.xx.xx (147.75.xx.xx) port 443
* ALPN: curl offers h2,http/1.1
* TLSv1.3 (OUT), TLS handshake, Client hello (1):
*  CAfile: /etc/ssl/certs/ca-certificates.crt
*  CApath: /etc/ssl/certs
* TLSv1.3 (IN), TLS handshake, Server hello (2):
* TLSv1.3 (IN), TLS handshake, Encrypted Extensions (8):
* TLSv1.3 (IN), TLS handshake, Certificate (11):
* TLSv1.3 (IN), TLS handshake, CERT verify (15):
* TLSv1.3 (IN), TLS handshake, Finished (20):
* TLSv1.3 (OUT), TLS change cipher, Change cipher spec (1):
* TLSv1.3 (OUT), TLS handshake, Finished (20):
* SSL connection using TLSv1.3 / TLS_AES_128_GCM_SHA256 / X25519 / RSASSA-PSS
* ALPN: server accepted h2
* Server certificate:
*  subject: O=mkcert development certificate; OU=dmatrenichev@MacBook-Pro-M1-Lepage.local (Dmitry Matrenichev)
*  start date: Jun 21 07:40:33 2023 GMT
*  expire date: Sep 21 06:40:33 2025 GMT
*  subjectAltName: host "147.75.xx.xx" matched cert's IP address!
*  issuer: O=mkcert development CA; OU=dmatrenichev@MacBook-Pro-M1-Lepage.local (Dmitry Matrenichev); CN=mkcert dmatrenichev@MacBook-Pro-M1-Lepage.local (Dmitry Matrenichev)
*  SSL certificate verify ok.
*   Certificate level 0: Public key type RSA (2048/112 Bits/secBits), signed using sha256WithRSAEncryption
*   Certificate level 1: Public key type RSA (3072/128 Bits/secBits), signed using sha256WithRSAEncryption
* TLSv1.3 (IN), TLS handshake, Newsession Ticket (4):
* using HTTP/2
* [HTTP/2] [1] OPENED stream for https://147.75.xx.xx/
* [HTTP/2] [1] [:method: GET]
* [HTTP/2] [1] [:scheme: https]
* [HTTP/2] [1] [:authority: 85fx58-147.proxy-us.75.xx.xx]
* [HTTP/2] [1] [:path: /]
* [HTTP/2] [1] [user-agent: curl/8.5.0]
* [HTTP/2] [1] [accept: */*]
* [HTTP/2] [1] [cookie: publicKeyId=c5d7c58c7f92e11b16c3aaabf7ac398ab6dc28fb;publicKeyIdSignatureBase64=wrsEABYKAG0Fgme/ZeMJkPesOYq23Cj7RRQAAAAAABwAIHNhbHRAbm90YXRpb25zLm9wZW5wZ3Bqcy5vcmcSPvNPEnAezhLyus2x2iNeMgLvWA4j5FAe28qAjAwcLRYhBMXXxYx/kuEbFsOqq/esOYq23Cj7AACXlgD/YMjAg6D+gApwaL/yMYwbFI52CMeAxv+XWXpCSzn2/FMBAIsvDSySjBKKJ+7XxoeJZH+Lvoj/bgFFAsfNK2/4RaIO]
> GET / HTTP/2
> Host: 85fx58-147.proxy-us.75.xx.xx
> User-Agent: curl/8.5.0
> Accept: */*
> Cookie: publicKeyId=c5d7c58c7f92e11b16c3aaabf7ac398ab6dc28fb;publicKeyIdSignatureBase64=wrsEABYKAG0Fgme/ZeMJkPesOYq23Cj7RRQAAAAAABwAIHNhbHRAbm90YXRpb25zLm9wZW5wZ3Bqcy5vcmcSPvNPEnAezhLyus2x2iNeMgLvWA4j5FAe28qAjAwcLRYhBMXXxYx/kuEbFsOqq/esOYq23Cj7AACXlgD/YMjAg6D+gApwaL/yMYwbFI52CMeAxv+XWXpCSzn2/FMBAIsvDSySjBKKJ+7XxoeJZH+Lvoj/bgFFAsfNK2/4RaIO
>
< HTTP/2 200
< accept-ranges: bytes
< content-type: text/html
< date: Wed, 26 Feb 2025 19:05:38 GMT
< etag: "67a38123-267"
< last-modified: Wed, 05 Feb 2025 15:17:55 GMT
< server: nginx/1.26.3
< content-length: 615
<
<!DOCTYPE html>
<html>
<head>
<title>Welcome to nginx!</title>
<style>
html { color-scheme: light dark; }
body { width: 35em; margin: 0 auto;
font-family: Tahoma, Verdana, Arial, sans-serif; }
</style>
</head>
<body>
<h1>Welcome to nginx!</h1>
<p>If you see this page, the nginx web server is successfully installed and
working. Further configuration is required.</p>

<p>For online documentation and support please refer to
<a href="http://nginx.org/">nginx.org</a>.<br/>
Commercial support is available at
<a href="http://nginx.com/">nginx.com</a>.</p>

<p><em>Thank you for using nginx.</em></p>
</body>
</html>
* Connection #0 to host 147.75.xx.xx left intact
* 
```

internal/backend/workloadproxy/reconcile_utils.go
smira: I'm actually going crazy about it. `uniqueSlice` is inefficient as it requires full iteration all the time. 

This is yet another layer of abstraction, specifically bad container implementation just for sake of what?

aliasToCluster is what? `map[string]string`, that is `alias -> (clusterID, port)`, and that's it? 

and a separate `map[string]*Cluster` which is `alias -> Cluster`? 

if we need more lookups for whatever reason, we can also have `cluster -> []alias`, like `map[string]map[string]struct{}` ? 

but something manageable without `pair` with ugly F1/F2, maps instead of bad unique slice container?

all it has to do is to act on reconcile events, nothing more, nothing less.

DmitriyMV: Okay, lets try different approach:
- Replaced `uniqueSlice` with `map[resource.ID]*clusterData`.
- Replaced `pair.Pair` with the actual struct with proper names.
- Removed `uniqueSlice` type and tests.

Does it look better?

cmd/integration-test/make-cookies/main.go
```diff
@@ -0,0 +1,60 @@
...
+
+	// don't forget to build this with the -tags=sidero.debug
+	if len(os.Args) != 2 {
+		return fmt.Errorf("usage: %s <endpoint>", os.Args[0])
+	}
+
+	cfg := clientconfig.New(os.Args[1])
+	defer cfg.Close() //nolint:errcheck
+
+	client, err := cfg.GetClient(context.Background())
```
Unix4ever, line 38:
```suggestion
	client, err := cfg.GetClient(ctx)
```

Unix4ever (approved):

Issue #803, opened 2024-12-23
Rework machine categories UI


Unix4ever:
One idea: ![Image](https://github.com/user-attachments/assets/238888e0-8ed2-4579-b934-a36372717d3e)

Another idea is to try using the collapsible side menu for the machine filtering presets.
Save the search query/labels in the browser query.
Save the pagination in the browser query.

Show the provider configuration instructions when there are no machines of a selected type.

Show pending machines on the bare metal provider category page.

PR #941, opened 2025-02-18
feat: implement the machine categories UI for the sidebar

Unix4ever:
Display the categories as the tree view, where machines expands into `self-managed` and `auto-provisioned`.

Then `auto-provisioned` expands into each infra provider category. Infra provider can expand if it has multiple infra providers of the same type connected to the system.

Fixes: https://github.com/siderolabs/omni/issues/803


Unix4ever:
the final UI
https://github.com/user-attachments/assets/339a914c-0846-4004-ba6c-80d587043b1a

DmitriyMV (approved):

PR #943, opened 2025-02-19
feat: add a flag to enable secure join token flow

Unix4ever:
With the feature flag it is now possible to use the old flow.

The new secure join flow is not stable yet, so the default mode is legacy, which doesn't enable node unique token generation for all machines, not only the ones using too old Talos versions.

### DIFF
--- a/cmd/omni/cmd/cmd.go
+++ b/cmd/omni/cmd/cmd.go
@@ -607,13 +607,10 @@ var initOnce = sync.OnceValue(func() *cobra.Command {
 		"enable Stripe machine usage reporting",
 	)
 
-	rootCmd.Flags().BoolVar(
-		&config.Config.DisableLegacyJoinTokens,
-		"disable-legacy-join-tokens",
-		config.Config.DisableLegacyJoinTokens,
-		"disables joining the machines running Talos below 1.6 "+
-			"by forcing strict policy that uses join token only for the initial machine join, "+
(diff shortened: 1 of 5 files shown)


cmd/omni/cmd/cmd.go
```diff
@@ -607,13 +607,10 @@ var initOnce = sync.OnceValue(func() *cobra.Command {
...
-		&config.Config.DisableLegacyJoinTokens,
-		"disable-legacy-join-tokens",
-		config.Config.DisableLegacyJoinTokens,
-		"disables joining the machines running Talos below 1.6 "+
-			"by forcing strict policy that uses join token only for the initial machine join, "+
-			"then uses randomly generated secret stored in the machine META partition",
+	rootCmd.Flags().Var(
+		&config.Config.JoinTokensMode,
+		"join-tokens-mode",
+		"configures Talos machine join flow to use secure node tokens",
```
utkuozdemir, line 613: Maybe let's print here the possible values as well...

utkuozdemir (approved):

cmd/omni/cmd/cmd.go
Unix4ever: it prints them actually, as it's a custom Flag implementation.
That's how it looks like in the CLI:
```
--join-tokens-mode [legacy,legacyAllowed,strict]          configures Talos machine join flow to use secure node tokens (default legacy)
```

PR #950, opened 2025-02-24
chore: remove deprecated linter

Orzelius:
The linter "[exportloopref](https://github.com/kyoh86/exportloopref?tab=readme-ov-file)" is deprecated and deactivated. It should be removed from the list of disabled linters. https://golangci-lint.run/product/roadmap/#linter-deprecation-cycle

As of Go 1.22 the exportloopref problem no longer occurs and fixed by Go team, see [here](https://go.dev/blog/loopvar-preview)

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -144,7 +144,6 @@ linters:
     - perfsprint # complains about us using fmt.Sprintf in non-performance critical code, updating just kres took too long
     - goimports # same as gci
     - musttag # seems to be broken - goes into imported libraries and reports issues there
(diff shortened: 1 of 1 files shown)


utkuozdemir:
Thank for for the PR. However, this file is not supposed to be edited, as it is auto-generated by our project scaffolding tool https://github.com/siderolabs/kres. The change needs to be done there instead, and after it is merged, here, we need to do `make rekres`.

Orzelius:
Ah my bad. Failed to notice that the file is autogenerated. Sorry for the trouble!
Here's the equivalent in [siderolabs/kres](https://github.com/siderolabs/kres) https://github.com/siderolabs/kres/pull/495

Issue #938, opened 2025-02-17
[feature] omnictl should warn if cluster doesn’t exist in omnictl talosconfig command


UnknownBlunders:
### Problem Description

Omni automatically prepends the Omni instance name to the kubectl and talosctl context name. For example, since our Omni instance name is "Ursa-Major", our cluster named "dash" has the kubectl context "Ursa-Major-dash".
 
We don't often create new clusters, just work on the existings ones, so I work with the context name in kubectl much more than the cluster name in omni. Because of this I've come to think of our clusters as being named "Ursa-Major-dash", "Ursa-Major-heavy" etc rather than just "dash", "heavy".
 
So when I created a new cluster called "rook-ceph-testing" I went to omnictl and ran `omnictl talosconfig -c Ursa-Major-rook-ceph-testing` which cheerfully merged the context "Ursa-Major-Ursa-Major-rook-ceph-testing" into my talosconfig. I noticed the problem, removed that context, and ran `omnictl talosconfig -c rook-ceph-testing`.
 
This is not a big deal, but I wish that omni had warned my that "Ursa-Major-rook-ceph-testing" was not a cluster it had registered. The command doesn't have to fail, I think it should just warn me

### Solution
 ``` bash
$ omnictl talosconfig -c nonexistent-cluster
WARNING: The cluster “nonexistent-cluster” does not currently exist
$
```

Unix4ever:
should be easy to fix, thanks for the suggestion

PR #951, opened 2025-02-24
chore: warn if cluster doesn't exist in omnictl talosconfig command

Orzelius:
If the cluster flag is provided and cluster is not found on the omni instance warn the user when retrieving taloscinfig via the `omnictl talosconfig` command.

fixes https://github.com/siderolabs/omni/issues/938

Please let me know if there's a better approach to fixing the issue or if it should error instead of printing a warning.

### DIFF
--- a/client/pkg/omnictl/talosconfig.go
+++ b/client/pkg/omnictl/talosconfig.go
@@ -10,12 +10,15 @@ import (
 	"os"
 	"path/filepath"
 
+	"github.com/cosi-project/runtime/pkg/safe"
+	"github.com/cosi-project/runtime/pkg/state"
 	"github.com/siderolabs/talos/pkg/machinery/client/config"
 	"github.com/spf13/cobra"
 
 	"github.com/siderolabs/omni/client/pkg/client"
 	"github.com/siderolabs/omni/client/pkg/client/management"
 	"github.com/siderolabs/omni/client/pkg/constants"
+	"github.com/siderolabs/omni/client/pkg/omni/resources/omni"
 	"github.com/siderolabs/omni/client/pkg/omnictl/internal/access"
 )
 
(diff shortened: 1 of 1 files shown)


client/pkg/omnictl/talosconfig.go
```diff
@@ -108,6 +112,13 @@ func getTalosconfig(args []string) func(ctx context.Context, client *client.Clie
 		if talosconfigCmdFlags.cluster == "" {
 			data, err = client.Management().Talosconfig(ctx, opts...)
 		} else {
+			_, err = client.Omni().State().Get(ctx, resource.NewMetadata(resources.DefaultNamespace, omni.ClusterType, talosconfigCmdFlags.cluster, resource.VersionUndefined))
+			if state.IsNotFoundError(err) {
+				fmt.Printf(`warning: cluster "%s" does not exist%s`, talosconfigCmdFlags.cluster, "\n")
+			} else if err != nil {
+				return err
+			}
```
Unix4ever, line 120: LGTM in general, but I'd suggest to change a couple of things here:

```suggestion
			_, err = safe.ReaderGetByID[*omni.Cluster](ctx, client.Omni().State(), talosconfigCmdFlags.cluster)
			if err != nil {
				if !state.IsNotFoundError(err) {
					return err
				}
				
				fmt.Fprintf(os.Stderr, "warning: cluster %q does not exist\n", talosconfigCmdFlags.cluster)
			}
```

1. We have the `safe` package which supports a shorter syntax for the resource operations.
2. More Go way would be to have returns and less nesting/else/else if blocks.
3. `%q` can help to omit explicit quotes and there's no need to inject `\n`
4. The log should go to stderr

Orzelius: Thank you for the feedback! I've amended the commit. I also extracted the `getLocalTalosconfigPath` function as the `gocyclo` linter started complaining about the complexity of the `getTalosconfig` function.

Unix4ever (approved):

Unix4ever:
yep, looks good to me now. Thanks for the contribution!

PR #955, opened 2025-02-24
fix: properly read the `siderolink-disable-last-endpoint` flag

Unix4ever:
The condition was inverted :facepalm:

### DIFF
--- a/internal/pkg/siderolink/manager.go
+++ b/internal/pkg/siderolink/manager.go
@@ -537,7 +537,7 @@ func (manager *Manager) pollWireguardPeers(ctx context.Context) error {
 						spec.Connected = sinceLastHandshake < wireguard.PeerDownInterval
 					}
 
-					if !config.Config.SiderolinkDisableLastEndpoint {
+					if config.Config.SiderolinkDisableLastEndpoint {
 						spec.LastEndpoint = ""
 
 						return nil


utkuozdemir (approved):

utkuozdemir:
So do we want to revisit this and enable it properly, after doing the potential fix for the wave of updates?

smira (approved):

Unix4ever:
> So do we want to revisit this and enable it properly, after doing the potential fix for the wave of updates?

The flag is already enabled in SaaS. So when we roll the new Omni out it will cause another wave of updates.
But the next release shouldn't cause any.

utkuozdemir:
> > So do we want to revisit this and enable it properly, after doing the potential fix for the wave of updates?
> 
> The flag is already enabled in SaaS. So when we roll the new Omni out it will cause another wave of updates. But the next release shouldn't cause any.

Yes but don't we want to make it work properly (do what is necessary so the wave of updates won't happen)?
I mean, it might mean changes in our infra and not in Omni - what I'm asking is, do we want to chase that?

Unix4ever:
> > > So do we want to revisit this and enable it properly, after doing the potential fix for the wave of updates?
> > 
> > 
> > The flag is already enabled in SaaS. So when we roll the new Omni out it will cause another wave of updates. But the next release shouldn't cause any.
> 
> Yes but don't we want to make it work properly (do what is necessary so the wave of updates won't happen)? I mean, it might mean changes in our infra and not in Omni - what I'm asking is, do we want to chase that?

I don't think you can make `lastEndpoint` work properly in SaaS, that's the thing.
That's why we had this flag, which apparently wasn't working.

Issue #889, opened 2025-01-30
[bug] minor: Pre-Install Extentions list hidden when browser window too small


TimJones:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

The extension list in the "Download Installation Media" window gets hidden when the browser window is too short.

E.G.

![Image](https://github.com/user-attachments/assets/30e515f8-9876-47c0-8090-dc5980068387)

### Expected Behavior

That the list remain visible as the overlay window is scrollable.

### Steps To Reproduce

1. Navigate to an Omni instance home page
2. Click "Download Installation Media" button
3. Shrink height of browser window

PR #957, opened 2025-02-25
fix: extentions list hidden on small screens

Orzelius:
I got the omni fe dev env working and decided to submit a small bug fix as a reward to myself :)

Fix extensions scroll list being hidden on horizontally tight screens by adding a small minimal height that allows to see about two rows.

fixes https://github.com/siderolabs/omni/issues/889

https://github.com/user-attachments/assets/b939437e-d06c-4fd7-a014-85430971f136

### DIFF
--- a/frontend/src/views/omni/Extensions/ExtensionsPicker.vue
+++ b/frontend/src/views/omni/Extensions/ExtensionsPicker.vue
@@ -5,7 +5,7 @@ Use of this software is governed by the Business Source License
 included in the LICENSE file.
 -->
 <template>
-  <div class="flex flex-col gap-2 overflow-hidden">
+  <div class="flex flex-col gap-2 overflow-hidden min-h-36">
     <t-input icon="search" v-model="filterExtensions"/>
 
     <div class="grid grid-cols-4 bg-naturals-N4 uppercase text-xs text-naturals-N13 pl-2 py-2">


frontend/src/views/omni/Extensions/ExtensionsPicker.vue
```diff
@@ -5,7 +5,7 @@ Use of this software is governed by the Business Source License
 included in the LICENSE file.
 -->
 <template>
-  <div class="flex flex-col gap-2 overflow-hidden">
+  <div class="flex flex-col gap-2 overflow-hidden min-h-36">
```
Orzelius, line 8: Everywhere else in the codebase minimal height is defined in css styles and in pixels. Is there any reason for why that is? Please let me know if this should be changed to match other definitions.

Unix4ever: What you did here is correct. Pixels is more of a legacy code which we need to fix some day.

Unix4ever:
Looking at the scroll we have there makes me think that we should make the area between the header and the buttons on the bottom scrollable instead.
So that close button on the top and actions on the bottom will be always visible. But that's for some other PR I guess.

Unix4ever (approved):

PR #970, opened 2025-02-27
fix: fix exposed service links on the sidebar

utkuozdemir:
Changes introduced in c662c2e0305dddd0cafe56a782a88b887bc59001 broke the regular links from opening in a new tab, as the click handler was preventing default behavior for all links. This fix ensures that regularLink entries are not affected while maintaining proper behavior for expandable menu items.


Unix4ever (approved):

Unix4ever:
Good catch :)

smira (approved):

utkuozdemir:
> Good catch :)

It was @TimJones :)

Issue #966, opened 2025-02-26
config patch migration fixes (UpdateInputs "hack")


smira:
For Omni main:
 
* [x] Only migrate `ConfigPatch`, not other resources (`ClusterMachineConfigPatch`, etc.).
* [x] Don't re-compress already compressed resources (minor). 
* [ ] `UpdateInputs` "hack" for dependent resources (so that dependent resources think that ConfigPatch hasn't been changed). (there might be a helper/example in migrations already, verify match against controllers).
  
---

For Omni v0.47.1:

* [x] disable this migration

DmitriyMV:
Partially done in https://github.com/siderolabs/omni/pull/972 and https://github.com/siderolabs/omni/pull/985

To be fully done when solution to https://github.com/siderolabs/omni/issues/967 is found.

Unix4ever:
I think this is not relevant anymore as we dropped the random timestamp from the config resource.

PR #972, opened 2025-02-27
chore: disable compressConfigsAndMachinePatches

DmitriyMV:
Migration takes a lot of time before booting Omni.
The PR that only migrates `ConfigPatch` and no other resources will be the next one in chain.

For #966


internal/backend/runtime/omni/migration/migrations.go
```diff
@@ -1365,14 +1360,27 @@ func compressUncompressed[
...
+				zap.String("resource", md.String()),
+				zap.String("owner", md.Owner()),
+			)
+
+			l.Info("compressing resources", zap.Int("total_num", total))
+		}
+
 		for val := range items.All() {
 			if val.Metadata().Phase() != resource.PhaseRunning {
+				l.Info("skipping migration for resource in non-running phase")
```
smira, line 1382: I think with 200k resources, these per-resource log lines would produce crazy amount of logs.

If we do want logs, here, I would rather do a summary line: `processed %d resources, %d skipped as already compressed, %d re-compressed, %d skipped as non-running`

DmitriyMV: 👍 https://github.com/siderolabs/omni/pull/972/files#diff-7a3b7babb5a5155c2301d949a7201d047f946d824771b60402cd3014c6b3096dR1409

smira (approved):
updateInputs issue is not solved (follow-up PR)?

DmitriyMV:
> updateInputs issue is not solved (follow-up PR)?

Yes, so we don't block here.

Issue #944, opened 2025-02-19
Support PXE booted machine and machines running from the ISO


Unix4ever:

Unix4ever:
In the next release of Talos we might want to try to address that.
There might be a potential solution for saving this data in the bootloader.

PR #974, opened 2025-02-27
 fix: properly support the PXE and ISO machines in the secure tokens flow

Unix4ever:
The unique token flow was reworked to support machines running from PXE and ISO.

As they do not support META persistence, Omni doesn't enforce secure tokens for them.
But to distinguish machines and make the UUID conflict resolution to work, Omni now calculates the node fingerprints out of the mac addresses of all physical interfaces on the node.

So now each unique token consists of two parts:

- fingerprint.
- a random string.

Omni detects Talos installation on the machine in the following way:

- check if the pending machine status exists and it detected the system disk.
- overwrite the previous check if the existing link was labeled with the Talos being installed.
- lastly if the `MachineStatus` exists, overwrite all checks with the installed label from it (ensures bare-metal provider workflow which goes to installed to not installed and PXE booted).

Then when a machine joins Omni with some token, Omni checks if the random part is equal. If it is equal, the machine is immediately accepted.

If the random part is different and the fingerprint matches:
- if Talos is installed - reject the machine and log the warning in the logs.
- if Talos is not installed - replace the existing link with the new one (only if the request has a valid join token).

Then if nothing matches, the UUID conflict resolution kicks in. Provisioner creates a `PendingMachine` which is marked with UUID conflict label and `PendingMachineStatus` controller generates a random UUID for the node.

Fixes: https://github.com/siderolabs/omni/issues/944


internal/pkg/siderolink/provision.go
```diff
@@ -398,15 +487,63 @@ func (h *ProvisionHandler) buildProvisionContext(ctx context.Context, req *pb.Pr
...
+	if link != nil {
+		_, forceValidUniqueToken = link.Metadata().Annotations().Get(siderolink.ForceValidNodeUniqueToken)
+
+		linkNodeUniqueToken, err = jointoken.ParseNodeUniqueToken(link.TypedSpec().Value.NodeUniqueToken)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	machineStatus, err := safe.ReaderGetByID[*system.ResourceLabels[*omni.MachineStatus]](ctx, h.state, req.NodeUuid)
```
Unix4ever, line 506: We can use InfraMachineStatus last wipe time?

Change the infra machine status resource to have the last wiped token value.
Populate it this way Link -> InfraMachine -> InfraMachineStatus.

internal/pkg/siderolink/provision.go
```diff
@@ -124,20 +126,28 @@ func (h *ProvisionHandler) Provision(ctx context.Context, req *pb.ProvisionReque
...
 }
 
-func ensureResource[T res](ctx context.Context, st state.State, provisionContext *provisionContext) (T, error) {
+// ensureResource creates the link resource (PendingMachine/Link) if it doesn't exist.
+// if exists, it updates the resource, but only if the link node unique token is either:
+// - has matching random part.
+// - has matching fingerprint and Talos is not installed.
+//
+//nolint:gocognit
+func ensureResource[T res](ctx context.Context, logger *zap.Logger,
```
Unix4ever, line 148: Try to split it to parts

internal/pkg/siderolink/provision.go
```diff
@@ -178,58 +188,121 @@ func ensureResource[T res](ctx context.Context, st state.State, provisionContext
...
+
+	// the token has the same fingerprint, but the random part doesn't match
+	// return the error
+	// this case might happen if there is a hardware failure, so that the node
+	// has lost it's META partition contents
+	if linkToken.IsSameFingerprint(provisionContext.requestNodeUniqueToken) {
+		logger.Warn("machine connection rejected: the machine has the correct fingerprint, but the random token part doesn't match")
+
+		return nil, status.Error(codes.PermissionDenied, "unauthorized")
+	}
```
Unix4ever, line 298: Can move this part inside the ensureResource method probably.

internal/backend/runtime/omni/controllers/omni/pending_machine_status.go
```diff
@@ -86,13 +91,68 @@ func (handler *pendingMachineStatusHandler) reconcileRunning(
...
+	}
+
+	macAddresses := make([]string, 0, links.Len())
+
+	for link := range links.All() {
+		if !link.TypedSpec().Physical() {
+			continue
+		}
+
+		macAddresses = append(macAddresses, string(link.TypedSpec().PermanentAddr))
```
Unix4ever, line 119: can be empty, if not set use `HardwareAddr`.

internal/backend/runtime/omni/controllers/omni/pending_machine_status.go
```diff
@@ -86,13 +91,68 @@ func (handler *pendingMachineStatusHandler) reconcileRunning(
...
+	}
+
+	slices.Sort(macAddresses)
+
+	hash := sha256.New()
+
+	for _, addr := range macAddresses {
+		hash.Write([]byte(addr))
+	}
```
DmitriyMV, line 128:
```suggestion
	macAddresses := make([][]byte, 0, links.Len())

	for link := range links.All() {
		if !link.TypedSpec().Physical() {
			continue
		}

		macAddresses = append(macAddresses, link.TypedSpec().PermanentAddr)
	}

	slices.SortFunc(macAddresses. bytes.Compare)

	hash := sha256.New()

	for _, addr := range macAddresses {
		hash.Write(addr)
	}
```

internal/pkg/siderolink/provision.go
```diff
@@ -124,20 +126,28 @@ func (h *ProvisionHandler) Provision(ctx context.Context, req *pb.ProvisionReque
...
 		return nil, status.Errorf(
 			codes.FailedPrecondition,
-			"Talos version %s is not supported on this Omni instance as '--disable-legacy-join-tokens' is set",
+			"Talos version %s is not supported on this Omni instance as '--join-tokens-mode' is set to 'strict'",
 			pointer.SafeDeref(req.TalosVersion),
 		)
 	}
 
-	return h.updateLink(ctx, provisionContext)
+	return h.provision(ctx, provisionContext)
```
Unix4ever, line 134: Sanitize errors returned here.

internal/backend/runtime/omni/controllers/helpers/helpers.go
```diff
@@ -198,7 +199,7 @@ func HandleInput[T generic.ResourceWithRD, S generic.ResourceWithRD](ctx context
 
 // GetTalosClient for the machine id.
 // Automatically pick secure or insecure client.
-func GetTalosClient[T generic.ResourceWithRD](ctx context.Context, r controller.Reader, address string, machine T) (*client.Client, error) {
+func GetTalosClient[T generic.ResourceWithRD](ctx context.Context, r controller.Reader, address string, machineResource T) (*client.Client, error) {
```
DmitriyMV, line 202:
```suggestion
func GetTalosClient[T interface {
	*V
	generic.ResourceWithRD
}, V any](ctx context.Context, r controller.Reader, address string, machineResource T) (*client.Client, error) {
```

internal/backend/runtime/omni/controllers/helpers/helpers.go
```diff
@@ -212,11 +213,20 @@ func GetTalosClient[T generic.ResourceWithRD](ctx context.Context, r controller.
 			)...)
 	}
 
-	if reflect.ValueOf(machine).IsNil() {
+	if reflect.ValueOf(machineResource).IsNil() {
```
DmitriyMV, line 216:
```suggestion
	if machineResource == nil {
```

DmitriyMV (approved):

PR #979, opened 2025-02-28
chore: add `omni_machine_config_patch_size` metrics

DmitriyMV:
Add metric which describes the sizes of `uncompressed` and `compressed` config patches.

### DIFF
--- a/internal/backend/runtime/omni/controllers/omni/config_patch_metrics.go
+++ b/internal/backend/runtime/omni/controllers/omni/config_patch_metrics.go
@@ -1,0 +1,113 @@
+// Copyright (c) 2025 Sidero Labs, Inc.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+
+package omni
+
+import (
+	"context"
+	"sync"
+	"time"
+
+	"github.com/cosi-project/runtime/pkg/controller"
+	"github.com/cosi-project/runtime/pkg/safe"
+	"github.com/prometheus/client_golang/prometheus"
+	"go.uber.org/zap"
+
+	"github.com/siderolabs/omni/client/pkg/omni/resources"
+	"github.com/siderolabs/omni/client/pkg/omni/resources/omni"
+)
+
+// ConfigPatchMetricsController publishes metrics on config patches.
+//
+//nolint:govet
+type ConfigPatchMetricsController struct {
+	metricsOnce sync.Once
+	patchSizes  *prometheus.HistogramVec
+}
+
+// Name implements controller.Controller interface.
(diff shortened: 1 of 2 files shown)


Unix4ever (approved):

internal/backend/runtime/omni/controllers/omni/machine_set_status.go
```diff
@@ -125,6 +140,12 @@ func NewMachineSetStatusController() *MachineSetStatusController {
 			// config patch to machine set if the machine is allocated, checks by different layers, if is on the cluster layer,
 			// matches all machine sets
 			func(ctx context.Context, _ *zap.Logger, r controller.QRuntime, patch *omni.ConfigPatch) ([]resource.Pointer, error) {
+				if data := patch.TypedSpec().Value.GetData(); len(data) > 0 {
```
smira, line 143: I'm not sure this metric would work great way exactly, as it is skewed towards reconciles, and on no reconciles it would do exactly nothing, and prefer updated patches or non-updated. 

I don't think it's a great idea in this place.

smira: See how other metrics are done in controllers - if we _want_ to do this, we need a controller which every e.g. 30 seconds lists all config patches and updates the histogram.

DmitriyMV: Done in https://github.com/siderolabs/omni/pull/979/files#diff-3a5c8c6112d9405bfd55d2edd58f32971d18564903cd232af42762ebfd4569f7R70

internal/backend/runtime/omni/controllers/omni/config_patch_metrics.go
```diff
@@ -0,0 +1,113 @@
...
+
+// Outputs implements controller.Controller interface.
+func (ctrl *ConfigPatchMetricsController) Outputs() []controller.Output {
+	return nil
+}
+
+func (ctrl *ConfigPatchMetricsController) initMetrics() {
+	ctrl.metricsOnce.Do(func() {
+		ctrl.patchSizes = prometheus.NewHistogramVec(prometheus.HistogramOpts{
+			Name: "omni_machine_config_patch_size",
```
smira, line 54:
```suggestion
			Name: "omni_config_patch_size",
```

This is not specifically any type of a config patch, but rather all config patches.

smira (approved):

internal/backend/runtime/omni/controllers/omni/config_patch_metrics.go
DmitriyMV: 👍

PR #986, opened 2025-03-04
chore: replace slices with iteration filter in `Helper.Get`

DmitriyMV:
This PR is mostly to ensure this concept works and we can drop unnessesary slices.


utkuozdemir (commented):
Can be more optimized, but IMO way less readable. I leave approval to other people :)

smira (commented):
I agree with Utku, and also a slice of pointers is tiny, totally not worth the effort here.

PR #990, opened 2025-03-05
fix: remove force unique token annotation from the link on wipe

Unix4ever:
The problem with the previous approach is that any consequent node reboot will mess up the connection flow for the machines managed by the bare metal infra provider.

So in addition to allowing to join the nodes with the different join token for one time, reset the annotation on the link. Which makes the machine go to the uninstalled mode once again, until it is added to a cluster.

### DIFF
--- a/internal/pkg/siderolink/provision.go
+++ b/internal/pkg/siderolink/provision.go
@@ -57,6 +57,7 @@ type provisionContext struct {
 	nodeUniqueTokensEnabled   bool
 	forceValidNodeUniqueToken bool
 	supportsSecureJoinTokens  bool
+	tokenWasWiped             bool
 }
 
 func (pc *provisionContext) isAuthorizedLegacyJoin() bool {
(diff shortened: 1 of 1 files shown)


internal/pkg/siderolink/provision.go
```diff
@@ -345,14 +378,14 @@ func (h *ProvisionHandler) removePendingMachine(ctx context.Context, pendingMach
 	return nil
 }
 
-func establishLink[T res](ctx context.Context, logger *zap.Logger, st state.State, provisionContext *provisionContext, annotations ...string) (*pb.ProvisionResponse, error) {
-	link, err := createResource[T](ctx, st, provisionContext, annotations...)
+func establishLink[T res](ctx context.Context, logger *zap.Logger, st state.State, provisionContext *provisionContext, callbacks ...func(resource.Resource) error) (*pb.ProvisionResponse, error) {
```
utkuozdemir, line 381: Should we make this code less smart/generic? Something like `addAnnotations, removeAnnotations []string` as last 2 args? And same for `createResource`?

Unix4ever: sure, if you think it's better

Unix4ever: rgr

utkuozdemir (approved):

PR #1000, opened 2025-03-09
fix: duplicate resources declaration in helm deployment

nniehoff:
Moves hard coded duplicated limits from deployment template to values for helm.


TimJones:
Hi @nniehoff and many thanks for the contribution.

TimJones (approved):

Issue #1005, opened 2025-03-11
[bug] Talos being installed on the Talos Boot USB instead of local disk


ncresswell:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

When provisioning bare metal devices at the edge (eg Intel NUC's), its common practice to boot the device from a USB key that has an ISO image flashed on it... when i do this with the Omni provided ISO, the Talos installation defaults to installing on the USB drive rather than the local disk

### Expected Behavior

I expect the installer to be smart enough to know that its booted from a USB ISO and default to installing on the local disk. I shouldnt need to create a machine patch forcing this.

### Steps To Reproduce

Download an ISO from Omni
Use Etcher to burn this onto a USB key
Boot the device from the key
Try to install Talos OS, note that it defaults to installing on the USB, not the local disk

Unix4ever:
USB stick isn't detected as the ISO, but we can change the priority on the automatic disk selector for the recent Talos versions.

If we see the disk to be the USB stick, put it lower in the priority.

PR #1009, opened 2025-03-12
fix: do not select USB sticks by default

Unix4ever:
USB sticks priority is now lower than any other devices, virtual or physical, but having the other kinds of transport.

Fixes: https://github.com/siderolabs/omni/issues/1005


Unix4ever:
Depends on https://github.com/siderolabs/talemu/pull/33

utkuozdemir (approved):

internal/backend/runtime/omni/controllers/omni/machine_config_gen_options.go
```diff
@@ -79,18 +83,44 @@ func GenInstallConfig(machineStatus *omni.MachineStatus, clusterMachineTalosVers
...
+		if other.Size < selected.Size {
+			return true
+		}
+
+		return false
+	}
 
 	if installDisk == "" {
 		for _, disk := range machineStatus.TypedSpec().Value.Hardware.Blockdevices {
 			if disk.Readonly || disk.Type == storage.Disk_CD.String() {
```
smira, line 112: I would rewrite this into two statements which are both testable and (I think) easier to understand:

```
candidates := allBlockDevices()

// absolute filtering
candidates := Filter(candidates, !read_only, !CD, size > min_disk_size)

candidates := Sort(candidates, prefer !USB, sort by size asc)

pickedDisk := candidates[0]
```

smira (approved):

PR #1016, opened 2025-03-14
chore: add GOEXPERIMENT env to vscode config

Orzelius:
A small followup to [#998](https://github.com/siderolabs/omni/pull/998)

Without this option set the language server complains about synctest package not existing and in-editor compilation and test features stop working.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,4 +1,7 @@
 {
     "css.lint.unknownAtRules": "ignore",
-    "vue3snippets.enable-compile-vue-file-on-did-save-code": false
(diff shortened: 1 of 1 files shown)


utkuozdemir:
Thank you, LGTM.

One remark though - we don't really aim for matching people's environments/editors. You can see that fact from our `.gitignore` files etc. as well. As long as they are configurable on the developer side without editing project files, they do not necessarily need to be committed.

Just as a small piece of information, so that you don't spend much time on such things :).

utkuozdemir (approved):

PR #1026, opened 2025-03-17
fix: sanitize redirect URLs that can be used in the auth flow

Unix4ever:
Only allow redirects to the exposed services. Reject any other links starting from `http` and `https`.


frontend/test/unit/auth.spec.ts
```diff
@@ -0,0 +1,81 @@
...
+    allowed: boolean
+  }[] = [
+    {
+      name: "existing exposed service https",
+      url: `https://${existing}-user.proxy-us.siderolabs.io/`,
+      allowed: true
+    },
+    {
+      name: "existing exposed service legacy",
+      url: `https://p-${existing}.user.siderolabs.io/`,
```
TimJones, line 51: Just FYI, we don't support 'legacy' format expose service URLs on the SaaS now if you want to refactor this out.

Unix4ever: gotcha

Unix4ever:
I will redo this PR. It's still not perfect.

Issue #1015, opened 2025-03-14
[feature] Allow more control over exposed service URL


rdenouden:
### Problem Description

At the moment the workload exposer only allows to set a name, icon and a port.

This results in a proxy url that has some randomness and tenant name in it which might change on redeployment.

In large environments this causes that to really expose a service this way in multiple clusters running on edge nodes that to implement the URL in an enterprise system to every time someone needs access to the service a process should make an API call to omni to assure what the URL is for that service.

### Solution

Please add an annotation that depicts a custom prefix to replace the randomness in the URL for an exposed service so the url will be fully predictable and managed within the the deployment pipelines.

rdenouden:
Thank you for following up so quickly! Keep up the good work!

utkuozdemir:
You are welcome, I'll also update the documentation soon.

PR #1030, opened 2025-03-19
feat: allow setting exposed service prefixes explicitly

utkuozdemir:
Allow specifying the desired prefix instead of getting assigned a random one using `omni-kube-service-exposer.sidero.dev/prefix` annotation on the Kubernetes services.

Closes https://github.com/siderolabs/omni/issues/1015.


internal/backend/runtime/omni/controllers/omni/internal/exposedservice/exposedservice.go
```diff
@@ -0,0 +1,87 @@
+// Copyright (c) 2025 Sidero Labs, Inc.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+
+// Package exposedservice provides helpers for controllers to manage ExposedServices.
+package exposedservice
```
utkuozdemir, line 7: Moved most of the exposed service logic into this package to keep the controller package sane.

internal/backend/runtime/omni/controllers/omni/internal/exposedservice/reconciler.go
```diff
@@ -0,0 +1,300 @@
...
+
+	exposedServices map[resource.ID]*omni.ExposedService
+	usedAliases     map[string]resource.ID
+
+	cluster                string
+	workloadProxySubdomain string
+	advertisedAPIURL       string
+
+	services []*corev1.Service
+}
```
utkuozdemir, line 46: Here is no logical change - I simply gather all required variables into `Reconciler` to avoid passing them as function args.

internal/backend/runtime/omni/controllers/omni/internal/exposedservice/reconciler.go
```diff
@@ -0,0 +1,300 @@
...
+
+			return nil
+		}
+	case !hasGeneratedAlias: // no alias, generate one
+		if alias, err = reconciler.generateExposedServiceAlias(); err != nil {
+			return fmt.Errorf("error generating exposed service alias: %w", err)
+		}
+	default: // keep the existing alias
+		alias, _ = res.Metadata().Labels().Get(omni.LabelExposedServiceAlias)
+	}
```
utkuozdemir, line 174: This is the actual change in the logic:
- Check if user requested an explicit prefix (set the annotation on the k8s service)
- If they did, use that alias instead of generating a new one
- When using that alias, make sure that it is not already taken. If it is, save the error on the resource. First claimer of the prefix gets to keep it.
- When user opts out from an explicit alias (removes the annotation), figure that out by using the new annotation `omni.LabelHasGeneratedExposedServiceAlias`, and generate a fresh one.

Unix4ever (approved):

Issue #1020, opened 2025-03-15
[bug] Incorrect status display and message on Control Plane subpanel


steverfrancis:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

On a cluster with 2 control plane nodes, Etcd state shows as Not Ready, with the mouse over message of "Etcd members count doesn't match quorum, expected odd number, got 2":

<img width="550" alt="Image" src="https://github.com/user-attachments/assets/17511e85-aafd-41db-b739-25fc2034d5ef" />

### Expected Behavior

etcd is in fact in a Ready state, and working, so the "Not Ready" Status is wrong.
It should not show read "Not Ready", but a yellow "Issue detected" label.

It is also incorrect to say "etcd member count doesn't match quorum." The probably is that the member count **does** equal quorum, so there is no room for any failures.
The message should say "Etcd members is equal to the minimum quorum. The cluster cannot tolerate the loss of any members."
If there are more than the quorum, but an even number of etcd members, the message should say "Etcd should have an odd number of members for the greatest fault tolerance."

### Steps To Reproduce

Add a cluster with 2 control plane nodes.

PR #1036, opened 2025-03-20
fix: show warning instead of error for etcd quorum being at min

Unix4ever:
Fixes: https://github.com/siderolabs/omni/issues/1020

### DIFF
--- a/internal/backend/runtime/omni/pkg/check/etcd.go
+++ b/internal/backend/runtime/omni/pkg/check/etcd.go
@@ -93,12 +93,24 @@ func Etcd(ctx context.Context, r controller.Reader, clusterName string) error {
 		}
 	}
 
-	if (len(members)-1)%2 != 0 {
+	quorum := len(members)/2 + 1
+	optimalMembers := quorum*2 - 1
+
+	switch {
+	case len(members) == 2:
 		return newError(
 			specs.ControlPlaneStatusSpec_Condition_Warning,
 			false,
-			"Etcd members count doesn't match quorum, expected odd number, got %d",
+			"Etcd members count is equal to the minimum quorum %d. The cluster cannot tolerate the loss of any members",
+			quorum,
+		)
+	case len(members) != optimalMembers:
+		return newError(
+			specs.ControlPlaneStatusSpec_Condition_Warning,
+			false,
+			"Etcd members count %d is not optimal. Recommended count is %d",
 			len(members),
+			optimalMembers,
 		)
 	}
 
--- a/frontend/src/components/common/TInput/TInput.vue
+++ b/frontend/src/components/common/TInput/TInput.vue
@@ -30,9 +30,9 @@ included in the LICENSE file.
         @blur="blurHandler"
         :placeholder="placeholder"
       />
-      <div v-if="type === 'number'" class="flex flex-col select-none">
-        <t-icon class="hover:text-naturals-N14 w-2" icon="arrow-up" @click="updateValue(numberValue + step)"/>
-        <t-icon class="hover:text-naturals-N14 w-2" icon="arrow-down"  @click="updateValue(numberValue - step)"/>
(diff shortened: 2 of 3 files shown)


utkuozdemir (approved):

internal/backend/runtime/omni/pkg/check/etcd.go
```diff
@@ -97,7 +97,7 @@ func Etcd(ctx context.Context, r controller.Reader, clusterName string) error {
 		return newError(
```
smira, line 97: I think the condition `(len(members)-1)%2 != 0` is wrong, and we should re-do the check (and probably messaging):

* n = 1 => OK (while it can't tolerate any loss)
* n = 2 => NOK 
* n = 3 => OK
* n = 4 => NOK (while it can tolerate a single member loss still, quorum is 3)
* n = 5 => OK

Unix4ever: I don't think the condition is wrong. Message might be not accurate though.

Unix4ever: updated the message

internal/backend/runtime/omni/pkg/check/etcd.go
```diff
@@ -93,12 +93,28 @@ func Etcd(ctx context.Context, r controller.Reader, clusterName string) error {
...
 
-	if (len(members)-1)%2 != 0 {
+	quorum := len(members)/2 + 1
+
+	switch {
+	case len(members) == quorum && len(members) != 1:
+		return newError(
+			specs.ControlPlaneStatusSpec_Condition_Warning,
+			false,
+			"Etcd members count is equal to the minimum quorum %d. The cluster cannot tolerate the loss of any members",
```
smira, line 103: I don't think this condition is reachable except for N = 2.

Unix4ever: yeah, can simplify

internal/backend/runtime/omni/pkg/check/etcd.go
```diff
@@ -93,12 +93,28 @@ func Etcd(ctx context.Context, r controller.Reader, clusterName string) error {
 		}
 	}
 
-	if (len(members)-1)%2 != 0 {
+	quorum := len(members)/2 + 1
```
smira, line 96: I think we should calculate the value which makes sense: `loss_we_can_tolerate = len(members) - quorum`,
`optimal_members = quorum*2-1`.

if `loss_we_can_tolerate == 0` => ??? show a warning that we can't tolerate any single member loss

if `len(members) == optimal_members` => OK

if `optimal_members < len(members)` => you can tolerate loss of %d members, but you can improve by adding X members to tolerate loss of %d members


I'm not a content writer person, but the math should be correct imho

Unix4ever: > if loss_we_can_tolerate == 0 => ??? show a warning that we can't tolerate any single member loss

But only if the count of members is 2.

I don't see much difference of what I have and what you suggest.

On wording I guess we can ask @steverfrancis :)

internal/backend/runtime/omni/pkg/check/etcd.go
```diff
@@ -93,12 +93,31 @@ func Etcd(ctx context.Context, r controller.Reader, clusterName string) error {
...
+
+	switch {
+	case len(members) == 2:
+		return newError(
+			specs.ControlPlaneStatusSpec_Condition_Warning,
+			false,
+			"Etcd members count is equal to the minimum quorum %d. The cluster cannot tolerate the loss of any members",
+			quorum,
+		)
+	case len(members) < quorum:
```
smira, line 107: is this reachable? how can members be less than quorum?

Unix4ever: oh right, it's not 🤦

smira (approved):

Issue #1048, opened 2025-03-28
[feature] Add Helm Chart to a public registry


Ca-moes:
### Problem Description

A Helm Chart is available on the repository (https://github.com/siderolabs/omni/tree/main/deploy/helm/omni) but since it is not on a public registry, there isn't a way to install it, apart from having a local copy.

### Solution

Using Github Pages, host the Chart there and make it available in https://artifacthub.io/

PR #1049, opened 2025-03-28
feat: add helm chart release workflow

Ca-moes:
Related to https://github.com/siderolabs/omni/issues/1048

Needs an admin to enable GitHub pages to host the Helm Chart. After this the GitHub Action can run correctly and the chart will be available from https://siderolabs.github.io/omni/

### DIFF
--- a/.github/workflows/chart-release.yaml
+++ b/.github/workflows/chart-release.yaml
@@ -1,0 +1,31 @@
+name: release-chart
+
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  release:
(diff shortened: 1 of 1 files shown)


frezbo:
GitHub Actions for all siderolabs projects are autogenerated using [kres](https://github.com/siderolabs/kres). So kres needs to support Helm releases first

smira:
@frezbo is going to look into this

frezbo:
we'll update this to use oci helm releases and avoid using GitHub pages

Issue #1054, opened 2025-04-01
Use `<platform>-installer` image


smira:
Prepend the `<platform>` arg to the `installer` image name, this can be done for all versions of Talos.

For Talos < 1.10 it would be no-op, but for Talos > 1.10 it's critical to preserve proper kernel command line args.

PR #1057, opened 2025-04-01
feat: use `<platform>-installer[-secureboot]` as the installer image

utkuozdemir:
Make use of the change introduced in https://github.com/siderolabs/image-factory/pull/210: Switch from using `factory.talos.dev/installer[-secureboot]/...` to `factory.talos.dev/<platform>-installer[-secureboot]/...`.

Closes https://github.com/siderolabs/omni/issues/1054.


frezbo (commented):
I guess this should only happen for upgrades only, otherwise we'd just try to do a no-op upgrade

utkuozdemir:
> I guess this should only happen for upgrades only, otherwise we'd just try to do a no-op upgrade

oh right that's a really good point. Let me do that.

internal/backend/runtime/omni/controllers/omni/cluster_machine_config.go
```diff
@@ -459,9 +459,13 @@ func buildInstallImage(imageFactoryHost string, resID resource.ID, installImage
 		return "", fmt.Errorf("machine %q has no secure boot status set", resID)
 	}
 
-	installerName := "installer"
+	if installImage.Platform == "" {
```
smira, line 462: if we want to avoid re-generating machine config, in theory we could do it only for Talos 1.10+

smira: or specifically populate `.platform` only for 1.10+

utkuozdemir: changed the pr to do that

utkuozdemir:
> I guess this should only happen for upgrades only, otherwise we'd just try to do a no-op upgrade

I changed it to change the install image only for 1.10. But even without it, it wouldn't cause an upgrade call to Talos anyway, as it doesn't get detected as outdated, because there is no change in Talos version or the schematic ID. I verified that.

smira:
> I changed it to change the install image only for 1.10. But even without it, it wouldn't cause an upgrade call to Talos anyway, as it doesn't get detected as outdated, because there is no change in Talos version or the schematic ID. I verified that.

yep, my concern was that it would cause outdated config

PR #1058, opened 2025-04-02
fix: restore timeout in `OmniSuite.SetupTest`

DmitriyMV:
The change from `context.WithTimeout` was unintentional, so lets revert that.

### DIFF
--- a/internal/backend/runtime/omni/controllers/omni/etcd_backup_test.go
+++ b/internal/backend/runtime/omni/controllers/omni/etcd_backup_test.go
@@ -57,9 +57,7 @@ import (
 func TestEtcdBackupControllerSuite(t *testing.T) {
 	t.Parallel()
 
-	synctest.Run(func() {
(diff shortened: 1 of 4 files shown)


internal/backend/runtime/omni/controllers/omni/etcd_backup_test.go
```diff
@@ -82,6 +80,8 @@ func (suite *EtcdBackupControllerSuite) register2(ctrl controller.Controller, er
 }
 
 func (suite *EtcdBackupControllerSuite) SetupTest() {
+	suite.OmniSuite.ctx, suite.OmniSuite.ctxCancel = context.WithCancel(context.Background()) //nolint:fatcontext
```
smira, line 83: I would put some comment here and in the config patch cleanup to explain that we need a timeout-less context in the synctest-based tests

smira (approved):

utkuozdemir (approved):

PR #1062, opened 2025-04-03
chore: change upload and download limits flags to mbps

DmitriyMV:
This is easier to process for the cli operator, since we now force it being at least 1 Mbps.

It also fixes the problem of limiter incorrectly assuming that the reader will use the entirety of a passed buffer,
resulting in incorrect limiting.

### DIFF
--- a/cmd/omni/cmd/cmd.go
+++ b/cmd/omni/cmd/cmd.go
@@ -516,18 +516,18 @@ var initOnce = sync.OnceValue(func() *cobra.Command {
 		"maximal interval between etcd backups",
 	)
 
-	rootCmd.Flags().Int64Var(
-		&config.Config.EtcdBackup.UploadLimit,
-		"etcd-backup-upload-limit",
-		config.Config.EtcdBackup.UploadLimit,
-		"throughput limit for etcd backup uploads",
+	rootCmd.Flags().Uint64Var(
+		&config.Config.EtcdBackup.UploadLimitMbps,
+		"etcd-backup-upload-limit-mbps",
+		config.Config.EtcdBackup.UploadLimitMbps,
+		"throughput limit in Mbps for etcd backup uploads, zero means unlimited",
 	)
 
-	rootCmd.Flags().Int64Var(
-		&config.Config.EtcdBackup.DownloadLimit,
-		"etcd-backup-download-limit",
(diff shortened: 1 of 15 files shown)


cmd/omni/cmd/cmd.go
```diff
@@ -516,18 +516,18 @@ var initOnce = sync.OnceValue(func() *cobra.Command {
 		"maximal interval between etcd backups",
 	)
 
-	rootCmd.Flags().Int64Var(
+	rootCmd.Flags().Float64Var(
 		&config.Config.EtcdBackup.UploadLimit,
 		"etcd-backup-upload-limit",
 		config.Config.EtcdBackup.UploadLimit,
-		"throughput limit for etcd backup uploads",
+		"throughput limit in MiB for etcd backup uploads, that can be xx.yy",
```
utkuozdemir, line 523: MiB feels too non-granular to me, maybe KiB would be the sweet spot?

DmitriyMV: You can specify things like 0.5 or 0.02. But I wonder if it would be better to specify in Mbit (bits) rather than bytes. Usually when we talk about throughput we talk 100 Mbits or 1 Gbit.

utkuozdemir: Another idea - use "byte size" type? Introduce a new cobra flag type, use one of the byte size libraries out there (maybe [this](https://github.com/zpatrick/go-bytesize))? So users can do `10MiB` etc.

hack/release.toml
```diff
@@ -33,5 +33,5 @@ User-managed config patches not associated with an existing target (e.g., cluste
 [notes.custom-etcd-backup-throughput]
 title = "Custom Etcd Backup Throughput"
 description = """\
-The throughput for etcd backup uploads/downloads can now be limited using the `--etcd-backup-upload-limit` and `--etcd-backup-download-limit` flags.
+The throughput for etcd backup uploads/downloads can now be limited using the `--etcd-backup-upload-limit-mbps` and `--etcd-backup-download-limit-mpbs` flags.
```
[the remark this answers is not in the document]
utkuozdemir: good bot

frezbo (approved):

smira (approved):

PR #1063, opened 2025-04-04
feat: stable tag fixes #483

### LINKED ISSUE #483: [feature] Add "stable" tag to released images
### Problem Description

Currently the image releases here in GH have a rolling "latest" tag and a version tag. It would be great if we could get a "stable" tag added that rolls with the latest-non-beta version of the image. This would allow for automatic stable release upgrading without version pinning.

### Solution

rolling "stable" tag assigned to latest-non-beta image

### Alternative Solutions

_No response_

### Notes

_No response_

FedotCompot:
Creates an image with stable tag if the tag is vX.X.X without alpha/beta etc...


utkuozdemir:
Thank you very muchfor the contribution 🙏🏻.

However this is not the right place to do it - as you can see at the top of the file, this file is automatically generated by our project scaffolding tool [kres](https://github.com/siderolabs/kres) and is not meant to be edited.

The change should be done in kres instead.

Also, please open an issue for it in the kres repo first, before writing any code. We can evaluate in team and make a decision and let you know before you actually implement it.

Issue #779, opened 2024-12-11
Provide a way to forcefully reset control planes which are not fully joined to the etcd cluster


Unix4ever:
Current implementation of the reset doesn't allow removing the control plane node if it gets stuck in booting phase.
This is done to protect from races in the etcd join flow.

We can provide a way to forcefully remove the node when the user knows that the machine will never be able to join etcd.

Some labels can be used in the `MachineSetNode` (?) to force deletion.

PR #1073, opened 2025-04-10
feat: allow force-destroying a node in booting state

utkuozdemir:
Implement force-destroy logic to allow machines to be reset and removed from the cluster even if they are stuck in the booting phase. This is useful in cases like a node not being able to join to etcd, and it is known by the cluster operator that it will never be able to.

Closes https://github.com/siderolabs/omni/issues/779.


internal/backend/runtime/omni/state_validation.go
```diff
@@ -1120,3 +1120,20 @@ func infraMachineConfigValidationOptions(st state.State) []validated.StateOption
 		})),
 	}
 }
+
+func nodeForceDestroyRequestValidationOptions(st state.State) []validated.StateOption {
```
Unix4ever, line 1124: We probably need the unit test for that.

internal/backend/runtime/omni/controllers/omni/cluster_machine_config_status_test.go
```diff
@@ -384,6 +384,93 @@ func (suite *ClusterMachineConfigStatusSuite) TestResetUngraceful() {
...
+				}
+
+				return nil
+			},
+		)
+
+		suite.Require().NoError(err)
+	}
+
+	time.Sleep(time.Second * 1)
```
Unix4ever, line 466: Can we avoid having this timeout here? Maybe some more reliable way to wait?

internal/backend/runtime/omni/controllers/omni/cluster_machine_config_status.go
```diff
@@ -689,6 +704,18 @@ func (h *clusterMachineConfigStatusControllerHandler) reset(
...
+	if err != nil && !state.IsNotFoundError(err) {
+		return fmt.Errorf("failed to get node force delete request %q: %w", machineConfig.Metadata().ID(), err)
+	}
+
+	forceDestroy := nodeForceDestroyRequest != nil
+	if forceDestroy {
+		logger.Info("node is requested to be force destroyed")
+
+		graceful = false
+	}
```
Unix4ever, line 717: Maybe lets extract that to a separate method?

Unix4ever (commented):
I'd also suggest to add CLI commands to force destroy the node.

internal/backend/runtime/omni/state_validation.go
utkuozdemir: good point, added

internal/backend/runtime/omni/controllers/omni/cluster_machine_config_status.go
utkuozdemir: good idea, done

internal/backend/runtime/omni/controllers/omni/cluster_machine_config_status_test.go
utkuozdemir: actually it was not needed, so I removed it

utkuozdemir:
> I'd also suggest to add CLI commands to force destroy the node.

Done, just added `omnictl cluster machine delete <id> [--force]`

While doing it, I moved the logic of the `omnictl delete <restype> <id>` command to a common place, its own package, and started using it from both commands.

One additional thing: I made this destroy command to log human-readable logs to `stderr` instead of previously `stdout`. I think `stderr` is actually the right target for it, but let me know if you think otherwise - currently we have both in different commands and are not consistent. (I even have better ideas for our CLIs outputs).

Unix4ever (approved):

PR #1075, opened 2025-04-10
fix: fix crash in the workload proxy controller

Unix4ever:
If alias with the same cluster ID is found in the aliases map, reuse it.


Unix4ever:
Will revert it as a whole until we're sure it's stable enough.

Issue #1068, opened 2025-04-08
[feature] Make exposed service issues easier to troubleshoot


utkuozdemir:
### Problem Description

- Add/improve logs when an exposed service is created/updated
- Add/improve logs when it is accessed (be careful about excessive logging)
- Include related resources (i.e., `ExposedServices`) in the support bundles.

PR #1079, opened 2025-04-10
fix: mark all exposed services to have a non-explicit alias

utkuozdemir:
Mark all existing exposed services to have non-explicit aliases (i.e., "generated aliases"), since all old ones have generated aliases. This will prevent their aliases from being wrongly regenerated.

Additionally, add a log statement if an exposed service is updated to facilitate troubleshooting.

Small part of https://github.com/siderolabs/omni/issues/1068.

(I reproduced the alias regeneration issue locally, and verified that this fix actually addresses that correctly).


internal/backend/runtime/omni/controllers/omni/internal/exposedservice/reconciler.go
```diff
@@ -136,6 +142,22 @@ func (reconciler *Reconciler) reconcileService(ctx context.Context, exposedServi
 		return false, fmt.Errorf("error updating exposed service: %w", err)
 	}
 
+	updatedVersion := updatedService.Metadata().Version()
+	cluster, _ := updatedService.Metadata().Labels().Get(omni.LabelCluster)
+	_, isGeneratedAlias := updatedService.Metadata().Labels().Get(omni.LabelHasGeneratedExposedServiceAlias)
+
+	if !updatedVersion.Equal(version) {
```
utkuozdemir, line 149: To log only the actual updates.

Unix4ever (approved):

utkuozdemir:
@Unix4ever I changed the PR to use a field and not a label anymore, and inverted the boolean logic, so a migration is not needed anymore. Tested it as well. Can you pls have another look?

Unix4ever (approved):

Unix4ever:
Yes, I like this more:)

PR #1082, opened 2025-04-11
fix: fix existing alias check for exposed services

utkuozdemir:
Yet another fix to exposed services: for the brand new exposed services, we wrongly assumed that they already had an alias generated. Fix this by checking if they already have an alias.

### DIFF
--- a/cmd/integration-test/pkg/tests/workload_proxy.go
+++ b/cmd/integration-test/pkg/tests/workload_proxy.go
@@ -8,6 +8,7 @@ package tests
 import (
 	"compress/gzip"
 	"context"
+	"crypto/tls"
 	_ "embed"
 	"encoding/base64"
 	"errors"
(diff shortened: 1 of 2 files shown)


cmd/integration-test/pkg/tests/workload_proxy.go
```diff
@@ -94,11 +95,18 @@ func AssertWorkloadProxy(testCtx context.Context, apiClient *client.Client, clus
 			svcURL = r.TypedSpec().Value.Url
 		})
 
+		defaultTransport, ok := http.DefaultTransport.(*http.Transport)
```
smira, line 98: I would suggest to use https://github.com/hashicorp/go-cleanhttp instead when you want to build a transport.

You can build a pooled transport as well, just make sure to CloseIdleConnections when done

smira: I hate `DefaultTransport`, as if something touches it somewhere, it might have wild effects on other parts of the code which are hard to track.

utkuozdemir: Yep, I would never modify it - clone it, then modify that one. But cleanhttp indeed seems cleaner.

smira: I even mean we import e.g. some library (bad one), and it modifies `DefaultTransport` as a side-effect

utkuozdemir: Oh right, then the clone will have that modification as well.

Issue #907, opened 2025-02-05
Join Token Management


smira:
Make join tokens a resource, migrate existing join token in `ConnectionParams` to a non-expiring token.

Token:

```yaml
metadata:
   id: <token>
spec:
   expirationTime: <> # optional
   revoked: true/false # could be a label?
```

This is a user-managed resource.

We need UI to manage join tokens: create new ones, revoke existing tokens, see a list of tokens.

We could also keep track of last used time for a token. (`TokenStatus` or label/annotation on a resource?)

Download media should allow to pick a token to be used (including CLI).

Provider HMAC-based token (JWT?) should be reworked:

* can we have a per-provider secret which is used to sign JWT-style join tokens for providers
* a unique secret is generated when a provider is registered with Omni
* this secret is removed if the provider is removed from Omni
* secret can be an ECDSA/RSA public/private key.

As a migration path from providers today using "default" join token as an HMAC secret, we can mark the migrated join token as "default", and as long as it hasn't been revoked, we can keep it using it.

PR #1089, opened 2025-04-11
feat: implement multiple token support and token management

Unix4ever:
Added new pages for join token management.
The token can be created, revoked and deleted.
Also support tokens with the expiration.

Introduced new resources:
- `JoinToken` - keeps the token, it's ID is the token string.
- `JoinTokenStatus` - is generated by the controller, it calculates the information about the current token state: active, revoked or expired. And also has the information about if the token is default.
- `DefaultJoinToken` - is the singleton resource that keeps the current default token id.

The behavior of siderolink manager was changed to create a default `JoinToken` resource out of whatever is currently stored in the `siderolink.Config` resource.

`siderolink.ConnectionParams` is now generated by the controller. It's using the default token from the `DefaultJoinToken` resource.

Infra providers will get their own unique tokens, but they won't use it
until the library is updated.
So they will still rely on the default join token until updated.
    
Dropped `siderolink.ConnectionParams` usage in most of the places. This
resource is kept only for the backward compatibility reasons.

Fixes: https://github.com/siderolabs/omni/issues/907


Unix4ever:
The UI:

https://github.com/user-attachments/assets/301fc807-38b2-49cf-ab94-88b3f78d8363

Unix4ever:
Need to add join token usage resource as the input to the cluster machine config controller.
And need to fix the join token list on the download installation media modal window.

frontend/src/views/omni/Settings/JoinTokens.vue
```diff
@@ -0,0 +1,204 @@
...
+            <div class="tokens-grid flex-1">
+              <div class="flex gap-2 items-center">
+                <span class="truncate">{{ item.spec.name ?? 'initial token' }}</span>
+                <div v-if="item.spec.is_default"
+                  class="px-2 py-1 rounded bg-primary-P3 bg-opacity-10 text-primary-P3">
+                  Default
+                </div>
+              </div>
+              <div class="truncate">
+                {{ item.metadata.id }}
```
Unix4ever: Make it hidden by default. Reveal on click

internal/backend/runtime/omni/controllers/omni/join_token_status.go
```diff
@@ -0,0 +1,102 @@
...
+					joinTokenStatus.TypedSpec().Value.State = specs.JoinTokenStatusSpec_ACTIVE
+
+					joinTokenStatus.Metadata().Labels().Set(auth.LabelTokenActive, "")
+				}
+
+				joinTokenStatus.TypedSpec().Value.UseCount = useCount
+				joinTokenStatus.TypedSpec().Value.ExpirationTime = joinToken.TypedSpec().Value.ExpirationTime
+				joinTokenStatus.TypedSpec().Value.Name = joinToken.TypedSpec().Value.Name
+
+				return nil
```
Unix4ever, line 92: If expiration time is not nil use `qtransform.RequeueAfter()` to reconcile the state.

Unix4ever:
We need to URL encode join token passed in the kernel args.

Unix4ever:
Create token generation API function, make the join token resource readonly.
In the token generation API always generate the Join token ID in the same format.

Optionally allow setting token length/strength.

internal/backend/runtime/omni/state_validation.go
```diff
@@ -1119,3 +1119,97 @@ func nodeForceDestroyRequestValidationOptions(st state.State) []validated.StateO
 		})),
 	}
 }
+
+const (
+	// tsgen:MaxJoinTokenNameLength
+	maxJoinTokenNameLength = 16
+	// tsgen:MinJoinTokenLength
+	minJoinTokenLength = 8
```
utkuozdemir: can be dropped

internal/backend/runtime/omni/migration/migrations.go
```diff
@@ -1829,3 +1829,58 @@ func createProviders(ctx context.Context, st state.State, logger *zap.Logger, _
...
+	links, err := safe.ReaderListAll[*siderolink.Link](ctx, st)
+	if err != nil {
+		return err
+	}
+
+	for link := range links.All() {
+		usage := siderolink.NewJoinTokenUsage(link.Metadata().ID())
+		usage.TypedSpec().Value.TokenId = connectionParams.TypedSpec().Value.JoinToken
+
+		if err = createOrUpdate(ctx, st, usage, func(res *siderolink.JoinTokenUsage) error {
```
utkuozdemir: can use st.modify if you want now :)

internal/backend/runtime/omni/controllers/omni/connection_params.go
```diff
@@ -0,0 +1,97 @@
...
+			},
+		},
+		qtransform.WithExtraMappedInput(
+			func(context.Context, *zap.Logger, controller.QRuntime, *siderolinkres.DefaultJoinToken) ([]resource.Pointer, error) {
+				return []resource.Pointer{
+					siderolinkres.NewConfig().Metadata(),
+				}, nil
+			},
+		),
+		qtransform.WithExtraMappedInput(
```
utkuozdemir: do we need this input?

internal/backend/runtime/omni/controllers/omni/connection_params.go
```diff
@@ -0,0 +1,97 @@
...
+					siderolink.WithJoinToken(spec.JoinToken),
+					siderolink.WithGRPCTunnel(spec.UseGrpcTunnel),
+				)
+				if err != nil {
+					return err
+				}
+
+				spec.Args = strings.Join(opts.GetKernelArgs(), " ")
+
+				logger.Info(fmt.Sprintf("add this kernel argument to a Talos instance you would like to connect: %s", spec.Args))
```
utkuozdemir: let's drop this

internal/backend/runtime/omni/controllers/omni/join_token_status.go
```diff
@@ -0,0 +1,106 @@
...
+			},
+			TransformFunc: func(ctx context.Context, r controller.Reader, _ *zap.Logger, joinToken *siderolink.JoinToken, joinTokenStatus *siderolink.JoinTokenStatus) error {
+				defaultJoinToken, err := safe.ReaderGetByID[*siderolink.DefaultJoinToken](ctx, r, siderolink.DefaultJoinTokenID)
+				if err != nil && !state.IsNotFoundError(err) {
+					return err
+				}
+
+				joinTokenStatus.TypedSpec().Value.IsDefault = defaultJoinToken != nil && defaultJoinToken.TypedSpec().Value.TokenId == joinToken.Metadata().ID()
+
+				links, err := safe.ReaderListAll[*siderolink.JoinTokenUsage](ctx, r)
```
utkuozdemir:
```suggestion
				usages, err := safe.ReaderListAll[*siderolink.JoinTokenUsage](ctx, r)
```

internal/backend/runtime/omni/controllers/omni/join_token_status.go
```diff
@@ -0,0 +1,106 @@
...
+					joinTokenStatus.TypedSpec().Value.State = specs.JoinTokenStatusSpec_ACTIVE
+
+					joinTokenStatus.Metadata().Labels().Set(auth.LabelTokenActive, "")
+				}
+
+				joinTokenStatus.TypedSpec().Value.UseCount = useCount
+				joinTokenStatus.TypedSpec().Value.ExpirationTime = joinToken.TypedSpec().Value.ExpirationTime
+				joinTokenStatus.TypedSpec().Value.Name = joinToken.TypedSpec().Value.Name
+
+				if joinTokenStatus.TypedSpec().Value.ExpirationTime != nil {
```
utkuozdemir, line 84: do not requeue if it is already expired

client/pkg/omni/resources/siderolink/join_token.go
```diff
@@ -0,0 +1,49 @@
...
+// JoinTokenExtension providers auxiliary methods for JoinToken resource.
+type JoinTokenExtension struct{}
+
+// ResourceDefinition implements [typed.Extension] interface.
+func (JoinTokenExtension) ResourceDefinition() meta.ResourceDefinitionSpec {
+	return meta.ResourceDefinitionSpec{
+		Type:             JoinTokenType,
+		Aliases:          []resource.Type{},
+		DefaultNamespace: resources.DefaultNamespace,
+		PrintColumns:     []meta.PrintColumn{},
```
utkuozdemir, line 47: maybe add name as print column, and potentially other fields

internal/backend/runtime/omni/state_validation.go
```diff
@@ -1119,3 +1119,97 @@ func nodeForceDestroyRequestValidationOptions(st state.State) []validated.StateO
...
+
+	return []validated.StateOption{
+		validated.WithUpdateValidations(validated.NewUpdateValidationForType(func(_ context.Context, old, res *siderolink.JoinToken, _ ...state.UpdateOption) error {
+			if old.TypedSpec().Value.Name == res.TypedSpec().Value.Name {
+				return nil
+			}
+
+			return validateJoinTokenName(res)
+		})),
+		validated.WithCreateValidations(validated.NewCreateValidationForType(func(_ context.Context, res *siderolink.JoinToken, _ ...state.CreateOption) error {
```
utkuozdemir: no need for the create validation for all, as the mgmt api will bypass this validation. we need to do the length validation in mgmt api.

internal/pkg/siderolink/provision.go
```diff
@@ -388,6 +388,25 @@ func establishLink[T res](ctx context.Context, logger *zap.Logger, st state.Stat
 		}
 	}
 
+	tokenUsage := siderolink.NewJoinTokenUsage(link.Metadata().ID())
+
+	tokenUsage.TypedSpec().Value.TokenId = pointer.SafeDeref(provisionContext.request.JoinToken)
+
+	if err = st.Create(ctx, tokenUsage); err != nil {
```
utkuozdemir: can do modify instead

utkuozdemir (approved):

PR #1094, opened 2025-04-14
fix: config patch cleanup

smira:
Check all labels, and treat as orphaned if all labels say it's orphaned.

### DIFF
--- a/internal/backend/runtime/omni/controllers/omni/config_patch_cleanup.go
+++ b/internal/backend/runtime/omni/controllers/omni/config_patch_cleanup.go
@@ -11,6 +11,7 @@ import (
 	"time"
 
 	"github.com/cosi-project/runtime/pkg/controller"
+	"github.com/cosi-project/runtime/pkg/controller/generic"
 	"github.com/cosi-project/runtime/pkg/safe"
 	"github.com/cosi-project/runtime/pkg/state"
 	"github.com/hashicorp/go-multierror"
@@ -134,7 +135,7 @@ func (ctrl *ConfigPatchCleanupController) processConfigPatch(ctx context.Context
 		return nil
 	}
 
-	logger.Info("destroy orphaned config patch")
+	logger.Info("destroy orphaned config patch", zap.Any("labels", configPatch.Metadata().Labels().Raw()))
 
 	destroyReady, err := r.Teardown(ctx, configPatch.Metadata(), controller.WithOwner(""))
 	if err != nil {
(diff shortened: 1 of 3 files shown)


internal/backend/runtime/omni/controllers/omni/config_patch_cleanup.go
```diff
@@ -171,54 +172,59 @@ func (ctrl *ConfigPatchCleanupController) initDefaults() {
...
-		return false, err
+	if time.Since(configPatch.Metadata().Updated()) < ctrl.DeleteOlderThan {
+		return false, nil
 	}
 
-	if cluster, ok := configPatch.Metadata().Labels().Get(omni.LabelCluster); ok {
-		_, err := safe.ReaderGetByID[*omni.Cluster](ctx, r, cluster)
-		if err != nil && state.IsNotFoundError(err) {
-			return true, nil
+	for _, labelCheck := range []func() (bool, error){
```
smira, line 204: this changes the logic - a patch is orphaned only if it's orphaned on all labels (not a single label like before)

internal/backend/runtime/omni/controllers/omni/config_patch_cleanup_test.go
```diff
@@ -57,6 +57,10 @@ func (suite *ConfigPatchCleanupSuite) TestReconcile() {
 	clusterPatch := omni.NewConfigPatch(resources.DefaultNamespace, "test-cluster-patch")
 	clusterPatch.Metadata().Labels().Set(omni.LabelCluster, cluster.Metadata().ID())
 
+	clusterWithNonExistentMachinePatch := omni.NewConfigPatch(resources.DefaultNamespace, "test-cluster-no-machine-patch")
```
smira, line 60: a specific patch which is orphaned by Machine, but not orphaned by Cluster

internal/backend/runtime/omni/controllers/omni/internal/configpatch/configpatch.go
```diff
@@ -54,8 +54,11 @@ func (h *Helper) Get(machine *omni.ClusterMachine, machineSet *omni.MachineSet)
 	for patch := range clusterPatchList.All() {
 		machineSetName, machineSetOk := patch.Metadata().Labels().Get(omni.LabelMachineSet)
 		clusterMachineName, clusterMachineOk := patch.Metadata().Labels().Get(omni.LabelClusterMachine)
+		_, machineOk := patch.Metadata().Labels().Get(omni.LabelMachine)
 
 		switch {
+		// skip it, it's a machine patch (and it also has cluster label)
```
smira, line 60: this is a breaking (?) change potentially, as it skips invalid patches which have both Machine & Cluster labels

Unix4ever (approved):

utkuozdemir (approved):

Issue #1095, opened 2025-04-14
Omni v0.48.{0,1}: machine config generation using wrong Talos version contract


smira:
The problem line: 

https://github.com/siderolabs/omni/blob/b44c3ec8b0430dac5f4001a7e26ec9e3b39de6c2/internal/backend/runtime/omni/controllers/omni/cluster_machine_config.go#L283-L284

This should be `clusterConfigVersion`, not a `cluster`.

The change was introduced in https://github.com/siderolabs/omni/pull/1057, between v0.47.1 and v0.48.0. 

Any version <= v0.47.1 is not affected. Versions v0.48.0 and v0.48.1 are affected.

Version v0.48.2 will be released with a fix.

The impact depends on two variables:

* the Talos version the cluster was created with (the initial version of Talos)
* current Talos version of the cluster

Depending on these two variables, Omni would have applied a change which updates the machine configuration with the version contract being the current version of Talos instead of the initial version.

Talos upgrades might be broken if going across versions which causes new backwards incompatible changes.

## `1.2` -> `1.3`

```diff
13a14
>         disableManifestsDirectory: true
19,24d19
<     registries:
<         mirrors:
<             k8s.gcr.io:
<                 endpoints:
<                     - https://registry.k8s.io
<                     - https://k8s.gcr.io
27a23
>         apidCheckExtKeyUsage: true
41d36
<     aescbcEncryptionSecret: yQyiM0JJH8ldLlbhaxu1cvZKunUhzWRsS83dH0I2RPQ=
71a67,71
>         auditPolicy:
>             apiVersion: audit.k8s.io/v1
>             kind: Policy
>             rules:
>                 - level: Metadata
```

* `disableManifestsDirectory` - should be fine
* `k8s.gcr.io` - should not matter, as k8s.gcr.io doesn't contain any recent Kubernetes images
* `apidCheckExtKeyUsage` - triggers a reboot, should be fine, as Talos got upgraded to 1.3+
* `aescbcEncryptionSecret` - this is bad if it got dropped, going to make `kube-apiserver` broken
* `auditPolicy` - should be noop

## `1.3` -> `1.4`

No changes.

## `1.4` -> `1.5`

```diff
>         diskQuotaSupport: true # Enable XFS project quota support for EPHEMERAL partition and user disks.
```

(this triggers a reboot)

## `1.5` -> `1.6`

```diff
>         # KubePrism - local proxy/load balancer on defined port that will distribute
>         kubePrism:
[...]

PR #1098, opened 2025-04-14
fix: use correct version contract for machine config generation

smira:
See #1095

### DIFF
--- a/internal/backend/runtime/omni/controllers/omni/cluster_machine_config.go
+++ b/internal/backend/runtime/omni/controllers/omni/cluster_machine_config.go
@@ -272,22 +272,33 @@ type clusterMachineConfigControllerHelper struct {
 	imageFactoryHost string
 }
 
-const talosVersionLatest = "latest"
-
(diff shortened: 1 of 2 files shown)


internal/backend/runtime/omni/controllers/omni/cluster_machine_config.go
```diff
@@ -272,22 +272,33 @@ type clusterMachineConfigControllerHelper struct {
...
+	// this is the version of Talos at the moment the cluster got created
+	//
+	// [NOTE]: this should be kept a constant for the lifetime of the cluster,
+	// as it dictates the Talos machinery config generation defaults.
+	// If this value is changed, it will cause the machine configuration to be regenerated
+	// with new version contract (defaults), and might cause unexpected issues.
+	//
+	// The desired version of Talos for this machine (not for config generation), but for the
+	// e.g. install image is stored in MachineConfigGenOptions.
+	initialTalosVersion := clusterConfigVersion.TypedSpec().Value.Version
```
smira, line 290: this is the line of the actual fix, everything else is to simplify/make the code more clear so we don't make this mistake in the future

Unix4ever (approved):

Issue #1097, opened 2025-04-14
revert the machine config version contract change


smira:
Revert the change which introduced the bug, but figure out a plan to avoid Talos reboots:

* `diskQuotaSupport` feature shouldn't change to avoid a reboot.
* `apidCheckExtKeyUsage` feature shouldn't change to avoid a reboot.

PR #1100, opened 2025-04-15
fix: create config patches to prevent reboot on version contract revert

utkuozdemir:
For the needed machines, create a config patch with a low priority to preserve the config changes which require the machines to be rebooted.

Prevents the reboot due to the fix in https://github.com/siderolabs/omni/issues/1097.
Part of the fix for https://github.com/siderolabs/omni/issues/1095.

### DIFF
--- a/internal/backend/runtime/omni/migration/migration_test.go
+++ b/internal/backend/runtime/omni/migration/migration_test.go
@@ -10,6 +10,7 @@ import (
 	"crypto/rand"
 	"errors"
 	"fmt"
+	"slices"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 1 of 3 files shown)


internal/backend/runtime/omni/migration/migrations.go
```diff
@@ -1537,3 +1538,117 @@ func moveEtcdBackupStatuses(ctx context.Context, st state.State, logger *zap.Log
...
+// to preserve the config changes the revert would cause which would require a reboot, namely:
+// - preserve machine.features.diskQuotaSupport setting
+// - preserve machine.features.apidCheckExtKeyUsage setting.
+func createVersionContractRevertConfigPatch(ctx context.Context, st state.State, logger *zap.Logger) error {
+	clusterConfigVersionList, err := safe.StateListAll[*omni.ClusterConfigVersion](ctx, st)
+	if err != nil {
+		return fmt.Errorf("failed to list cluster config versions: %w", err)
+	}
+
+	clusterMachineConfigVersionMap := make(map[string]string, clusterConfigVersionList.Len())
```
Unix4ever, line 1555:
```suggestion
	clusterConfigVersionMap := make(map[string]string, clusterConfigVersionList.Len())
```

internal/backend/runtime/omni/migration/migration_test.go
```diff
@@ -1798,6 +1798,138 @@ func (suite *MigrationSuite) TestMoveEtcdBackupStatuses() {
...
+`)))
+
+	// no patch expected, since it will be in tearing down phase
+	cluster14config3 := omni.NewClusterMachineConfig(resources.DefaultNamespace, "cluster-14-machine-3")
+	cluster14config3.Metadata().Labels().Set(omni.LabelCluster, "cluster-14")
+
+	suite.Require().NoError(cluster14config1.TypedSpec().Value.SetUncompressedData([]byte(`machine:
+  features:
+	foo: bar
+    diskQuotaSupport: true
```
Unix4ever, line 1870: I wonder if that's how displays it, or the idents are messed up here?

internal/backend/runtime/omni/migration/migration_test.go
```diff
@@ -1798,6 +1798,138 @@ func (suite *MigrationSuite) TestMoveEtcdBackupStatuses() {
...
+    quux: corge
+`)))
+
+	// patch expected: **only** diskQuotaSupport needs to be preserved
+	cluster14config1 := omni.NewClusterMachineConfig(resources.DefaultNamespace, "cluster-14-machine-1")
+	cluster14config1.Metadata().Labels().Set(omni.LabelCluster, "cluster-14")
+
+	suite.Require().NoError(cluster14config1.TypedSpec().Value.SetUncompressedData([]byte(`machine:
+  features:
+	foo: bar
```
Unix4ever, line 1849: Probably you have tabs here

internal/backend/runtime/omni/migration/migration_test.go
```diff
@@ -1798,6 +1798,138 @@ func (suite *MigrationSuite) TestMoveEtcdBackupStatuses() {
...
+
+				buf, err := patch.TypedSpec().Value.GetUncompressedData()
+				suite.Require().NoError(err)
+
+				dataStr := string(buf.Data())
+
+				var m map[string]any
+
+				// assert that it is a valid yaml
+				suite.Require().NoError(yaml.Unmarshal([]byte(dataStr), &m))
```
Unix4ever, line 1907: I'd suggest to decode the file using talos config parser.

Unix4ever (approved):
LGTM with minor comments

internal/backend/runtime/omni/migration/migration_test.go
utkuozdemir: Yep, thanks, copy/pasting messed it up in a few places. Now enabled displaying whitespace in my editor so it won't happen in the future.

internal/backend/runtime/omni/migration/migration_test.go
utkuozdemir: Changed it to use `omni.ValidateConfigPatch(...)`, which does exactly that under the hood.

internal/backend/runtime/omni/migration/migrations.go
```diff
@@ -1537,3 +1538,116 @@ func moveEtcdBackupStatuses(ctx context.Context, st state.State, logger *zap.Log
...
+			patchData += "    apidCheckExtKeyUsage: true\n"
+
+			numApidCheckExtKeyUsageEnabled++
+		}
+
+		configPatch := omni.NewConfigPatch(resources.DefaultNamespace, fmt.Sprintf("000-%s-preserve-version-contract", uuid.NewString()))
+
+		configPatch.Metadata().Labels().Set(omni.LabelCluster, cluster)
+		configPatch.Metadata().Labels().Set(omni.LabelClusterMachine, clusterMachineConfig.Metadata().ID())
+		configPatch.Metadata().Labels().Set(omni.LabelSystemPatch, "")
```
utkuozdemir, line 1634: @Unix4ever what do you think about this? I thought also about leaving it out, so it'd be more obvious to the users, and if they wanted, they could delete it on the UI.

frezbo:
are we reading the existing machine config to see what's the correct diskquota and apidCheckExtKeyUsage values?

utkuozdemir:
> are we reading the existing machine config to see what's the correct diskquota and apidCheckExtKeyUsage values?

Yep, it is happening here: https://github.com/siderolabs/omni/pull/1100/files#diff-7a3b7babb5a5155c2301d949a7201d047f946d824771b60402cd3014c6b3096dR1606-R1607

internal/backend/runtime/omni/migration/migrations.go
Unix4ever: I don't know. It has downsides either way.
Maybe we can remove the patch and add meaningful name and description annotations on the resources.

It can even contain the URL to the issue in the description.

internal/backend/runtime/omni/migration/migrations.go
```diff
@@ -1537,3 +1538,116 @@ func moveEtcdBackupStatuses(ctx context.Context, st state.State, logger *zap.Log
...
+	if err != nil {
+		return fmt.Errorf("failed to list cluster config versions: %w", err)
+	}
+
+	clusterConfigVersionMap := make(map[string]string, clusterConfigVersionList.Len())
+	for clusterConfigVersion := range clusterConfigVersionList.All() {
+		clusterConfigVersionMap[clusterConfigVersion.Metadata().ID()] = clusterConfigVersion.TypedSpec().Value.Version
+	}
+
+	clusterMachineConfigList, err := safe.StateListAll[*omni.ClusterMachineConfig](ctx, st)
```
smira, line 1560: one thing to note here is that this is the last generated config, not the one last applied yet, but I don't want to overcomplicate this migration even further

internal/backend/runtime/omni/migration/migrations.go
utkuozdemir: changed it to do that.

smira (approved):

internal/backend/runtime/omni/migration/migrations.go
utkuozdemir: discussed with @Unix4ever and changed it to check the last applied config version, skip if it is not up-to-date. IT is not 100% buletproof, but probably will give better results (particularly for disconnected machines).

Unix4ever (approved):

internal/backend/runtime/omni/migration/migration_test.go
```diff
@@ -1798,6 +1799,183 @@ func (suite *MigrationSuite) TestMoveEtcdBackupStatuses() {
...
+	cluster14config4.Metadata().Labels().Set(omni.LabelCluster, "cluster-14")
+
+	suite.Require().NoError(cluster14config4.TypedSpec().Value.SetUncompressedData([]byte(`machine:
+  features:
+    foo: bar
+    diskQuotaSupport: true
+`)))
+
+	cluster14configStatus4 := omni.NewClusterMachineConfigStatus(resources.DefaultNamespace, "cluster-14-machine-4")
+	cluster14configStatus4.TypedSpec().Value.ClusterMachineConfigVersion = "0" // mark the v0 to be applied, but v1 was not applied
```
utkuozdemir, line 1903: This simulates a config which was generated but not applied (e.g., a disconnected machine).
We assert that it doesn't get the patch.

Issue #1099, opened 2025-04-14
[bug] Schema issue in Auto-Provision Machine Class


pierrop:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

Hello,
We're working on a custom Infra Provider. Recently, we've encountered an issue when creating or editing an Auto-Provision Machine Class. Specifically, we are unable to modify our provider's custom settings, only the Talos configurations are editable via the UI:

![Image](https://github.com/user-attachments/assets/8b9d46ca-6266-4e7f-ac8c-425452475292)

Inspecting Chrome DevTools shows an error related to schema compilation from the provider: 

![Image](https://github.com/user-attachments/assets/f5448003-82e8-41e4-ba8a-d87ee17a7696)

### Expected Behavior

We expect to see and edit the provider-specific custom settings as defined in our Provider Config schema when creating or updating an Auto-Provision Machine Class.

### Steps To Reproduce

- Connect a custom infrastructure provider to Omni with the following Provider Config schema:
```json
{
  "type": "object",
  "properties": {
    "location": {
      "type": "string",
      "enum": ["EU: Montreuil 1"]
    },
    "instance_type": {
      "type": "string",
      "enum": ["bm7.hmm.8"]
    },
    "talos_extensions": {
      "type": "string"
    },
    "talos_schematic": {
      "type": "string"
    }
  },
  "required": [
    "location",
    "instance_type"
  ]
}
```
- Attempt to create/edit an Auto-Provision Machine Class to observe that custom provider settings cannot be configured.

### What browsers are you seeing the problem on?

Chrome

PR #1109, opened 2025-04-15
fix: move JSON schema forms validation to backend

Unix4ever:
AJV library is using unsafe evals inside. When we enabled CSP for Omni, it got broken.
The simplest way around that is to delegate JSON forms validation to the backend.

Fixes: https://github.com/siderolabs/omni/issues/1099


internal/backend/grpc/validate.go
```diff
@@ -70,3 +83,88 @@ func validatePGPPublicKey(armored []byte, opts ...pgp.ValidationOption) (publicK
...
+	s.jsonSchemaCacheMu.Lock()
+	defer s.jsonSchemaCacheMu.Unlock()
+
+	shaSum := sha256.Sum256([]byte(request.Schema))
+	shaSumString := hex.EncodeToString(shaSum[:])
+
+	var err error
+
+	if s.jsonSchemaCache == nil {
+		s.jsonSchemaCache = map[string]*jsonschema.Schema{}
```
utkuozdemir, line 101: I know that we require auth for this function, but a potential attack vector could be posting many different large schemas to this endpoint. Anything I post would be kept in memory forever. Maybe we should limit the cache size? (maybe use a library for it - like the cost-based eviction here or sth https://github.com/hypermodeinc/ristretto#features)

internal/backend/runtime/helpers/jsonschema.go
```diff
@@ -0,0 +1,49 @@
...
+		return nil, err
+	}
+
+	return compiler.Compile(filename)
+}
+
+// ValidateSchema validates the YAML file against the JSON schema.
+func ValidateSchema(data string, schema *jsonschema.Schema) error {
+	// NaN type causes jsonschema validator to crash with nil reference error
+	data = regexp.MustCompile(`(?i)\.nan`).ReplaceAllString(data, "null")
```
utkuozdemir, line 37: We can extract `regexp.MustCompile(`(?i)\.nan`)` as package-level variable.

internal/backend/grpc/validate.go
Unix4ever: used expirable.LRU without expiration but with the size limit.
We use it for the Talos clients already.

internal/backend/grpc/validate.go
```diff
@@ -70,3 +83,81 @@ func validatePGPPublicKey(armored []byte, opts ...pgp.ValidationOption) (publicK
 		expiration: expiration,
 	}, nil
 }
+
+func (s *managementServer) ValidateJSONSchema(ctx context.Context, request *management.ValidateJsonSchemaRequest) (*management.ValidateJsonSchemaResponse, error) {
```
smira, line 87: It almost feels like it should have been a dry-run Update/Create which would run the validation and report errors back? 


(I don't want to block this PR, just saying)

internal/backend/grpc/management.go
```diff
@@ -72,6 +74,29 @@ type JWTSigningKeyProvider interface {
...
+	return &managementServer{
+		omniState:             omniState,
+		jwtSigningKeyProvider: jwtSigningKeyProvider,
+		logHandler:            logHandler,
+		logger:                logger,
+		dnsService:            dnsService,
+		imageFactoryClient:    imageFactoryClient,
+		auditor:               auditor,
+		omniconfigDest:        omniconfigDest,
+		jsonSchemaCache:       expirable.NewLRU[string, *jsonschema.Schema](64, nil, 0),
```
utkuozdemir, line 96: Probably not a big deal, but this size (64) is for the number of items I believe, not caring about the size of a single entry. That's why I suggested something like a [cost-based eviction](https://github.com/hypermodeinc/ristretto), each item's cost would be the size of their raw schema bytes. I don't know if there's a way to do it with this LRU.

internal/backend/runtime/helpers/jsonschema.go
```diff
@@ -0,0 +1,51 @@
+// Copyright (c) 2025 Sidero Labs, Inc.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+
+package helpers
```
utkuozdemir, line 6: Don't want to nitpick much, but since we also want to organize things a bit better, what do you think about creating a `jsonschema` package and moving these into it? The function names would then be `Parse` and `Validate`.

internal/backend/grpc/management.go
smira: I'd say just removing free-form JSON schema removes the whole problem, and probably cache isn't actually needed?

We validate a specific provider schema

internal/backend/grpc/validate.go
Unix4ever: It reports the errors back but they're not in the machine readable format.

internal/backend/grpc/management.go
Unix4ever: The component for the JsonForms is generic. I'd rather validate it on the frontend, but as that's not possible anymore I moved it to backend.

Of course it's doable, we can lock it to the providers only. But don't really want to do that.

internal/backend/runtime/helpers/jsonschema.go
Unix4ever: makes sense, let's do it

internal/backend/grpc/management.go
Unix4ever: We could implement a generic resource viewer and editor based on this JsonForm component in the future.
If we lock it to the provider schema, it won't be possible.

utkuozdemir (approved):

PR #1142, opened 2025-04-26
fix: update to fix "make mkcert-install" 

miledxz:
When I run on mac `make mkcert-install`  
the issue is,  `Makefile:94: *** missing separator.  Stop.`
https://github.com/siderolabs/omni/blob/main/Makefile#L94

@smira What do you think about this update does this works on Linux for you maybe (just asking I don't know on what OS are you)

or maybe it's better just to make HELP_MENU_HEADER unexported ? 
as it's not closing this, https://github.com/siderolabs/omni/blob/main/Makefile#L135

cheers !

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -87,6 +87,8 @@ REMOVE_VOLUMES ?= false
 # help menu
 
 export define HELP_MENU_HEADER
+
+define BUILD_SETUP_GUIDE
 # Getting Started
 
 To build this project, you must have the following installed:


smira:
The `Makefile` is auto-generated by [kres](https://github.com/siderolabs/kres), so if the fix is needed, it is needed in `kres`.

In general, I think Darwin `make` (which should be BSD make by its origins) is not that compatible with GNU make which is more used/tested with `Makefile`s. So my 2 cents is to install/use GNU make, and also if there are trivial compatibility issues, we can/should fix them in kres.

PR #1146, opened 2025-04-28
fix: remove "Generated by Omni..." comment from machine config

utkuozdemir:
We should not need this comment at all - while we cannot recall the original reason for adding it, removing this should probably be fine.


smira (approved):
😨

Unix4ever (approved):

internal/backend/runtime/omni/controllers/omni/cluster_machine_config_status.go
```diff
@@ -195,22 +195,17 @@ func NewClusterMachineConfigStatusController(imageFactoryHost string) *ClusterMa
...
-					if grpcSt != nil && grpcSt.Code() == codes.InvalidArgument {
-						configStatus.TypedSpec().Value.LastConfigError = grpcSt.Message()
+							return nil
+						}
 
-						return nil
+						return fmt.Errorf("failed to apply config to machine '%s': %w", machineConfig.Metadata().ID(), err)
 					}
-
-					return fmt.Errorf("failed to apply config to machine '%s': %w", machineConfig.Metadata().ID(), err)
```
utkuozdemir, line 213: @Unix4ever @smira for visibility - this was a bug surfaced by this change - it was causing the machine status to be stuck in the `Reconfiguring` phase, as we were short circuiting without updating its versions/sha etc. if the config was up to date.

It seems there is no other issues with his change - I also tested manually.

smira (approved):

Issue #1086, opened 2025-04-11
[feature] ability to opt out of integrity behaviors with AWS S3 SDK


brantgurga:
### Problem Description

We use Backblaze as the S3-compatible storage for the etcd backups in Omni. This worked until yesterday when 0.48.0 seemed to roll out for our managed instance. Backblaze acknowledges the incompatibility at https://www.backblaze.com/docs/cloud-storage-use-the-aws-sdk-for-go-with-backblaze-b2

I am trying to see if they will be supporting those headers.

### Solution

It'd be great if there was a way to opt out of those integrity behaviors so the backups still work. There is a comment at https://github.com/aws/aws-cli/issues/9214#issuecomment-2596130576 that indicates the SDK can be invoked with a WHEN_REQUIRED parameter and not automatically do the integrity behaviors when it's an endpoint not supporting them.

### Alternative Solutions

I'm trying to get a concrete answer from Backblaze on if they'll support the integrity behavior. That's certainly the ideal path, but even with that path, there's some duration with no backups unless we switch the backup storage provider.

JinxCappa:
This is an issues with other S3-compatible services. Backup is also broken on IDrive E2 (I personally verified) and Cloudflare R2 (was told by someone). Being able to disable would be great.

brantgurga:
Though no indication of timeline was indicated, Backblaze indicated they intend to change to be compatible. Their response was copying the notice on https://www.backblaze.com/docs/cloud-storage-use-the-aws-sdk-for-go-with-backblaze-b2 and emphasizing "yet".

brantgurga:
For the interaction with Backblaze, I tried to get any sense of timeline from them on supporting the headers. Even saying I'm just looking for a basic ballpark of this week, this month, this year, or longer, they gave no sense of timeline. To me that communicates they have no real plan to fix it on their side and are just telling people to downgrade the AWS SDK.

JinxCappa:
I'm not sure when this will be addressed so in the interim I built my own omni image and rolled back the sdk.

PR #1151, opened 2025-04-28
fix: loosen s3 integrity check for etcd backups

utkuozdemir:
S3 Go SDK introduced a breaking change which broke uploads to S3-compatible APIs like Backblaze. Revert it by setting request checksum calculation to "when required".

For more info, see the discussion and the comment here: https://github.com/aws/aws-sdk-go-v2/discussions/2960#discussioncomment-12077210

Fixes https://github.com/siderolabs/omni/issues/1086.


smira:
Promote to e2e-backups?

Unix4ever (approved):

PR #1158, opened 2025-04-29
chore: rewrite MachineSetNode controller to use qcontrollers

### LINKED ISSUE #556: Rewrite MachineSetNode controller to use qcontrollers

Orzelius:
Fixes #556


internal/backend/runtime/omni/controllers/omni/machine_set_node.go
```diff
@@ -28,195 +30,246 @@ import (
 	"github.com/siderolabs/omni/client/pkg/omni/resources/system"
 )
 
+// Questions for pr
+// 1. which data is stored in the spec and which in the labels?
+
```
Unix4ever, line 35: This resource is used to extract the labels from any resource type.
In this case it's `omni.MachineStatus` labels.

As this controller doesn't use anything from the `omni.MachineStatus` spec, it can rely on the stripped down version of the resource.

internal/backend/runtime/omni/controllers/omni/machine_set_node.go
```diff
@@ -28,195 +30,246 @@ import (
...
+		if err != nil {
+			return nil, err
+		}
 
-		err = list.ForEachErr(func(machineSet *omni.MachineSet) error {
-			return ctrl.reconcileMachineSet(ctx, r, machineSet, allMachineStatuses, allMachineSetNodes, machineStatusMap, visited, logger)
-		})
+		machineSetId, ok := machineSetNode.Metadata().Labels().Get(omni.LabelMachineSet)
+		if !ok {
+			// maybe should log a warning. Ask in pr
```
Unix4ever, line 198: Logging would be nice to have. Can also return error here. It will also be logged.

Unix4ever:
In general: the trick part about the QControllers is that all inputs for the controller should have finalizers added by the QController.

So if the input is running we add the finalizer, then if the resource triggers reconcile and is tearing down, we should remove the finalizer.

The problem is that without the finalizers it is not guaranteed that the delete operation comes to the controller. It will be unnoticed by it.
I can explain that in a call if you want. This is tricky :)

Orzelius:
Thank you for the review! Seems like this is not too far from the finish line so I'll take some time to try to finish this pr.

> In general: the tricky part about the QControllers is that all inputs for the controller should have finalizers added by the QController.

Okay I see, this was the part I was confused about. I'll try to figure it out on my own (will learn more that way :p) and implement the changes needed and will ping you again for feedback. Thank you for taking the time to explain!

internal/backend/runtime/omni/controllers/omni/machine_set_node.go
Orzelius: Okay I see, but is there a reason for why, for example, the evicted label shouldn't be an `isEvicted` field on the `omni.MachineStatus` spec?

Also, currently it seems that references to owner resources seem to be done using labels instead of spec fields. 
For example:
* machineSets have a `LabelCluster` instead of having a `clusterId` field in the spec
* machines resources have a `LabelMachineSet` label instead of a a `MachineSetId` or a similar field in the spec 

Is there a reason for that? Is it due to avoid having confusing zero values in the spec as the labels can be deleted?

frontend/src/views/cluster/ClusterMachines/MachineRequest.vue
```diff
@@ -40,7 +40,9 @@ const stage = computed(() => {
 </script>
 
 <template>
-  <div class="col-span-full grid grid-cols-subgrid items-center p-2 pr-4 text-xs text-naturals-n14">
+  <div
+    class="col-span-full grid grid-cols-subgrid items-center py-3 pr-4 pl-2 text-xs text-naturals-n14"
```
Slessi, line 44: out of curiosity why the padding change? 👀

Unix4ever: I spotted that the height of normal machines and the ones which are currently being provisioned is different.

Slessi: yea, its the `[...]` actions button thing that usually grows the row being missing, i guess this is ok if it fixes it

Issue #1070, opened 2025-04-09
handle non-secureboot UKIs


utkuozdemir:
Handle schematic (use FullId) for UKI-booted Talos same way as SecureBoot.

PR #1162, opened 2025-04-29
fix: collect and handle UKI boot information

utkuozdemir:
With Talos v1.10, even non-secure boot machines can be booted with a UKI. Replace the secure boot status collection and handling logic to take UKI boot into consideration instead.

This affects which schematic ID we use to compare the desired vs actual schematics.

Closes https://github.com/siderolabs/omni/issues/1070.

### DIFF
--- a/client/api/omni/specs/omni.proto
+++ b/client/api/omni/specs/omni.proto
@@ -18,10 +18,12 @@ message MachineSpec {
   reserved 4;
 }
 
-// SecureBootStatus describes the status of the SecureBoot feature.
-message SecureBootStatus {
-  // Enabled is true if SecureBoot is detected to be available and enabled.
-  bool enabled = 1;
+// SecurityState describes the UKI and secure boot information of the machine.
+message SecurityState {
+  // SecureBoot is true if secure boot is detected to be available and enabled.
+  bool secure_boot = 1; // use existing logic
+  // BootedWithUki is true if the machine is booted with a UKI. Always true if secure boot is enabled.
+  bool booted_with_uki = 2;
 }
 
 message Overlay {
@@ -238,8 +240,7 @@ message MachineStatusSpec {
   string initial_talos_version = 16;
 
   reserved 17;
-
-  SecureBootStatus secure_boot_status = 18;
+  reserved 18;
 
   repeated Diagnostic diagnostics = 19;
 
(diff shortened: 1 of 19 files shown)


internal/backend/runtime/omni/controllers/omni/cluster_machine_config.go
```diff
@@ -468,13 +468,13 @@ func buildInstallImage(imageFactoryHost string, resID resource.ID, installImage
...
-	if secureBootStatus == nil { // should never happen - must have been handled before entering this function
+	bootType := installImage.BootType
+	if bootType == nil { // should never happen - must have been handled before entering this function
 		return "", fmt.Errorf("machine %q has no secure boot status set", resID)
 	}
 
 	installerName := "installer"
-	if secureBootStatus.Enabled {
+	if bootType.IsSecureBoot {
 		installerName = "installer-secureboot"
```
frezbo, line 478: @smira should we get back into having a unified installer from factory, so we could simplify this?

smira: yes, we should probably at least create an issue for it

client/api/omni/specs/omni.proto
```diff
@@ -18,10 +18,12 @@ message MachineSpec {
...
-// SecureBootStatus describes the status of the SecureBoot feature.
-message SecureBootStatus {
-  // Enabled is true if SecureBoot is detected to be available and enabled.
-  bool enabled = 1;
+// BootType describes the UKI and secure boot information of the machine.
+message BootType {
+  // IsSecureBoot is true if secure boot is detected to be available and enabled.
+  bool is_secure_boot = 1; // use existing logic
+  // IsFromUki is true if the machine is booted from UKI. Always true if secure boot is enabled.
+  bool is_from_uki = 2;
```
Unix4ever, line 26: It's `BootedWithUKI` in Talos resources, lets make name it the same way here for consistency?

utkuozdemir: Sure, will do. Btw, what do you think about storing the whole SecurityState resource spec as-is?

utkuozdemir: Oh, nvm, probably we don't want to wake up on other fields' changes.

PR #1163, opened 2025-05-01
fix: use the correct sort order for the automatic install disk selection

Unix4ever:
It was inverted, so it was picking the biggest disk.

### DIFF
--- a/internal/backend/runtime/omni/controllers/omni/machine_config_gen_options.go
+++ b/internal/backend/runtime/omni/controllers/omni/machine_config_gen_options.go
@@ -6,6 +6,7 @@
 package omni
 
 import (
+	"cmp"
 	"context"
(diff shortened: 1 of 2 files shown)


utkuozdemir (approved):

internal/backend/runtime/omni/controllers/omni/machine_config_gen_options.go
```diff
@@ -103,7 +103,7 @@ func GenInstallConfig(machineStatus *omni.MachineStatus, clusterMachineTalosVers
 			}
 
 			if a.Size < b.Size {
```
smira, line 105:
```suggestion
			return cmp.Compare(a.Size, b.Size)
```

smira (approved):

Issue #1053, opened 2025-04-01
Talos 1.10 readiness


smira:
A list of issues to update Omni to be ready for 1.10.

PR #1164, opened 2025-05-01
feat: update machinery and specs to Talos 1.10.0

smira:
See #1053

Update also JSON schema for the config, use `UpgradeWithOptions` API.


frezbo (approved):

internal/backend/runtime/omni/controllers/omni/cluster_machine_config_status.go
```diff
@@ -465,7 +465,12 @@ func (h *clusterMachineConfigStatusControllerHandler) syncInstallImageAndSchemat
 		return false, err
 	}
 
-	_, err = c.Upgrade(upgradeCtx, image, !maintenance, stageUpgrade, false)
+	_, err = c.UpgradeWithOptions(upgradeCtx,
```
smira, line 468: this is the most tricky change.

@frezbo please don't change machinery's client interfaces - that is going to break consumers -- even if Talos itself doesn't use some option, it doesn't mean that there are no other consumers

frezbo: should I revert?

smira: yes, let's do that please for 1.10.1

smira: this Omni change should be fine though - using better API

utkuozdemir (approved):

PR #1174, opened 2025-05-05
fix: correctly encode exposed service redirect URL after auth

### LINKED ISSUE #1168: [bug] workload proxy authentication requires Omni UI authentication

### Current Behavior

If you have set up a service to be workload proxied, that works fine when you connect to it after you load the omni home page, but copy the URL:

Image
and attempt to connect to the URL from another browser or incognito browser, then you will be asked to "authenticate workload proxy access" via the Omni log in flow, but it then errors out with "Bad Request"

If, in that browser session, you then authenticate to your omni site first, the same flow works.

### Expected Behavior

That you could log in to services that have been proxied without first having to log in to the Omni web ui.

### Steps To Reproduce

Set up a workload proxyied service in Omni.
Copy the URL that is linked to.
Open it in a new browser.
Authenticate successfully,
Get "Bad Request" error.

_No response_

### Anything else?

_No response_

utkuozdemir:
We had a URL encoding issue in the `redirect` query parameter in the new workload proxy auth/redirect flow. On the authentication page, we URL-decoded the redirect URL to be used after the authentication, but forgot to re-encode it before issuing the redirect afterwards, which submits it to the backend.

Backend then was unable to base64-decode the signed redirect URL, resulting in a bad request.

Closes #1168.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,6 +1,6 @@
 # THIS FILE WAS AUTOMATICALLY GENERATED, PLEASE DO NOT EDIT.
 #
-# Generated on 2025-04-22T09:44:01Z by kres fd5cab0.
+# Generated on 2025-05-05T10:42:38Z by kres 1a0156b.
 
 version: "2"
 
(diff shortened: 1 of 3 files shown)


utkuozdemir:
> hopefully with our new tests work we can catch these earlier

Yes certainly we should - I'm working on the new tests.

smira (approved):

Unix4ever (approved):

Issue #1131, opened 2025-04-22
add more frontend unit-tests


smira:
Improve coverage of frontend code.

PR #1184, opened 2025-05-06
test: add unit test for nextAvailableClusterName function

kalio007:
Addresses #1131
Some updates in this PR
- Added a  Return Type `Promise<string>`  to the nextAvailableClusterName function.
- Unit test to cover some edge cases for the function.


kalio007:
Hello @smira @Unix4ever , I would appreciate your feedback on this PR. Thanks

Unix4ever (commented):
Thank you for the PR!

I'd suggest to drop all irrelevant changes and keep only the test updates.

kalio007:
Thanks for the review @Unix4ever , with this, I should write other unit tests faster.

PR #1186, opened 2025-05-07
feat(ci): support releasing helm charts

frezbo:
Support releasing Helm charts as signed oci artifacts.

Fixes: #1049


deploy/helm/omni/Chart.yaml
frezbo, line 1: >  `make generate` before a release now also updates `Chart.yaml`

Unix4ever (approved):

Issue #894, opened 2025-02-03
Screen with the list of infra providers


smira:

utkuozdemir:
I don't have much input atm other than maybe showing the online/offline/health status using green/red/gray indicators or something like that. You probably already have that in mind.

About grouped/plain list, I think starting with a plain list is perfectly fine, but as you wish.

PR #1211, opened 2025-05-16
feat: allow managing infra providers in the UI

Unix4ever:
No longer show the infra provider service accounts on the service accounts page.
Instead, move the infra provider service account creation to the infra providers page.
They are now always created in pair with `infra.Provider` resources.

If `infra.Provider` resource doesn't exist the provider is not allowed to connect.
Also `infra.Provider` resource controls the lifecycle of the `infra.ProviderStatus`, `infra.ProviderHealthStatus` and the service account created for the provider: when the `infra.Provider` is deleted everything is cleaned up.

`infra.ProviderStatus` and `infra.ProviderHealthStatus` are now combined into `infra.ProviderCombinedStatus` which represents the current connection state of the provider: if it runs or not and if it has any errors.

`infra.ProviderCombinedStatus` resources are now shown on the new page in the Omni UI.

`omnictl` got new commands for managing and viewing infra providers in the system: `omnictl infraprovider [create|delete|renewkey|list]`.

Fixes: https://github.com/siderolabs/omni/issues/894


Unix4ever:
Screenshot 2025-05-16 at 15 20 38

client/pkg/omni/resources/infra/provider.go
```diff
@@ -0,0 +1,58 @@
...
+		DefaultNamespace: resources.InfraProviderNamespace,
+		PrintColumns: []meta.PrintColumn{
+			{
+				Name:     "Name",
+				JSONPath: "{.name}",
+			},
+			{
+				Name:     "Description",
+				JSONPath: "{.description}",
+			},
```
utkuozdemir, line 55: These are obsolete, remove & maybe replace with correct ones.

internal/backend/runtime/omni/controllers/omni/service_account_status.go
```diff
@@ -0,0 +1,103 @@
...
+			UnmapMetadataFunc: func(r *auth.ServiceAccountStatus) *auth.Identity {
+				return auth.NewIdentity(resources.DefaultNamespace, r.Metadata().ID())
+			},
+			TransformFunc: func(ctx context.Context, r controller.Reader, _ *zap.Logger, identity *auth.Identity, status *auth.ServiceAccountStatus) error {
+				user, err := safe.ReaderGetByID[*auth.User](ctx, r, identity.TypedSpec().Value.UserId)
+				if err != nil {
+					return err
+				}
+
+				if role.Role(user.TypedSpec().Value.Role) == role.InfraProvider {
```
Unix4ever, line 52: drop it, after adding the check in the map metadata optional

internal/backend/runtime/omni/migration/migrations.go
```diff
@@ -1760,3 +1761,40 @@ func dropMachineClassStatusFinalizers(ctx context.Context, st state.State, logge
 
 	return nil
 }
+
+// createProviders creates infra.Provider resource for each existing infra.ProviderStatus.
+// This migration is required to avoid breaking user's providers connection which were created before 0.50.
+func createProviders(ctx context.Context, st state.State, logger *zap.Logger, _ migrationContext) error {
+	providers, err := safe.ReaderListAll[*infra.ProviderStatus](ctx, st)
```
Unix4ever, line 1768: Use `auth.Identity` with prefix `access.InfraProviderPrefix` instead. 
This way it will be more resilient. Any providers ever registered will be able to connect after we roll out the change.

Unix4ever: Also test the providers connection manually.

utkuozdemir (approved):

PR #1214, opened 2025-05-20
fix: use the correct schematic IP for maintenance upgrades

utkuozdemir:
Maintenance upgrades triggered from the UI were using the wrong schematic ID, causing the machines which use UKI to lose siderolink kernel args and disconnect.

Since we have a complex logic to build the correct install image including the schematic, move it to a central place.

Add a new management endpoint for the maintenance upgrades. UI now calls this endpoint instead of calling the Talos API directly. The new endpoint builds the install image correctly using the common logic and issues the upgrade.


Unix4ever:
I've spotted an issue not related to this PR.
Can you please fix the test name in the `.kres.yaml`?

Here:
https://github.com/siderolabs/omni/blob/74a9ad36f4eadca0681bf4dca2884d6a4b4256fb/.kres.yaml#L252

And here:
https://github.com/siderolabs/omni/blob/74a9ad36f4eadca0681bf4dca2884d6a4b4256fb/.kres.yaml#L191

utkuozdemir:
> I've spotted an issue not related to this PR. Can you please fix the test name in the `.kres.yaml`?
> 
> Here:
> 
> https://github.com/siderolabs/omni/blob/74a9ad36f4eadca0681bf4dca2884d6a4b4256fb/.kres.yaml#L252
> 
> And here:
> 
> https://github.com/siderolabs/omni/blob/74a9ad36f4eadca0681bf4dca2884d6a4b4256fb/.kres.yaml#L191
> 
> It should be `MaintenanceUpgrade`, not downgrade.

Oh, nice catch. Did it just now.

frezbo (approved):

Unix4ever (approved):

Unix4ever:
we might need to bump the limits for the metrics. With the new test it goes over 11000 operations.

Unix4ever:
Lets bump this [timeout](https://github.com/siderolabs/omni/blob/0020ee3d79592585f883275f832a9826b7f82c2c/cmd/integration-test/pkg/tests/talos.go#L732) to 10 minutes. I think the test fails just because it takes too long

PR #1215, opened 2025-05-20
release(v0.50.0): prepare release

Unix4ever:
This is the official v0.50.0 release.


utkuozdemir (approved):

frezbo (commented):
`make generate` is missing, since the helm chart file didn't get updated

frezbo:
> > make generate is missing, since the helm chart file didn't get updated
> 
> This `make generate` has the same issue as in tag, as it's using the latest git tag, which isn't yet created when we submit the PR.
> 
> Maybe that all should be a part of the release.sh.
> 
> I've updated the helm chart manually.

I thought the process was a `make generate ABBREV_TAG=<new tag to be released>`

Unix4ever:
> I thought the process was a make generate ABBREV_TAG=<new tag to be released>

Ahh, that also works I guess.

Issue #1171, opened 2025-05-05
[test] turn integration tests into regular Go tests & run Omni from tests


utkuozdemir:
Instead of having [this workaround](https://github.com/siderolabs/omni/blob/main/cmd/integration-test/pkg/tests/utils.go), make Omni tests follow the regular Go test file format `<file>_test.go` with the structure like:
- A root test does the initialization of Omni and the Talos machines
  - Here, call the Omni entrypoint by the test itself by default and start Omni, passing `zaptest.NewLogger(t)` as its logger and registering its deferred cleanup logic
  - This behavior should be overridable - tests should be able to run against an external Omni (since the integration tests are used by other places, e.g., by the infra providers)
- It runs the regular tests as sub-tests via `t.Run(...)`
- To be able to pass the custom Omni endpoint and potentially other input, integration tests must still be able to accept input (command line args or env vars etc.)
- This integration tests should be compiled to a binary and published the same way we publish them today (e.g., `ghcr.io/siderolabs/omni-integration-test`)

These will bring the following benefits:
- Omni integration tests will look and run more naturally
- We will be compiling & running Omni with `-race` when we compile the tests with `race`, as we will be building Omni along with tests and run it from there
- The logs will be properly indented test logs, and for the non-failing tests, the Omni logs will not be printed when `-v` is not specified
- We will be able to get coverage information from Omni integration tests

PR #1239, opened 2025-06-02
test: use `go test` to build and run Omni integration tests

Unix4ever:
All test modules were moved under `integration` tag and are now in `internal/integration` folder: no more `cmd/integration-test` executable.

New Kres version is able to build the same executable from the tests directory instead.

All Omni related flags were renamed, for example `--endpoint` -> `--omni.endpoint`.

2 more functional changes:

- Enabled `--test.failfast` for all test runs.
- Removed finalizers, which were running if the test has failed.

Both of these changes should make it easier to understand the test failure: Talos node logs won't be cluttered with the finalizer tearing down the cluster.

Fixes: https://github.com/siderolabs/omni/issues/1171


internal/integration/talos_test.go
```diff
@@ -636,9 +638,11 @@ func AssertTalosUpgradeIsCancelable(testCtx context.Context, st state.State, clu
...
 		events := make(chan state.Event)
 
-		require.NoError(t, st.WatchKind(ctx, omni.NewClusterMachineStatus(resources.DefaultNamespace, "").Metadata(), events),
-			state.WithLabelQuery(resource.LabelEqual(omni.LabelCluster, clusterName)),
-		)
+		t.Logf("watching for the machines in cluster %q", clusterName)
+
+		require.NoError(t, st.WatchKind(ctx, omni.NewClusterMachineStatus(resources.DefaultNamespace, "").Metadata(), events,
+			state.WatchWithLabelQuery(resource.LabelEqual(omni.LabelCluster, clusterName)),
+		))
```
Unix4ever, line 645: oh, I like errors like this. The difference is so subtle: the brackets were placed incorrectly, so `state.WatchWithLabelQuery(resource.LabelEqual(omni.LabelCluster, clusterName)` ended up as the second argument to `require.NoError`, so the watch was actually ignoring the labels 🤦 

This error was there like since forever

utkuozdemir: wow :D

internal/integration/integration_test.go
```diff
@@ -0,0 +1,348 @@
...
+		t.Run("KubernetesUpgrades", testKubernetesUpgrades(testOptions))
+		t.Run("EtcdBackupAndRestore", testEtcdBackupAndRestore(testOptions))
+		t.Run("MaintenanceUpgrade", testMaintenanceUpgrade(testOptions))
+		t.Run("Auth", testAuth(testOptions))
+		t.Run("ClusterTemplate", testClusterTemplate(testOptions))
+		t.Run("WorkloadProxy", testWorkloadProxy(testOptions))
+		t.Run("StaticInfraProvider", testStaticInfraProvider(testOptions))
+	})
+
+	postRunHooks(t, testOptions)
```
smira, line 211: `t.Cleanup` here and move it up?

smira (approved):

internal/integration/integration_test.go
Unix4ever: I think there was something with the context already being canceled at the moment of cleanup. But let me check once again

utkuozdemir (approved):
❤️

internal/integration/integration_test.go
Unix4ever: yeah, it looks like `t.Context()` is canceled at the moment of cleanup

smira: I think in any case you want a detached context in cleanup with some timeout

Unix4ever: not a big fan of the detached context. It kinda makes sense there though. Updated

Unix4ever: This doesn't work either. I wanted to do t.Run in the post hooks and it doesn't work in cleanup. I think I'll keep what I had initially.

PR #1242, opened 2025-06-03
test: improve workload proxying tests

### LINKED ISSUE #1129: add more integration tests for workload proxy
Workload Proxy: more services, more clusters, more transitions (enabling/disable service).

### LINKED ISSUE #1172: [test] generate a support bundle when integration test fails
Generate a support bundle before tearing down the cluster, attach it to the CI build artifacts.

utkuozdemir:
Add many more services and test scenarios to the workload proxying feature:
- Use two clusters, a 1+1 and a 1+2.
- Use multiple nginx workloads.
- Each workload serving its name in its `index.html` and it being asserted (i.e., we assert that we hit the correct service).
- Multiple exposed services per workload.
- Multiple parallel requests per exposed service.
- Toggle the feature off and on, assert service accessibility.
- Toggle an exposed service off and on by removing/readding the k8s service annotation, assert accessibility.
- Test explicit prefixes.

Additionally:
- Fix two bugs in workload services:
  - Check the cookies before returning 404 for a non-existing exposed service prefix.
  - Add timeouts to `inmem` proxy transport, so requests do not potentially hang forever.
- Bring back the logic the saving of a support bundle when an integration test fails, and fix its save path.

Closes #1129, closes #1172.


internal/integration/workloadproxy/workloadproxy.go
```diff
@@ -0,0 +1,668 @@
...
+		}
+	}, 3*time.Minute, 5*time.Second)
+
+	return deployment, services
+}
+
+func doGzip(t *testing.T, input []byte) []byte {
+	t.Helper()
+
+	var buf strings.Builder
```
smira:
```suggestion
	var buf bytes.Buffer
```

This would simplify and make things more correct

smira (approved):

internal/integration/workloadproxy/workloadproxy.go
utkuozdemir: Oh, this was not intentional since the beginning, an oversight.

PR #1246, opened 2025-06-04
feat: rewrite Omni config management 

Unix4ever:
Omni can now be configured via a config file instead of the command line
flags.
The flags `--config-path` will now read the config provided in the YAML
format.
The config structure was completely changed. It was not public before,
so it's fine to ignore backward compatibility.
The command line flags were not changed.


cmd/omni/cmd/cmd.go
```diff
@@ -332,68 +464,294 @@ var initOnce = sync.OnceValue(func() *cobra.Command {
...
+	rootCmd.Flags().BoolVar(
+		&config.Config.Features.DisableControllerRuntimeCache,
+		"disable-controller-runtime-cache",
+		config.Config.Features.DisableControllerRuntimeCache,
+		"disable watch-based cache for controller-runtime (affects performance)",
+	)
+}
+
+func defineDebugFlags() {
+	// debug
```
utkuozdemir: intentional?

internal/backend/grpc/configs_test.go
```diff
@@ -92,10 +92,10 @@ func TestGenerateConfigs(t *testing.T) {
 		require.Equal(t, codes.PermissionDenied, status.Code(err), err)
 	})
 
-	config.Config.EnableBreakGlassConfigs = true
+	config.Config.Features.EnableBreakGlassConfigs = true
```
utkuozdemir, line 95: not related to this PR, but imo we should avoid doing this, basically accessing/modifying config directly in code or tests. We should inject the config values to the component which needs it, and in the tests we inject the mock/test value.

internal/integration/integration_test.go
```diff
@@ -80,6 +86,10 @@ func TestIntegration(t *testing.T) {
 		OutputDir:                artifactsOutputDir,
 	}
 
+	if runEmbeddedOmni {
+		runOmni(t)
+	}
```
utkuozdemir: 🔥

internal/pkg/auth/config_test.go
```diff
@@ -43,8 +43,8 @@ func TestEnsureAuthConfigResource(t *testing.T) {
 		},
 		{
 			name: "enable auth0",
-			initialConfig: config.AuthParams{
-				Auth0: config.Auth0Params{
+			initialConfig: config.Auth{
+				Auth0: config.Auth0Config{
```
utkuozdemir: nit, but since `Auth0Config` is already in the `config` package, should it simply be `Auth0`? Just like `AuthParams` -> `Auth`, it would be `Auth0Params` -> `Auth0`.

Same goes for other types as well.

utkuozdemir:
So far looks good to me. My only suggestion is to get rid of all `Params` and `Config` suffixes from the types.

internal/backend/grpc/configs_test.go
Unix4ever: I'd be happy to avoid using global config variable all over the places.
I need to limit the scope of the change though :)

cmd/omni/cmd/cmd.go
Unix4ever: no longer needed

internal/pkg/auth/config_test.go
Unix4ever: yeah, can do that

internal/integration/integration_test.go
Unix4ever: I'll revert that for now. Will come in a separate PR

go.mod
```diff
@@ -113,6 +113,11 @@ require (
 	sigs.k8s.io/controller-runtime v0.20.4
 )
 
+require (
+	dario.cat/mergo v1.0.2
```
smira: need to clean this up?

cmd/omni/cmd/run.go
```diff
@@ -8,46 +8,64 @@ package cmd
...
 // RunService starts the main Omni server.
-func RunService(ctx context.Context, logger *zap.Logger, params *config.Params) error {
+func RunService(ctx context.Context, logger *zap.Logger, paramsList ...*config.Params) error {
 	cfg := config.InitDefault()
 
-	raw, err := yaml.Marshal(params)
-	if err != nil {
-		return err
+	for _, params := range paramsList {
+		if err := mergo.Merge(cfg, params, mergo.WithOverride); err != nil {
```
smira: we do have `merge` in Talos `machinery` package, you might want to use this instead of pulling a dependency

internal/pkg/config/testdata/config-full.yaml
```diff
@@ -0,0 +1,121 @@
...
+    enabled: true
+    address: 0.0.0.0:9988
+  pprof:
+    enabled: true
+    address: 0.0.0.0:8124
+
+auth:
+  keyPruner:
+    interval: 1m
+  joinTokensMode: strict
```
Unix4ever: move it to siderolink

internal/pkg/config/testdata/config-full.yaml
```diff
@@ -0,0 +1,121 @@
...
+    enabled: true
+    subdomain: "proxy-us"
+
+debug:
+  server:
+    enabled: true
+    address: 0.0.0.0:9988
+  pprof:
+    enabled: true
+    address: 0.0.0.0:8124
```
Unix4ever:
```suggestion
    enabled: true
    endpoint: 0.0.0.0:9988
  pprof:
    enabled: true
    endpoint: 0.0.0.0:8124
```

internal/pkg/config/services.go
```diff
@@ -0,0 +1,221 @@
...
+		return ks.AdvertisedURL
+	}
+
+	return fmt.Sprintf("https://%s", ks.BindEndpoint)
+}
+
+// SiderolinkService manages WireGuard connections to the Talos machines connected to Omni.
+type SiderolinkService struct {
+	WireGuardBindEndpoint       string `yaml:"wireGuardEndpoint"`
+	WireGuardAdvertisedEndpoint string `yaml:"wireGuardAdvertisedEndpoint"`
```
Unix4ever:
```suggestion
        WireGuard struct {
           BindEndpoint string `yaml:"endpoint"`
           AdvertisedEndpoint string `yaml:"advertisedEndpoint"`
        }
```

internal/pkg/config/testdata/config-full.yaml
```diff
@@ -0,0 +1,121 @@
...
+    boltdb:
+      path: "_out/omni.db"
+    etcd:
+      endpoints:
+        - http://localhost:2379
+      dialKeepAliveTime: 30s
+      dialKeepAliveTimeout: 5s
+      caPath: etcd/ca.crt
+      certPath: etcd/client.crt
+      keyPath: etcd/client.key
```
Unix4ever:
```suggestion
      caFile: etcd/ca.crt
      certFile: etcd/client.crt
      keyFile: etcd/client.key
```

internal/pkg/config/storage.go
```diff
@@ -0,0 +1,57 @@
...
+// BoltDB defines boltdb storage configs.
+type BoltDB struct {
+	Path string `yaml:"path"`
+}
+
+// EtcdParams defines etcd storage configs.
+type EtcdParams struct { ///nolint:govet
+	// External etcd: list of endpoints, as host:port pairs.
+	Endpoints            []string      `yaml:"endpoints"`
+	DialKeepAliveTime    time.Duration `yaml:"dialKeepAliveTime"`
```
Unix4ever, line 44:
```suggestion
	DialKeepAliveInterval    time.Duration `yaml:"dialKeepAliveInterval"`
```

cmd/omni/cmd/run.go
Unix4ever: are you talking about https://github.com/siderolabs/talos/blob/81ca27949427c546f43b0409b56f733becabc2f6/pkg/machinery/config/configpatcher/strategic.go#L31 ?
This one is too specific for the configs

internal/pkg/config/storage.go
Unix4ever: Going to keep this weird name, as it's the same in the etcd library.

Unix4ever:
Additionally check if the config has extra fields. This helps to identify typos.

utkuozdemir (approved):

utkuozdemir:
Also closes #1250.

PR #1251, opened 2025-06-09
Run omni inside integration test

Unix4ever:


.github/workflows/ci.yaml
```diff
@@ -1,6 +1,6 @@
 # THIS FILE WAS AUTOMATICALLY GENERATED, PLEASE DO NOT EDIT.
 #
-# Generated on 2025-06-06T17:20:38Z by kres fc6afbe-dirty.
```
Unix4ever, line 3: run rekres

internal/backend/runtime/omni/state.go
```diff
@@ -39,75 +39,176 @@ import (
...
+	case err := <-s.defaultPersistentState.errors:
+		return err
+	case <-ctx.Done():
+	}
+
+	return nil
+}
+
+// NewTestState creates a new test state using the in-memory storage and no virtual one.
+func NewTestState(logger *zap.Logger) *State {
```
Unix4ever: move it to `state_test.go`

hack/test/integration.sh
```diff
@@ -269,21 +321,26 @@ if [[ "${ENABLE_SECUREBOOT}" == "true" ]]; then
...
+  --omni.embedded \
+  --omni.config-path ${OMNI_CONFIG} \
+  --omni.log-output=${TEST_OUTPUTS_DIR}/omni-integration.log \
   --test.failfast \
+  --test.coverprofile=${ARTIFACTS}/coverage-integration.txt \
   --test.v \
   ${INTEGRATION_TEST_ARGS:-}
 
 if [ "${INTEGRATION_RUN_E2E_TEST:-true}" == "true" ]; then
+  SIDEROLINK_DEV_JOIN_TOKEN="${JOIN_TOKEN}" \
```
Unix4ever, line 358: run with the cmd line flags here.
use limited config, and pass half as flags.

internal/backend/runtime/omni/state.go
Unix4ever: won't work as we use this method in the test modules outside of this one

.github/workflows/ci.yaml
Unix4ever: I think Kres doesn't touch the files which are not changed. So running rekres doesn't update it

utkuozdemir (approved):

PR #1254, opened 2025-06-10
fix: fix exposed service prefix conflict resolution

utkuozdemir:
Exposed services with user-specified (explicit) prefixes had a problem with their reconciliation in the following scenario:
- service A has explicit prefix X
- service B also has explicit prefix X -> service B is marked as in error state because the prefix X is already owned by another service
- service A is updated to have a new prefix, Y
- service B is still in the error state, until a new reconciliation is triggered (e.g., a service C is introduced or any other change)

Fix this by removing the freed prefix (X) from the "used prefixes" list during reconciliation. Add a unit test for it.

Rework the `exposedservice.Reconciler` to make it testable by simplifying it.

### DIFF
--- a/internal/backend/runtime/omni/controllers/omni/internal/exposedservice/reconciler.go
+++ b/internal/backend/runtime/omni/controllers/omni/internal/exposedservice/reconciler.go
@@ -8,7 +8,6 @@ package exposedservice
 import (
 	"bytes"
 	"compress/gzip"
-	"context"
 	"encoding/base64"
 	"encoding/xml"
 	"errors"
@@ -18,9 +17,7 @@ import (
 	"strconv"
 	"strings"
 
-	"github.com/cosi-project/runtime/pkg/controller"
 	"github.com/cosi-project/runtime/pkg/resource"
-	"github.com/cosi-project/runtime/pkg/safe"
 	"github.com/siderolabs/gen/optional"
 	"go.uber.org/zap"
 	corev1 "k8s.io/api/core/v1"
@@ -33,7 +30,6 @@ import (
 
 // Reconciler is the reconciler for ExposedService resources.
 type Reconciler struct {
-	r      controller.ReaderWriter
 	logger *zap.Logger
 
 	exposedServices map[resource.ID]*omni.ExposedService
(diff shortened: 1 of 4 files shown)


internal/backend/runtime/omni/controllers/omni/internal/exposedservice/reconciler.go
```diff
@@ -171,7 +136,19 @@ func (reconciler *Reconciler) updateExposedService(res *omni.ExposedService, exp
...
+
+	// if the service has an explicit alias, but it is now getting a new alias,
+	// we need to free it here, so that it can be reused by another service
+	abandoningCurrentExplicitAlias := hasExistingAlias && hasExplicitAlias &&
+		(!explicitAliasRequested || requestedExplicitAlias != currentAlias)
+	if abandoningCurrentExplicitAlias {
+		if ownerID, ok := reconciler.usedAliases[currentAlias]; ok && ownerID == res.Metadata().ID() {
+			delete(reconciler.usedAliases, currentAlias)
+		}
+	}
```
utkuozdemir, line 151: Actual fix of the issue in the description.

internal/backend/runtime/omni/controllers/omni/internal/exposedservice/reconciler.go
```diff
@@ -136,32 +120,13 @@ func (reconciler *Reconciler) reconcileService(ctx context.Context, exposedServi
...
-			zap.String("url", updatedService.TypedSpec().Value.Url),
-			zap.Uint32("port", updatedService.TypedSpec().Value.Port),
-			zap.String("label", updatedService.TypedSpec().Value.Label),
-			zap.String("error", updatedService.TypedSpec().Value.Error),
-		)
-	}
-
-	alias, _ := updatedService.Metadata().Labels().Get(omni.LabelExposedServiceAlias)
-	reconciler.usedAliases[alias] = updatedService.Metadata().ID()
-	reconciler.exposedServices[updatedService.Metadata().ID()] = updatedService
```
utkuozdemir, line 164: Moved into the controller, as we do not do actual operations on the state in this component anymore. It is now side-effect free.

smira (approved):

internal/backend/runtime/omni/controllers/omni/internal/exposedservice/reconciler.go
```diff
@@ -171,7 +136,19 @@ func (reconciler *Reconciler) updateExposedService(res *omni.ExposedService, exp
...
-	_, hasExistingAlias := res.Metadata().Labels().Get(omni.LabelExposedServiceAlias)
+	currentAlias, hasExistingAlias := res.Metadata().Labels().Get(omni.LabelExposedServiceAlias)
+	requestedExplicitAlias = strings.ToLower(requestedExplicitAlias)
+	currentAlias = strings.ToLower(currentAlias)
+
+	// if the service has an explicit alias, but it is now getting a new alias,
+	// we need to free it here, so that it can be reused by another service
+	abandoningCurrentExplicitAlias := hasExistingAlias && hasExplicitAlias &&
+		(!explicitAliasRequested || requestedExplicitAlias != currentAlias)
+	if abandoningCurrentExplicitAlias {
```
Unix4ever, line 147: I wish we had a simpler condition here... But I can't propose anything from the top of my head, so it's fine to keep that.

utkuozdemir: I tried hard to simplify it :) this is as much as I could. the possible states are not few...

PR #1260, opened 2025-06-13
release(v0.51.0-beta.0): prepare release

utkuozdemir:
This is the official v0.51.0-beta.0 release.


smira:
should we do beta.0 first?

utkuozdemir:
> should we do beta.0 first?

Yep, turned the PR into that.

PR #1262, opened 2025-06-13
fix: do more fixes on config backward-compatibility

utkuozdemir:
- Make sure to write the account ID passed via `--account-id` into the correct struct so it will land into the final config as expected.

- Add `merge:"replace"` directives on a few `[]string` fields in the config, so that when the user sets overrides, they won't be merged with the base/default values. This is specifically important for the etcd endpoints, as with the default config merge behavior, we end up with an endpoint set of default + command-line args, like: `["http://localhost:2379","https://omni-etcd-0.omni-etcd.svc:2379","https://omni-etcd-1.omni-etcd.svc:2379","https://omni-etcd-2.omni-etcd.svc:2379"]`.

### DIFF
--- a/cmd/omni/cmd/cmd.go
+++ b/cmd/omni/cmd/cmd.go
@@ -107,11 +107,13 @@ var rootCmdArgs struct {
 // RootCmd returns the root command.
 func RootCmd() *cobra.Command { return initOnce() }
 
+var cmdConfig = config.InitDefault()
+
 var initOnce = sync.OnceValue(func() *cobra.Command {
 	rootCmd.Flags().BoolVar(&rootCmdArgs.debug, "debug", false, "enable debug logs.")
 
-	rootCmd.Flags().StringVar(&config.Config.Account.ID, "account-id", config.Config.Account.ID, "instance account ID, should never be changed.")
-	rootCmd.Flags().StringVar(&config.Config.Account.Name, "name", config.Config.Account.Name, "instance user-facing name.")
+	rootCmd.Flags().StringVar(&cmdConfig.Account.ID, "account-id", cmdConfig.Account.ID, "instance account ID, should never be changed.")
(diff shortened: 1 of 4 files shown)


internal/pkg/config/storage.go
```diff
@@ -40,7 +40,7 @@ type BoltDB struct {
 // EtcdParams defines etcd storage configs.
 type EtcdParams struct { ///nolint:govet
 	// External etcd: list of endpoints, as host:port pairs.
-	Endpoints            []string      `yaml:"endpoints"`
+	Endpoints            []string      `yaml:"endpoints" merge:"replace"`
```
utkuozdemir, line 43: Important one

cmd/omni/cmd/cmd.go
```diff
@@ -107,11 +107,13 @@ var rootCmdArgs struct {
...
 func RootCmd() *cobra.Command { return initOnce() }
 
+var cmdConfig = config.InitDefault()
+
 var initOnce = sync.OnceValue(func() *cobra.Command {
 	rootCmd.Flags().BoolVar(&rootCmdArgs.debug, "debug", false, "enable debug logs.")
 
-	rootCmd.Flags().StringVar(&config.Config.Account.ID, "account-id", config.Config.Account.ID, "instance account ID, should never be changed.")
-	rootCmd.Flags().StringVar(&config.Config.Account.Name, "name", config.Config.Account.Name, "instance user-facing name.")
+	rootCmd.Flags().StringVar(&cmdConfig.Account.ID, "account-id", cmdConfig.Account.ID, "instance account ID, should never be changed.")
```
utkuozdemir, line 115: Important one.

smira (approved):
seems to be a disaster

utkuozdemir:
> seems to be a disaster

yep, we should have done a smaller testing in staging before the beta.0.

PR #1273, opened 2025-06-18
release(v0.51.0): prepare release

Unix4ever:
This is the official v0.51.0 release.


hack/release.toml
```diff
@@ -6,7 +6,7 @@ github_repo = "siderolabs/omni"
 match_deps = "^github.com/((talos-systems|siderolabs)/[a-zA-Z0-9-]+)$"
 
 pre_release = true
```
utkuozdemir:
```suggestion
pre_release = false
```

Unix4ever: good catch :)

Issue #1255, opened 2025-06-10
[feature] bring back new workload proxy implementation


utkuozdemir:
Find the issue, revert [this revert](https://github.com/siderolabs/omni/pull/1077).

PR #1278, opened 2025-06-18
feat: add jitter and lazy health checks to exposed services

utkuozdemir:
- Add jitter to the exposed service health checks, so they spread evenly even when the services are all reconciled at the exact same time.
- Add the "lazy" logic to the current workload proxy health checks by wrapping the "regular" LB with a lazy LB wrapper. With this, we gain:
  - Health checks are started only when an exposed service is attempted to be accessed ("dialed").
  - They are stopped after 5 minutes of inactivity.

Depends on siderolabs/go-loadbalancer#24.

Part of #1255.

### DIFF
--- a/cmd/omni/cmd/cmd.go
+++ b/cmd/omni/cmd/cmd.go
@@ -655,6 +655,13 @@ func defineFeatureFlags() {
 		"workload proxying subdomain.",
 	)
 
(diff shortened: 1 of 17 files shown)


Unix4ever (approved):

utkuozdemir:
Holding this during holidays.

PR #1279, opened 2025-06-19
fix: do not try running debug server in the prod builds

Unix4ever:
Otherwise the debug server stops returning `nil` and that stops the entire Omni instance.
Also refactor the fns list in the `server.go` to give more idea on which subsystem is stopped first. Otherwise it's really hard to understand what caused Omni to stop.


utkuozdemir (approved):
cool, similar to what I did here https://github.com/siderolabs/omni-infra-provider-bare-metal/blob/main/internal/provider/provider.go#L219-L223

Issue #1132, opened 2025-04-22
automated upgrade testing


smira:
1. Run "previous" Omni, create a Talos cluster (we need to define a version of Talos to use?).
2. Prepare a cluster template.
3. Run `omnictl` matching Omni, apply the template, assert the cluster is ready.
4. Any additional tests, like workload service proxy is working, `kubectl` access, ...
5. Stop "previous" Omni, run `HEAD` Omni.
6. Assert that:
  * the cluster is healthy, everything still works
  * machines haven't been rebooted
  * machine configuration hasn't been changed
  * Omni controllers don't fail repeatedly

Blackbox approach: don't use current Omni client code, use matching `omnictl`, we can run integration test of matching version.

Test matrix:

* what is `previous` - one or two stable releases (?)
* what is `HEAD` - current version, last stable release?

Example:

* `v0.48.3` -> `HEAD`
* `v0.47.1` -> `v0.48.3`

PR #1281, opened 2025-06-20
test: test updating from old Omni version to the current

Unix4ever:
CI will run integration tests from the previous release first, then run them from the current commit using other set of tests.

Fixes: https://github.com/siderolabs/omni/issues/1132

### DIFF
--- a/hack/test/integration.sh
+++ b/hack/test/integration.sh
@@ -22,6 +22,7 @@ echo "127.0.0.1 my-instance.localhost" | tee -a /etc/hosts
 
 TALOS_VERSION=1.10.2
 ENABLE_TALOS_PRERELEASE_VERSIONS=false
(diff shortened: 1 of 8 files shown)


hack/test/integration.sh
```diff
@@ -22,6 +22,7 @@ echo "127.0.0.1 my-instance.localhost" | tee -a /etc/hosts
 
 TALOS_VERSION=1.10.2
 ENABLE_TALOS_PRERELEASE_VERSIONS=false
+ANOTHER_OMNI_VERSION="${ANOTHER_OMNI_VERSION:-latest}"
```
Unix4ever, line 25: latest for now, will need to update it to the next prod release which has upgrade tests enabled

utkuozdemir (approved):

PR #1282, opened 2025-06-20
test: save a support bundle when a test suite has failed

utkuozdemir:
Always save it, not only in DestroyCluster. It is not used as finalizer anymore.

### DIFF
--- a/internal/backend/runtime/omni/controllers/omni/infra_machine.go
+++ b/internal/backend/runtime/omni/controllers/omni/infra_machine.go
@@ -240,6 +240,8 @@ func (helper *infraMachineControllerHelper) modify(ctx context.Context, infraMac
 		return err
 	}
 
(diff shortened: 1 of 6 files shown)


Unix4ever (commented):
I think it's better to just collect all support bundles after all tests if some failed. In the `integration_test.go`. It can be a post run hook. See how the provisioned machines are removed

utkuozdemir:
> I think it's better to just collect all support bundles after all tests if some failed. In the `integration_test.go`. It can be a post run hook. See how the provisioned machines are removed

ah that's a good idea, let me give that a try.

utkuozdemir:
> I think it's better to just collect all support bundles after all tests if some failed. In the `integration_test.go`. It can be a post run hook. See how the provisioned machines are removed

Done that, you can have another look.

Unix4ever (approved):
awesome

PR #1283, opened 2025-06-20
feat: allow encoding the machine request ID into the join tokens

Unix4ever:
Now the infra providers can be configured to encode the request ID right into the Join token.
Request ID encoding must be used in pair with `WithoutConnectionParams` flag for the schematics generation.


internal/backend/runtime/omni/controllers/omni/machine_status_snapshot.go
```diff
@@ -111,6 +111,8 @@ func (ctrl *MachineStatusSnapshotController) MapInput(ctx context.Context, _ *za
 		if state.IsNotFoundError(err) {
 			return nil, nil
 		}
+
+		return nil, err
```
utkuozdemir, line 115: good catch

client/pkg/omni/resources/infra/machine_registration.go
```diff
@@ -0,0 +1,45 @@
...
+	"github.com/cosi-project/runtime/pkg/resource"
+	"github.com/cosi-project/runtime/pkg/resource/meta"
+	"github.com/cosi-project/runtime/pkg/resource/typed"
+
+	"github.com/siderolabs/omni/client/pkg/omni/resources"
+	"github.com/siderolabs/omni/client/pkg/omni/resources/system"
+)
+
+// NewMachineRegistration creates a new MachineRegistration resource.
+func NewMachineRegistration(id string) *MachineRegistration {
```
utkuozdemir, line 18: Is this type only for the static infra providers?

client/pkg/omni/resources/infra/machine_registration.go
```diff
@@ -0,0 +1,45 @@
...
+
+const (
+	// MachineRegistrationType is the type of MachineRegistration resource.
+	//
+	// tsgen:MachineRegistrationType
+	MachineRegistrationType = resource.Type("InfraMachineRegistrations.omni.sidero.dev")
+)
+
+// MachineRegistration contains the mininum data extracted from the machine resource required by the infra providers.
+type MachineRegistration = typed.Resource[system.Null, MachineRegistrationExtension]
```
utkuozdemir: Should we still declare a spec type with no fields? In case, for whatever reason, we add a field in the future? So we can evolve the API in a nicer way.

client/pkg/infra/provision/context.go
```diff
@@ -216,6 +218,12 @@ func (context *Context[T]) GenerateSchematicID(ctx context.Context, logger *zap.
 	}
 
 	if !schematicOptions.skipConnectionParams {
+		if context.ConnectionParams.CustomDataEncoded {
+			return "", errors.New(`the provider is configured to encode machine request ID into the join tokens,
+so generating the schematics with the embedded Omni connection params is not allowed.
+Please use WithoutConnectionParams option. The Omni connection params should be injected the other way, which depends on the provider`)
+		}
```
utkuozdemir: Tried to rephrase this a bit.

```suggestion
		if context.ConnectionParams.CustomDataEncoded {
			return "", errors.New(`the provider is configured to embed connection parameters into the schematic, but it also includes a machine request ID, which is not allowed in the connection parameters in the schematic. if the machine request ID must be part of the connection parameters, provide them to the machine through another mechanism using the infrastructure provider.`)
		}
```

Also I think starting the second sentence also lowercase feels more consistent, but that's a nit.

client/pkg/infra/options.go
```diff
@@ -84,3 +85,15 @@ func WithHealthCheckInterval(interval time.Duration) Option {
 		o.healthCheckInterval = interval
 	}
 }
+
+// WithEncodeRequestIDsIntoTokens enables encoding the request IDs into tokens.
+// This elliminates the need for setting the node UUID on the MachineRequestStatus.
```
utkuozdemir:
```suggestion
// This eliminates the need for setting the node UUID on the MachineRequestStatus.
```

client/pkg/omni/resources/infra/machine_registration.go
Unix4ever: yep

client/pkg/omni/resources/infra/machine_registration.go
Unix4ever: makes sense

client/pkg/omni/resources/infra/machine_registration.go
Unix4ever: It's actually for any providers

internal/backend/runtime/omni/controllers/omni/machine.go
```diff
@@ -133,20 +147,38 @@ func (h *machineControllerHelper) handleStaticInfraProvider(ctx context.Context,
...
-	if !ok {
+	machineRequest, err := safe.ReaderGetByID[*infra.MachineRequest](ctx, r, machineRequestID)
+	if err != nil {
 		return xerrors.NewTaggedf[qtransform.SkipReconcileTag](
-			"the link is created by the infra provider, but doesn't have the machine request set label yet",
+			"the link is created by the infra provider, but the machine request doesn't exist",
 		)
 	}
 
+	machineRequestStatus, err := safe.ReaderGetByID[*infra.MachineRequestStatus](ctx, r, machineRequestID)
```
utkuozdemir, line 164: curious, do we need to put a finalizer to this?

utkuozdemir (approved):

internal/backend/runtime/omni/controllers/omni/machine.go
Unix4ever: not really, we don't care about the resource removal

PR #1287, opened 2025-06-24
docs: add guide for development on darwin

Orzelius:
Add some points about the development setup on darwin.

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -1,12 +1,19 @@
 # Developing Omni
 
+This guide assumes a MacOS or a Linux system.
+Using [OrbStack](https://orbstack.dev/) is reccommended on Macos.
+
+Make sure you have a running [Talos development environment](https://www.talos.dev/latest/advanced/developing-talos/) before proceeding with this guide.
+
 Decide on the CIDR for the local set of QEMU Talos machines.
 In this document, we are going to use the `172.20.0.0/24` CIDR, but you can use any CIDR you want.
 
-With the CIDR `172.20.0.0/24`, the bridge IP is going to be `172.20.0.1`, so we are going to use the bridge IP
+With the CIDR `172.20.0.0/24`, the bridge IP is going to be `172.20.0.1`, so we are going to use that IP
 as the Omni endpoint QEMU VMs can reach.
 
 ## Build Omni and omnictl
+
+For example for `linux` on `amd64`.
 
 ```shell
 make omni-linux-amd64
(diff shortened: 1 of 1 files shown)


DEVELOPMENT.md
```diff
@@ -1,13 +1,20 @@
 # Developing Omni
 
+This guide assumes a MacOs or a Linux system.
```
utkuozdemir:
```suggestion
This guide assumes a MacOS or a Linux system.
```

DEVELOPMENT.md
```diff
@@ -1,13 +1,20 @@
...
+Using [OrbStack](https://orbstack.dev/) is reccommended on Macos.
+
+Make sure you have a running [Talos development environment](https://www.talos.dev/latest/advanced/developing-talos/) before proceeding with this guide..
```
utkuozdemir:
```suggestion
Make sure you have a running [Talos development environment](https://www.talos.dev/latest/advanced/developing-talos/) before proceeding with this guide.
```

DEVELOPMENT.md
```diff
@@ -56,6 +63,15 @@ This should result in the following (git-ignored) files:
...
 - `hack/compose/docker-compose.override.yml`
 
+Depending on your environment you'll need to rewrite or add some options to the `hack/compose/docker-compose.override.yml` file.
+
+- If you are on MacOs the following options have to be set due to networking limitations on the platform:
+  - Set `--machine-api-bind-addr` to `'0.0.0.0:8090'`
+  - Set `--siderolink-wireguard-bind-addr` to `'0.0.0.0:50180'`
+  - Set `--siderolink-wireguard-advertised-addr` to `172.20.0.1:50180`
+  - Set `--event-sink-port` to `8091`
+  - Set `--machine-api-advertised-url` to `'grpc://172.20.0.1:8090'`
```
utkuozdemir:
```suggestion
- If you are on MacOs the following options have to be set due to networking limitations on the platform:
  - Set `--machine-api-bind-addr` to `0.0.0.0:8090`
  - Set `--siderolink-wireguard-bind-addr` to `0.0.0.0:50180`
  - Set `--siderolink-wireguard-advertised-addr` to `172.20.0.1:50180`
  - Set `--event-sink-port` to `8091`
  - Set `--machine-api-advertised-url` to `grpc://172.20.0.1:8090`
```

utkuozdemir:
I would suggest adding some kind of warning/note to point out that Linux need to be used for the best development experience, as Omni is designed to run on Linux. Not sure how to phrase that.

Orzelius:
> I would suggest adding some kind of warning/note to point out that Linux need to be used for the best development experience, as Omni is designed to run on Linux. Not sure how to phrase that.

In case of omni wouldn't the dev experience be nearly identical on all systems as long as the initial networking related hurdles are overcome? Is there more linux specific functionality or caveats I'm unaware of?

utkuozdemir:
> > I would suggest adding some kind of warning/note to point out that Linux need to be used for the best development experience, as Omni is designed to run on Linux. Not sure how to phrase that.
> 
> In case of omni wouldn't the dev experience be nearly identical on all systems as long as the initial networking related hurdles are overcome? Is there more linux specific functionality or caveats I'm unaware of?

Not really I guess, just we mostly treat non-Linux OSes as non-first class citizens (I admit it makes much more sense for Talos, but still). So I could imagine people trying to develop in different OSes and having all kinds of unexpected behavior.

But I guess it's fine, it can have its own benefits like improving the developer experience over time. Let's leave it as-is.

PR #1299, opened 2025-07-01
fix: update SAML library to the forked version with the ACS parser fix

Unix4ever:
The fix properly propagates the `xmlns` attributes from the parent elements to the children.

Fusion Auth doesn't set `xmlns` attribute on the `<Assertion>` element, but it's set on the `<Response>` which is the parent element.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,8 @@ module github.com/siderolabs/omni
 go 1.24.4
 
 replace (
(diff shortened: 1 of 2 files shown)


utkuozdemir (approved):

smira (approved):

Unix4ever:
> did we propose a fix upstream?

yep

PR #1303, opened 2025-07-03
chore: rewrite join config generation

Unix4ever:
Now the machine join config is always generate when there's a `machine` resource. It will automatically populate the correct parameters for the machine API URL, logs and events.
If the machine is managed by an infra provider it will populate it's request ID too.

The default provider join config is also generated, but it is not used in the common infra provider library, because it's easier to just generate the config at the moment it's going to be used.

The code for the siderolink join config generation was unified in all the places, and is now in `client/pkg/siderolink`.

The new management API introduced for downloading the join config in the UI `GetMachineJoinConfig`.


client/pkg/siderolink/siderolink.go
```diff
@@ -0,0 +1,266 @@
...
+// Without the token it will try to use tokenless access to Omni.
+func WithJoinToken(token string) JoinConfigOption {
+	return func(opts *JoinConfigOptions) {
+		opts.joinToken = token
+	}
+}
+
+// JoinOptions is the intermediate struct used in the kernel args and the partial config generation.
+type JoinOptions struct {
+	APIURL            *url.URL
```
Unix4ever: Make this struct private

client/pkg/siderolink/siderolink.go
```diff
@@ -0,0 +1,266 @@
...
+	}
+}
+
+// WithLogServerPort specifies log server port.
+// Without the port it will not render the KmsgLog config section.
+func WithLogServerPort(value int) JoinConfigOption {
+	return func(opts *JoinConfigOptions) {
+		opts.logServerPort = value
+	}
+}
```
Unix4ever, line 111: Make these always required

client/pkg/siderolink/siderolink.go
```diff
@@ -0,0 +1,266 @@
...
+
+// WithMachineAPIURL sets the machine API URL.
+// Without the API URL it will not render the SideroLink config section.
+func WithMachineAPIURL(value string) JoinConfigOption {
+	return func(opts *JoinConfigOptions) {
+		opts.machineAPIURL = value
+	}
+}
```
Unix4ever, line 95: Make this required, introduce `WithoutMachineAPIURL` option.

client/pkg/siderolink/siderolink.go
```diff
@@ -0,0 +1,266 @@
...
+
+// WithJoinToken adds the join token to the machine API URL.
+// Without the token it will try to use tokenless access to Omni.
+func WithJoinToken(token string) JoinConfigOption {
+	return func(opts *JoinConfigOptions) {
+		opts.joinToken = token
+	}
+}
```
Unix4ever, line 130: Make this required

internal/backend/runtime/omni/controllers/omni/machine_join_config.go
```diff
@@ -0,0 +1,79 @@
...
+				}
+
+				config, err := siderolink.RenderJoinConfig(joinOptions)
+				if err != nil {
+					return err
+				}
+
+				kernelArgs := siderolink.GetKernelArgs(joinOptions)
+
+				providerJoinConfig.TypedSpec().Value.Config = &specs.JoinConfig{
```
Unix4ever:
```suggestion
				machineJoinConfig.TypedSpec().Value.Config = &specs.JoinConfig{
```

Issue #1102, opened 2025-04-15
some Talos machine logs are lost when Omni is restarted


smira:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

We have multiple evidences of Talos logs (several lines) being lost in Omni around the time Omni restarts.

The lost logs are _after_ Omni restarts.

### Expected Behavior

The actual bug is not clear - is it Omni or Talos?

Talos should retry sending logs, so it should not lose logs when Omni is down/restarted.

### Steps To Reproduce

.

PR #1304, opened 2025-07-04
fix: gracefully handle logServer shutdown

oguzkilcan:
Refactor the siderolink manager to improve context handling for Wireguard and its associated services.
This ensures proper shutdown and cleanup of resources.
Add a linger option to TCP connections in the log receiver to allow the OS a grace period to close the connection.

Fixes: #1102

### DIFF
--- a/internal/pkg/siderolink/manager.go
+++ b/internal/pkg/siderolink/manager.go
@@ -265,17 +265,36 @@ func (manager *Manager) Run(
 	trustdPort string,
 	logServerPort string,
 ) error {
-	eg, groupCtx := errgroup.WithContext(ctx)
+	eg, ctx := errgroup.WithContext(ctx)
 
-	if err := manager.startWireguard(groupCtx, eg, manager.serverAddr); err != nil {
+	if err := manager.runWireguard(ctx, eg, listenHost, eventSinkPort, trustdPort, logServerPort); err != nil {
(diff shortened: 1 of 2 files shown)


internal/pkg/siderolink/manager.go
```diff
@@ -265,17 +265,36 @@ func (manager *Manager) Run(
...
+}
 
-	eg.Go(func() error {
-		return manager.provisionServer.runCleanup(groupCtx)
-	})
+func (manager *Manager) runWireguard(ctx context.Context, eg *errgroup.Group, listenHost, eventSinkPort, trustdPort, logServerPort string) error {
+	wireguardCtx, wireguardCancel := context.WithCancel(context.Background())
+	defer wireguardCancel()
+
+	if err := manager.startWireguard(wireguardCtx, eg, manager.serverAddr); err != nil { //nolint:contextcheck
```
smira, line 281: what does shutdown `startWireguard` now?

oguzkilcan: `startWireguard` is now receiving a cancellable context and a different error group, so we can wait for all wireguard dependent services(incuding logServer) to finish separately before attempting to tore down wireguard connection.

utkuozdemir (approved):
LGTM

PR #1324, opened 2025-07-10
fix: add validation of k8s version

### LINKED ISSUE #1284: [bug] Cluster template sync allows k8s upgrades across > 1 minor version

### Current Behavior

- UI only allows upgrades to the latest patch of the next minor version (e.g., from `1.28.15` to `1.29.15`).
- Cluster templates allow setting higher unsupported minor versions (e.g., `1.30.5`), which bypasses this protection.
- Sync fails only if the version is invalid (e.g., `1.29.16`).

### Expected Behavior

Cluster templates should enforce the same Kubernetes version upgrade restrictions as the UI.

### Steps To Reproduce

`omnictl cluster template export <cluster> -o cluster-template.yaml`

update k8s version > 2 minor versions

`omnictl cluster template sync -f cluster-template.yaml`

cluster will start upgrading to unsupported version

_No response_

### Anything else?

I believe downgrading also possible through this

shanduur:
We lack K8s upgrade path validation - this allowed upgrades using API to jump over minor Kubernetes releases, which is not a supported upgrade pattern. This PR adds new validation and reworks some of the tests.

Fixes #1284

### DIFF
--- a/internal/backend/runtime/omni/state_validation.go
+++ b/internal/backend/runtime/omni/state_validation.go
@@ -21,6 +21,7 @@ import (
 	"github.com/cosi-project/runtime/pkg/safe"
 	"github.com/cosi-project/runtime/pkg/state"
 	"github.com/hashicorp/go-multierror"
+	"github.com/siderolabs/go-kubernetes/kubernetes/upgrade"
 
 	"github.com/siderolabs/omni/client/api/omni/specs"
 	"github.com/siderolabs/omni/client/pkg/cosi/labels"
@@ -69,6 +70,16 @@ func clusterValidationOptions(st state.State, etcdBackupConfig config.EtcdBackup
 
 		if skipKubernetesVersion {
 			return nil
+		}
+
+		var currentKubernetesVersion string
+
+		if existingRes != nil {
+			currentKubernetesVersion = existingRes.TypedSpec().Value.KubernetesVersion
+		}
+
+		if err = validateKubernetesVersion(currentKubernetesVersion, res.TypedSpec().Value.KubernetesVersion); err != nil {
+			return err
 		}
 
 		for _, compatibleKubernetesVersion := range talosVersion.TypedSpec().Value.CompatibleKubernetesVersions {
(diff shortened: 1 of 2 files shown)


internal/backend/runtime/omni/state_validation.go
```diff
@@ -156,6 +167,7 @@ func clusterValidationOptions(st state.State, etcdBackupConfig config.EtcdBackup
 
 			skipTalosVersion := existingRes.TypedSpec().Value.TalosVersion == newRes.TypedSpec().Value.TalosVersion
 			skipKubernetesVersion := existingRes.TypedSpec().Value.KubernetesVersion == newRes.TypedSpec().Value.KubernetesVersion
+			// skipKubernetesVersion := skipTalosVersion && existingRes.TypedSpec().Value.KubernetesVersion == newRes.TypedSpec().Value.KubernetesVersion
```
shanduur: I'm not sure about this. There was a test case that confirmed that previous behavior was desired, however as we discussed with @smira this should only be true when the Talos version check is skipped.

smira: so my reasoning is the following: if I trigger a Talos upgrade which will make Kubernetes version not compatible with new Talos, we should ignore such request

Unix4ever: I'm struggling to understand the reason behind this condition.
I think it's fine to drop it.

The validation should check if the new Talos version (or current if not changed) supports the new Kubernetes version (or current).

Unix4ever: > so my reasoning is the following: if I trigger a Talos upgrade which will make Kubernetes version not compatible with new Talos, we should ignore such request

Ahh, yeah. Now I remember. This check is probably required there.
We need to keep validating the kubernetes version even if it's not changed when we try to update Talos.

shanduur: I adjusted the code to include the check suggested by @smira, thanks!

Unix4ever (approved):

PR #1332, opened 2025-07-14
feat: add support for imported cluster secrets

### LINKED ISSUE #1312: implement `ImportedClusterSecrets`
A user-controlled resource, if created before the `Cluster` is created, blocks generation of fresh `ClusterSecrets`, and instead `ClusterSecrets` is copied from `ImportedClusterSecrets`.

Cluster should be tainted (`ClusterStatus`? labeled/annotated) if `ImportedClusterSecrets` were used until the PKI is rotated in Omni.

(hide behind a feature gate)

oguzkilcan:
Introduce new resource `ImportedClusterSecrets` for importing an existing secrets bundle.
Add new field `imported` to `ClusterSpec` for utilizing resource `ImportedCreatedSecrets`.
Add new field `imported` to `ClusterSecrets` for pointing out source of the secrets bundle.

This is a feature-gated feature to allow using an existing secrets bundle (`talos gen secrets`) while creating a new Cluster. Cluster created with this method are marked as `tainted`. This feature is part of a story to facilitate importing existing talos clusters to omni.

Fixes #1312


internal/integration/blocks_test.go
```diff
@@ -231,26 +231,23 @@ func AssertBlockKubernetesDeploymentCreateAndRunning(ctx context.Context, manage
 }
 
 // AssertClusterCreateAndReady is a reusable group of tests that can be used to verify that a cluster is created and ready.
-func AssertClusterCreateAndReady(ctx context.Context, rootClient *client.Client, name string, options ClusterOptions) []subTest { //nolint:nolintlint,revive
-	clusterName := "integration-" + name
-	options.Name = clusterName
-
+func AssertClusterCreateAndReady(ctx context.Context, rootClient *client.Client, options ClusterOptions) []subTest { //nolint:nolintlint,revive
```
utkuozdemir, line 234: For the reviewer: This refactor was my suggestion, so that we have a single way of passing cluster name. Not related otherwise to the PR.

utkuozdemir (commented):
LGTM, but I'll leave the approval to someone else, as I was already involved in it.

internal/backend/runtime/omni/state_validation.go
```diff
@@ -128,6 +128,14 @@ func clusterValidationOptions(st state.State, etcdBackupConfig config.EtcdBackup
 		return nil
 	}
 
+	validateClusterImport := func(res *omni.Cluster) error {
+		if !clusterImportEnabled && res.TypedSpec().Value.Imported {
```
smira: I think `imported` field should be immutable in the `Cluster` resource, and it doesn't look like it is this way today

utkuozdemir: I was skeptical about this, since immutable fields cause a worse UX - for example, if we don't add this "imported" flag to the cluster templates, apply won't be possible (imagine you are migrating the cluster to be from manually to template-managed), as it'll try to set it back to `false`.

This field only matters at the moment of secrets creation, not later. So I think being able to set it to false later doesn't harm. Just my 2 cents.

smira: In that case the field shouldn't be there in the first place - just use ImportedClusterSecrets if available, otherwise generate new ones. 

As otherwise it becomes kind of weird one way or another if this field can be changed later, but has no effect.

internal/backend/runtime/omni/controllers/omni/secrets.go
```diff
@@ -65,6 +67,33 @@ func NewSecretsController(etcdBackupStoreFactory store.Factory) *SecretsControll
 					return err
 				}
 
+				if cluster.TypedSpec().Value.Imported {
+					var ics *omni.ImportedClusterSecrets
```
oguzkilcan: remove this line

internal/backend/runtime/omni/controllers/omni/secrets.go
```diff
@@ -65,6 +67,33 @@ func NewSecretsController(etcdBackupStoreFactory store.Factory) *SecretsControll
...
+					if icsErr != nil {
+						return fmt.Errorf("failed to decode imported cluster secrets: %w", icsErr)
+					}
+
+					data, icsErr := json.Marshal(bundle)
+					if icsErr != nil {
+						return fmt.Errorf("error marshaling secrets: %w", icsErr)
+					}
+
+					secrets.TypedSpec().Value.Imported = true
```
oguzkilcan, line 88: set imported to false in normal flow

internal/backend/runtime/omni/state_validation.go
oguzkilcan: Removed `imported` field from ClusterSpec.

PR #1342, opened 2025-07-18
feat: implement the API for reading resources and their dependency graph

Unix4ever:
Then another Tool can represent it in a nice way.
This will be a good education tool for the start, and then we can add monitoring there too.


Unix4ever:
https://github.com/user-attachments/assets/1e92644d-5a47-4db0-80d6-d774b524ccd4

internal/backend/server.go
```diff
@@ -207,7 +207,7 @@ func (s *Server) Run(ctx context.Context) error {
 
 	servicesServer := grpcomni.MakeServiceServers(
 		s.state.Default(),
-		s.omniRuntime.CachedState(),
+		s.omniRuntime,
```
smira, line 210: is this intentional?

Unix4ever: yeah, I pass the runtime as the whole to the resource server now. It needs to get the dependency graph from it

Orzelius:
This looks soo nice! 🔥

client/pkg/client/client.go
```diff
@@ -135,6 +136,11 @@ func (c *Client) Talos() *talos.Client {
 	return talos.NewClient(c.conn)
 }
 
+// Resource provides access to the resources API.
+func (c *Client) Resource() *resource.Client {
```
utkuozdemir: This name might be a bit confusing for the API consumers - I'd need to know that to get resources I need to use `.COSI()`, and `.Resource()` does something else. Maybe we find a better name?

client/api/omni/resources/resources.proto
```diff
@@ -105,4 +140,7 @@ service ResourceService {
   rpc Delete(DeleteRequest) returns (DeleteResponse);
   rpc Teardown(DeleteRequest) returns (DeleteResponse);
   rpc Watch(WatchRequest) returns (stream WatchResponse);
+
+  rpc Controllers(google.protobuf.Empty) returns (ControllersResponse);
```
utkuozdemir: Maybe let's not use `Empty` and have an empty request message type instead? Just in case for the future.

internal/backend/grpc/dependency_graph.go
```diff
@@ -0,0 +1,243 @@
...
+	"github.com/cosi-project/runtime/pkg/state"
+	"github.com/siderolabs/gen/xslices"
+	"github.com/siderolabs/omni/client/api/omni/resources"
+	"github.com/siderolabs/omni/internal/pkg/auth/actor"
+	"github.com/siderolabs/talos/pkg/machinery/proto"
+	"google.golang.org/protobuf/types/known/emptypb"
+)
+
+func (s *ResourceServer) Controllers(ctx context.Context, _ *emptypb.Empty) (*resources.ControllersResponse, error) {
+	graph, err := s.runtime.GetDependencyGraph()
```
utkuozdemir, line 39: Shouldn't we do an authz check here?

internal/backend/grpc/dependency_graph.go
```diff
@@ -0,0 +1,243 @@
...
+			source string
+			target string
+			style  string
+		)
+
+		switch edge.EdgeType {
+		case controller.EdgeOutputExclusive:
+			source = nodesMap[edge.ControllerName]
+			target = nodesMap[edge.ResourceType]
+			style = "bold"
```
utkuozdemir: This is more of a nit, but should the backend tell how it should be rendered? Maybe instead we return the edge type as-is, and frontend can apply the style it wants. Backend would just return the data, no style.

client/pkg/client/client.go
Unix4ever: Let me try that

internal/backend/grpc/dependency_graph.go
Unix4ever: Yeah, I guess. This API is available for the debug builds only, so not a big deal, but I will add that anyways

client/pkg/client/client.go
Unix4ever: Actually I don't need this for the inspector, so I'll just drop this for now

PR #1352, opened 2025-07-22
fix: preserve encoded join token version

utkuozdemir:
If a machine is using a v1 token, generate its join config to also use the v1 version, do not promote it to be v2. This is needed for the existing bare-metal provider machines to not lose connectivity, because their tokens now are validated against v2 logic.

@Unix4ever I did this fix to make bare metal provider able to work again as-is, with their v1 tokens. But not exactly sure how much sense I am making. Can you check if I understood the problem right?

### DIFF
--- a/internal/backend/runtime/omni/controllers/omni/machine_join_config.go
+++ b/internal/backend/runtime/omni/controllers/omni/machine_join_config.go
@@ -14,6 +14,7 @@ import (
 	"go.uber.org/zap"
 
 	"github.com/siderolabs/omni/client/api/omni/specs"
+	"github.com/siderolabs/omni/client/pkg/jointoken"
 	"github.com/siderolabs/omni/client/pkg/omni/resources"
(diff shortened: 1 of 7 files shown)


internal/backend/runtime/omni/controllers/omni/machine_join_config.go
```diff
@@ -51,6 +57,7 @@ func NewMachineJoinConfigController() *MachineJoinConfigController {
 					siderolink.WithLogServerPort(int(siderolinkAPIConfig.TypedSpec().Value.LogsPort)),
 					siderolink.WithMachineAPIURL(siderolinkAPIConfig.TypedSpec().Value.MachineApiAdvertisedUrl),
 					siderolink.WithMachine(machine),
+					siderolink.WithVersion(parsedTokenUsage.Version),
```
utkuozdemir, line 60: This is the fix

PR #1356, opened 2025-07-23
fix: ignore `MachineStatus` having no TalosVersion in DNS service

Unix4ever:
We can probably tolerate it being empty in the Talos request proxying. Otherwise the behavior is confusing: the node is connected, but it's not possible to reach it with `talosctl` using the node UUID.

### DIFF
--- a/internal/backend/dns/service.go
+++ b/internal/backend/dns/service.go
@@ -272,8 +272,6 @@ func (d *Service) updateEntryByMachineStatus(res *omni.MachineStatus) {
 	version := res.TypedSpec().Value.TalosVersion
 	if version == "" {
 		d.logger.Warn("no Talos version in the machine status", zap.String("id", res.Metadata().ID()))
-
-		return
 	}
 
 	d.lock.Lock()


smira:
What would happen if we keep it empty? It looks like this check was there for some reason?

smira:
What I mean, is that we could replace it with "latest" talos version as a wild guess, but leaving it empty looks against the initial intention of the author.

Unix4ever:
> What I mean, is that we could replace it with "latest" talos version as a wild guess, but leaving it empty looks against the initial intention of the author.

I'd say we should let it use the lowest possible version instead. As this field is used to identify which role to impersonate for Talos.
Talos below 1.4 didn't have Operator role, so we were using Admin for it.

If we do not populate the version, it will be unknown and then the code will fallback to the lowest.

smira (approved):

smira:
Looks like from https://github.com/siderolabs/omni/issues/1357 that it's not talos version issue.

utkuozdemir (approved):

Unix4ever:
> Looks like from https://github.com/siderolabs/omni/issues/1357 that it's not talos version issue.

yeah, I'm looking

Unix4ever:
In any case I don't think this is a bad change. So I will merge it

PR #1362, opened 2025-07-24
release(v1.1.0): prepare release

Unix4ever:
This is the official v1.1.0 release.


Unix4ever:
> idk about 1.1.0 vs. 1.0.0 :) but I don't think it's important enough

Yeah, same

utkuozdemir:
> > idk about 1.1.0 vs. 1.0.0 :) but I don't think it's important enough
> 
> 
> 
> Yeah, same

I thought the 1.1.0 thing was a joke tbh 🙂. It feels a bit weird to do 1.1 before 1.0, and also, making 1.0 betas and then release 1.1. I'd vote for doing 1.0.

Unix4ever:
> I thought the 1.1.0 thing was a joke tbh 🙂. It feels a bit weird to do 1.1 before 1.0, and also, making 1.0 betas and then 

Well, if we care that much then lets just cut 1.0

PR #1365, opened 2025-07-25
chore: run inspector in the dev docker-compose

Unix4ever:
The inspector will start on `0.0.0.0:12000` if the compose is started `WITH_DEBUG=true`.


Unix4ever:
depends on: https://github.com/siderolabs/omni-inspector/pull/3

utkuozdemir (approved):

Issue #1093, opened 2025-04-14
Join Token CLI support


smira:
Equivalent to Omni UI operations - revoke, create, ...

PR #1377, opened 2025-07-28
feat: implement join token management CLI

Unix4ever:
The commands added:
```
omnictl jointoken create
omnictl jointoken delete
omnictl jointoken renew
omnictl jointoken revoke
omnictl jointoken unrevoke
omnictl jointoken make-default
```

Fixes: https://github.com/siderolabs/omni/issues/1093


client/pkg/omnictl/jointoken.go
```diff
@@ -0,0 +1,274 @@
...
+
+	joinTokenCmd.AddCommand(joinTokenCreateCmd)
+	joinTokenCmd.AddCommand(joinTokenListCmd)
+	joinTokenCmd.AddCommand(joinTokenDeleteCmd)
+	joinTokenCmd.AddCommand(joinTokenRevokeCmd)
+	joinTokenCmd.AddCommand(joinTokenMakeDefaultCmd)
+	joinTokenCmd.AddCommand(joinTokenUnrevokeCmd)
+	joinTokenCmd.AddCommand(joinTokenRenewCmd)
+
+	joinTokenCreateCmd.Flags().DurationVarP(&joinTokenCreateFlags.ttl, "ttl", "t", 0, "TTL for the join token")
```
smira, line 269: is zero reasonable default here?

Unix4ever: yeah. 0 will make this token never expire

smira: sorry, I commented on the wrong one, but the one below - the renew TTL

Unix4ever: I think it's fine there too as with 0 it will fail the command and then the user will have to set some value

Unix4ever: I marked that flag as required for that command

Issue #1348, opened 2025-07-21
migration: generate node unique tokens for all old machines connected to Omni


Unix4ever:
The current implementation will generate the node unique tokens only when the machine calls the provision API, which won't happen for the machines which were already connected and running before the `strict` join token mode was enabled.

Reconnect might happen eventually but that hard to predict.
So it's better to have some migration that will go over each connected machine in on the account and generate the node unique token for it.

PR #1379, opened 2025-07-29
feat: compute unique token status for each node

Unix4ever:
The node token status can be used to check if the machine has the unique token generated. It also shows the exact token state:

- `NONE` - token is supported, but not yet generated.
- `UNSUPPORTED` - Talos is < 1.6.x, so token won't be generated.
- `EPHEMERAL` - token is generated, but is not persistent, so join token rotation and machine reboot will make the node to disconnect.
- `PERSISTENT` - token is generated and is persisted to the `META` partition.

If the node unique status is `NONE` the same controller will try to generate node unique token.

Fixes: https://github.com/siderolabs/omni/issues/1348


Unix4ever:
New controller schema:

Screen Recording 2025-07-29 at 17 04 30(1)

internal/backend/runtime/omni/controllers/omni/node_unique_token_status.go
```diff
@@ -0,0 +1,211 @@
...
+	ctx, cancel := context.WithTimeout(ctx, time.Second*10)
+	defer cancel()
+
+	logger.Info("proactively generating node unique token")
+
+	return handler.generateUniqueNodeToken(ctx, c, logger)
+}
+
+// getMachineFingerprint gets all network devices and calculates the checksum out of their mac addresses.
+func (handler *nodeUniqueTokenStatusHandler) getMachineFingerprint(ctx context.Context, c *client.Client) (string, error) {
```
utkuozdemir: should we reuse this from `pendingMachineStatusHandler`? move to a common place?

Unix4ever: yep, can do

Issue #1380, opened 2025-07-30
[feature] provide a summary for the token if it can be rotated or not


Unix4ever:
The summary should have all machines which are using the token, their usage status.
The text representation should be presented to the user when he tries to revoke the token via UI or CLI.

PR #1384, opened 2025-07-30
feat: warn user about join token being in use during revoke/delete

Unix4ever:
The UI now shows warning if the token revoke/delete operation will render some machines not able to connect.
Same thing was done for the CLI commands.

Fixes: https://github.com/siderolabs/omni/issues/1380


Unix4ever:
Screenshot 2025-07-30 at 20 59 23
Screenshot 2025-07-30 at 20 59 42

PR #1400, opened 2025-08-06
fix: stop enforcing talos version check on machine allocation

### LINKED ISSUE #1358: [bug] do not enforce major.minor version check on machine allocation if the machine is (or was) running in agent mode

### Current Behavior

When creating a cluster using bare metal provider, if the agent mode major.minor was different than the cluster version, we get this:
```
machines which are running Talos without installation can be added only to Talos clusters with the same major and minor versions
```

### Expected Behavior

Do not enforce this check somehow, as we will PXE moot the machine into Talos (non-agent) with the correct version as the first step, then apply the config.

### Steps To Reproduce

Run the provider with agent mode talos version 1.9, try to create a cluster with 1.10.

_No response_

### Anything else?

_No response_

oguzkilcan:
Do not enforce major.minor version check for talos on machine allocation if the machine is running in agent mode

Fixes #1358

### DIFF
--- a/frontend/src/views/omni/Clusters/Management/ClusterCreate.vue
+++ b/frontend/src/views/omni/Clusters/Management/ClusterCreate.vue
@@ -301,6 +301,7 @@ const detectVersionMismatch = (machine: Resource<MachineStatusSpec>) => {
   const machineVersion = semver.parse(machine.spec.talos_version);
 
   const installed = machine.metadata.labels?.[MachineStatusLabelInstalled] !== undefined;
+  const inAgentMode = !!machine.spec.schematic?.in_agent_mode;
 
   if (!machineVersion || !clusterVersion) {
     return null;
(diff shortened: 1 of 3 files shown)


frontend/src/views/omni/Clusters/Management/ClusterCreate.vue
```diff
@@ -301,6 +301,7 @@ const detectVersionMismatch = (machine: Resource<MachineStatusSpec>) => {
   const machineVersion = semver.parse(machine.spec.talos_version);
 
   const installed = machine.metadata.labels?.[MachineStatusLabelInstalled] !== undefined;
+  const inAgentMode = machine.spec.schematic?.in_agent_mode !== undefined;
```
Slessi: If `in_agent_mode` is `false`, do you still want to `return null` later? (Same question for the other file)

If `false` is not possible, and/or you only care about `true`, I would do it like this for readability & clarity.

```suggestion
  const inAgentMode = !!machine.spec.schematic?.in_agent_mode;
```

oguzkilcan: > If in_agent_mode is false, do you still want to return null later?

No,  at this point I don't plan on using this variable further. Basically we have two cases:
1. `in_agent_mode` is `true`: We do not show the warning and allow allocation of the machine.
2. `in_agent_mode` is `false` or `null`: We show the warning and do not allow allocation of the machine.

Since I only care about the `true` value for my if condition. I'll update the code as you suggested.

smira (approved):

PR #1401, opened 2025-08-06
feat: rework breadcrumbs for clusters page

### LINKED ISSUE #1391: [UI] more widespread breadcrumbs usage
Currently breadcrumbs work only on a single cluster level:

```
Cluster / Machine / Service Name
```

We can make it also include:

```
Clusters / Cluster / Machine / Service Name
```

Slessi:
Replace "All Nodes" with "Clusters" and link to clusters page instead of the all nodes page. In the original issue it was not completely clear to me what the expected end result is meant to be.

Additionally, whilst this makes it work it more nicely for the clusters page, it feels a bit weird for the nodes page. When you navigate from there, it will now appear as if your trail of breadcrumbs is from the clusters page.

Thoughts?

Fixes #1391


smira:
I'd ping @rothgar on this one, and probably if you could attach some screen capture to make it easier to see how it looks now.

I think one of the issues was a flow like:

* open cluster overview page
* click on any machine
* breadcrumbs point back to "all nodes", but this is not where I came from.

smira:
Or should we push "all nodes" as a path from `Cluster A` -> `All Nodes` -> `Machine A`

(also confusingly nodes and machines, as `All Nodes` shows Kubernetes nodes, and other pages show machines).

Slessi:
@smira Here's a video

I navigate to clusters, click through the breadcrumbs to show the trail, then go to the nodes, and show the confusing trail change

https://github.com/user-attachments/assets/f78fed6f-61a6-4881-9d2e-795801afa2af

smira:
yep, it feels wrong still to me (I can also do a cluster overview vs. clusters list before I click on a machine)

Slessi:
I could try go for something that makes two breadcrumb paths, depending on the path you took to arrive to those pages, i.e.

Cluster / machine-xxx / machine logs
All Nodes / machine-xxx / machine logs

But it would have to be manual for each case for now. Eventually can look into a more "automatic" breadcrumb trail.

That said there might be some deeper issue here with where things are located in general in the sidebar, but I'll hold on till I get a better understanding of machine/node/cluster and all of that jazz before making suggestions to this effect

rothgar:
IMO breadcrumbs should show me the hierarchy, not "how I got here"

Similar to a file path in file browser. It doesn't matter how I got into a folder, the hierarchy always shows parent folder(s). I can use browser back button to return from where I came.

So in that case I think the breadcrumbs should always show

Cluster (the cluster name, not generic "cluster") / Machine (name)

Having additional path shortcuts for All clusters and All nodes doesn't make sense with how other interfaces (web browsers, files, etc) show paths

Currently, the machine logs tab shows an extra breadcrumb path, but the other tabs don't. I'm not sure why that is and maybe we should remove that breadcrumb.

image

image

Slessi:
I agree with the hierarchy system and that the tabs are extra. I changed it to align with this, no "All Nodes" or "Clusters" or "Machines". There is now in effect only a breadcrumb on a node, and only if it is part of a cluster.

See vid:

https://github.com/user-attachments/assets/a36ff259-159c-4e1a-a345-5590f73e93bd

rothgar (approved):

PR #1405, opened 2025-08-07
fix: keep control plane status up to date

### LINKED ISSUE #1404: [bug] Etcd Status does not get updated after it's set as Not Ready

### Current Behavior

ControlPlaneStatusController does not keep Etcd status updated. When the status is set as not ready, it stays as is unless one of the inputs of the controller gets updated. Controller doesn't actually continue to monitor the status.

### Expected Behavior

ControlPlaneStatusController keeps the Etcd status up to date regardless of it's inputs. It should continue to monitor Etcd status in regular basis in case there are changes. Its shouldn't only depend on it's inputs.

### Steps To Reproduce

1. Restart omni.
2. Introduce network latency to the control plane nodes, so wireguard connection doesn't get established before ControlPlaneStatusController reconciles automatically (because of omni restart).
3. Fix the network latency. 
4. Go to cluster machine details for each control plane node and see Etcd is running and healthy
5. See on the Cluster details screen Control Plane and Etcd is not ready and continues to stay as long as there are no updates to Cluster.
6. Make a non-trivial change to cluster (e.g cluster level config patch to add node labels). See that Control Plane and Etcd is reported as Ready/OK. 

_No response_

### Anything else?

_No response_

oguzkilcan:
- Regularly wake up ControlPlaneStatusController to ensure the status
- Add ControlPlaneStatusType to SupportBundle
- Decrease the wait time if ControlPlaneStatus has errors, so we can reflect the actual state faster
 
Fixes #1404

### DIFF
--- a/internal/backend/runtime/omni/controllers/omni/control_plane_status.go
+++ b/internal/backend/runtime/omni/controllers/omni/control_plane_status.go
@@ -30,8 +30,11 @@ import (
 // ControlPlaneStatusController creates ControlPlaneStatus resource for controlplane MachineSets.
 type ControlPlaneStatusController = qtransform.QController[*omni.MachineSet, *omni.ControlPlaneStatus]
 
-// ControlPlaneCheckTimeout is the timeout for the controlplane checks.
-const ControlPlaneCheckTimeout = 5 * time.Minute
+const (
+	// ControlPlaneCheckTimeout is the timeout for the controlplane checks.
+	ControlPlaneCheckTimeout = 5 * time.Minute
+	requeueTimeout           = 30 * time.Minute
+)
 
 // NewControlPlaneStatusController initializes ControlPlaneStatusController.
 func NewControlPlaneStatusController() *ControlPlaneStatusController {
@@ -81,6 +84,7 @@ func NewControlPlaneStatusController() *ControlPlaneStatusController {
 					interrupted     bool
 				)
 
+				var checkErr error
 				for _, handler := range handlers {
 					if interrupted {
 						spec.SetCondition(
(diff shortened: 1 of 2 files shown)


internal/backend/runtime/omni/controllers/omni/control_plane_status.go
```diff
@@ -112,7 +118,7 @@ func NewControlPlaneStatusController() *ControlPlaneStatusController {
 					spec.SetCondition(handler.condition, specs.ControlPlaneStatusSpec_Condition_Ready, specs.ControlPlaneStatusSpec_Condition_Info, "")
 				}
 
-				return nil
+				return controller.NewRequeueInterval(requeueTimeout)
```
smira, line 124: I would suggest to return the error itself on error condition:

* first, it would be visible in the logs
* second, controller-runtime will automatically do exponential backoff retries

We could aggregate all errors from all checks via `errors.Join` and send the aggregate if it's non-nil.

On "success" we can do retry on healthy like above 👍

smira: (the fear to return errors predates QControllers)

internal/backend/runtime/omni/controllers/omni/control_plane_status.go
```diff
@@ -112,7 +118,11 @@ func NewControlPlaneStatusController() *ControlPlaneStatusController {
 					spec.SetCondition(handler.condition, specs.ControlPlaneStatusSpec_Condition_Ready, specs.ControlPlaneStatusSpec_Condition_Info, "")
 				}
 
-				return nil
+				if checkErr != nil {
+					return controller.NewRequeueError(checkErr, requeueOnErrTimeout)
```
smira: please just do `return checkErr` to let the runtime handle retries

smira (approved):
let's backport to 1.0 and create 1.0.1 👍

PR #1412, opened 2025-08-11
test: fix asserting etcd members test

oguzkilcan:
In test `AssertEtcdMembershipMatchesOmniResources` we compare existing etcd members with ClusterMachines with retry, however we weren't refreshing etcd member list after the first fetch and this lead to test failure if the member list info was constructed incorrectly.


internal/integration/talos_test.go
```diff
@@ -318,6 +318,7 @@ func AssertEtcdMembershipMatchesOmniResources(testCtx context.Context, client *c
 		require.NoError(err)
 
 		err = retry.Constant(time.Minute*2, retry.WithUnits(time.Second)).Retry(func() error {
+			resp, _ = c.EtcdMemberList(ctx, &machine.EtcdMemberListRequest{})
```
utkuozdemir: Should we swallow errors here? Maybe do retry.ExpectedError instead (only for expected types ofc, checking via errors.Is and/or GRPC's status.Code etc.). Otherwise hard fail.

oguzkilcan: We already check error for this on the first call and fail fast if it returns error. I assumed that this error check can be omitted on subsequent calls.

smira: I think Utku is right in the sense that if we don't check for error, we might get a confusing error about `resp` being nil for example.

Also (unrelated to this specific change) we could instead of ugly `retry.Constant` use `assert.Eventually*` family of functions to do proper assertions within the test with better behavior.

utkuozdemir: Agree, just also be careful with Eventually* stuff with context cancellation - check if context has error on each iteration, and fail fast if there is.

PR #1432, opened 2025-08-14
feat: support commands for getting kernel args and join configs in CLI

Unix4ever:
Two new commands for getting the partial machine join config and kernel args for a particular token ID:
- `omnictl jointoken machine-config <token-id>`
- `omnictl jointoken kernel-args <token-id>`


shanduur (approved):

rothgar:
This is all that's needed for #1433, right?

Unix4ever:
Yes, that's it. A bit different from what you suggested, but your suggestion might not work as the user should always specify the token to use.

PR #1434, opened 2025-08-14
feat: make sidebar collapsible on mobile

### LINKED ISSUE #1369: [UI] collapsible sidebar menu
The sidebar should be automatically collapsed for the small screens.
It will be shown when clicking the hamburger button.

Slessi:
Make the sidebar collapsible on small screen sizes

Fixes #1369 

https://github.com/user-attachments/assets/3e9d9850-8e09-4b7b-8e98-c6e490f950a0


Slessi:
@Unix4ever Just to confirm, what screen size is that? right now its collapsible at <768px, and I think an ipad mini on the narrowest orientation is 768px so its on the edge of deciding _not_ to shrink.

I think we can definitely improve the content side of things to make it looks good with sidebar open at 768px, personally I would opt for that, but I can also increase the collapse tier now if you like

Unix4ever:
> I think we can definitely improve the content side of things to make it looks good with sidebar open at 768px, personally I would opt for that, but I can also increase the collapse tier now if you like

If we can do it without collapsing the sidebar then I'm up for it. I'm just not 100% sure.

PR #1440, opened 2025-08-15
feat: change default location for storing Omni config and user PGP keys

### LINKED ISSUE #1375: [feat] Store `omniconfig` and pgp keys under `~/.talos/omni` instead of XDG directory specification
Using XDG paths proved to be confusing for many users: It spreads the files across different directories, and to different paths on different OSes (for example, keys are under ~/.local/share/omni/keys/ on Linux and ~/Library/Application\ Support/omni/keys/ under MacOS).

It also makes some use cases difficult, such as mounting all Omni config inside a container and make it use user auth. If they were instead under `~/.talos/omni`, we could simply do:
```
docker run -v ~/.talos/omni/:/root/.talos/omni:ro ...
```

Let's change the config path to be `~/.talos/omni/`, and migrate the existing files from their XDG dirs to the new location on the first attempt to load them (might need to do changes in https://github.com/siderolabs/go-api-signature and Omni).

oguzkilcan:
Default location for Omni configuration and user PGP keys were under respectively `XDG_CONFIG_HOME` and `XDG_DATA_HOME`. Even though this environment variables are customizable, default values for these hugely differ based on the operating system. We want to move towards `kubectl` approach of using Users' home directory to store Omni configuration and user PGP keys.

This change is OS independent. All the locations below are written in Unix style for brevity. On Windows, User's home directory is under `%USERPROFILE%` and folder separators are denoted with `\` instead of `/`.

- Change the default location for Omni configuration from `$XDG_CONFIG_HOME/omni/config` to `$HOME/.talos/omni/config`. Also change the default location for user PGP keys from `$XDG_DATA_HOME/omni/keys/*` to `$HOME/.talos/keys/*`.
- Add new `omnictl` flag `--siderov1-key-dir` (also configurable via `SIDEROV1_KEYS_DIR` env var) to allow customizing the directory to use for PGP keys
- Update `omnictl` behavior when it comes to interacting with deprecated default Omni config location explained in detail below.

`omnictl` will attempt to copy existing Omni configuration from old default location to new default location, only if:
- `OMNICONFIG` env var is not set
- `--omniconfig` flag is not provided
- `$HOME/.talos/omni/config` doesn't contain a config file
- `$XDG_CONFIG_HOME/omni/config` contains a config file and file is accessible(readable)
- `$HOME/.talos` is writable
This should offer users a seamless switch between old default location to new default location.

Fixes #1375


client/pkg/omnictl/config/config.go
```diff
@@ -194,11 +200,112 @@ func (c *Config) Merge(additionalConfigPath string) ([]Rename, error) {
 	return renames, nil
 }
 
-func defaultPath() (string, error) {
+type accessType int32
```
utkuozdemir: Just use a Boolean?

client/pkg/omnictl/config/config.go
```diff
@@ -9,16 +9,22 @@ import (
...
+	"github.com/siderolabs/talos/pkg/machinery/constants"
 	"gopkg.in/yaml.v3"
 )
 
 const (
 	// OmniConfigEnvVar is the environment variable to override the default config path.
 	OmniConfigEnvVar = "OMNICONFIG"
 
-	relativePath       = "omni/config"
+	omniConfigFolder   = "omni"
```
utkuozdemir: I recommend using "directory" as folder is the Windows term :) omniConfigDir should be fine here

client/pkg/omnictl/config/config.go
```diff
@@ -194,11 +200,112 @@ func (c *Config) Merge(additionalConfigPath string) ([]Rename, error) {
...
+//nolint:gocognit
+func defaultPath(access accessType) (string, error) {
 	path := os.Getenv(OmniConfigEnvVar)
 	if path != "" {
 		return path, nil
 	}
 
-	return xdg.ConfigFile(relativePath)
+	baseDir, err := config.GetTalosDirectory()
+	if err == nil {
```
utkuozdemir: I would flip this condition - return early if there is an error.

client/pkg/omnictl/config/config.go
```diff
@@ -194,11 +200,112 @@ func (c *Config) Merge(additionalConfigPath string) ([]Rename, error) {
...
+
+	if access == READ && fileutils.FileExists(filepath.Join(xdg.ConfigHome, OmniRelativePath)) {
+		return xdg.ConfigFile(OmniRelativePath)
+	}
+
+	return "", err
+}
+
+func ensurePath(path string) (string, error) {
+	err := os.MkdirAll(path, os.ModeDir|0o700)
```
utkuozdemir: I can suggest doing inline error returns on calls which return a single value which is an error. There are multiple.

client/pkg/omnictl/config/config.go
```diff
@@ -194,11 +200,112 @@ func (c *Config) Merge(additionalConfigPath string) ([]Rename, error) {
...
+
+// CustomSideroV1KeysDirPath returns the custom SideroV1 auth keys directory path if it's provided as command line flag or with environment variable.
+// If not provided, it returns an empty string.
+func CustomSideroV1KeysDirPath(path string) string {
+	if path != "" {
+		return path
+	}
+
+	path = os.Getenv(constants.SideroV1KeysDirEnvVar)
+	if path != "" {
```
utkuozdemir: You can just return the path. No need for the variable or the check.

client/pkg/omnictl/root.go
```diff
@@ -28,11 +30,23 @@ var RootCmd = &cobra.Command{
...
+			filepath.Join("$HOME", constants.TalosDir, config.OmniRelativePath),
+			filepath.Join("$XDG_CONFIG_HOME", config.OmniRelativePath),
 		))
 	RootCmd.PersistentFlags().StringVar(&access.CmdFlags.Context, "context", "",
 		"The context to be used. Defaults to the selected context in the omniconfig file.")
 	RootCmd.PersistentFlags().BoolVar(&access.CmdFlags.InsecureSkipTLSVerify, "insecure-skip-tls-verify", false,
 		"Skip TLS verification for the Omni GRPC and HTTP API endpoints.")
+	RootCmd.PersistentFlags().StringVar(
+		&access.CmdFlags.SideroV1KeysDir,
+		"siderov1-keys-dir",
```
utkuozdemir, line 45: I feel like this flag name could be confusing, users don't know much about the siderov1 thing. Maybe do simply "keys-dir" instead?

client/pkg/client/options.go
```diff
@@ -37,27 +38,42 @@ func WithInsecureSkipTLSVerify(insecureSkipTLSVerify bool) Option {
...
 func WithServiceAccount(serviceAccountBase64 string) Option {
 	return func(options *Options) {
-		options.AuthInterceptor = signatureAuthInterceptor("", "", serviceAccountBase64)
+		options.AuthInterceptor = signatureAuthInterceptor("", "", "", serviceAccountBase64)
 	}
 }
 
 // WithUserAccount is used for accessing Omni by a human.
-func WithUserAccount(contextName, identity string) Option {
+func WithUserAccount(contextName, identity, customKeysDir string) Option {
```
utkuozdemir: This is a breaking change in the client library. We probably should add `WithCustomKeysDir` function instead.

client/pkg/client/options.go
```diff
@@ -37,27 +38,42 @@ func WithInsecureSkipTLSVerify(insecureSkipTLSVerify bool) Option {
...
+		UserKeyProvider:      GetNewKeyProvider(customKeysDir),
 		ContextName:          contextName,
 		Identity:             identity,
 		ClientName:           version.Name + " " + version.Tag,
 		ServiceAccountBase64: serviceAccountBase64,
 	})
 }
 
+// GetNewKeyProvider returns a new KeyProvider for Omni client.
+func GetNewKeyProvider(customKeysDir string) *client.KeyProvider {
```
utkuozdemir: Is this function exported intentionally? Because the moment you do that, it becomes part of the public API.

client/pkg/omnictl/config/config.go
```diff
@@ -194,11 +200,89 @@ func (c *Config) Merge(additionalConfigPath string) ([]Rename, error) {
...
+func defaultPath(readOnly bool) (string, error) {
 	path := os.Getenv(OmniConfigEnvVar)
 	if path != "" {
 		return path, nil
 	}
 
-	return xdg.ConfigFile(relativePath)
+	baseDir, err := config.GetTalosDirectory()
+	if err != nil {
+		// TODO: This is a temporary workaround to continue supporting the old default Omni config directory (XDG) while not allowing further updates. We should remove this in the future.
```
utkuozdemir: I'm not sure if we should remove this ever. If you think we should, the comment should say "we should remove this when ... happens".

client/pkg/omnictl/config/config.go
```diff
@@ -194,11 +200,89 @@ func (c *Config) Merge(additionalConfigPath string) ([]Rename, error) {
...
+		if readOnly && fileutils.FileExists(filepath.Join(xdg.ConfigHome, OmniRelativePath)) {
+			fmt.Fprintf(os.Stderr, "WARN: Default Omni config location: '%s' is not writable, falling back to deprecated Omni config location: '%s' for reading.\n",
```
utkuozdemir: We only failed to read the Talos directory here, we don't know if "Default Omni config location: '%s' is not writable" is actually the case. I think the message probably should simply say "failed to determine Talos directory..."

client/pkg/omnictl/config/config.go
```diff
@@ -194,11 +200,89 @@ func (c *Config) Merge(additionalConfigPath string) ([]Rename, error) {
...
+		}
+
+		return "", err
+	}
+
+	return evaluatePath(readOnly, baseDir)
+}
+
+func evaluatePath(readOnly bool, baseDir string) (string, error) {
+	// TODO: This is a temporary workaround to support migrating away from the old default Omni config directory (XDG). We should remove this in the future.
```
utkuozdemir: Again, we can never know if the migration happened for everyone. I'd say we keep this code forever.

PR #1445, opened 2025-08-18
feat: home page redesign

### LINKED ISSUE #1275: [UI] home page redesign
Make the charts more meaningful.
Add release notes, add recent machines.

Image

Slessi:
Redesign the home page as per [figma](https://www.figma.com/design/4qvWHbfWdFKcDc29wPRn3l/Sidero-UI?node-id=3432-13317&t=GToKCVteAkrynjYv-1)

> [!NOTE]
> Left out the change log in this version as we didn't have a consensus on it yet

Fixes #1275

| Mobile   |      Tablet      |  Desktop |
|:--------:|:-------------:|:-----:|
| image |  image | image |


frontend/src/views/omni/Extensions/ExtensionsPicker.vue
```diff
@@ -64,7 +64,6 @@ const updateExtension = (extension: TalosExtensionsSpecInfo) => {
 
     <Watch
       v-if="talosVersion"
-      class="flex-1 overflow-x-hidden overflow-y-auto"
```
Unix4ever, line 67: I hope this doesn't break anything

frontend/src/views/omni/Home/HomeGeneralInformationCopyable.vue
```diff
@@ -0,0 +1,52 @@
+<!--
```
Unix4ever, line 1: I'd maybe propose to make it a global component instead, as we have the same buttons in other views.

frontend/src/views/omni/Home/HomeGeneralInformationCopyable.vue
```diff
@@ -0,0 +1,52 @@
...
+        :class="{ 'font-mono': secret }"
+        :title="displayValue"
+        @click="secretVisible = !secretVisible"
+      >
+        {{ displayValue }}
+      </span>
+
+      <button
+        aria-label="copy"
+        class="shrink-0 rounded p-0.5 text-primary-p2 hover:bg-naturals-n5 hover:text-primary-p1"
```
Unix4ever, line 45: Missing `cursor-pointer`.

frontend/src/views/omni/Extensions/ExtensionsPicker.vue
Unix4ever: Yes, it actually broke the scrolling there.

frontend/src/views/omni/Home/HomeGeneralInformationCopyable.vue
Slessi: I _can_ add it, but this is actually somewhat of a heated topic that tailwind changed by default in v4: https://github.com/tailwindlabs/tailwindcss/issues/8961

If we disagree with tailwind, we can do what they suggested:
https://tailwindcss.com/docs/upgrade-guide#buttons-use-the-default-cursor

```cs
@layer base {
  button:not(:disabled),
  [role="button"]:not(:disabled) {
    cursor: pointer;
  }
}
```

WDYT? Personally I get both sides of the argument... don't have a preference yet

frontend/src/views/omni/Home/HomeGeneralInformationCopyable.vue
Slessi: This isn't just the button though, its the whole title/value/copy setup, so this:

image

Do we have this particular setup in more places than the homepage?

frontend/src/views/omni/Modals/RoleEdit.vue
```diff
@@ -88,7 +88,6 @@ const close = () => {
           },
           runtime: Runtime.Omni,
         }"
-        class="flex-1"
```
Unix4ever, line 91: This broke the alignment of the `Update Role` button to the right. The dropdown is flex-1 to occupy the whole width - the update role button size.

Unix4ever:
Screenshot 2025-08-27 at 00 51 29
This `1 Node` should be in the same column. Now it's shifted.

frontend/src/views/omni/Extensions/ExtensionsPicker.vue
Slessi: Fixed!

frontend/src/views/omni/Home/HomeGeneralInformationCopyable.vue
Unix4ever: I'd say bring it back. We had it like that before, I think no one likes the changes :D

frontend/src/views/omni/Home/HomeGeneralInformationCopyable.vue
Unix4ever: Yeah, I think so. We will have it for the machines list too. When we get the new design used there

frontend/src/views/omni/Modals/RoleEdit.vue
Slessi: Fixed!

frontend/src/views/omni/Home/HomeGeneralInformationCopyable.vue
Slessi: Brought it back 👍

Unix4ever (approved):

Issue #1413, opened 2025-08-11
[bug] auth.saml.metadata not present in Helm Chart


ErikThorsell:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

If the snippet below is used in the `values.yaml` file, when generating manifests from the Helm Chart:

```yaml
auth:
  auth0:
    enabled: false
  saml:
    enabled: true
    metadata: /etc/saml/idp-metadata.xml
```

The `auth.saml.metadata` option is ignored and a dummy value for `auth.saml.url` is used instead. The resulting `- args:` section in the deployment looks as follows:

```yaml
      - args:
        - [...]
        - --auth-saml-enabled=true
        - --auth-saml-url="https://www.saml.example"
        - [...]
```

### Expected Behavior

The documentation reads as follows:

> * --auth-saml-url is the URL we saved earlier in the configuration of Keycloak.
> 
>    * --auth-saml-metadata may also be used if you would like to pass it as a file instead of a URL and can be used if using self-signed certificates for Keycloak.
> 
> ([Reference](https://omni.siderolabs.com/tutorials/install-airgapped-omni?utm_source=chatgpt.com#install-omni))

Since `auth.saml.metadata` is mentioned as an argument in the documentation for air-gapped deployments, I would expect this argument to also be supported in the Helm Chart. An air-gapped deployment is (as is also mentioned in the documentation) the exact kind of deployment which would benefit from having the SAML Metadata file stored locally at deploy time.

I have checked the deployment template and there is nothing about `auth.saml.metadata` there:

```yaml
        {{- if .Values.auth.saml.enabled }}
          - --auth-saml-enabled=true
        {{- if .Values.auth.saml.url }}
          - --auth-saml-url="{{ .Values.auth.saml.url }}"
        {{- end }}
```

https://github.com/siderolabs/omni/blob/928b7d894868c845d7db2bdcababc4621d01b5ea/deploy/helm/omni/templates/deployment.yaml#L83

### Steps To Reproduce

Use the following snippet as part of the values.yaml file when generating manifests from the Helm Chart:

```yaml
auth:
  auth0:
    enabled: false
  saml:
    enabled: true
    metadata: /etc/saml/idp-metadata.xml
```

Issue #1423, opened 2025-08-12
[bug] Omni Helm chart has wrong indentation of label app.kubernetes.io/component


drexedlu-dr:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

The Helm chart for Omni contains wrong indentation in [deployment.yaml](https://github.com/siderolabs/omni/blob/a268b8fdbcdb1242e1d727f6fcae0dbf87b40869/deploy/helm/omni/templates/deployment.yaml#L29
)

The code looks like this:

```
  template:
    metadata:
    {{- if .Values.deployment.podAnnotations }}
      annotations:
      {{- toYaml .Values.deployment.annotations | nindent 6 }}
    {{- end }}
      labels: 
      {{- include "omni.labels" . | nindent 8 }}
      app.kubernetes.io/component: omni
```

The last line must be indented by 8 and not by 6.

### Expected Behavior

Correct yaml indentation:

```
  template:
    metadata:
    {{- if .Values.deployment.podAnnotations }}
      annotations:
      {{- toYaml .Values.deployment.annotations | nindent 6 }}
    {{- end }}
      labels: 
      {{- include "omni.labels" . | nindent 8 }}
        app.kubernetes.io/component: omni
```

### Steps To Reproduce

Command for generating and validating yaml:

`helm template oci://ghcr.io/siderolabs/charts/omni | kubectl apply -f - --dry-run=server`

Output:

`Error from server (BadRequest): error when creating "STDIN": Deployment in version "v1" cannot be handled as a Deployment: strict decoding error: unknown field "spec.template.metadata.app.kubernetes.io/component"
`

Issue #1414, opened 2025-08-11
[bug] Quotations are incorrectly added to the auth.saml.url in the Helm Chart


ErikThorsell:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

When specifying `auth.saml.url`, quotations are added around the value by the template:

```yaml
- --auth-saml-url="{{ .Values.auth.saml.url }}"
```

If a URL is provided sans protocol, the following error occurs:

```sh
Error: failed to run server: Get "%22saml.example.com/saml/metadata%22": unsupported protocol scheme ""
2025/08/11 10:32:47 failed to run Omni: failed to run server: Get "%22saml.example.com/saml/metadata%22": unsupported protocol scheme ""
```

Initially, this lead me to believe that one must provide the URL with the protocol. However, adding `https://` to the URL gives a slightly more cryptic error:

```sh
Error: failed to run server: parse "\"https://saml.example.com/saml/metadata\"": first path segment in URL cannot contain colon
2025/08/11 10:30:10 failed to run Omni: failed to run server: parse "\"https://saml.example.com/saml/metadata\"": first path segment in URL cannot contain colon
```

Turns out, one shall use `https://`, but the bigger issue is that the quotation marks cause an error. The issue is remedied by manually removing the quotations from the generated manifests, but that's not a feasible solution long term.

### Expected Behavior

Adding `auth.saml.url` to the values.yaml file and generating manifests should result in manifests that yield no syntactical errors.

### Steps To Reproduce

Generate manifests using Helm, with the option: `auth.saml.url` present. Attempt to deploy Omni and the error should be present.

PR #1447, opened 2025-08-19
feat: add custom volume and custom volume mount support for omni helm chart

oguzkilcan:
- Add custom volume and custom volume mount support for omni helm chart
- Fix indentation issue with deployment labels
- Remove extra quotes from saml url argument

Fixes: #1413, #1414 and #1423

### DIFF
--- a/deploy/helm/omni/values.yaml
+++ b/deploy/helm/omni/values.yaml
@@ -41,6 +41,7 @@ resources:
 extraArgs:
   # - --debug
   # - --image-factory-address=factory.talos.dev
+  # - --auth-saml-metadata=/etc/omni/saml/metadata.xml
 volumeMounts:
   tls:
     mountPath: null # "/etc/omni/tls"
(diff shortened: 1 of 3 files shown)


smira (approved):

deploy/helm/omni/values.yaml
```diff
@@ -56,3 +57,18 @@ volumes:
...
+## Add custom volumes
+customVolumes:
+#  - name: saml-metadata
+#    configMap:
+#      name: saml-metadata
+#      items:
+#        - key: metadata.xml
+#          path: metadata.xml
+
+## Add custom volumesMounts
```
utkuozdemir, line 69:
```suggestion
## Add custom volumeMounts
```

utkuozdemir (approved):

deploy/helm/omni/values.yaml
utkuozdemir: let's leave it to the bigger overhaul.

PR #1460, opened 2025-08-21
chore: create a useWatch composable

Slessi:
Create a useWatch composable to allow us to get typed & watched data and without [Watch.vue](https://github.com/siderolabs/omni/blob/7ae7ea749e587123caf4b8b1801ce316465997a1/frontend/src/components/common/Watch/Watch.vue)'s styling.


frontend/src/components/common/Watch/useWatch.ts
```diff
@@ -0,0 +1,36 @@
...
+import type { Resource } from '@/api/grpc'
+import type { WatchJoinOptions, WatchOptions } from '@/api/watch'
+import Watch, { WatchJoin } from '@/api/watch'
+
+export function useWatch<TSpec = unknown, TStatus = unknown>(
+  opts: WatchJoinOptions[] | WatchOptions,
+) {
+  type WatchedResource = Resource<TSpec, TStatus>
+
+  const items = shallowRef<WatchedResource[]>([])
```
Slessi: the types for `api/watch.ts` are a bit all over the place but one quick way to make everyone happy is to use `shallowRef` instead of a `ref`

I think from what I glanced in `watch.ts` that this change is fine, but this is functionally the only real change.

frontend/src/components/common/Watch/Watch.vue
```diff
@@ -4,66 +4,23 @@ Copyright (c) 2025 Sidero Labs, Inc.
...
-      }
-
-      return opts.value as WatchOptions
-    }),
-  )
-}
 
-const err = resourceWatch!.err
-const loading = resourceWatch!.loading
+const { items, err, loading } = useWatch<any, any>(props.opts)
```
Slessi: inside a component, you could here do something like this and get typed `items`

```suggestion
const { items, err, loading } = useWatch<ClusterStatusSpec, any>(props.opts)
```

Image

That said, it is not clear to me what `status` inside a `Resource` is meant to be, so not sure what to type it with, but I kept the second generic around.

Unix4ever: status only applies to Kubernetes resources, not for Omni ones and is normally imported from the kubernetes openapi generated ts module.

Unix4ever (approved):

PR #1463, opened 2025-08-22
chore: improve useWatch typings

Slessi:
Improve useWatch typings to be able to narrow to a single item instead of an array of items when the appropriate options are passed

### DIFF
--- a/frontend/src/components/common/Watch/Watch.vue
+++ b/frontend/src/components/common/Watch/Watch.vue
@@ -4,14 +4,25 @@ Copyright (c) 2025 Sidero Labs, Inc.
 Use of this software is governed by the Business Source License
 included in the LICENSE file.
 -->
-<script setup lang="ts" generic="T = any">
-import type { WatchJoinOptions, WatchOptions } from '@/api/watch'
+<script
+  setup
+  lang="ts"
+  generic="
+    TSpec = any,
+    TStatus = any,
+    TOptions extends WatchJoinOptions[] | WatchOptions = WatchJoinOptions[] | WatchOptions
+  "
+>
+import { computed } from 'vue'
+
+import type { Resource } from '@/api/grpc'
+import type { WatchJoinOptions, WatchOptions, WatchOptionsSingle } from '@/api/watch'
 import TSpinner from '@/components/common/Spinner/TSpinner.vue'
 import { useWatch } from '@/components/common/Watch/useWatch'
 import TAlert from '@/components/TAlert.vue'
 
 type Props = {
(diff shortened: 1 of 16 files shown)


frontend/src/components/common/Watch/Watch.vue
```diff
@@ -20,7 +22,11 @@ type Props = {
 
 const props = defineProps<Props>()
 
-const { items, err, loading } = useWatch<T, any>(props.opts)
+const watch = useWatch<T, any>(props.opts)
+const { err, loading } = watch
+
+const data = 'items' in watch ? watch.items : watch.item
+const hasData = computed(() => (Array.isArray(data.value) ? !!data.value.length : !!data.value))
```
Slessi: This has to be this way because in `Watch.vue` we don't know what options type we received, so we figure it out from the response, but in practice when calling `useWatch` yourself directly, it will JustWork™ like so:

```ts
// Multi
const { items } = useWatch<SomeSpec>({
  resource: {
    namespace: SomeNamespace,
    type: SomeType,
  },
})

// Single
const { item } = useWatch<SomeSpec>({
  resource: {
    namespace: SomeNamespace,
    type: SomeType,
    id: SomeID,
  },
})
```

frontend/src/components/common/Watch/useWatch.ts
Slessi, line 1: behold, the ts-foo

frontend/src/components/common/Watch/Watch.vue
```diff
@@ -32,17 +38,17 @@ const { items, err, loading } = useWatch<T, any>(props.opts)
...
       >
       <slot name="norecords" />
     </template>
     <div
-      v-show="(!loading && !err && (items.length > 0 || !noRecordsAlert)) || displayAlways"
+      v-show="(!loading && !err && (!hasData || !noRecordsAlert)) || displayAlways"
       class="wrapper"
     >
-      <slot :items="items" />
+      <slot :items="Array.isArray(data) ? data : [data]" />
```
Slessi: This is to keep all consumers as receiving `Ref[]` instead of `Ref` - I'm not sure it is possible to overload vue components to make the component "know" what type its receiving based on its props (like `useWatch` is doing) so I left it like this for now

Unix4ever: templating the component is possible, but I'm not sure if you can pass the type there

Unix4ever:
As the bonus I think we could write some unit tests for the new `Watch.vue` stuff.

PR #1475, opened 2025-08-26
feat: support Userpilot reports

Unix4ever:
Omni now has `--user-pilot-app-token` flag. When set it will enable `Userpilot` integration in the UI.

### DIFF
--- a/frontend/src/methods/features.ts
+++ b/frontend/src/methods/features.ts
@@ -3,6 +3,7 @@
 // Use of this software is governed by the Business Source License
 // included in the LICENSE file.
 
+import { Userpilot } from 'userpilot'
 import type { Ref } from 'vue'
 import { computed, ref } from 'vue'
 
@@ -10,7 +11,8 @@ import { Runtime } from '@/api/common/omni.pb'
 import type { Resource } from '@/api/grpc'
 import { ResourceService } from '@/api/grpc'
 import type { FeaturesConfigSpec } from '@/api/omni/specs/omni.pb'
-import { withRuntime } from '@/api/options'
+import type { CurrentUserSpec } from '@/api/omni/specs/virtual.pb'
+import { withAbortController, withRuntime } from '@/api/options'
 import { DefaultNamespace, FeaturesConfigID, FeaturesConfigType } from '@/api/resources'
 import Watch from '@/api/watch'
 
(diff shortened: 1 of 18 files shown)


internal/frontend/handler.go
```diff
@@ -110,8 +110,8 @@ func (handler *StaticHandler) serveFile(w http.ResponseWriter, r *http.Request,
 		} else {
 			w.Header().Set("Referrer-Policy", "strict-origin-when-cross-origin")
 
-			w.Header().Set("Content-Security-Policy", "default-src 'self'; img-src * data: ; "+
-				";connect-src 'self' https://*.auth0.com ;font-src 'self' data: "+
+			w.Header().Set("Content-Security-Policy", "default-src 'self' https://*.userpilot.io; img-src * data: ; "+
+				";connect-src 'self' https://*.auth0.com https://*.userpilot.io wss://*.userpilot.io ;font-src 'self' data: "+
```
Slessi, line 114: arent we including the package directly from package.json? or does it still load JS code remotely?

frontend/src/main.ts
```diff
@@ -22,7 +22,7 @@ import AppUnavailable from '@/AppUnavailable.vue'
 import { AuthType, authType, suspended } from '@/methods'
 import router from '@/router'
 
-import { withRuntime } from './api/options'
+import { withRuntime } from './api/options';
```
Slessi: you have some stuff like this that you can fix with `npm run lint:fix`

frontend/src/methods/features.ts
```diff
@@ -31,6 +33,33 @@ export const setupWorkloadProxyingEnabledFeatureWatch = (): Ref<boolean> => {
   })
 }
 
+let userPilotInitialized = false
+
+export const initializeUserPilot = async () => {
+  if (!currentUser.value) {
+    return
+  }
```
Slessi: wouldn't it be nicer to pass the user as a param to the function and keep it a bit more pure? we have it in the place we call this initialise anyway

or is there a reason we do it this way?

frontend/src/methods/features.ts
```diff
@@ -31,6 +33,33 @@ export const setupWorkloadProxyingEnabledFeatureWatch = (): Ref<boolean> => {
...
+
+  if (!userPilotInitialized) {
+    const featuresConfig = await ResourceService.Get<Resource<FeaturesConfigSpec>>({
```
Slessi: is it possible since this is async that refreshUser is called twice and we initialize pilot twice or so?

internal/frontend/handler.go
Unix4ever: We load it from the package json, but then it does some requests to load more I guess 🤷

frontend/src/methods/features.ts
Unix4ever: makes sense, will update. No particular reason for that

internal/frontend/handler.go
Unix4ever: I've used their docs on the recommended way to install the lib using npm

frontend/src/methods/features.ts
Unix4ever: I will make it abort the call to the resource service if called again

Slessi (approved):

Issue #1495, opened 2025-09-01
[bug] copy kernel parameters button not working behind nginx


soostdijck:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

I am running a on-prem omni that runs in docker, behind a nginx instance. The nginx config is configured [as per the docs here](https://docs.siderolabs.com/omni/infrastructure-and-extensions/self-hosted/expose-omni-with-nginx-https#how-to-expose-omni-with-nginx-https).
The omni container runs with (amongst others) these options (replaced my domain with mydomain):
```
        - '--event-sink-port=8091'
        - '--bind-addr=127.0.0.1:8080'
        - '--machine-api-bind-addr=127.0.0.1:8090'
        - '--k8s-proxy-bind-addr=127.0.0.1:8100'
        - '--advertised-api-url=https://omni.mydomain.net'
        - '--advertised-kubernetes-proxy-url=https://kube.omni.mydomain.net:443/'
        - '--siderolink-api-advertised-url=https://api.omni.mydomain.net:443/'
        - '--siderolink-wireguard-advertised-addr=omni.mydomain.net:50180'
```

The buttons "copy kernel parameters" and "download machine join config" don't seem to be working. The container shows this message:

> `{"level":"info","ts":1756721888.9572775,"caller":"logging/handler.go:59","msg":"HTTP request done","component":"server","handler":"grpc_gateway","request_url":"/api/management.ManagementService/GetMachineJoinConfig","method":"POST","remote_addr":"127.0.0.1","duration":0.000221748,"status":400,"response_length":88}`

### Expected Behavior

I expect the join config and kernel params to work in the gui

### Steps To Reproduce

1. run omni in docker with listed options
2. run nginx in front of omni, terminate TLS on nginx
3. log into omni
4. try to download join config and kernel params

Unix4ever:
Thanks for the report. We will get it fixed shortly

Unix4ever:
Please use `/omni/machine/jointokens` page to use this functionality for now.
The fix will be backported to 1.1.3

PR #1500, opened 2025-09-01
chore: improve tbutton typings

Slessi:
Improve TS typing for TButton by inheriting the base button props to account for [fallthrough attributes](https://vuejs.org/guide/components/attrs). Also removed currently unused attributes.

This can prevent issues like:
- https://github.com/siderolabs/omni/issues/1495

### DIFF
--- a/frontend/src/components/common/Button/TButton.vue
+++ b/frontend/src/components/common/Button/TButton.vue
@@ -5,42 +5,23 @@ Use of this software is governed by the Business Source License
 included in the LICENSE file.
 -->
 <script setup lang="ts">
-import type { StyleValue } from 'vue'
-import { computed, toRefs } from 'vue'
+import type { ButtonHTMLAttributes } from 'vue'
 
 import type { IconType } from '@/components/common/Icon/TIcon.vue'
 import TIcon from '@/components/common/Icon/TIcon.vue'
 
-type Props = {
+interface Props extends /* @vue-ignore */ Omit<ButtonHTMLAttributes, 'type'> {
   type?: 'primary' | 'secondary' | 'subtle' | 'subtle-xs' | 'compact' | 'highlighted'
-  toggle?: boolean
-  disabled?: boolean
   icon?: IconType
-  danger?: boolean
-  iconPosition?: 'left' | 'middle' | 'right'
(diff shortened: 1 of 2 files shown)


frontend/src/components/common/Button/TButton.vue
```diff
@@ -5,42 +5,23 @@ Use of this software is governed by the Business Source License
...
+import type { ButtonHTMLAttributes } from 'vue'
 
 import type { IconType } from '@/components/common/Icon/TIcon.vue'
 import TIcon from '@/components/common/Icon/TIcon.vue'
 
-type Props = {
+interface Props extends Omit<ButtonHTMLAttributes, 'type'> {
   type?: 'primary' | 'secondary' | 'subtle' | 'subtle-xs' | 'compact' | 'highlighted'
-  toggle?: boolean
-  disabled?: boolean
```
Unix4ever, line 17: I didn't expect we don't use disabled anywhere 🤔

Slessi: We do use it, but its a default property of a button and did not need retyping.

It will automatically fallthrough like the other base button props.

frontend/src/components/common/Button/TButton.vue
```diff
@@ -5,42 +5,23 @@ Use of this software is governed by the Business Source License
...
 import type { IconType } from '@/components/common/Icon/TIcon.vue'
 import TIcon from '@/components/common/Icon/TIcon.vue'
 
-type Props = {
+interface Props extends /* @vue-ignore */ Omit<ButtonHTMLAttributes, 'type'> {
   type?: 'primary' | 'secondary' | 'subtle' | 'subtle-xs' | 'compact' | 'highlighted'
-  toggle?: boolean
-  disabled?: boolean
   icon?: IconType
-  danger?: boolean
```
smira, line 19: did `danger` go away?

smira (approved):

frontend/src/components/common/Button/TButton.vue
Slessi: It was not in use for this particular component so i dropped it, KISS and all that

Trivial to return if required

PR #1501, opened 2025-09-02
WIP: feat: add support for updating kernel args

utkuozdemir:
Allow updating kernel args, similar to the set of extensions.


internal/backend/runtime/omni/controllers/omni/machine_status_test.go
```diff
@@ -429,7 +454,7 @@ func (suite *MachineStatusSuite) TestMachineSchematic() {
 			},
 			expected: &specs.MachineStatusSpec_Schematic{
 				Id:               defaultSchematic,
-				InitialSchematic: defaultSchematic,
+				InitialSchematic: "",
```
utkuozdemir, line 457: I think this was a bug - if the machine is in agent mode, it means it is completely reset, we cannot talk about its initial state anymore. So now, we reset the initial state (the schematic id and the kernel args/extensions) to be empty.

The initial state was incorrect anyway, as we explicitly did not return any extensions when the machine was in agent mode.

internal/backend/runtime/omni/controllers/omni/machine_status.go
```diff
@@ -514,13 +515,26 @@ func (ctrl *MachineStatusController) handleNotification(ctx context.Context, r c
...
+			// Otherwise, we missed the moment to capture them.
+			if spec.Schematic.InitialState == nil && !spec.Schematic.Invalid && spec.Schematic.FullId == spec.Schematic.InitialSchematic {
+				spec.Schematic.InitialState = &specs.MachineStatusSpec_Schematic_InitialState{
+					Extensions: event.Schematic.Extensions,
+					KernelArgs: event.Schematic.KernelArgs,
+				}
+			}
+
+			// if the schematic is invalid or the machine is in agent mode, we reset the initial schematic information
+			if spec.Schematic.Invalid || spec.Schematic.InAgentMode {
```
utkuozdemir, line 534: About the `inAgentMode` check, see https://github.com/siderolabs/omni/pull/1501/files#r2315640127.

client/pkg/omni/resources/omni/labels.go
```diff
@@ -219,4 +219,9 @@ const (
 	// from which MachineExtensions resource was generated.
 	// tsgen:ExtensionsConfigurationLabel
 	ExtensionsConfigurationLabel = SystemLabelPrefix + "root-configuration"
+
+	// ExtraKernelArgsConfigurationLabel defines the source ExtraKernelArgs resource
+	// from which MachineExtraKernelArgs resource was generated.
+	// tsgen:ExtraKernelArgsConfigurationLabel
+	ExtraKernelArgsConfigurationLabel = SystemLabelPrefix + "root-configuration"
```
Unix4ever, line 235: The constant seems to be duplicate to the one above

client/pkg/omni/resources/omni/extensions_configuration.go
```diff
@@ -14,7 +14,7 @@ import (
 	"github.com/siderolabs/omni/client/pkg/omni/resources"
 )
 
-// NewExtensionsConfiguration creates new extensions configuration resource.
+// NewExtensionsConfiguration creates new Extensions configuration resource.
```
Unix4ever: Maybe if we change that it should be?
```suggestion
// NewExtensionsConfiguration creates new ExtensionsConfiguration resource.
```

internal/backend/runtime/omni/controllers/omni/machine_config_gen_options.go
```diff
@@ -43,27 +45,61 @@ func NewMachineConfigGenOptionsController() *MachineConfigGenOptionsController {
...
 	)
 }
 
 // GenInstallConfig creates a config patch with an automatically picked install disk.
-func GenInstallConfig(machineStatus *omni.MachineStatus, clusterMachineTalosVersion *omni.ClusterMachineTalosVersion, genOptions *omni.MachineConfigGenOptions) {
-	if clusterMachineTalosVersion != nil {
+//
+// InstallImage will be updated only if the kernel args are in sync, i.e., they are applied via the ClusterMachineConfig, and can be observed on the ClusterMachineConfigStatus.
+//
+// This ensures that any changes in the schematic are held back until the machine config has the correct .machine.install.extraKernelArgs.
```
frezbo, line 113: do we check `extraKernelArgs` are only used when not booted with a UKI?

internal/backend/runtime/omni/state_access.go
```diff
@@ -352,10 +355,6 @@ func verbToRole(verb state.Verb) role.Role {
 //
 //nolint:cyclop,gocyclo
 func filterAccess(ctx context.Context, access state.Access) error {
-	if actor.ContextIsInternalActor(ctx) {
```
frezbo, line 355: intentional?

client/pkg/omni/resources/omni/labels.go
utkuozdemir: It is, but I didn't want to rename the variable above since it's a public package. And to make the intention clear, I made a new variable. Their value happens to be the same - I can make this one point to the other one, but I think it is more explicit this way.

internal/backend/runtime/omni/state_access.go
utkuozdemir: nope, I messed up, reverted.

internal/backend/runtime/omni/controllers/omni/machine_config_gen_options.go
utkuozdemir: We have non-UKI handling code different from UKI handling code, but in both cases we set `.extraKernelArgs` for simplicity/consistency, and because it doesn't do harm.

internal/backend/runtime/omni/controllers/omni/cluster_machine_config.go
```diff
@@ -261,6 +261,7 @@ func reconcileClusterMachineConfig(
 		return err
 	}
 
+	machineConfig.TypedSpec().Value.ExtraKernelArgs = machineConfigGenOptions.TypedSpec().Value.KernelArgs
```
utkuozdemir:
```suggestion
	machineConfig.TypedSpec().Value.ExtraKernelArgs = machineConfigGenOptions.TypedSpec().Value.ExtraKernelArgs
```

oguzkilcan (approved):

PR #1509, opened 2025-09-03
chore: remove the usage of --input-dir flag in tests

Orzelius:
the flag is being removed in talos

### DIFF
--- a/hack/test/integration.sh
+++ b/hack/test/integration.sh
@@ -268,14 +268,16 @@ EOF
 )
 PARTIAL_CONFIG_DIR="${ARTIFACTS}/partial-config"
 mkdir -p "${PARTIAL_CONFIG_DIR}"
-echo "${PARTIAL_CONFIG}" >"${PARTIAL_CONFIG_DIR}/controlplane.yaml"
-echo "${PARTIAL_CONFIG}" >"${PARTIAL_CONFIG_DIR}/worker.yaml"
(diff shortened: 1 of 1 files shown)


hack/test/integration.sh
```diff
@@ -287,10 +287,15 @@ ${ARTIFACTS}/talosctl cluster create \
   --name test-1 \
   --cidr=172.20.0.0/24 \
   --no-masquerade-cidrs=172.21.0.0/24,172.22.0.0/24 \
-  --input-dir="${PARTIAL_CONFIG_DIR}" \
+  --skip-injecting-config \
+  --wait=false \
   --vmlinuz-path="https://factory.talos.dev/image/${SCHEMATIC_ID}/v${TALOS_VERSION}/kernel-amd64" \
   --initrd-path="https://factory.talos.dev/image/${SCHEMATIC_ID}/v${TALOS_VERSION}/initramfs-amd64.xz"
 
+${ARTIFACTS}/talosctl apply-config --insecure --nodes 172.20.0.2 --file "${PARTIAL_CONFIG_DIR}/controlplane.yaml"
```
utkuozdemir: I'd say better to do in a loop, looping for the total number of machines (extract machine counts to variables).

smira: I think it makes little sense to keep it controlplane/worker.yaml as it's same config (see on lines 271-272). so just make it `config.yaml` and apply to all machines?

Issue #1428, opened 2025-08-13
[UI] Move to file based routing


Slessi:
Using [file based routing](https://uvr.esm.is/guide/file-based-routing.html#file-based-routing) (which is how Nuxt/Next do it for their SSR frameworks) with `unplugin-vue-router`, we can get automatically generated and type safe frontend routing.

Docs
- https://router.vuejs.org/guide/advanced/typed-routes.html
- https://uvr.esm.is/introduction.html

PR #1512, opened 2025-09-04
chore: refactor routes to use vue-router better

Slessi:
Refactor routes to remove `withPrefix` and `route.meta.sidedbar` and instead use vue-router's  children/components.

Will also serve as prep work for #1428 later.

### DIFF
--- a/frontend/src/router/index.ts
+++ b/frontend/src/router/index.ts
@@ -5,8 +5,8 @@
 
 import { authGuard } from '@auth0/auth0-vue'
 import { Userpilot } from 'userpilot'
-import type { RouteLocation, RouteLocationRaw, RouteMeta, RouteRecordRaw } from 'vue-router'
-import { createRouter, createWebHistory } from 'vue-router'
+import type { RouteRecordRaw } from 'vue-router'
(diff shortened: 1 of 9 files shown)


frontend/src/router/index.ts
```diff
@@ -86,366 +86,308 @@ import OmniServiceAccounts from '@/views/omni/Users/ServiceAccounts.vue'
 import OmniUsers from '@/views/omni/Users/Users.vue'
 
 export const FrontendAuthFlow = 'frontend'
+const requireCookies = false
```
Slessi, line 89: this was a parameter on `checkAuthorized` which was never passed but I kept the logic incase it is going to be used for something later

Unix4ever:
This part of the sidebar stopped working:

Screenshot 2025-09-04 at 21 45 52


It's when you open the node which is a part of cluster.

Slessi:
@Unix4ever fixed it [here](https://github.com/siderolabs/omni/compare/2e4696d649e8cd8803f40f4c095e7feec4970e29..d8db45a79f870a97643c9d8071424db570108953)

PR #1514, opened 2025-09-04
feature: implement toast messages

Slessi:
Replace the notification banner feature with toast messages. Toasts are stackable and can be swiped/dismissed away.

https://github.com/user-attachments/assets/e800d051-07f4-4994-858d-998b18dfda37


Unix4ever:
I'd maybe only set lower timeout for the success messages. Something like 5 seconds

Unix4ever:
Only need the `dismiss all` button and it will be good to go.

Slessi:
@Unix4ever dismiss all can be implemented by using `toast.dismiss()` without passing a specific toast id. But design-wise where would the button go? The library does not provide a first-party feature for that so we need decide ourselves.

I do have some arguments against this button:

- We currently only ever show one notification per interaction and only for important interactions
- Too many notifications ends up being noise for users so maybe we should limit notifications rather than spamming with a "clear all spam" button
- If you have a mix of errors and successes, you may dismiss-all whilst looking at a success and unknowingly dismiss an error
- If a "dismiss all" button forces us to limit the number of on screen notifications, then we won't have enough notifications to merit a "dismiss all" 😄

---

Instead of dismiss all I would advocate for a no noise notification policy, which would limit notifications enough that we won't have enough to need a dismiss all but also that all information will be important enough to merit reading, low quantity high quality.

**Good notifications** -> unexpected, async, or push based information:
- **Omni update available** (push based)
- **Cluster X created** (async, takes time to prepare all the machines right?)
- **Cluster X is now offline** (unexpected)

**Bad notifications** -> obvious, or low value information:
- **Copied text** (low value, [should be an animation](https://github.com/siderolabs/omni/issues/1520) on the button)
- **Successfully logged in** (obvious, the UI changed)
- **The User X is created** (obvious, immediate sync action and user is visible in the user list)

Unix4ever:
I thought maybe something that appears on the top of the expanded list of the toasts.

https://github.com/user-attachments/assets/1f548a69-3fc5-41bd-b58e-658addc51981

If it's shown only when the whole list is expanded. It's easy to see if there are any important messages.

Slessi:
There isn't nice way that I can see to add a button above the expanded list. The library supports customising individual toasts but not controlling or knowing the expanded state. There isn't a way to inject non-toast components into the toast wrapper/container either.

In their examples for having a close all button they inject a custom component as a toast (instead of just styling classes like I did) and then render their a close all button on all toast messages. I could _maybe_ make it that it only shows on the first toast but it would be a time investment trying to work with their internals and could break during version upgrades.

WDYT? Close all on all toasts or leave things as is?

Unix4ever:
If that's so much of a trouble I'd say leave it as is for now.

PR #1517, opened 2025-09-04
refactor: drop extra input finalizers

smira:
Simplify controllers, add migrations to clear finalizers which are no longer needed.

See https://github.com/cosi-project/runtime/pull/633


internal/backend/runtime/omni/migration/migrations.go
```diff
@@ -1942,3 +1942,79 @@ func moveClusterTaintFromResourceToLabel(ctx context.Context, st state.State, _
 
 	return err
 }
+
+func dropExtraInputFinalizers(ctx context.Context, st state.State, logger *zap.Logger, _ migrationContext) error {
```
Unix4ever, line 1947: I think we need some test for the migration.

smira: yep, I can do that

smira: done!

PR #1541, opened 2025-09-09
chore: move /cluster routes to /omni/clusters

### LINKED ISSUE #1524: Normalise /cluster and /omni/cluster
We have both `/omni` and `/cluster` as root level routes, but it feels weird to have it this way. Clusters are a part of omni too.

That said `/omni` is also possibly redundant if you had e.g. `myomni.com/omni`. Maybe better would be a `/admin` or other such route, to indicate that this is the authenticated tree of the website, and everything authenticated goes under there. Non-authenticated routes can be on the `/` e.g. `/forbidden` or `/login`.

Slessi:
Move `/cluster` routes to `/omni/clusters`

Fixes #1524 

Open discussion: https://talossystems.slack.com/archives/C05348FENHG/p1757419832762469


frontend/src/router/index.ts
```diff
@@ -94,23 +94,27 @@ const routes: RouteRecordRaw[] = [
...
-        },
-      },
-    ],
+    path: '/authenticate',
+    name: 'Authenticate',
+    component: Authenticate,
+    beforeEnter: async (to) => {
+      return authType.value === AuthType.Auth0 ? await authGuard(to) : true
+    },
   },
```
Slessi, line 103: this `/omni/authenticate` route (now `/authenticate`) isn't a problem to change right? just to confirm

frontend/src/router/index.ts
```diff
@@ -94,23 +94,27 @@ const routes: RouteRecordRaw[] = [
...
 
+  // Redirects for legacy routes
+  { path: '/omni', redirect: '/' },
+  { path: '/omni/:catchAll(.*)', redirect: (to) => `/${to.params.catchAll}` },
+  { path: '/cluster', redirect: '/clusters' },
+  { path: '/cluster/:catchAll(.*)', redirect: (to) => `/clusters/${to.params.catchAll}` },
```
Slessi, line 109: just incase, so old routes will continue working for now

frontend/src/router/index.ts
smira: I guess we might have it in Auth0 settings?

Slessi: Nah the auth0 flow is all external and will just redirect back to the root page (`whatever.com/`)

This is just the page with the login button on our side that you use to regenerate some public key on the omni web side, not totally sure how it all works yet, but TL;DR not auth0 related

smira: ah... then it might be (?) the one we call from the `omnictl` for authentication... so it should be somewhere in the source code (we try to open it in a browser)

Slessi: mmm found some `/internal/backend` stuff (changed now) but couldn't see anything else

Unix4ever: SAML and OIDC have redirects to `/omni/authenticate`

Slessi: @Unix4ever Is that in omni code somewhere, or in the config of a remote provider? I assume thats the `/internal/backend` stuff I changed.

Regardless it will still work right now because my "legacy redirects" section makes it so `/omni/authenticate` will send you to `/authenticate`. Mainly I just don't want to rely on the redirects as part of the core app logic, but have them only as a fallback for a period of time (e.g. 1 year)

Unix4ever: I think changing `internal/backend` is enough 👍

Issue #1521, opened 2025-09-05
More lint/style rules for UI


Slessi:
- [x] `"htmlWhitespaceSensitivity": "ignore"` for prettier, for consistent HTML templates. We don't need collapsible whitespace behaviour.
- [x] ~https://github.com/vue-a11y/eslint-plugin-vuejs-accessibility~
  - Too many a11y issues for now
- [x] https://eslint.vuejs.org/rules/define-props-destructuring.html
- [x] ~see if `@vue/eslint-config-typescript` offers more beyond their `recommended` list~
  - They do, but it would just be re-implementing `tsc` which we do already 
- [x] Have another look at the uncategorised section of https://eslint.vuejs.org/rules/
- [x] ~Something to enforce type/interface decision in TS? Bit annoying to have `type Props` and `interface Props` mismatch in places~
    - Not important

PR #1559, opened 2025-09-10
chore: ignore html whitespace

Slessi:
Add a setting to prettier to ignore HTML whitespace differences. Its not a concern with our setup, so we prefer consistency.

Related https://github.com/siderolabs/omni/issues/1521

### DIFF
--- a/frontend/.prettierrc.json
+++ b/frontend/.prettierrc.json
@@ -4,6 +4,7 @@
   "singleQuote": true,
   "printWidth": 100,
   "plugins": ["prettier-plugin-tailwindcss"],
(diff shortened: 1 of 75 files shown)


frontend/.prettierrc.json
```diff
@@ -4,6 +4,7 @@
   "singleQuote": true,
   "printWidth": 100,
   "plugins": ["prettier-plugin-tailwindcss"],
+  "htmlWhitespaceSensitivity": "ignore",
```
Slessi, line 7: This is the change, and the rest is auto fixed by prettier

smira (approved):

frezbo (approved):

Issue #1446, opened 2025-08-18
[feature] Guided user onboarding


rothgar:
### Problem Description

We need a way to create dynamic, guided user onboarding flows in Omni. This will help them get familiar with Omni and know how to use it.

### Solution

Install [user pilot](https://userpilot.com/) javascript libraries in Omni so it can fetch guides at startup for users. This should be disabled by default from the Omni binary so we can enable it optionally in the SaaS for new users. 

It will not be used to create onboarding guides for on-prem users and should not be a blocker for the application if there is no internet access to the userpilot CDN.

### Alternative Solutions

It's possible we could build something ourselves, but we also want to be able to collect data on what features users are using in the SaaS and to help with new feature discovery.

Slessi:
Some open questions for me

1. Will need an account with them for this, what's the process for getting/creating external accounts for 3rd party services?
2. Does this package itself need to be omitted from non-saas binaries entirely? Or would we be fine with including it for all binaries and only doing runtime checks to determine if we should start the onboarding?
3. Is there an existing way to identify if we are in a saas enviromment (and is it available to the frontend)?

@Unix4ever mentioned there is currently something in the backend that could be used for detecting saas, but isn't currently exposed to the frontend and might need improvement (?)

Unix4ever:
I would assume that we can actually have the resource that the UI can request which will tell it to use the userpilot key.
Then if there's no such resource, userpilot will be disabled.

I guess the steps will be:
- install the library into the project.
- add https://docs.userpilot.com/developer/installation/SDK-APIs#identify-userid%2C-%5Bproperties%5D call to properly track user activity. The current user info is available in the frontend.
- add the `--user-pilot-api-key` Omni cmdline flag. If the flag is set, Omni will then create the resource with the key which will be then requested by the frontend when the page loads.
- change https://github.com/siderolabs/omni-controller to additionally pass `--user-pilot-api-key` to the Omni service.

Unix4ever:
We also need the consent screen for that.
If the user doesn't agree, the UserPilot library should be not enabled.

Unix4ever:
Mockups for the consent screen are done, only need to review the wording there. @rothgar you can review the mockups here https://www.figma.com/design/4qvWHbfWdFKcDc29wPRn3l/Sidero-UI?node-id=4250-37109&t=wpLhKGwNuthxp1qQ-1

PR #1565, opened 2025-09-11
feat: implement user consent form for the `UserPilot`

Unix4ever:
The form allows to agree or decline data collection. When declined UserPilot library is not activated.

Tracker state is saved in the local storage.

Fixes: https://github.com/siderolabs/omni/issues/1446

### DIFF
--- a/frontend/src/App.vue
+++ b/frontend/src/App.vue
@@ -10,6 +10,7 @@ import { useRouter } from 'vue-router'
 
 import AppToast from '@/components/common/AppToast/AppToast.vue'
 import TSuspended from '@/components/common/Suspended/TSuspended.vue'
+import UserConsent from '@/components/common/UserConsent/UserConsent.vue'
 import TSideBar from '@/components/SideBar/TSideBar.vue'
 import THeader from '@/components/THeader/THeader.vue'
 import TModal from '@/components/TModal.vue'
@@ -23,6 +24,7 @@ router.afterEach(() => (isSidebarOpen.value = false))
 
 <template>
   <main class="flex h-screen flex-col">
+    <UserConsent />
     <AppToast />
 
     <THeader
--- a/frontend/src/components/common/UserConsent/UserConsent.vue
+++ b/frontend/src/components/common/UserConsent/UserConsent.vue
@@ -1,0 +1,66 @@
+<!--
+Copyright (c) 2025 Sidero Labs, Inc.
+
+Use of this software is governed by the Business Source License
+included in the LICENSE file.
+-->
+<script setup lang="ts">
+import { ref, watch } from 'vue'
+
+import TButton from '@/components/common/Button/TButton.vue'
+import { currentUser } from '@/methods/auth'
+import { getUserPilotToken, initializeUserPilot } from '@/methods/features'
+import { trackingState } from '@/methods/features'
+
+const enableTracking = async () => {
+  if (!currentUser.value) {
+    return
+  }
+
+  trackingState.value = true
+
+  await initializeUserPilot(currentUser.value)
+}
+
+const trackerToken = ref<string | undefined>()
+
+const fetchTrackingToken = async () => {
+  if (!currentUser.value) {
+    return
+  }
+
+  trackerToken.value = await getUserPilotToken()
+}
+
+watch(() => currentUser.value, fetchTrackingToken)
+</script>
+
+<template>
+  <div v-if="trackingState === undefined && trackerToken">
+    <div class="absolute top-0 left-0 z-50 flex h-screen w-screen flex-col justify-stretch">
+      <div class="flex-1 bg-black opacity-25" />
+      <div class="border-t border-primary-p3 bg-naturals-n6 p-4">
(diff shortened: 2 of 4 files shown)


Unix4ever:
Screenshot 2025-09-11 at 23 50 45

frontend/src/components/common/UserConsent/UserConsent.vue
```diff
@@ -0,0 +1,48 @@
...
+  <div>
+    <div class="absolute top-0 left-0 z-50 flex h-screen w-screen flex-col justify-stretch">
+      <div class="flex-1 bg-black opacity-25" />
+      <div class="border-t border-primary-p3 bg-naturals-n6 p-4">
+        <p class="text-lg text-naturals-n13">Cookies for a Better Experience</p>
+        <div class="flex items-center justify-between">
+          <p class="text-sm">
+            We use cookies to improve our website's interface and onboarding experience. We don't
+            use them for ads or share your data with third parties.
+            <TButton type="subtle" class="inline" @click="openPrivacyPolicy">Learn more</TButton>
```
Slessi: Its preferrable to use an `<a>` tag for links with `rel="noopener noreferrer"` if its an external one. `list-item-link` might give u the styles u are looking for? and it will be `inline` by default also

Its better for a11y and writing tests, and better for UX so you can do things like mouse middle click to open in a new tab

frontend/src/App.vue
```diff
@@ -5,24 +5,40 @@ Use of this software is governed by the Business Source License
...
+
+const fetchTrackingToken = async () => {
+  if (!currentUser.value) {
+    return
+  }
+
+  trackerToken.value = await getUserPilotToken()
+}
+
+watch(() => currentUser.value, fetchTrackingToken)
```
Slessi: Why not put this inside of `UserConsent`?

frontend/src/methods/features.ts
```diff
@@ -36,24 +36,49 @@ export const setupWorkloadProxyingEnabledFeatureWatch = (): Ref<boolean> => {
 let userPilotInitialized = false
 let userPilotInitializeAbortController: AbortController | null = null
 
-export const initializeUserPilot = async (user: Resource<CurrentUserSpec>) => {
-  if (!userPilotInitialized) {
-    userPilotInitializeAbortController?.abort()
+export enum Tracking {
+  Enabled = 'enabled',
+  Disabled = 'disabled',
+}
```
Slessi: could just be a boolean instead of an enum

Unix4ever: It should be 3 states: not set (empty string), or enabled/disabled.
I guess I can achieve that with undefined though maybe

frontend/src/App.vue
Unix4ever: I'm fine with that. Is putting `v-if` on the top level element of the Vue template a good practice?

frontend/src/components/common/UserConsent/UserConsent.vue
```diff
@@ -0,0 +1,69 @@
...
+    return
+  }
+
+  trackingState.value = Tracking.Enabled
+
+  await initializeUserPilot(currentUser.value)
+}
+
+const openPrivacyPolicy = () => {
+  window.open('http://siderolabs.com/omni-cookie-policy', '_blank')?.focus()
```
Unix4ever: @rothgar please let me know if that's a correct URL

frontend/src/App.vue
Slessi: Generally speaking the least surprising DX is that a component always renders itself, unless the parent consuming it decides not to.

In this particular case, between the fragmented logic and the potential "why isnt it rendering" surprise, keeping the consent/pilot logic contained to a single file is the lesser evil IMO as its quite specific

As far as vue goes it won't really matter

frontend/src/components/common/UserConsent/UserConsent.vue
```diff
@@ -0,0 +1,69 @@
...
+        <p class="text-lg text-naturals-n13">Cookies for a Better Experience</p>
+        <div class="flex items-center justify-between">
+          <p class="text-sm">
+            We use cookies to improve our website's interface and onboarding experience. We don't
+            use them for ads or share your data with third parties.
+            <a
+              type="subtle"
+              class="list-item-link cursor-pointer"
+              rel="noopener noreferrer"
+              @click="openPrivacyPolicy"
```
Slessi: Nice that we have `a` now but is it still a `@click` handler instead of a `href` for any specific reason? 👀 

```suggestion
             href="http://siderolabs.com/omni-cookie-policy"
             target="_blank"
```

frontend/src/App.vue
Unix4ever: ✅

frontend/src/components/common/UserConsent/UserConsent.vue
rothgar: Please link it to https://www.siderolabs.com/privacy-policy/ There's a lot more information there and I'll add a specific paragraph about userpilot user tracking.

PR #1569, opened 2025-09-12
feat: abort ongoing cluster import process

### LINKED ISSUE #1313: take the cluster out of Omni management
In the context of the cluster import, it allows to "abort" ("revert") the import.

## Input

`Cluster`

## Result

* `Cluster` and associated resources (related to the cluster) are removed from Omni
* `Machine`/`Link` resources for the machines which are part of the cluster are removed from Omni (wihout wiping)

Omni doesn't attempt to revert any machine configuration changes, e.g. Kubernetes controlplane endpoint.

Omni assumes a user has PKI to access the cluster (in the future we might implement "take out" PKI).

oguzkilcan:
Added a new omnictl command for aborting cluster import process and removing the created resources (e.g `Cluster`, `MachineSet`s, `MachineSetNode`s without reseting the machines.

Closes #1313


client/pkg/omnictl/cluster/import.go
```diff
@@ -0,0 +1,88 @@
...
+	"github.com/siderolabs/omni/client/pkg/client"
+	"github.com/siderolabs/omni/client/pkg/omni/resources"
+	"github.com/siderolabs/omni/client/pkg/omni/resources/omni"
+	"github.com/siderolabs/omni/client/pkg/omnictl/internal/access"
+)
+
+var abortImportCmd = &cobra.Command{
+	Use:   "abort <cluster name>",
+	Short: "Abort an ongoing cluster import operation",
+	Long: `Abort an ongoing cluster import operation. This will clean up any resources created during the import process and
```
smira, line 25: my concern with this is that (I think) we discussed this should be available for any cluster as an option

so it's more like `cluster export` if you wish for any cluster (?)

if we need to taint it before doing so, it's fine as well

oguzkilcan: I was thinking of implementing an `omnictl cluster export` command that basically dumps the secret bundle, machine configs and/also talosconfig, and taints the cluster accordingly. Then this operation can be finalized with `omnictl cluster export finalize` command basically doing the same thing that aborting an import operation does.

smira: yes, I think that would be great, so it's useful outside of import process as well

client/pkg/omnictl/cluster/import.go
```diff
@@ -0,0 +1,88 @@
...
+
+	if _, ok := clusterStatus.Metadata().Labels().Get(omni.LabelClusterTaintedByImporting); !ok {
+		return fmt.Errorf("cluster %q is not tainted as importing", id)
+	}
+
+	if _, ok := cluster.Metadata().Annotations().Get(omni.ClusterLocked); !ok {
+		return fmt.Errorf("cluster %q is not locked", id)
+	}
+
+	fmt.Fprintf(os.Stderr, "Aborting import operation for cluster %q\n", id)
```
Orzelius: Should this and the following log really go to Stderr?

internal/backend/runtime/omni/migration/manager.go
```diff
@@ -236,6 +236,10 @@ func NewManager(state state.State, logger *zap.Logger) *Manager {
 				callback: populateNodeUniqueTokens,
 				name:     "populateNodeUniqueTokens",
 			},
+			{
+				callback: moveClusterTaintFromResourceToLabel,
+				name:     "moveClusterTaintFromResourceToLabel",
```
Orzelius, line 241: I'm wondering why it's better to have them as labels. Was there a discussion somewhere about this change?

client/pkg/omnictl/cluster/import.go
oguzkilcan: Changed it to Stdout

internal/backend/runtime/omni/migration/manager.go
oguzkilcan: I discussed this with Artem. Using the resource was to prevent users from removing the taint after it was set. Using labels on a Omni owned resource gives us the same capability and we don't have to maintain an extra resource.

internal/backend/runtime/omni/migration/migrations.go
```diff
@@ -1918,3 +1918,28 @@ func populateNodeUniqueTokens(ctx context.Context, st state.State, _ *zap.Logger
...
+	}
+
+	for taint := range clusterTaints.All() {
+		if err = safe.StateModify(ctx, st, omni.NewClusterStatus(resources.DefaultNamespace, taint.Metadata().ID()),
+			func(res *omni.ClusterStatus) error {
+				res.Metadata().Labels().Set(omni.LabelClusterTaintedByBreakGlass, "")
+
+				return nil
+			},
+		); err != nil {
```
Unix4ever: You need to add `WithPhaseAny` here to the modify call, otherwise clusters stuck in destroying will make this migration fail.

internal/backend/runtime/omni/migration/migrations.go
```diff
@@ -1918,3 +1918,28 @@ func populateNodeUniqueTokens(ctx context.Context, st state.State, _ *zap.Logger
...
+			return err
+		}
+
+		if err = st.TeardownAndDestroy(ctx, taint.Metadata()); err != nil {
```
Unix4ever, line 1938: I don't think the taint has any finalizers, but just for the future reference: as the controllers don't run during the migration waiting for them to remove finalizers is pointless as it will never complete

PR #1578, opened 2025-09-16
chore: rekres and fix e2e test runs

oguzkilcan:
Rekres to bring back retrieve PR labels step default job and use relative location of GH workspace instead of `/tmp` for local storage.


hack/test/integration.sh
```diff
@@ -431,6 +431,7 @@ if [ "${INTEGRATION_RUN_E2E_TEST:-true}" == "true" ]; then
   cd internal/e2e-tests/
   docker buildx build --load . -t e2etest
   docker run --rm \
+    -v ${TEST_OUTPUTS_DIR}/e2e/playwright-report:/tmp/test/playwright-report \
```
Slessi: this folder does not exist on this branch

.github/workflows/ci.yaml
```diff
@@ -769,14 +728,6 @@ jobs:
       - name: Set secrets for job
         run: |
           sops -d .secrets.yaml | yq -e '.secrets | to_entries[] | .key + "=" + .value' >> "$GITHUB_ENV"
-      - name: Download artifacts
-        uses: actions/download-artifact@v4
```
frezbo, line 773: we don't need this step anymore?

frezbo (approved):

.github/workflows/ci.yaml
oguzkilcan: We weren't using this step on cron e2e tests, so I presume it's also not needed here.

PR #1584, opened 2025-09-17
chore: strip comments from generated ClusterMachineConfig

### LINKED ISSUE #1581: [chore] do not include the comments to the generated ClusterMachineConfig
The comments might change from version to version.

To avoid updating the existing configs only the new calls to generate cluster machine config should use no comments.

oguzkilcan:
Do not include the comments to the generated ClusterMachineConfig to prevent triggering noop machine updates caused by changes on machine config comments

Fixes #1581

### DIFF
--- a/internal/backend/runtime/omni/controllers/omni/cluster_machine_config.go
+++ b/internal/backend/runtime/omni/controllers/omni/cluster_machine_config.go
@@ -6,6 +6,7 @@
 package omni
 
 import (
+	"bytes"
 	"context"
 	"errors"
 	"fmt"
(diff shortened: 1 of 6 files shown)


internal/backend/runtime/omni/controllers/omni/cluster_machine_config.go
```diff
@@ -257,8 +260,22 @@ func reconcileClusterMachineConfig(
 		return nil //nolint:nilerr
 	}
 
-	if err = machineConfig.TypedSpec().Value.SetUncompressedData(data); err != nil {
-		return err
+	skipUpdate := false
+
+	// skip comparing existing config to generated config if existing config has it's comments stripped to avoid unnecessary decompression/unmarshalling
```
utkuozdemir, line 265:
```suggestion
	// skip comparing existing config to generated config if existing config has its comments stripped to avoid unnecessary decompression/unmarshaling
```

Issue #1588, opened 2025-09-17
startup check for 'strict mode'


smira:
Ensure there are no machines on Talos < 1.6 before enabling strict mode.

If the condition is not met, crash Omni.

PR #1601, opened 2025-09-19
feat: check on start up if the account has Talos < 1.6 and strict tokens

Unix4ever:
If both conditions are true, make Omni crash with the error explaining that using Talos 1.6 and `strict` join tokens mode is not possible.

Fixes: https://github.com/siderolabs/omni/issues/1588


client/api/omni/specs/omni.proto
```diff
@@ -1222,6 +1222,8 @@ message MachineStatusMetricsSpec {
   uint32 connected_machines_count = 2;
   uint32 allocated_machines_count = 3;
   uint32 pending_machines_count = 4;
+
+  map<string, int32> versions_map = 5;
```
Unix4ever, line 1229: I added this as an optimization to avoid listing all machine statuses, but then realized that this resource is ephemeral and it won't exist on the Omni start up.

But I decided to keep it as it will be used in the versions breakdown screen anyways.

PR #1602, opened 2025-09-19
fix: update home page wording as SideroLink

rothgar:
Make it more clear with docs references the endpoint is for the SideroLink connection.

I might have missed other references to WireGuard but didn't see any obvious ones we'd want to replace.

### DIFF
--- a/frontend/src/views/omni/Home/HomeGeneralInformation.vue
+++ b/frontend/src/views/omni/Home/HomeGeneralInformation.vue
@@ -101,7 +101,7 @@ watch(apiConfigErr, (err) => err && showError(err))
       />
 
       <HomeGeneralInformationCopyable
-        title="WireGuard Endpoint"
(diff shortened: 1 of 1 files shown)


Slessi (approved):
LGTM for home page just gotta rebase ✅ 

There is also a mention on clusters page that you can change [here](https://github.com/siderolabs/omni/blob/dbe39ea1fcf811d3d66df48cc8a2bdfe2ce5dea4/frontend/src/views/cluster/Overview/components/OverviewRightPanel/OverviewRightPanelCondition.vue#L30) if it makes sense.


Link and piccu:
https://demo.omni.stage.managed.siderolabs.io/clusters/talos-default
Image

utkuozdemir (approved):

rothgar:
I left my yubikey at home so I won't be able to rebase this PR until next week.

PR #1606, opened 2025-09-21
feat: add more cluster and machine metrics

utkuozdemir:
Add metrics for enabled cluster features and for various machine properties:

```text
# HELP omni_cluster_features Number of clusters with specific features enabled.
# TYPE omni_cluster_features gauge
omni_cluster_features{feature="disk_encryption"} 1
omni_cluster_features{feature="embedded_discovery_service"} 0
omni_cluster_features{feature="workload_proxy"} 1

# HELP omni_machine_platforms Number of machines in the instance by platform.
# TYPE omni_machine_platforms gauge
omni_machine_platforms{platform="akamai"} 0
omni_machine_platforms{platform="aws"} 0
omni_machine_platforms{platform="azure"} 0
omni_machine_platforms{platform="cloudstack"} 0
omni_machine_platforms{platform="digital-ocean"} 0
omni_machine_platforms{platform="equinixMetal"} 0
omni_machine_platforms{platform="exoscale"} 0
omni_machine_platforms{platform="gcp"} 0
omni_machine_platforms{platform="hcloud"} 0
omni_machine_platforms{platform="metal"} 10
omni_machine_platforms{platform="nocloud"} 0
omni_machine_platforms{platform="opennebula"} 0
omni_machine_platforms{platform="openstack"} 0
omni_machine_platforms{platform="oracle"} 0
omni_machine_platforms{platform="scaleway"} 0
omni_machine_platforms{platform="upcloud"} 0
omni_machine_platforms{platform="vmware"} 0
omni_machine_platforms{platform="vultr"} 0

# HELP omni_machine_secure_boot_status Number of machines in the instance by secure boot status.
# TYPE omni_machine_secure_boot_status gauge
omni_machine_secure_boot_status{enabled="false"} 10
omni_machine_secure_boot_status{enabled="true"} 0
omni_machine_secure_boot_status{enabled="unknown"} 0

# HELP omni_machine_uki_status Number of machines in the instance by UKI (Unified Kernel Image) status.
# TYPE omni_machine_uki_status gauge
omni_machine_uki_status{booted_with_uki="false"} 0
omni_machine_uki_status{booted_with_uki="true"} 10
omni_machine_uki_status{booted_with_uki="unknown"} 0
```

Closes #22.

PR #1613, opened 2025-09-23
fix: fix the order in the grpc interceptor chain

utkuozdemir:
Change the interceptor order to move the Prometheus metrics collector earlier, so that we can get metrics for the calls that fail early. Related to siderolabs/omni#1606.

Additionally, ensure that `get` access to the `AuthConfig` resource does not require a GRPC signature.

### DIFF
--- a/internal/pkg/auth/interceptor/auth_config.go
+++ b/internal/pkg/auth/interceptor/auth_config.go
@@ -14,6 +14,8 @@ import (
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/metadata"
 
+	resapi "github.com/siderolabs/omni/client/api/omni/resources"
+	authres "github.com/siderolabs/omni/client/pkg/omni/resources/auth"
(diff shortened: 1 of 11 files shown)


internal/pkg/auth/interceptor/auth_config.go
```diff
@@ -36,7 +38,15 @@ func NewAuthConfig(enabled bool, logger *zap.Logger) *AuthConfig {
 // Unary returns a new unary GRPC interceptor.
 func (c *AuthConfig) Unary() grpc.UnaryServerInterceptor {
 	return func(ctx context.Context, req any, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (any, error) {
-		ctx = c.intercept(ctx, info.FullMethod)
+		isGetAuthConfigRequest := false
+
+		if req != nil && info.FullMethod == resapi.ResourceService_Get_FullMethodName {
```
frezbo: also check if `info != nil`?

PR #1625, opened 2025-09-23
feat(frontend): improve chart colors for more intuitive status visual…

niklasfrick:
Add custom color support to RadialBar component:

   - Update Clusters chart with intuitive colors (green=healthy, red=unhealthy)
   - Update Machines chart with intuitive colors (green=connected, red=disconnected)
   - Useing more intuitive colors for status: green for positive states, red for problems, yellow for warnings, blue for operations

### DIFF
--- a/frontend/src/components/common/Charts/RadialBar.vue
+++ b/frontend/src/components/common/Charts/RadialBar.vue
@@ -18,6 +18,7 @@ interface Props {
   items: {
     label: string
     value: number
+    color?: string
   }[]
   legendFormatter?: (value: number) => string
 }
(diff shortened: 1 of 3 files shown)


Slessi:
I understand the intent here but need to have an internal discussion about this I think, it covers a topic I have raised about using colours only to show intent, or maybe in this case branding, which I believe is the reason that the orange is so prominent. I do see how orange feels like a "danger" colour though, despite being used for the healthy bar.

Adding screenshots for reference, before and after
before

after

niklasfrick:
Thanks for the feedback and for adding the screenshots — noted on the internal discussion, no rush! 😉

Slessi:
@niklasfrick We've discussed this with the team and designers and we have some broader plans we are going to tackle regarding colour use in omni. For now will close this as it doesn't feel quite right in the current version of the page, but we hear you! New colours coming soon to an omni near you.

PR #1630, opened 2025-09-24
test: fix the data race in TestEtcdElectionsLost

utkuozdemir:
Wait for the goroutine to complete before the test exits, so its `StopElections` does not produce a log after the test has completed, which causes a data race.

### DIFF
--- a/internal/backend/runtime/omni/state_etcd_election_test.go
+++ b/internal/backend/runtime/omni/state_etcd_election_test.go
@@ -7,6 +7,7 @@ package omni_test
 
 import (
 	"context"
+	"sync"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


internal/backend/runtime/omni/state_etcd_election_test.go
```diff
@@ -15,6 +15,7 @@ import (
 	"github.com/stretchr/testify/require"
 	"go.uber.org/zap/zaptest"
 	"golang.org/x/sync/errgroup"
+	"gvisor.dev/gvisor/pkg/sync"
```
smira: are you sure about this import?

oguzkilcan: GoLand is at it again

utkuozdemir: yes this is much better one :P

oguzkilcan (approved):

Issue #1534, opened 2025-09-08
Break TList down into smaller components


Slessi:
`TList` is currently holding too many responsibilities and it is difficult to make changes without breaking other places that use it. We should break it down into smaller composable pieces.

PR #1646, opened 2025-09-26
feat: introduce storybook for omni frontend

### LINKED ISSUE #1526: Implement storybook into omni
With https://storybook.js.org/ we can create pages where individual components are rendered in isolation. These components can also be uploaded to a public URL as a free service provided by storybook.

This allows us to get design feedback on components without requiring users to checkout and build a branch. It also can help developers and designers explore our component library and see what low level components we have to create pages. This could be buttons, dropdowns, icon library, theme colours, etc.

It can also assist with local development, giving us a way to develop components before backend data is available, or in an iterative process. You can develop and preview components directly in storybook before integrating them into omni, and merge smaller PRs with incremental developments, and only integrate the component at the final stage of development.

There is also a _paid_ service they offer for [visual regression testing](https://www.chromatic.com/storybook) where they take snapshots of all of your components and compare them to the latest snapshots from `main` and report all visual changes. These can be approved or rejected on their UI and it will count as a failed CI job. Something we could look into later, but it isn't cheap.

Slessi:
Introduce storybook for omni frontend

Fixes #1526


Unix4ever:
spotted a bug after using the storybook ;)

Screenshot 2025-09-26 at 21 37 40

We use overflow hidden and it cuts the dropdown

Slessi:
look at that, instant value!

~I'll fix it separately cause its a `TList` issue so maybe as a part of #1534 - or maybe by adjusting how the dropdown component works~

Made an issue to fix it https://github.com/siderolabs/omni/issues/1649

frontend/.storybook/public/mockServiceWorker.js
Slessi, line 1: This is a generated file

Unix4ever (approved):

PR #1648, opened 2025-09-26
test: use bridge IP for WireGuard in CI

Unix4ever:
It was using local pod IP which was generating new schematic every time the test runs.


hack/test/integration.sh
```diff
@@ -68,6 +68,10 @@ chown -R ${SUDO_USER:-$(whoami)} ${ARTIFACTS}
 # Determine the local IP SideroLink API will listen on
 LOCAL_IP=$(ip -o route get to 8.8.8.8 | sed -n 's/.*src \([0-9.]\+\).*/\1/p')
 
+if [[ "${CI:-false}" == "true" ]]; then
+  LOCAL_IP=172.20.0.0
```
smira:
```suggestion
  LOCAL_IP=172.20.0.1
```

utkuozdemir:
good find!

PR #1651, opened 2025-09-29
chore: bump min Talos version

utkuozdemir:
Bump min supported Talos version to be `1.6.0`. Enforce it on cluster creation and cluster Talos version updates.


client/pkg/constants/constants.go
```diff
@@ -83,3 +83,6 @@ const CompressionThresholdBytes = 2048
 
 // InfraProviderHealthCheckInterval is the default interval of the infra provider health checks.
 const InfraProviderHealthCheckInterval = 30 * time.Second
+
+// MinTalosVersion is the minimum Talos version supported by Omni.
```
smira: I think we should clarify that it is the minimum for _new_ clusters

utkuozdemir: Ok, but to be clear, if you modify the version to a not supported version, e.g., try to go from 1.5.0 to 1.5.1, the validation would also block that. It is probably fine, right?

smira: no, that was meant to be for "new clusters only". so if you run 1.5 already, you can upgrade to 1.5.x any version

PR #1680, opened 2025-10-03
fix(frontend): fix locked icon not showing when cluster is locked

Slessi:
Fix locked icon not showing when cluster is locked

### DIFF
--- a/frontend/src/components/common/Icon/TIcon.vue
+++ b/frontend/src/components/common/Icon/TIcon.vue
@@ -15,8 +15,6 @@ import {
   CpuChipIcon,
   DocumentIcon,
   DocumentTextIcon,
-  LockClosedIcon,
-  LockOpenIcon,
   PauseCircleIcon,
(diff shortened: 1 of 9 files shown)


frontend/src/components/common/Icon/TIcon.vue
```diff
@@ -89,7 +87,7 @@ const icons = {
   'link-down': defineAsyncComponent(() => import('../../icons/IconLinkDown.vue')),
   loading: defineAsyncComponent(() => import('../../icons/IconLoading.vue')),
   'locked-toggle': defineAsyncComponent(() => import('../../icons/IconLockClosedToggle.vue')),
-  locked: LockClosedIcon,
+  locked: defineAsyncComponent(() => import('../../icons/IconLocked.vue')),
```
Slessi, line 90: there is a bug in heroicons that breaks aria-label, so i just inlined them. Thats more or less what they recommend anyway.

PR #1689, opened 2025-10-06
feat: return unsupported k8s versions

### LINKED ISSUE #1670: Show all kubernetes versions when updating
Currently we only show kubernetes versions that are upgradeable in the UI. We should show all kubernetes versions, including ones that cannot be upgraded to directly (as you cannot skip minor versions) but mark them disabled with a tooltip or message to indicate that you cannot upgrade to them _yet_ due to not being able to skip minor versions.

E.g. 1.32 is missing from this list

Image

Slessi:
Return unsupported k8s versions to be displayed in the frontend

Fixes #1670

### DIFF
--- a/frontend/msw/helpers.ts
+++ b/frontend/msw/helpers.ts
@@ -56,7 +56,7 @@ export function createWatchStreamHandler<T = unknown, S = unknown>({
       const events = resources.map((r, i) => createCreatedEvent(r, i + 1))
 
       if (!skipBootstrap) {
-        events.push(createBootstrapEvent(totalResults ?? initialResources.length))
+        events.push(createBootstrapEvent(totalResults ?? resources.length))
       }
 
       events.forEach((event) => controller.enqueue(encodeResponse(event)))
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -24,6 +24,7 @@
         "monaco-editor": "^0.52.2",
         "monaco-yaml": "^5.3.1",
         "pluralize": "^8.0.0",
+        "semver": "^7.7.2",
         "userpilot": "^1.4.1",
         "uuid": "^13.0.0",
         "vue": "^3.5.13",
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -41,6 +41,7 @@
     "monaco-editor": "^0.52.2",
     "monaco-yaml": "^5.3.1",
     "pluralize": "^8.0.0",
+    "semver": "^7.7.2",
     "userpilot": "^1.4.1",
     "uuid": "^13.0.0",
     "vue": "^3.5.13",
--- a/frontend/src/views/omni/Modals/MaintenanceUpdate.stories.ts
+++ b/frontend/src/views/omni/Modals/MaintenanceUpdate.stories.ts
@@ -5,6 +5,7 @@
 import { faker } from '@faker-js/faker'
 import { createWatchStreamHandler } from '@msw/helpers'
 import type { Meta, StoryObj } from '@storybook/vue3-vite'
+import * as semver from 'semver'
 
 import type { Resource } from '@/api/grpc'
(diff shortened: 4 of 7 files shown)


rothgar:
Because there are so many versions is there a way to minimize/hide versions on the list. I would guess the majority of people who open this list are trying to upgrade to the latest version that's available.

internal/backend/runtime/omni/controllers/omni/internal/kubernetes/versions.go
```diff
@@ -21,8 +21,14 @@ import (
 	"github.com/siderolabs/omni/client/pkg/omni/resources/omni"
 )
 
+type UpgradeVersions struct {
+	UpgradeVersions            []string
+	UpgradeVersionsWithoutPath []string
```
utkuozdemir: nit, to avoid repetition.

```suggestion
	Versions            []string
	VersionsWithoutPath []string
```

internal/backend/runtime/omni/controllers/omni/internal/kubernetes/versions.go
```diff
@@ -46,34 +54,46 @@ func CalculateUpgradeVersions(ctx context.Context, r controller.Reader, currentV
...
+	// skip versions which do not have an upgrade path from the current version
 	for version := range availableVersions {
-		k8sVersion, err := compatibility.ParseKubernetesVersion(version)
+		path, err := upgrade.NewPath(currentVersion, version)
 		if err != nil {
 			return nil, err
 		}
 
-		if k8sVersion.SupportedWith(talosVersion) != nil {
+		if !path.IsSupported() {
```
smira, line 47: I feel this might be a really long list, while all we need to show to the user is that there is a next version after supported ones, which is not available for one of two reasons. I don't think anybody needs a list of a hundred unavailable versions?

rothgar: I agree, we shouldn't show everything with notes about why. We should only show the available options for their version of Talos. It would be even better if there were an "Upgrade to latest" button that would automatically upgrade to whatever latest version they're allowed to upgrade to would be.

Slessi: There are ways to deal with long lists if necessary, but we do not need to show every single version. 

My inspiration for this was from this [:thread: thread](https://talossystems.slack.com/archives/C05348FENHG/p1759320174166889), WDYT? Only showing available options is what we currently do, but it caused that minor confusion.

Current version in this PR scrolls the current version into view on load, keeps major versions as sticky headers, and adds an extra note if necessary. You can play with it on [storybook](https://return-unsupported-k8s-versions--68d664245076e2bb623d98f3.chromatic.com/?path=/story/views-omni-modals-updatekubernetes--data)

https://github.com/user-attachments/assets/0b518b70-2891-4545-bff0-f08495d8b977

smira: My point is to avoid bloating resources with a list of hundreds unsupported versions (in theory), while we only need to show the _next_ one grayed out. We don't need other versions.

E.g. if I'm on 1.33.7, and my Talos only supports up to 1.33.x, then 1.34.0 is grayed out "unsupported by Talos".

If I'm on 1.32, Talos supports up to 1.34.x, I can upgrade to 1.33, but 1.34.0 is grayed out as "upgrade to 1.33.x first".

We could also show previous one grayed out saying that we don't support downgrades.

Slessi: I'm ok with that also. I saw there are min/max versions in the code that it uses to generate these versions, are those returned somewhere already? I could use them to create the version you are suggesting. If not returned yet I can try return them.

smira: Yes, they are in the code as well, but it's basically the first version it fails for you if you order them properly. You don't need the version itself.

PR #1692, opened 2025-10-07
fix: update CSP for userpilot script

rothgar:
Documentation for the fix pulled from https://docs.userpilot.com/developer/security/csp

### DIFF
--- a/internal/frontend/handler.go
+++ b/internal/frontend/handler.go
@@ -110,7 +110,7 @@ func (handler *StaticHandler) serveFile(w http.ResponseWriter, r *http.Request, 
 		} else {
 			w.Header().Set("Referrer-Policy", "strict-origin-when-cross-origin")
 
-			w.Header().Set("Content-Security-Policy", "default-src 'self' https://*.userpilot.io; img-src * data: ; "+
+			w.Header().Set("Content-Security-Policy", "script-src 'self' https://*.userpilot.io 'unsafe-inline'; img-src * data: ; "+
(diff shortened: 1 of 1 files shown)


internal/frontend/handler.go
```diff
@@ -110,7 +110,7 @@ func (handler *StaticHandler) serveFile(w http.ResponseWriter, r *http.Request,
 		} else {
 			w.Header().Set("Referrer-Policy", "strict-origin-when-cross-origin")
 
-			w.Header().Set("Content-Security-Policy", "default-src 'self' https://*.userpilot.io; img-src * data: ; "+
+			w.Header().Set("Content-Security-Policy", "script-src 'self' https://*.userpilot.io 'unsafe-inline'; img-src * data: ; "+
```
Slessi, line 113: the script-src/default-src change shouldnt change anything (default is the faullback if no script, and we have no script)

what problem are you seeing @rothgar? cause `unsafe-inline` is bit _meh_ - unlikely to be an issue for us specifically but if avoidable preferable to avoid

Could also instead use the `nonce` version instead https://docs.userpilot.com/developer/security/csp#using-the-nonce-attribute-in-csp-rules-for-userpilot

rothgar: The problem is the userpilot script isn't loading in the test environment. I'm not exactly sure why, but I haven't been able to create any onboarding flows and test them without the script loaded.

This change is what their support suggested from their docs. If there's a better way I'm happy to consider it.

PR #1694, opened 2025-10-08
feat: show cpus if they have cores or frequency

### LINKED ISSUE #1688: [bug] stop filtering out processors with unset CPU frequency

### Current Behavior

See https://github.com/siderolabs/omni/discussions/1633#discussioncomment-14503670 for the context.

### Expected Behavior

Oxide SMBIOS data doesn't contain frequency, but it has core count (vCPUs). Omni should show CPUs/cores, but skip frequency data.

### Steps To Reproduce

See above.

I verified that SMBIOS data doesn't contain frequency, so it's not a bug in parsing SMBIOS data.

_No response_

### Anything else?

_No response_

Slessi:
Return processors in the MachineStatusSpec if they have cores or frequency, rather than cores and frequency

Fixes #1688

### DIFF
--- a/frontend/src/views/omni/Machines/MachineDetailsPanel.vue
+++ b/frontend/src/views/omni/Machines/MachineDetailsPanel.vue
@@ -26,59 +26,45 @@ defineEmits<{
   close: []
 }>()
 
-const machineName = computed(() => {
-  return machine?.spec.message_status?.network?.hostname ?? machine?.metadata.id
-})
+const machineName = computed(
+  () => machine?.spec.message_status?.network?.hostname ?? machine?.metadata.id,
+)
 
-const processors = computed(() => {
-  const processors = machine?.spec.message_status?.hardware?.processors || []
+const processors = computed(() =>
+  machine?.spec.message_status?.hardware?.processors?.map(
+    ({ frequency, core_count, description }) => {
+      const parts: string[] = []
 
-  const format = (proc: { frequency?: number; core_count?: number; description?: string }) => {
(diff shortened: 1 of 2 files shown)


internal/backend/runtime/omni/controllers/omni/internal/task/machine/poll.go
```diff
@@ -223,7 +223,7 @@ func pollProcessors(ctx context.Context, c *client.Client, info *Info) error {
 		hardware.NamespaceName,
 		hardware.ProcessorType,
 		func(r *hardware.Processor) error {
-			if r.TypedSpec().CoreCount == 0 || r.TypedSpec().MaxSpeed == 0 {
+			if r.TypedSpec().CoreCount == 0 && r.TypedSpec().MaxSpeed == 0 {
 				return nil
 			}
```
Slessi, line 228: so before you needed both to get a CPU return, now you need only one

frontend/src/views/omni/Machines/MachineDetailsPanel.vue
```diff
@@ -26,59 +26,45 @@ defineEmits<{
...
+      const parts: string[] = []
+
+      if (frequency) parts.push(`${frequency / 1000} GHz`)
+      if (core_count) parts.push(pluralize('core', core_count, true))
+      if (description?.trim()) parts.push(description.trim())
+
+      return parts.join(', ')
+    },
+  ),
+)
```
Slessi, line 45: frontend updated to not show rubbish like `0 GHz, 2 cores`

Issue #1315, opened 2025-07-08
`omnictl cluster import` CLI


smira:
## Why CLI?

CLI needs direct access to Talos API before SideroLink is established, so it might be beneficial to have CLI which can be put to a location which has direct access to Talos API, and Omni API is available as well.

## Prerequisites

Admin-level access to Talos API: `talosconfig`.

Access to Omni API (e.g. a service account/interactive auth).

A list of machines to be imported (or guess using Talos discovery).

## Outline

1. Run pre-checks via Talos API.
2. Backup Talos machine configuration.
3. Inject SideroLink config into the machine configuration.
4. Reconstruct machine config patches.
5. Create Omni resources: ImportedClusterSecrets, Cluster, MachineSet, MachineSetNode, ConfigPatch. The cluster is locked.

### Pre-checks

* All machines are healthy.
* All machines run same version of Talos.
* All machines run same version of Kubernetes.
* Talos should either: run a vanilla Talos image, run an Image Factory image (the same that is linked to the Image Factory)
* If Talos runs an image with a schematic, it should be present in the Image Factory linked to this Omni instance.
* Talos version is >= X.Y.Z.

Allow to skip pre-checks.

### Backup Machine Configuration

Backup Talos machine configuration to files (allows to revert to a previous state if aborted).

###  Inject SideroLink Config

Patch Talos machine configuration with Omni join config, wait for `Machine` resources to appear in Omni in invalid state.

### Reconstruct Machine Config Patches

1. Guess or ask user about the initial version contract.
2. Extract root PKI from the controlplane machine config.
3. Extract Kubernetes version from the machine config.
4. Extract cluster name and controlplane endpoint.
5. Build "default/base" machine configs with secrets, version contract, and Kubernetes version, cluster name, controlplane endpoint: outputs base controlplane and worker machine configs.
6. Build config patches:
   a. User submits config patches (if they have them).
   b. Figure out a per-machine config patch by comparing base generated and existing machine configs (both "additive" and "delete" patches).
 7. Validate machine config patches: apply base + patch, compare with existing (ignore some diffs, e.g. `machine.install`).

### Create Omni Resources

Using the PKI extracted in the previous step, create `ImportedClusterSecrets`.

Using the information gathered in the previous step (name, version contract, Kubernetes version, Talos version) create a `Cluster` resource in a *locked* state.

Create two `MachineSet`s: controlplane and workers.

Create `ConfigPatch`es for all future `ClusterMachine`s.

[...]

PR #1695, opened 2025-10-08
feat: implement cluster import

oguzkilcan:
Allow importing existing talos cluster to Omni using `omnictl cluster import`

Closes: #1315


client/pkg/clusterimport/clusterimport.go
```diff
@@ -0,0 +1,857 @@
...
+	importContext.workerMachineSet = &workerMachineSet
+	importContext.machineSetNodesMap = machineSetNodesMap
+	importContext.configPatchesMap = configPatchesMap
+
+	input.logf("importing cluster %q with %d nodes (%d control planes, %d workers)", input.Cluster, len(input.Nodes), len(controlPlanes), len(workers))
+
+	return importContext, nil
+}
+
+var ErrValidation = errors.New("validation error")
```
utkuozdemir, line 405: interesting that linter didn't complain about this not having docs.

client/pkg/clusterimport/clusterimport.go
```diff
@@ -0,0 +1,857 @@
...
+			machineSetNodesMap[machineSetNode.Metadata().ID()] = &machineSetNode
+		}
+
+		genOutput, nodeErr := machineconfig.Generate(machineconfig.GenerateInput{
+			ClusterID:                input.Cluster,
+			MachineID:                info.machineID,
+			InitialTalosVersion:      input.InitialTalosVersion,
+			InitialKubernetesVersion: input.InitialKubernetesVersion,
+			IsControlPlane:           info.isControlPlane,
+			SiderolinkEndpoint:       "", // No need to set the real value, it's value will not be overwritten by the diff.
```
utkuozdemir: maybe simply remove the line?

client/pkg/clusterimport/clusterimport.go
```diff
@@ -0,0 +1,857 @@
...
+	if c.input.DryRun {
+		c.input.logf(" > skipped in dry-run")
+
+		return nil
+	}
+
+	if err := c.omniState.Modify(
+		ctx,
+		c.importedClusterSecrets,
+		func(res resource.Resource) error { return nil },
```
utkuozdemir: nit: this can be more compact. applies to many other places as well.

client/pkg/omnictl/cluster/import.go
```diff
@@ -6,19 +6,33 @@ package cluster
...
+	"github.com/siderolabs/omni/client/pkg/infra/imagefactory"
 	"github.com/siderolabs/omni/client/pkg/omni/resources"
 	"github.com/siderolabs/omni/client/pkg/omni/resources/omni"
+	"github.com/siderolabs/omni/client/pkg/omni/resources/siderolink"
 	"github.com/siderolabs/omni/client/pkg/omnictl/internal/access"
 )
 
+const (
+	flagImportForce = "force"
+)
```
utkuozdemir, line 26: can be single line

client/pkg/omnictl/cluster/import.go
```diff
@@ -68,21 +90,180 @@ func abortImport(ctx context.Context, client *client.Client, id resource.ID) err
...
+
+	c, err := imagefactory.NewClient(imagefactory.ClientOptions{FactoryEndpoint: featuresConfig.TypedSpec().Value.ImageFactoryBaseUrl})
+	if err != nil {
+		return nil, fmt.Errorf("failed to set up image factory client: %w", err)
+	}
+
+	return c, nil
+}
+
+type realTalosClient struct {
```
utkuozdemir:
```suggestion
type talosClient struct {
```

I guess it was an artifact from the previous experiment, here we can simply call it `talosClient`.

client/pkg/clusterimport/client.go
```diff
@@ -0,0 +1,299 @@
...
+
+		if msg.GetMetadata().GetError() != "" {
+			return fmt.Errorf("healthcheck error: %s", msg.GetMetadata().GetError())
+		}
+
+		fmt.Fprintf(logWriter, " > %s\n", msg.GetMessage()) //nolint:errcheck
+	}
+}
+
+func (c *talosClientWrapper) getSchematic(extensionStatuses []*runtime.ExtensionStatus) (string, *schematic.Schematic, error) {
```
Unix4ever, line 193: I wonder if we could extract this code to a common place for Omni's machine status controller and here.

client/pkg/clusterimport/client.go
```diff
@@ -0,0 +1,299 @@
...
+
+	schematicInfo, err := schematic.Unmarshal([]byte(schematicSpec.TypedSpec().Metadata.ExtraInfo))
+	if err != nil {
+		return "", nil, fmt.Errorf("failed to unmarshal schematic extra info: %w", err)
+	}
+
+	return schematicSpec.TypedSpec().Metadata.Version, schematicInfo, nil
+}
+
+func extractVersion(imageRef string) (string, error) {
```
Unix4ever: Same for this code. I think we had some place using the same logic 🤔

internal/backend/runtime/omni/state_validation.go
```diff
@@ -163,10 +163,19 @@ func clusterValidationOptions(st state.State, etcdBackupConfig config.EtcdBackup
...
+			}
 
+			_, wasLocked := existingRes.Metadata().Annotations().Get(omni.ClusterLocked)
 			_, stillLocked := newRes.Metadata().Annotations().Get(omni.ClusterLocked)
-			if wasLocked && stillLocked {
+			_, wasImporting := existingRes.Metadata().Annotations().Get(omni.ClusterImportIsInProgress)
+
+			_, stillImporting := newRes.Metadata().Annotations().Get(omni.ClusterImportIsInProgress)
+			if wasLocked && stillLocked && !wasImporting && !stillImporting {
+				fmt.Printf("phase %s", newRes.Metadata().Phase())
```
Unix4ever: debug code?

client/pkg/clusterimport/client.go
oguzkilcan: It seems we are using `image.GetTag`. I'll switch to using that.

internal/backend/runtime/omni/controllers/omni/cluster_machine_config_status.go
```diff
@@ -680,14 +686,18 @@ func (h *clusterMachineConfigStatusControllerHandler) reset(
...
+
+	if locked && importing && cluster.Metadata().Phase() == resource.PhaseTearingDown {
 		// Aborting Cluster import, means that we should just let it go
 		logger.Info("removed without reset")
 
 		return nil
 	}
 
-	if _, ok := clusterStatus.Metadata().Labels().Get(omni.LabelClusterTaintedByExporting); ok {
+	if locked && exporting && cluster.Metadata().Phase() == resource.PhaseTearingDown {
```
Unix4ever, line 700: Maybe lets define the condition as a human readable flag and reuse it in 2 places.

internal/backend/runtime/omni/state_validation.go
oguzkilcan: Yes, good catch.

internal/backend/runtime/omni/controllers/omni/cluster_machine_config_status.go
oguzkilcan: Can you elaborate?

cmd/omni/cmd/cmd.go
```diff
@@ -703,8 +703,8 @@ func defineFeatureFlags() {
 	rootCmd.Flags().BoolVar(
 		&cmdConfig.Features.EnableClusterImport,
 		"enable-cluster-import",
-		cmdConfig.Features.EnableClusterImport,
-		"enable cluster import feature.",
+		true,
```
Unix4ever: We should set it to true in the `config.go` when we create the default config instead.
And keep using `cmdConfig.Features.EnableClusterImport` here.

client/pkg/omnictl/cluster/import.go
```diff
@@ -32,57 +42,147 @@ will only work if the cluster is locked and tainted as "importing"`,
...
+		return access.WithClient(func(ctx context.Context, client *client.Client) error {
+			return importCluster(ctx, client, args[0])
+		})
+	},
 }
 
 func init() {
 	clusterCmd.AddCommand(importCmd)
 	importCmd.AddCommand(abortImportCmd)
+	importCmd.Flags().StringSliceVarP(&input.Nodes, flagNodes, "n", []string{}, "all nodes to import")
```
Unix4ever: Maybe we could explain that it should be endpoints:
```suggestion
	importCmd.Flags().StringSliceVarP(&input.Nodes, flagNodes, "n", []string{}, "endpoints of all nodes to import")
```

oguzkilcan: That's even better.

client/pkg/clusterimport/client.go
oguzkilcan: Not sure about this. I checked machine status controller we do lots of heavy lifting there to make sure schematic is suitable for omni. Here we are just converting fake extension created by image factory to schematic to be able to ensure it with image factory.

Unix4ever: Ok, maybe it's fine if it doesn't have intersections.

internal/backend/runtime/omni/controllers/omni/cluster_machine_config_status.go
Unix4ever: ah, sorry it's two different conditions :) NVM then

client/pkg/clusterimport/clusterimport.go
```diff
@@ -0,0 +1,864 @@
...
+		return nil, fmt.Errorf("failed to get members from node %q: %w", firstNode, err)
+	}
+
+	if len(members) == 0 {
+		input.logf("failed to discover any members from the cluster. skipping member validation...\n > assuming discovery service is disabled and all the nodes have been provided as input")
+	} else if len(members) != nodeCount {
+		return nil, fmt.Errorf("number of members in cluster (%d) does not match number of provided nodes (%d)", len(members), nodeCount)
+	}
+
+	defaultJoinToken, err := safe.ReaderGetByID[*siderolinkres.DefaultJoinToken](ctx, omniState, siderolinkres.NewDefaultJoinToken().Metadata().ID())
```
Unix4ever, line 87: Potentially we might want to make the join token customizable. Maybe not for this PR.

client/pkg/clusterimport/clusterimport.go
```diff
@@ -0,0 +1,864 @@
...
+	clusterID := info.TypedSpec().ClusterName
+
+	input.logf("importing cluster %q to Omni", clusterID)
+
+	existingCluster, err := safe.ReaderGetByID[*omni.Cluster](ctx, omniState, clusterID)
+	if err != nil {
+		if !state.IsNotFoundError(err) {
+			return nil, err
+		}
+	}
```
Unix4ever:
```suggestion
	if err != nil && !state.IsNotFoundError(err) {
		return nil, err
	}
```

client/pkg/clusterimport/clusterimport.go
oguzkilcan: I agree but let's keep this version simple. We can improve the experience after it gets some usage.

client/pkg/clusterimport/clusterimport.go
```diff
@@ -0,0 +1,864 @@
...
+	secretsBundle := secrets.NewBundleFromConfig(secrets.NewFixedClock(time.Now()), nodeInfoMap[controlPlanes[0]].machineConfig.Provider())
+
+	bundleYaml, err := yaml.Marshal(secretsBundle)
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal secrets bundle: %w", err)
+	}
+
+	ics.TypedSpec().Value.Data = string(bundleYaml)
+
+	cluster := *omni.NewCluster(resources.DefaultNamespace, clusterID)
```
Unix4ever: What is the reason to de-reference pointer here and below?

oguzkilcan: No reason, will fix it.

client/pkg/clusterimport/clusterimport.go
```diff
@@ -0,0 +1,864 @@
...
+
+	if err := c.omniState.Modify(ctx, c.controlPlaneMachineSet, func(res resource.Resource) error { return nil }); err != nil {
+		return fmt.Errorf("failed to create control plane machine set: %w", err)
+	}
+
+	return nil
+}
+
+func (c *Context) createCluster(ctx context.Context) error {
+	c.input.logf("creating cluster")
```
Unix4ever: maybe it would be nice to add the cluster name here

client/pkg/clusterimport/clusterimport.go
```diff
@@ -0,0 +1,864 @@
...
+		}
+	}
+
+	return nil
+}
+
+func (c *Context) createMachineSetNodes(ctx context.Context) error {
+	c.input.logf("creating %d machine set nodes", len(c.machineSetNodesMap))
+
+	if c.input.DryRun {
```
Unix4ever, line 783: Maybe it would be cool to log the nodes list it's going to connect in dry run mode

client/pkg/clusterimport/clusterimport.go
```diff
@@ -0,0 +1,864 @@
...
+		}
+	}
+
+	return os.OpenFile(file, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0o644)
+}
+
+func (c *Context) createConfigPatches(ctx context.Context) error {
+	c.input.logf("creating %d config patches", len(c.machineSetNodesMap))
+
+	if c.input.DryRun {
```
Unix4ever, line 761: Could be nice to log the config patch names here

client/pkg/clusterimport/clusterimport.go
```diff
@@ -0,0 +1,864 @@
...
+func (c *Context) createMachineSetControlPlanes(ctx context.Context) error {
+	c.input.logf("creating machine set for control planes")
+
+	if c.input.DryRun {
+		c.input.logf(" > skipped in dry-run")
+
+		return nil
+	}
+
+	if err := c.omniState.Modify(ctx, c.controlPlaneMachineSet, func(res resource.Resource) error { return nil }); err != nil {
```
Unix4ever: if we extract `func(res resource.Resource) error { return nil });` to a common method and use it everywhere in this file it might improve readability.

Something like that:
```suggestion
	if err := c.omniState.Modify(ctx, c.controlPlaneMachineSet, noop); err != nil {
```

Unix4ever:
well done 💯

client/pkg/clusterimport/clusterimport.go
Unix4ever: yep, just an idea for the future maybe

PR #1696, opened 2025-10-08
chore(frontend): update dependencies

Slessi:
Update frontend dependencies


.github/workflows/chromatic.yml
oguzkilcan, line 1: I think this file should be created/managed by kres

Slessi: I don't know that it will ever be relevant to other projects but if it is we can cross that bridge when we get to it

oguzkilcan: It would benefit from `rekres` for all the upgrades etc. In the end this is a GH action and we manage those via kres. It's contents might not be used anywhere else that's ok.

PR #1699, opened 2025-10-09
feat(frontend): clarify information inside update kubernetes modal

Slessi:
Show only minor versions and omit patches for versions which arent upgradeable and show why they arent upgradeable

image


rothgar:
nit: capitalize Talos

PR #1705, opened 2025-10-13
chore: bump default Talos version, deps, rekres, re-generate

utkuozdemir:
Bump everything to appropriate versions. Remove some unused imports.


frontend/src/schemas/config.schema.json
utkuozdemir, line 1: @Slessi probably this shouldn't have been formatted. I ran `make lint-fmt` and it didn't re-format this, so I assume it was done manually once, and hopefully will stay this way in the future.

utkuozdemir: Oh, it needs manual editing. I'll have a look into this.

Slessi: @utkuozdemir its already [ignored](https://github.com/siderolabs/omni/blob/a9986eab1ed19838d082ecfef49141c37831b5a5/frontend/.prettierignore#L3), wont ever be ([and wasnt ever](https://github.com/siderolabs/omni/commits/main/frontend/src/schemas/config.schema.json)) formatted

utkuozdemir: nvm it got messed up because of our new website.

PR #1706, opened 2025-10-13
refactor(tbutton): separate type and size styles in tbutton

Slessi:
Separate the type and size styles in `TButton`.

This gets rid of `compact` and `subtle-xs` as you can replicate them with `type="primary" size="sm"` or `type="subtle" size="xxs"` (had to be `xxs` because `subtle` itself was shrinking itself already, so that is now `type="subtle" size="xs"`)

[📚 Storybook](https://button-seperate-size-type--68d664245076e2bb623d98f3.chromatic.com/?path=/story/components-common-button-tbutton--all-buttons)

image

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -83,6 +83,7 @@
         "vite-plugin-node-polyfills": "^0.24.0",
         "vite-plugin-vue-devtools": "^8.0.2",
         "vitest": "^3.2.4",
+        "vue-component-type-helpers": "^3.1.1",
         "vue-tsc": "^3.1.1",
         "yaml": "^2.8.1"
       },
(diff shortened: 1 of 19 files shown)


Unix4ever:
I think button looks a bit weird if you select xs/xss options. I think it should have padding.

Screenshot 2025-10-13 at 22 43 32

Slessi:
@Unix4ever I do agree but that would be a style change, the PR currently doesn't change any buttons in use anywhere, except for `compact` which was so close to `primary` that I think it made more sense to just be a smaller version of `primary`.

`xs` and `xxs` are only used by `subtle` buttons which are text-only, so for those it looks ok. I think they should either be links, or changed to be one of the other button types (to look like buttons) depending on their use cases, but maybe not in this PR?

Unix4ever:
> but maybe not in this PR?

Yeah. Maybe we need to discuss that a bit.

Ok, lets keep it as is for now.

PR #1709, opened 2025-10-13
refactor(routes): normalise /machine and /machines into /machines

Slessi:
Normalise machine routes to always be under /machines and not /machine


frontend/src/router/index.ts
```diff
@@ -201,62 +203,102 @@ export const routes: RouteRecordRaw[] = [
...
+                    component: () => import('@/views/cluster/Config/PatchEdit.vue'),
+                  },
+                ],
+              },
+            ],
+          },
+        ],
       },
       {
         path: 'machine-classes',
```
Slessi, line 281: i do also wonder if `/machine-classes` makes sense as opposed to something like `/machines/classes` or something, for hierarchy

Issue #1276, opened 2025-06-18
[feature] use the wizard when downloading the installation media from Omni


Unix4ever:
Same as in the image factory.

Related to #506

AlexB-Shorepine:
Redesign is made and stored in this [Figma section](https://www.figma.com/design/4qvWHbfWdFKcDc29wPRn3l/Sidero-UI?node-id=3737-37434&t=GToKCVteAkrynjYv-1)
As discussed with Artem it is turned from 1 pager with all the options to a staged wizard-like presentation with several steps, where content of each step would be what it can/should be based previous choices in previous steps, in attempt to make it more clear.
This redesign also contains visuals for saving previously made media flow (issue-506)

PR #1710, opened 2025-10-13
feat(installation-media): add route for installation media page

Slessi:
Add route for installation media page

Related to #1276


smira:
I would appreciate PR descriptions to be more descriptive:

* what I guess is it adds a stub for the future page which is hidden by default
* it fixes some other small issues. 

But it's not obvious from the PR description which basically repeats what the change itself is, while giving zero clue about why.

PR #1715, opened 2025-10-14
feat(frontend): create a generic table component

Slessi:
Create a generic table component to share table styling amongst pages. Currently we only have one instance of this, other pages are using `grid` layouts instead of tables. Will update them case-by-case.

> `table` is preferred over `grid` for tabular data for a11y and easier alignment

### DIFF
--- a/frontend/src/components/common/Table/Table.stories.ts
+++ b/frontend/src/components/common/Table/Table.stories.ts
@@ -1,0 +1,59 @@
+// Copyright (c) 2025 Sidero Labs, Inc.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+import { faker } from '@faker-js/faker'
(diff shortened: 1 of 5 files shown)


Unix4ever:
We mainly use grid to make the elements always occupy the same size, even if the column is empty.
I think table aligns column by the max size of the element there.

I don't have any objections against using tables, but only if the size of columns won't bounce every time something changes in size.

Slessi:
@Unix4ever By default tables will definitely auto-resize, is it the bouncing when elements change size that is a bother?

its possible to force table columns widths to be fixed, usually the auto-resizing is a feature though, wouldnt you want to fit as much content as possible? if you have 4 columns of 25% width, you might end up with for example 1 truncated column and 3 blank columns, where the 1st column couldve had enough space to show all its content if allowed

Unix4ever:
I'd probably experiment with that to see how it looks first on the different pages before changing that.

Issue #1704, opened 2025-10-13
[bug] omni does not honor locked/unlocked machines during kubernetes updates


fsgh42:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

When rolling out a kubernetes update via omni, omni will not honor the order in which a user unlocks machines.

The order omni wants to do, seems to be based on alphanum sort of the machine's friendly names (hostnames).

### Expected Behavior

Say I have `n` worker machines in a cluster, all of them are locked.
I want to unlock a random worker machine and the update on this particular machine is triggered.

### Steps To Reproduce

1. create a talos/omni cluster with k8s v1.32 (3cp, 3 workers), lock all worker nodes
2. trigger update to k8s v1.33
3. wait for controlplanes to finish
4. omni will show "waiting for node xyz to be unlocked"
5. unlock another node -> update will stall

PR #1717, opened 2025-10-15
fix: honor lock status for machines during kubernetes upgrade

oguzkilcan:
Currently we do not check machine lock status while calculating the upgrade path for kubernetes, we do it during the update process. Since order is important for updates, this leads to locked machines being higher up in the upgrade plan then unlocked machines and blocking the update from proceeding.

This PR will move the check for machine lock status to upgrade plan calculation, eliminating locked machines blocking the whole kubernetes update process.

Fixes: #1704

### DIFF
--- a/internal/backend/runtime/omni/controllers/omni/internal/kubernetes/upgrade_test.go
+++ b/internal/backend/runtime/omni/controllers/omni/internal/kubernetes/upgrade_test.go
@@ -455,6 +455,194 @@ func TestUpgradePath(t *testing.T) {
 				},
 			},
 		},
+		{
+			name: "upgrade plan with locked nodes",
+			machineMap: &kubernetes.MachineMap{
+				ControlPlanes: map[string]string{
+					"cp1": "node-cp-1",
+				},
+				Workers: map[string]string{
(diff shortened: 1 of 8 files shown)


internal/backend/runtime/omni/controllers/omni/internal/kubernetes/upgrade_test.go
```diff
@@ -455,13 +455,171 @@ func TestUpgradePath(t *testing.T) {
...
+					},
+				},
+			}),
+			desiredVersion: "1.33.2",
+			expected: &kubernetes.UpgradePath{
+				ClusterID:          "cluster5",
+				AllComponentsReady: true,
+				AllNodes:           []string{"cp1", "worker2"},
+				AllNodesToRequiredImages: map[string][]string{
+					"cp1":     requiredImages("1.33.2", true),
```
smira, line 526: Images should still be pulled I think even if the node is locked, otherwise it makes upgrade slow for no reason

smira:
Should we have an integration test?

utkuozdemir (approved):
LGTM

PR #1722, opened 2025-10-16
feat(auth): make auth0 tokens only be valid for 2 minutes

### LINKED ISSUE #1716: Preventing re-use of auth tokens
We currently use long lived JWT tokens when creating pgp keys. This means leaked JWTs could be used to generate new pgp keys.

As we only need the JWT when creating keys, the endpoint for this should only allow a fresh JWT or some kind of one time challenge proof. This will however mean users will need to re-authenticate every 8 hours.

Related to #1713

Slessi:
Make auth0 auth tokens only be valid for 2 minutes. After this time the user will have to reauthenticate if he does not have any non-expired PGP keys.

Fixes #1716

### DIFF
--- a/frontend/src/main.ts
+++ b/frontend/src/main.ts
@@ -6,6 +6,7 @@
 import '@/index.css'
 
 import { createAuth0 } from '@auth0/auth0-vue'
+import { milliseconds, millisecondsToSeconds } from 'date-fns'
 import editorWorker from 'monaco-editor/esm/vs/editor/editor.worker?worker'
 import yamlWorker from 'monaco-yaml/yaml.worker?worker'
 import { createApp } from 'vue'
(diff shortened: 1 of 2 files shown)


frontend/src/main.ts
```diff
@@ -73,6 +74,7 @@ const setupApp = async () => {
         clientId: authConfigSpec!.auth0!.client_id!,
         authorizationParams: {
           redirect_uri: window.location.origin,
+          max_age: millisecondsToSeconds(milliseconds({ minutes: 2 })),
```
Slessi, line 77: n.b. about the API i wrote here I just _like_ the clarity of using date-fns's `milliseconds` API over doing things like `60 * 2 * 1000` or whatever.

`max_age` is a value in seconds.

Unix4ever (approved):

Issue #1703, opened 2025-10-13
[bug] cancel button should not be shown in omni UI on kubernetes minor updates


fsgh42:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

When triggering a minor `kubernetes` update (e.g. from `1.32.x` for `1.33.y`), a `cancel` button is shown in the UI.

To the best of my knowledge, it is not possible to cancel such an update, as downgrading minor versions it not possible.

### Expected Behavior

No cancel button is shown on `kubernetes` minor updates.

### Steps To Reproduce

1. create talos/omni cluster with k8s `v1.32.x`
2. trigger update to `1.33.y`

PR #1724, opened 2025-10-17
fix: hide cancel button on minor kubernetes upgrades

oguzkilcan:
Hide cancel button on minor Kubernetes upgrades as downgrading from one minor to another is not supported in Omni

Fixes: #1703

### DIFF
--- a/frontend/src/views/cluster/Overview/components/OverviewContent.vue
+++ b/frontend/src/views/cluster/Overview/components/OverviewContent.vue
@@ -5,6 +5,7 @@ Use of this software is governed by the Business Source License
 included in the LICENSE file.
 -->
 <script setup lang="ts">
+import * as semver from 'semver'
 import type { Ref } from 'vue'
 import { computed, onMounted, ref, toRefs, watch } from 'vue'
 
@@ -173,6 +174,12 @@ const machineLockedForKubernetesUpgrade = computed(() => {
 onMounted(async () => {
   isEmbeddedDiscoveryServiceAvailable.value = await embeddedDiscoveryServiceFeatureAvailable()
 })
+
+const isMinorDowngrade = (fromVersion?: string, toVersion?: string): boolean => {
+  if (!fromVersion || !toVersion) return true
+
+  return semver.minor(fromVersion) > semver.minor(toVersion)
+}
 </script>
 
 <template>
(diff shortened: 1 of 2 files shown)


frontend/src/views/cluster/Overview/components/OverviewContent.vue
```diff
@@ -173,6 +174,15 @@ const machineLockedForKubernetesUpgrade = computed(() => {
 onMounted(async () => {
   isEmbeddedDiscoveryServiceAvailable.value = await embeddedDiscoveryServiceFeatureAvailable()
 })
+
+const isMinorDowngrade = (
+  fromVersion: string | undefined,
+  toVersion: string | undefined,
+): boolean => {
+  if (fromVersion == undefined || toVersion == undefined) return true
```
Slessi: always always <ins>___always___</ins> triple equals in javascript `===`

that said, you can just do this here
```suggestion
  if (!fromVersion || !toVersion) return true
```

frontend/src/views/cluster/Overview/components/OverviewContent.vue
```diff
@@ -173,6 +174,15 @@ const machineLockedForKubernetesUpgrade = computed(() => {
 onMounted(async () => {
   isEmbeddedDiscoveryServiceAvailable.value = await embeddedDiscoveryServiceFeatureAvailable()
 })
+
+const isMinorDowngrade = (
+  fromVersion: string | undefined,
+  toVersion: string | undefined,
```
Slessi: Same but arguably preferred is below

```suggestion
  fromVersion?: string,
  toVersion?: string,
```

smira:
I don't think we should remove "Cancel" button, as otherwise how would you abort an upgrade? 

Yes, downgrading is not supported, but if the upgrade goes "wrong" and it hangs e.g. on the first node (api-server doesn't start), there should be a way to abort. 

"Cancel" != downgrade imho.

oguzkilcan:
This PR was just to bring the UI to the same level as Omni backend. When upgrade starts we set new version on `Cluster` resource, cancel button basically changes it to the previous version. 

Trying to set `KubernetesVersion` on `Cluster` resource to previous minor version fails with validation error. This applies to all the sources where Cluster resource can be modified. We were just not handling this properly on UI. Cancel button wasn't doing anything. 

We should discuss on how to implement abort functionality to upgrades.

PR #1725, opened 2025-10-17
chore: bump Talos to 1.11.3, reorder CI workflow jobs

utkuozdemir:
Make unit tests and lint run before the integration tests.

image


smira (approved):

client/go.mod
```diff
@@ -10,6 +10,7 @@ require (
 	github.com/ProtonMail/gopenpgp/v2 v2.9.0
 	github.com/adrg/xdg v0.5.3
 	github.com/blang/semver/v4 v4.0.0
+	github.com/containers/image/v5 v5.36.2
```
frezbo, line 13: oh we need to fix this, @shanduur machinery shouldn't have a need to import containers/image

Issue #1519, opened 2025-09-04
[bug] Cannot update service account role


noe-charmet:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

through the Omni UI, I try to update the permissions of a service account and get this error message

>Failed to Update Service Account
> User id is not defined

<img width="1394" height="708" alt="Image" src="https://github.com/user-attachments/assets/c304baaf-4697-4abf-8056-7c7acf8adf05" />

### Expected Behavior

That my service account role gets update without an error

### Steps To Reproduce

1. Create a service account
2. Attempt to update its permissions

Unix4ever:
Thanks for the report, we'll take a look.

PR #1731, opened 2025-10-20
fix: correctly fetch user ID for service accounts on the role edit page

Unix4ever:
Fixes: https://github.com/siderolabs/omni/issues/1519

### DIFF
--- a/frontend/src/views/omni/Modals/RoleEdit.vue
+++ b/frontend/src/views/omni/Modals/RoleEdit.vue
@@ -6,12 +6,16 @@ included in the LICENSE file.
 -->
 <script setup lang="ts">
 import type { Ref } from 'vue'
-import { ref } from 'vue'
+import { onBeforeMount, ref } from 'vue'
 import { useRoute, useRouter } from 'vue-router'
 
 import { Runtime } from '@/api/common/omni.pb'
+import { type Resource, ResourceService } from '@/api/grpc'
+import type { IdentitySpec, UserSpec } from '@/api/omni/specs/auth.pb'
+import { withRuntime } from '@/api/options'
 import {
   DefaultNamespace,
+  IdentityType,
   RoleAdmin,
   RoleNone,
   RoleOperator,
@@ -20,7 +24,7 @@ import {
 } from '@/api/resources'
 import TButton from '@/components/common/Button/TButton.vue'
 import TSelectList from '@/components/common/SelectList/TSelectList.vue'
-import Watch from '@/components/common/Watch/Watch.vue'
+import { useWatch } from '@/components/common/Watch/useWatch'
 import { canManageUsers } from '@/methods/auth'
 import { updateRole } from '@/methods/user'
 import { showError, showSuccess } from '@/notification'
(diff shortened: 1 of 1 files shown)


oguzkilcan (approved):

frontend/src/views/omni/Modals/RoleEdit.vue
```diff
@@ -79,27 +113,16 @@ const close = () => {
...
+      <template v-if="user">
+        <TSelectList
+          v-if="user?.spec?.role"
+          class="h-full grow"
+          title="Role"
+          :values="roles"
+          :default-value="user.spec?.role"
+          @checked-value="(value) => (role = value)"
+        />
+      </template>
```
Slessi: you dont need the template, `v-if` on the child is enough
```suggestion
      <TSelectList
        v-if="user?.spec.role"
        class="h-full grow"
        title="Role"
        :values="roles"
        :default-value="user.spec.role"
        @checked-value="(value) => (role = value)"
      />
```

frontend/src/views/omni/Modals/RoleEdit.vue
```diff
@@ -35,20 +39,50 @@ const role: Ref<string | undefined> = ref()
...
+  }),
+)
+
+onBeforeMount(async () => {
+  if (!route.query.serviceAccount) {
     return
   }
 
-  if (!route.query.user) {
-    showError(`Failed to Update ${object}`, 'User id is not defined')
```
Slessi, line 45: idk the page enough but is this check no longer needed? i see its moved to `userID` but will it always have a value?

frontend/src/views/omni/Modals/RoleEdit.vue
```diff
@@ -35,20 +39,50 @@ const role: Ref<string | undefined> = ref()
...
+      resource: {
+        id: userID.value,
+        namespace: DefaultNamespace,
+        type: UserType,
+      },
+      runtime: Runtime.Omni,
+      skip: !userID.value,
+    }
+  }),
+)
```
Slessi: for passing stuff to composables you can skip computed, a simple [getter function](https://vuejs.org/guide/reusability/composables.html#accepting-reactive-state) is ok too 


```suggestion
const { data: user } = useWatch<UserSpec>(() => ({
  resource: {
    id: userID.value,
    namespace: DefaultNamespace,
    type: UserType,
  },
  runtime: Runtime.Omni,
  skip: !userID.value,
}))
```

frontend/src/views/omni/Modals/RoleEdit.vue
```diff
@@ -35,20 +39,50 @@ const role: Ref<string | undefined> = ref()
...
+    const identity: Resource<IdentitySpec> = await ResourceService.Get(
+      {
+        id: route.query.serviceAccount as string,
+        namespace: DefaultNamespace,
+        type: IdentityType,
+      },
+      withRuntime(Runtime.Omni),
+    )
+
+    userID.value = identity.spec.user_id!
```
Slessi, line 69: idk the backend but if its possibly undefined, instead forcing it to be defined why not:
```suggestion
    if (identity.spec.user_id) {
      userID.value = identity.spec.user_id
    }
```

Slessi (approved):
LGTM just nits

frontend/src/views/omni/Modals/RoleEdit.vue
Unix4ever: It shouldn't be undefined. That's the way protobuf generator does it

frontend/src/views/omni/Modals/RoleEdit.vue
Unix4ever: `role` will be empty if we don't have the user, so I guess now this check is redundant.

Unix4ever: but actually lets keep it

frontend/src/views/omni/Modals/RoleEdit.vue
Slessi: that is unfortunate that we cannot rely on types then D:

Unix4ever: I think it's by nature of gRPC there. It can be unset, but the server never does it this way.

PR #1733, opened 2025-10-21
feat: implement extra kernel args support

### LINKED ISSUE #1732: [feature] Reimplement Extra Kernel Args
Redo the reverted https://github.com/siderolabs/omni/commit/049ab877e9a332d5ddd7aa867f5afc4a72d35fcd.

As discussed and agreed:
- Only for: BootedWithUKI OR (Talos 1.12 in maintenance mode OR GrubUseUKICmdline==true) (related: https://github.com/siderolabs/talos/pull/12029)
- Per-machine, no layered approach like config patches/extensions.
- No cluster templates support.
- Supports maintenance mode.
- Extract the extra args when machine is discovered then "detach" from them, i.e., leave their management to the user.

utkuozdemir:
(Re)implement the kernel args support functionality in the following way:
- Only support UKI or UKI-like (>=1.12 with GrubUseUKICmdline) systems.
- In `MachineStatusController`:
  - When we see a machine for the first time, do a one-time operation of extracting of the extra kernel args from it and store them in the newly introduced `KernelArgs` resource. This resource is user-owned from that point on.
  - Mark the `MachineStatus` with an annotation as "its kernel args are initialized".
  - Start storing the the raw schematic.
  - Take a one-time snapshot of the extensions on the machine and set them as "initial extensions". They might not be the "actual initial", i.e., the set of extensions when we actually seen the machine for the first time, but we do this in a best-effort basis. We need this, since now we cannot simply go back to the initial schematic ID when all extensions are removed - kernel args are also included in the schematic.
  - Start collecting the kernel cmdline from Talos machines as well.
- Adapt the `SchematicConfiguration` controller to not revert to the initial schematic ID ever - it now always computes the needed schematic - when it wants to revert to the initial set of extensions, it uses the new field on the `MachineStatus`.
- Introduce the resource `MachineUpgradeStatus` and its controller `MachineUpgradeStatusController`, which handles the maintenance mode upgrades when kernel args are updated. The controller is named this way, since our long-term plan is to centralize all upgrade calls to be done from this controller. Currently, it does not change Talos version or the set of extensions. It works only in maintenance mode, only for kernel args changes (when supported).
- Introduce the resource `KernelArgsStatus` and its controller `KernelArgsStatusController`, which provides information about the kernel args updates. Its status is reliable in both maintenance and non-maintenance modes.
- Build a UI to update these args.

Closes #1732.


frontend/src/views/omni/Modals/UpdateKernelArgs.vue
```diff
@@ -0,0 +1,198 @@
...
+const argsWatch = new Watch(status)
+
+argsWatch.setup({
+  runtime: Runtime.Omni,
+  resource: {
+    id: route.query.machine as string,
+    namespace: DefaultNamespace,
+    type: KernelArgsStatusType,
+  },
+})
```
Slessi: There is an easier (& preferred) thing for this now `useWatch`

```suggestion
const { data: status } = useWatch<KernelArgsStatusSpec>({
  runtime: Runtime.Omni,
  resource: {
    id: route.query.machine as string,
    namespace: DefaultNamespace,
    type: KernelArgsStatusType,
  },
})
```

frontend/src/views/omni/Modals/UpdateKernelArgs.vue
```diff
@@ -0,0 +1,198 @@
...
+
+  showSuccess(`Kernel args are updated`)
+
+  close()
+}
+
+const updateKernelArgs = async () => {
+  let existing: Resource<KernelArgsSpec> | undefined = undefined
+  try {
+    existing = await ResourceService.Get<Resource<KernelArgsSpec>>(md, withRuntime(Runtime.Omni))
```
Slessi: dont we have it already? from the `onMounted` call? if it might have changed since then, should we `watch` instead of `get`? or is there something else at play here?

frontend/src/views/omni/Modals/UpdateKernelArgs.vue
```diff
@@ -0,0 +1,198 @@
...
+    <div class="heading">
+      <h3 class="text-base text-naturals-n14">
+        Update Kernel Args of Machine {{ route.query.machine }}
+      </h3>
+      <CloseButton @click="close" />
+    </div>
+
+    <template v-if="unmetConditions.length > 0">
+      <div class="pb-5">
+        <TAlert type="warn" title="Kernel args won't be updated, there are unmet conditions:">
```
Slessi: template & div look extra here

```suggestion
    <TAlert
      v-if="unmetConditions.length > 0"
      type="warn"
      title="Kernel args won't be updated, there are unmet conditions:" class="mb-5">
```

utkuozdemir: for context, I did frontend just like a POC :) wanted to get you involved to do it properly.

client/api/omni/specs/omni.proto
```diff
@@ -1067,6 +1082,7 @@ message MachineClassSpec {
     string provider_data = 4;
 
     GrpcTunnelMode grpc_tunnel = 5;
+    repeated string extensions = 6;
```
utkuozdemir: Remove, this not being there was intentional

internal/backend/runtime/omni/controllers/omni/machine_status.go
```diff
@@ -118,6 +118,10 @@ func (ctrl *MachineStatusController) Settings() controller.QSettings {
 				Kind: controller.OutputExclusive,
 				Type: omni.MachineStatusType,
 			},
+			{
```
utkuozdemir: remove

internal/backend/runtime/omni/controllers/omni/machineupgrade/status.go
```diff
@@ -0,0 +1,346 @@
...
+
+			return nil
+		}
+
+		return err
+	}
+
+	logger.Info("Talos upgrade initiated")
+
+	status.TypedSpec().Value.Phase = specs.MachineUpgradeStatusSpec_Updating
```
utkuozdemir: name it upgrade

internal/backend/runtime/omni/controllers/omni/machineupgrade/status.go
```diff
@@ -0,0 +1,346 @@
...
+	status.TypedSpec().Value.Status = "Talos upgrade initiated"
+	status.TypedSpec().Value.Error = ""
+
+	return nil
```
utkuozdemir: requeue interval

client/pkg/omni/resources/omni/kernel_args.go
```diff
@@ -0,0 +1,53 @@
...
+	)
+}
+
+const (
+	// KernelArgsType is the type of the KernelArgs resource.
+	// tsgen:KernelArgsType
+	KernelArgsType = resource.Type("KernelArgs.omni.sidero.dev")
+)
+
+// KernelArgs describes desired machine KernelArgs list for a particular machine, machine set or cluster.
```
Unix4ever: Probably only for a single machine now?

client/pkg/omni/resources/omni/kernel_args_status.go
```diff
@@ -0,0 +1,65 @@
...
+	)
+}
+
+const (
+	// KernelArgsStatusType is the type of the KernelArgsStatus resource.
+	// tsgen:KernelArgsStatusType
+	KernelArgsStatusType = resource.Type("KernelArgsStatuses.omni.sidero.dev")
+)
+
+// KernelArgsStatus describes desired machine KernelArgsStatus list for a particular machine, machine set or cluster.
```
Unix4ever: This one looks outdated.

client/pkg/omni/resources/omni/ongoing_task.go
```diff
@@ -40,8 +40,7 @@ type OngoingTaskExtension struct{}
 func (OngoingTaskExtension) ResourceDefinition() meta.ResourceDefinitionSpec {
 	return meta.ResourceDefinitionSpec{
 		Type:             OngoingTaskType,
-		Aliases:          []resource.Type{},
-		DefaultNamespace: resources.DefaultNamespace,
+		DefaultNamespace: resources.EphemeralNamespace,
```
Unix4ever, line 43: Good change, but nothing will drop the ongoing task resources from the etcd state.

Unix4ever:
My proposed UI:

https://github.com/user-attachments/assets/10ddea04-6cd5-4ca7-8a42-3d5cb39100e4

Unix4ever:
The diff:
```
diff --git a/frontend/src/views/omni/Modals/UpdateKernelArgs.vue b/frontend/src/views/omni/Modals/UpdateKernelArgs.vue
index 5e034f24..2f374e84 100644
--- a/frontend/src/views/omni/Modals/UpdateKernelArgs.vue
+++ b/frontend/src/views/omni/Modals/UpdateKernelArgs.vue
@@ -14,10 +14,10 @@ import { type Resource, ResourceService } from '@/api/grpc'
 import { type KernelArgsSpec, type KernelArgsStatusSpec } from '@/api/omni/specs/omni.pb'
 import { withRuntime } from '@/api/options'
 import { DefaultNamespace, KernelArgsStatusType, KernelArgsType } from '@/api/resources'
+import IconButton from '@/components/common/Button/IconButton.vue'
 import TButton from '@/components/common/Button/TButton.vue'
 import TInput from '@/components/common/TInput/TInput.vue'
 import { useWatch } from '@/components/common/Watch/useWatch'
-import TAlert from '@/components/TAlert.vue'
 import { showError, showSuccess } from '@/notification'
 import CloseButton from '@/views/omni/Modals/CloseButton.vue'

@@ -51,12 +51,15 @@ const md = {
 }

 let kernelArgs: Resource<KernelArgsSpec> | undefined = undefined
+let initialArgs: string

 onMounted(async () => {
   try {
     kernelArgs = await ResourceService.Get<Resource<KernelArgsSpec>>(md, withRuntime(Runtime.Omni))

     args.value = (kernelArgs.spec.args || []).map((arg) => arg.trim()).join(' ')
+
+    initialArgs = args.value
   } catch (e) {
     if (e.code !== Code.NOT_FOUND) {
       throw e
@@ -122,6 +125,8 @@ const updateKernelArgs = async () => {
   await ResourceService.Update(kernelArgs, kernelArgs.metadata.version, withRuntime(Runtime.Omni))
 }

+const editArgs = ref(false)
+
 let closed = false

 const close = () => {
@@ -136,7 +141,7 @@ const close = () => {
 </script>

 <template>
-  <div class="modal-window">
+  <div class="modal-window flex flex-col gap-2">
     <div class="heading">
       <h3 class="text-base text-naturals-n14">
         Update Kernel Args of Machine {{ route.query.machine }}
@@ -144,29 +149,47 @@ const close = () => {
       <CloseButton @click="close" />
     </div>

-    <TAlert
-      v-if="unmetConditions.length > 0"
-      type="warn"
-      title="Kernel args won't be updated, there are unmet conditions:"
-    >
-      <span v-for="(condition, index) in unmetConditions" :key="index" class="py-2 text-primary-p3">
-        {{ condition }}
-      </span>
-    </TAlert>
-
-    <div class="mb-5">
-      <TAlert type="info" title="Current Kernel Args">
-        <p class="font-mono break-words whitespace-pre-wrap">{{ currentArgs }}</p>
-      </TAlert>
-      <TAlert type="info" title="Current Kernel Cmdline">
-        <p class="font-mono break-words whitespace-pre-wrap">{{ currentCmdline }}</p>
-      </TAlert>
-    </div>
-
-    <div class="flex flex-wrap items-center gap-2">
-      <TInput v-model="args" title="Extra Kernel Args" class="h-full flex-1" placeholder="..." />
-      <TButton type="highlighted" class="h-9 w-32" @click="handleUpdateKernelArgs">Update</TButton>
-    </div>
+    <template v-if="unmetConditions.length > 0">
+      <div class="font-bold text-primary-p3">
+        Kernel args won't be updated immediately, there are unmet conditions
+      </div>
+      <div class="flex flex-col">
+        <span
+          v-for="(condition, index) in unmetConditions"
+          :key="index"
+          class="my-0.5 rounded border-l-3 border-l-yellow-y1 bg-naturals-n5 py-1 pl-3 text-xs text-naturals-n13"
+        >
+          {{ condition }}
+        </span>
+      </div>
+    </template>
+
+    <div class="text-sm font-semibold text-naturals-n14">Current Kernel Cmdline</div>
+    <code>{{ currentCmdline || 'none' }}</code>
+    <template v-if="!editArgs">
+      <div class="text-sm font-semibold text-naturals-n14">Current Kernel Args</div>
+      <div class="my-0.5 flex items-center gap-2 rounded bg-naturals-n6 pr-2">
+        <code class="flex-1">
+          {{ currentArgs || 'none' }}
+        </code>
+        <IconButton
+          v-if="currentArgs === initialArgs"
+          icon="edit"
+          @click="() => (editArgs = true)"
+        />
+      </div>
+    </template>
+
+    <template v-if="currentArgs !== initialArgs || editArgs">
+      <div class="text-sm font-semibold text-naturals-n14">New Kernel Args</div>
+      <div class="flex flex-wrap items-center gap-2">
+        <TInput v-model="args" class="h-full flex-1 font-mono" placeholder="none" />
+        <TButton v-if="editArgs" class="h-9 w-32" @click="() => (editArgs = false)">Cancel</TButton>
+        <TButton type="highlighted" class="h-9 w-32" @click="handleUpdateKernelArgs">
+          Update
+        </TButton>
+      </div>
+    </template>
   </div>
 </template>

@@ -180,4 +203,8 @@ const close = () => {
 .heading {
   @apply mb-5 flex items-center justify-between text-xl text-naturals-n14;
 }
+
+code {
+  @apply rounded bg-naturals-n6 px-1 px-2.5 py-0.5 py-2 font-mono text-xs text-naturals-n13;
+}
 </style>
```

frontend/src/views/omni/Modals/DownloadInstallationMedia.vue
```diff
@@ -313,6 +313,8 @@ const createSchematic = async () => {
       }
     }
 
+    schematic.extensions?.sort()
```
Unix4ever: I'd still propose to sort it on the backend and drop the sort here.

frontend/src/views/omni/Modals/UpdateKernelArgs.vue
utkuozdemir: you were right, I changed it to use that one. `get` is the correct semantics.

client/pkg/omni/resources/omni/ongoing_task.go
utkuozdemir: We were always explicitly setting it to be Ephemeral, so there is none. It was just messing with `omnictl get` not returning anything unless you explicitly passed the namespace. Just a cosmetic thing.

frontend/src/views/omni/Modals/DownloadInstallationMedia.vue
utkuozdemir: We already do, so this is redundant, removing.

utkuozdemir:
> My proposed UI:
> 
>  Screen.Recording.2025-10-27.at.13.21.53.mov

Thanks, much nicer, applied your patch.

client/pkg/omni/resources/omni/ongoing_task.go
Unix4ever: oh, I see

PR #1739, opened 2025-10-22
chore(readme): clarify readme and add a comment in vite.config about allowedHosts

Slessi:
Clarify the readme section for customising the dev server and include a reference to it in the vite config regarding allowedHosts

### DIFF
--- a/frontend/vite.config.ts
+++ b/frontend/vite.config.ts
@@ -41,6 +41,7 @@ export default defineConfig(({ command }) => {
     server: {
       port: 8121,
       host: '127.0.0.1',
+      // See customizing the dev server in the readme for allowedHosts
     },
   }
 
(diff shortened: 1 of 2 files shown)


frontend/vite.config.ts
```diff
@@ -41,6 +41,7 @@ export default defineConfig(({ command }) => {
     server: {
       port: 8121,
       host: '127.0.0.1',
+      // See customizing the dev server in the readme for allowedHostss
```
utkuozdemir:
```suggestion
      // See customizing the dev server in the readme for allowedHosts
```

Slessi: oh sure, the spellcheck complained about `customising` but this is ok 😆

PR #1741, opened 2025-10-22
feat(frontend): add a radio group component

Slessi:
Adds a radio group component for use in the installation media form

image

### DIFF
--- a/frontend/src/components/common/Radio/RadioGroup.vue
+++ b/frontend/src/components/common/Radio/RadioGroup.vue
@@ -1,0 +1,62 @@
+<!--
+Copyright (c) 2025 Sidero Labs, Inc.
+
+Use of this software is governed by the Business Source License
+included in the LICENSE file.
+-->
+<script setup lang="ts" generic="T extends string | number | boolean | Record<string, any>">
+import {
+  RadioGroup,
+  RadioGroupDescription,
+  RadioGroupLabel,
+  RadioGroupOption,
+} from '@headlessui/vue'
+
+defineProps<{
+  label: string
+  options: {
+    label: string
+    description?: string
+    value: T
+  }[]
+}>()
+
+const model = defineModel<T>()
+</script>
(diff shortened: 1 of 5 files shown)


smira (approved):

frontend/src/components/common/Radio/RadioGroup.vue
```diff
@@ -0,0 +1,62 @@
...
+</script>
+
+<template>
+  <RadioGroup v-model="model">
+    <RadioGroupLabel class="mb-4 block text-sm font-medium text-naturals-n14">
+      {{ label }}
+    </RadioGroupLabel>
+
+    <div class="space-y-3">
+      <RadioGroupOption
```
Unix4ever, line 34: I'd maybe suggest to have `cursor-pointer` on the whole element.

Unix4ever:
And also could be nice to have some color transition on the element when it gets activated.
Just opacity-0 -> opacity-1 would be good enough.

We have transitions on the checkbox so I thought it could be more consistent this way.

frontend/src/components/common/Radio/RadioGroup.vue
Slessi: There are gaps in-between options that don't select anything, you could see a pointer cursor for those gaps and not have anything happen when you click - I think it's better if the pointer cursor only shows on elements that actually have click listeners

Unix4ever: Not sure I follow. I can click anywhere in the `div` which has both the circle and the label and between them and it works.
So I guess we can use `cursor-pointer` on the whole element 🤔

Slessi: Ahhh yea it is meant to be how you are saying, I misunderstood the first time. Was playing with things and forgot to move cursor-pointer back to where it was, will fix

PR #1742, opened 2025-10-23
fix: make workload proxy cookies HTTP only

Unix4ever:
It required moving cookies creation/deletion to the backend. Cookies are now updated on `ConfirmPublicKey` and `RevokePublicKey` calls.

Introduced the `RevokePublicKey` method as `auth0` flow didn't have any backend calls in the end of the logout.

Added a way for propagating cookies from the gRPC service to the gRPC gateway.

### DIFF
--- a/frontend/src/methods/auth.ts
+++ b/frontend/src/methods/auth.ts
@@ -3,6 +3,7 @@
 // Use of this software is governed by the Business Source License
 // included in the LICENSE file.
 
+import { type Auth0VueClient } from '@auth0/auth0-vue'
 import type { ComputedRef, Ref } from 'vue'
 import { computed, onBeforeMount, ref, watch } from 'vue'
 
@@ -17,6 +18,7 @@ import type {
 } from '@/api/omni/specs/virtual.pb'
 import { withRuntime } from '@/api/options'
 import {
+  AuthFlowQueryParam,
   ClusterPermissionsType,
   CurrentUserID,
   CurrentUserType,
@@ -26,8 +28,10 @@ import {
   PermissionsType,
   VirtualNamespace,
 } from '@/api/resources'
-import { identity } from '@/methods/key'
+import { identity, resetKeys, revokePublicKey } from '@/methods/key'
+import { FrontendAuthFlow } from '@/router'
 
+import { AuthType, authType } from '.'
 import { initializeUserPilot } from './features'
 
 export const currentUser: Ref<Resource<CurrentUserSpec> | undefined> = ref()
@@ -236,4 +240,30 @@ const updateToken = async (tokenID: string, update: (token: Resource<JoinTokenSp
   update(token)
 
   await ResourceService.Update(token, token.metadata.version, withRuntime(Runtime.Omni))
+}
+
+const redirectToURL = (url: string) => {
+  if (window.top) {
+    window.top.location.href = url
+  } else {
+    window.location.href = url
+  }
+}
+
+export const logout = async (auth0?: Auth0VueClient) => {
+  await revokePublicKey()
+
+  await auth0?.logout({
+    logoutParams: {
+      returnTo: window.location.origin,
+    },
+  })
+
+  resetKeys()
+
+  currentUser.value = undefined
+
+  if (authType.value !== AuthType.Auth0) {
+    redirectToURL(`/logout?${AuthFlowQueryParam}=${FrontendAuthFlow}`)
+  }
 }
--- a/frontend/src/methods/key.ts
+++ b/frontend/src/methods/key.ts
@@ -24,8 +24,6 @@ import {
   SignatureHeaderKey,
   SignatureVersionV1,
   TimestampHeaderKey,
-  workloadProxyPublicKeyIdCookie,
-  workloadProxyPublicKeyIdSignatureBase64Cookie,
 } from '@/api/resources'
 
 let interceptorsRegistered = false
@@ -33,6 +31,7 @@ let keysReloadTimeout: number
 
 const privateKeyArmored = useLocalStorage<string>('privateKey', null)
 const publicKeyArmored = useLocalStorage<string>('publicKey', null)
+const publicKeyID = useLocalStorage<string>('publicKeyID', null)
 
 export const identity = useLocalStorage<string>('identity', null)
 export const fullname = useLocalStorage<string>('fullname', null)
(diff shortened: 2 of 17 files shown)


frontend/src/methods/key.ts
```diff
@@ -257,7 +274,11 @@ const registerInterceptors = () => {
     request: async (url, config?: { headers?: Headers; method?: string }) => {
       url = encodeURI(url)
 
-      if (!/^\/(api|image)/.test(url) || url.indexOf('/api/auth.') !== -1) {
+      if (
+        !/^\/(api|image)/.test(url) ||
+        (url.indexOf('/api/auth.') !== -1 &&
+          url.indexOf('/api/auth.AuthService/RevokePublicKey') === -1)
```
Slessi: if its in the middle of the URL
```suggestion
        !url.includes('/api/auth.')
```

if its at the start of the URL
```suggestion
        !url.startsWith('/api/auth.')
```

but i guess if its at the start wouldnt `/^\/(api|image)/` already cover it?

frontend/src/methods/auth.ts
```diff
@@ -237,3 +241,29 @@ const updateToken = async (tokenID: string, update: (token: Resource<JoinTokenSp
...
+
+const redirectToURL = (url: string) => {
+  if (window.top) {
+    window.top.location.href = url
+  } else {
+    window.location.href = url
+  }
+}
+
+export const logout = async (auth0: Auth0VueClient | undefined) => {
```
Slessi: prefer `?:` so that you dont need to do `logout(undefined)` in places where u _know_ there is no `auth0` and can just do `logout()`

```suggestion
export const logout = async (auth0?: Auth0VueClient) => {
```

frontend/src/methods/auth.ts
```diff
@@ -237,3 +241,29 @@ const updateToken = async (tokenID: string, update: (token: Resource<JoinTokenSp
...
+    logoutParams: {
+      returnTo: window.location.origin,
+    },
+  })
+
+  resetKeys()
+
+  currentUser.value = undefined
+
+  if (authType.value !== AuthType.Auth0) {
```
Slessi, line 266: wasnt it before `(authType.value === AuthType.OIDC)`? and `SAML` did `navigateToLogin`

Unix4ever: yeah, it got changed

frontend/src/methods/key.ts
Unix4ever: we need to add signature to all requests that start with `/api` and `/image`, but not including:
- `/api/auth.AuthService/ConfirmPublicKey`
- `/api/auth.AuthService/RegisterPublicKey`

I agree that it would be nice to write it cleaner, but the condition you suggest is different.

Unix4ever: Another option to write this is this:
```ts
      if (!url.startsWith('/api') || !url.startsWith('/image')) {
        return [url, config]
      }

      if (url.startsWith('/api/auth.AuthService/') && !url.startsWith('/api/auth.AuthService/RevokePublicKey')) {
        return [url, config]
      }
```

Unix4ever:
>  I'm not seeing any set-cookie responses from backend, or any cookies in my dev tools - I should though, no?
>
> Frontend works regardless of this

It should be in the `ConfirmPublicKey` request:

Screenshot 2025-10-23 at 13 40 40

frontend/src/methods/key.ts
Unix4ever: Used `startsWith` where possible

internal/backend/grpc/auth.go
```diff
@@ -197,6 +226,65 @@ func (s *authServer) AwaitPublicKeyConfirmation(ctx context.Context, request *au
...
+	// resetting the workload proxy cookies
+	if err = cookies.Set(
+		ctx,
+		&http.Cookie{
+			Name:     workloadproxy.PublicKeyIDCookie,
+			Path:     "/",
+			Expires:  time.Unix(0, 0),
+			HttpOnly: true,
+			Secure:   true,
+			Domain:   s.workloadProxyCookieDomain,
```
Slessi, line 265: The domain has to match the domain you want to set the cookie for, I am getting this. Is it a problem in my setup?

Image

internal/backend/grpc/auth.go
```diff
@@ -197,6 +226,65 @@ func (s *authServer) AwaitPublicKeyConfirmation(ctx context.Context, request *au
...
+
+	// resetting the workload proxy cookies
+	if err = cookies.Set(
+		ctx,
+		&http.Cookie{
+			Name:     workloadproxy.PublicKeyIDCookie,
+			Path:     "/",
+			Expires:  time.Unix(0, 0),
+			HttpOnly: true,
+			Secure:   true,
```
Slessi, line 264: AFAIK we also want `SameSite: "Strict"` so that is only sent for requests originating from the `Domain` we set

Slessi (dismissed):

internal/backend/grpc/auth.go
Unix4ever: you use `omni.local:8099` as the domain to open Omni, right?

internal/backend/grpc/auth.go
Unix4ever: If the Omni domain is `user.omni.siderolabs.io`, the cookies should also work for `abcdef-user.proxy-us.siderolabs.io`.
If setting `SameSite: "Strict"` won't break it I can change it.

Slessi: Its unrelated to that. If you want the cookie to work for multiple subdomains, you set the domain for that to the parent domain (`siderolabs.io` if i understand that case) and it will work.

`SameSite: "Strict"` means that if the browser makes a request to `*.siderolabs.io` it will only include the cookies for that domain if the request _also_ comes from `*.siderolabs.io`.

I.e. if someone creates a page `evil.com` and on that page it makes a request to `siderolabs.io`, it won't include the cookies. Its a [CSRF](https://developer.mozilla.org/en-US/docs/Web/Security/Attacks/CSRF) thing.

internal/backend/grpc/auth.go
Slessi: yea, for that the domain should be `omni.local` - I imagine there is maybe some subdomain logic going on in backend but it cant set a cookie on `*.local` AFAIK. Need at least 2 parts.

Unix4ever: I couldn't find a working way to set these cookies so far :/ 
It's not a big deal as in SaaS we always have the domain of >=3 level.

This issue was there always, so I can merge it as is for now and then we can figure out a solution if we need to.

Slessi: sure, i thought it would be as simple as "last 2 parts" of the domain, e.g. `4.3.2.1`, `3.2.1` or `2.1` all return `2.1`, or if u only have `1` then just return `1`

PR #1745, opened 2025-10-23
feat(installation-media): add initial page for creating installation media

Slessi:
Add the initial page to create installation media.

image


Unix4ever:
Shouldn't the bottom bar stick to the bottom of the container here?
Screenshot 2025-10-24 at 16 09 10

Right now it feels like it's aligned by the size of the content above it

Slessi:
@Unix4ever yea, it is actually that way if you visit in browser `/machines/installation-media/create` but in storybook there isn't a parent container for the `h-full` to grow to so it takes the minimum content size

image

Unix4ever (approved):

Issue #1272, opened 2025-06-18
[feature] Allow force-deletion of machines managed by a (static) infra provider


utkuozdemir:
### Problem Description

- Your Omni instance has some machines managed by the bare-metal infra provider.
- Bare-metal infra provider instance is shut down/removed completely, while its machines are still present in Omni.
- Try to delete these machines (delete the machine from the UI / delete via `omnictl delete link`)
- The resources will get stuck because the bare-metal infra provider still has its finalizers on them.
- The only way out of this scenario at this moment is to bring back the provider. The provider ID has to match, it can be a dummy provider, it just needs to be able to attempt to finalize the machines (i.e., wipe their disks). Provider's finalization is already best-effort (i.e., even if it cannot wipe, it will remove its finalizer from the Omni resources `Link`, `InfraMachine` etc.)

The user, when they know that the bare-metal provider or the machines are gone for good (formatted/disconnected forever etc.), should be able to bring Omni to the clear, fresh state.

### Solution

Outcome of in-team discussion:
- It must not be possible to remove an `InfraProvider` as long as it has any machines that are not in `tearingDown` phase. In other words, it should either have no machines or the machines which were attempted to be removed. Do this via validation.
- User needs to explicitly attempt to remove all machines of this provider. They might get stuck in the `tearingDown` phase.
- Then, the user attempts to remove the `InfraProvider` on Omni.
- `InfraProvider` goes into `tearingDown` phase. In this phase, it won't let any new machine registrations for its ID.
- A new controller watches `InfraProvider`s as input, setting its finalizer on them. When an input goes `tearingDown`, it should clean up the finalizers of this `InfraProvider` from the respective resources (e.g., `Link`s, `InfraMachine`s and so on. Check for more). Effectively, acting as a "stub" of the infra provider being deleted, freeing these machines further deletion.
  - Idea: we should also remove the own namespace of the provider, to not leave any traces of it.
- After this new controller is done with its job, it removes its finalizer from `InfraProvider`, so it gets deleted as well.

utkuozdemir:
@Unix4ever can you please read through and see if I missed anything or got wrong? Feel free to edit.

Unix4ever:
looks good to me

PR #1748, opened 2025-10-24
feat: allow force-deletion of infra machines

oguzkilcan:
Allow force-deletion of `InfraMachine`s managed by a static `InfraProvider`.

* Add logic for removing a `Link` resource  for an associated `InfraMachine` to trigger deletion of finalizer `InfraMachineController` from a tearing down `ClusterMachine`, allowing ClusterMachine to be successfully deleted
* Add finalizer `InfraMachineController` to `Link` resource for associated `InfraMachine` to keep `Link` from prematurely deleted before `InfraMachine` is cleaned-up
* Clean up `Cluster` related labels from `InfraMachine` after it was removed from the cluster
* Clean up associated `Link` resources after `InfraProvider` is deleted
* Clean up finalizers set by `InfraProvider` on `InfraMachine` resource after `InfraProvider` is deleted
* Clean up associated `InfraMachineStatus` after `InfraProvider` is deleted
* Prevent deletion of `InfraProvider` when there are `Machine`s in running status managed by it
* Prevent deletion of `MachineStatus` until `Machine` has only `MachineStatusController` finalizer

Closes: #1272


utkuozdemir:
We realized that we can rework this PR a bit, so that the resources managed by controllers (`InfraMachine`, `ClusterMachine`) will be instead "freed (from the infra provider finalizer)" by their respective controllers, not in the cleanup controller.

So, let's hold this until @oguzkilcan is back.
cc: @Unix4ever

internal/backend/runtime/omni/controllers/omni/machine_status.go
```diff
@@ -372,6 +372,11 @@ func (ctrl *MachineStatusController) updateMachineConnectionStatus(machine *omni
 func (ctrl *MachineStatusController) reconcileTearingDown(ctx context.Context, r controller.QRuntime, logger *zap.Logger, machine *omni.Machine) error {
 	ctrl.runner.StopTask(logger, machine.Metadata().ID())
 
+	// Hold tearing down MachineStatus until the Machine has only this controller as a finalizer.
```
oguzkilcan: May not be needed anymore

PR #1749, opened 2025-10-24
feat(installation-media): add a stepper component

Slessi:
Add a stepper component for use in the installation media screens.

Also adds [Reka UI](https://reka-ui.com/) as a headless UI library dependency, [HeadlessUI](https://headlessui.com/) from tailwind isn't really maintained for vue and has a smaller selection of components.

### DIFF
--- a/frontend/eslint.config.ts
+++ b/frontend/eslint.config.ts
@@ -65,13 +65,13 @@ export default defineConfigWithVueTs(
       'vue/eqeqeq': 'error',
       'vue/match-component-file-name': 'error',
       'vue/match-component-import-name': 'error',
+      'vue/multi-word-component-names': 'off',
       'vue/next-tick-style': 'error',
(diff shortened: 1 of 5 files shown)


frontend/eslint.config.ts
```diff
@@ -65,13 +65,13 @@ export default defineConfigWithVueTs(
       'vue/eqeqeq': 'error',
       'vue/match-component-file-name': 'error',
       'vue/match-component-import-name': 'error',
+      'vue/multi-word-component-names': 'off',
```
Slessi, line 68: I give up on bothering with this, I get their [point](https://eslint.vuejs.org/rules/multi-word-component-names#rule-details) but the use case will never apply to us and when the component is `Button` or in this case `Stepper` I don't have anything good to add other than `TButton` or `TStepper` 🫠

PR #1754, opened 2025-10-26
Fix: Removing https from URL in values file for auth0

niklasfrick:
Double https:// leads to authentication with auth0 not working


Unix4ever:
Thanks for the contribution!

PR #1758, opened 2025-10-27
fix: make sidebar menus which do not open routes expand the submenus

Unix4ever:
This brings back the way it was working initially.


frontend/src/components/common/MenuItem/TMenuItem.vue
```diff
@@ -36,6 +36,10 @@ const vueroute = useRoute()
 const { subItems, level } = toRefs(props)
 
 const toggleSubmenu = () => {
+  if (props.route) {
+    return
+  }
```
Slessi: This seems to break the "Machines" item, the one that is both a link & submenu. I can no longer expand the submenu for that one.

The rest are unchanged by this and function the same

Slessi (approved):

PR #1759, opened 2025-10-27
fix: make sure that machine state is never `nil` in deprovision calls

Unix4ever:
Make it more predictable.
Even if the machine wasn't fully provisioned the state will be there, but will not have anything populated (which is fine).

### DIFF
--- a/client/pkg/infra/controllers/provision.go
+++ b/client/pkg/infra/controllers/provision.go
@@ -223,7 +223,7 @@ func (ctrl *ProvisionController[T]) Reconcile(ctx context.Context,
 	})
 }
 
-//nolint:gocognit
+//nolint:gocognit,gocyclo,cyclop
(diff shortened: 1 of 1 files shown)


client/pkg/infra/controllers/provision.go
```diff
@@ -470,6 +478,10 @@ func (ctrl *ProvisionController[T]) reconcileTearingDown(ctx context.Context, r
 		}
 	}
 
+	if err = ctrl.provisioner.Deprovision(ctx, logger, t, machineRequest); err != nil {
```
smira, line 481: is it okay that it got moved before other teardown actions?

Unix4ever: I think it should be fine. In case if it fails we still keep the machine request.

PR #1767, opened 2025-10-28
feat(installation-media): add placeholder steps

Slessi:
Add placeholder steps for the installation media wizard connected to a form state which is persisted to session storage.

https://github.com/user-attachments/assets/b9aa615d-4fda-4186-9025-02010c3ce7c8

### DIFF
--- a/frontend/src/components/common/TInput/TInput.vue
+++ b/frontend/src/components/common/TInput/TInput.vue
@@ -47,12 +47,10 @@ defineExpose({
 })
 
 const numberValue = computed(() =>
-  typeof modelValue.value === 'number' ? modelValue.value : parseFloat(modelValue.value),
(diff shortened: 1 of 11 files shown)


frontend/src/components/common/TInput/TInput.vue
```diff
@@ -47,12 +47,10 @@ defineExpose({
...
 
 const numberValue = computed(() =>
-  typeof modelValue.value === 'number' ? modelValue.value : parseFloat(modelValue.value),
+  typeof modelValue.value === 'number' ? modelValue.value : parseFloat(modelValue.value ?? ''),
 )
 
-const modelValue = defineModel<T>({
-  required: true,
-})
+const modelValue = defineModel<T>()
```
Slessi, line 53: made it support undefined so u can pass in models which are `string | undefined`

PR #1778, opened 2025-10-29
feat(installation-media): replace modal with link

Slessi:
Replace the installation media modal with a link to the wizard if feature flag is enabled


Unix4ever:
Probably it's better to direct the user to the list of saved presets instead.

Unix4ever (approved):

PR #1783, opened 2025-10-30
fix: remove non-machinery Talos import, fix changelog

utkuozdemir:
Changelog included the whole of Talos due to the unwanted dependency of `github.com/siderolabs/talos` brought in by a test. Remove that dependency, and re-generate the changelog for `v1.3.0-beta.0`.

Also, bump the Talos machinery version and rekres, which also bumps Go version to `1.25.3`.


internal/integration/import_test.go
```diff
@@ -20,7 +21,6 @@ import (
 	"github.com/cosi-project/runtime/pkg/state"
 	"github.com/siderolabs/gen/xslices"
 	"github.com/siderolabs/talos/pkg/machinery/client"
-	"github.com/siderolabs/talos/pkg/provision/providers/vm"
```
utkuozdemir, line 23: This was the culprit bringing Talos in.

go.mod
```diff
@@ -116,11 +117,6 @@ require (
 	sigs.k8s.io/controller-runtime v0.22.3
 )
 
-require (
-	github.com/siderolabs/talos v1.11.3
```
utkuozdemir, line 120: This is removed, and was causing the changelog to explode.

PR #1787, opened 2025-10-31
refactor(frontend): refactor tselectlist with reka-ui

Slessi:
Refactor `TSelectList` from `@headlessui` to `reka-ui` which is more testable out of the box and has collision detection. Also added tests, stories, and the `overheadTitle` variant for the installation media page.

### DIFF
--- a/frontend/src/components/common/Form/EnumRenderer.vue
+++ b/frontend/src/components/common/Form/EnumRenderer.vue
@@ -29,7 +29,6 @@ const values = computed(() => {
     <TSelectList
       :id="control.id + '-input'"
       class="h-6"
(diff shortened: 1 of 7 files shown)


frontend/src/components/common/Form/EnumRenderer.vue
```diff
@@ -29,7 +29,6 @@ const values = computed(() => {
     <TSelectList
       :id="control.id + '-input'"
       class="h-6"
-      menu-align="right"
```
Slessi, line 32: this property is no longer needed because it has automatic collision detection and will move the dropdown accordingly if it would've otherwise gone offscreen

PR #1791, opened 2025-11-03
fix(frontend): ignore invalid signature error on logout

Slessi:
Ignore invalid signature errors on logout which are expected if your saved keys do not exist inside etcd.

### DIFF
--- a/frontend/src/methods/auth.ts
+++ b/frontend/src/methods/auth.ts
@@ -8,6 +8,7 @@ import type { ComputedRef, Ref } from 'vue'
 import { computed, onBeforeMount, ref, watch } from 'vue'
 
 import { Runtime } from '@/api/common/omni.pb'
+import { Code } from '@/api/google/rpc/code.pb'
 import type { Resource } from '@/api/grpc'
 import { ResourceService } from '@/api/grpc'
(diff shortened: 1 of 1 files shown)


smira (approved):

frontend/src/methods/auth.ts
```diff
@@ -251,7 +251,12 @@ const redirectToURL = (url: string) => {
 }
 
 export const logout = async (auth0?: Auth0VueClient) => {
-  await revokePublicKey()
+  try {
+    await revokePublicKey()
+  } catch (error) {
+    // If the error is that we have bad keys, we can ignore it
+    if (error.message !== 'invalid signature') throw error
```
Unix4ever: I think this will be more reliable
```suggestion
    if (error.code !== Code.UNAUTHENTICATED) throw error
```

Unix4ever (approved):

frontend/src/methods/auth.ts
Slessi: Code is better, will change.

Did not realise we are [throwing an object](https://github.com/siderolabs/omni/blob/516d2debf11fc6dcf9730a2f5ba999284be91a87/frontend/src/api/fetch.pb.ts#L180) here, that is kinda unexpected to me

PR #1792, opened 2025-11-03
feat: centralize Schematic ID computation

utkuozdemir:
We compute the schematic id for a machine in two different places: in the `SchematicConfigurationController` for allocated machines, and in `MachineUpgradeStatusController` for maintenance mode machines.

Centralize this computation to be done only in `SchematicConfigurationController`.

Change the lifecycle of the `SchematicConfiguration` resource to be bound to a machine, not to a cluster.

### DIFF
--- a/internal/backend/runtime/omni/controllers/omni/schematic_configuration.go
+++ b/internal/backend/runtime/omni/controllers/omni/schematic_configuration.go
@@ -9,6 +9,7 @@ import (
 	"context"
 	"errors"
 	"slices"
+	"strings"
 
 	"github.com/blang/semver/v4"
 	"github.com/cosi-project/runtime/pkg/controller"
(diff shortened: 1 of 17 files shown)


frontend/src/components/common/OngoingTasks/OngoingTasks.vue
```diff
@@ -35,6 +35,15 @@ const resource = {
   type: OngoingTaskType,
 }
 
+const truncateSchematicId = (schematicId?: string) => {
+  const limit = 10
+  if (schematicId && schematicId.length > limit) {
+    return `${schematicId.slice(0, limit)}...`
```
Slessi: If this truncation is for aesthetics only, better to do it in CSS so we only truncate visually and as needed. Then it is still visible to screen readers and tests, and will remain fully copyable with keyboard. Also nice to show the full value on hover with a `<Tooltip>` component or `title=""` attribute.

internal/backend/runtime/omni/migration/migrations.go
```diff
@@ -2019,3 +2019,45 @@ func dropExtraInputFinalizers(ctx context.Context, st state.State, logger *zap.L
 
 	return nil
 }
+
+func dropSchematicConfigFinalizerFromClusterMachines(ctx context.Context, s state.State, _ *zap.Logger, _ migrationContext) error {
```
Unix4ever, line 2053: We need a test for that.

internal/backend/runtime/omni/controllers/omni/schematic_configuration.go
```diff
@@ -81,28 +81,43 @@ func NewSchematicConfigurationController(imageFactoryClient ImageFactoryClient)
...
+				}
+
+				ids := make([]resource.Pointer, 0, cmList.Len())
+
+				for clusterMachine := range cmList.All() {
+					msPtr := omni.NewMachineStatus(resources.DefaultNamespace, clusterMachine.Metadata().ID())
+					ids = append(ids, msPtr.Metadata())
+				}
+
+				return ids, nil
```
Unix4ever, line 111:
```suggestion
				return slices.Collect(cmList.Pointers()), nil
```

internal/backend/runtime/omni/migration/migrations.go
utkuozdemir: done

internal/backend/runtime/omni/controllers/omni/schematic_configuration.go
utkuozdemir: actually, we map them to MachineStatus pointers, so this doesn't work.

frontend/src/components/common/OngoingTasks/OngoingTasks.vue
utkuozdemir: I tried a few things and failed, so I created a separate task for it https://github.com/siderolabs/omni/issues/1882

and took the liberty of assigning it to you, hope that's ok :)

PR #1794, opened 2025-11-04
refactor(frontend): create a docs link generator

Slessi:
Create a docs link generator for omni/talos/k8s-guide docs.

For talos docs, it will default to the [DefaultTalosVersion](https://github.com/siderolabs/omni/blob/084e813a0e869cf9529c3c25443b628b92dbe977/frontend/src/api/resources.ts#L32) if you do not specifically specify a talos version.

### DIFF
--- a/frontend/src/methods/index.ts
+++ b/frontend/src/methods/index.ts
@@ -3,8 +3,9 @@
 // Use of this software is governed by the Business Source License
 // included in the LICENSE file.
 import type { Node as V1Node } from 'kubernetes-types/core/v1'
-import type { ComputedRef, Ref } from 'vue'
-import { computed, ref } from 'vue'
+import { coerce } from 'semver'
+import type { ComputedRef, MaybeRefOrGetter, Ref } from 'vue'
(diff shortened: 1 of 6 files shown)


frontend/src/methods/index.ts
```diff
@@ -34,47 +36,6 @@ export const getStatus = (item: V1Node) => {
   return NodesViewFilterOptions.NOT_READY
 }
 
-export const getServiceHealthStatus = (service) => {
```
Slessi, line 37: these deleted functions are all unused, so just cleaning

frontend/src/views/omni/Modals/UpdateKubernetes.vue
```diff
@@ -268,10 +262,7 @@ const upgradeClick = async () => {
     <p class="text-xs">
       Downgrading minor versions is not supported. You can not skip minor version upgrades. You can
       only upgrade to versions supported by your talos version. See the
```
smira:
```suggestion
      only upgrade to versions supported by your Talos version. See the
```

(or Talos Linux)

smira (approved):

PR #1797, opened 2025-11-05
fix(frontend): never add install disk for default disk

Slessi:
> Context: [Slack :thread:](https://talossystems.slack.com/archives/C05348FENHG/p1762189904234169?thread_ts=1761919202.636059&cid=C05348FENHG)
> When creating a cluster, an install disk config patch might be added even though you have selected the default install disk

Ensure that during cluster creation we never add an install disk config patch for the default install disk path.

Did a bit of a refactor to get clear access to the default disk value, and also it seems the "reset" functionality wasn't working, and clicking "Cancel" didn't do anything, but it does now.

### DIFF
--- a/frontend/src/views/omni/Clusters/Management/ClusterMachineItem.vue
+++ b/frontend/src/views/omni/Clusters/Management/ClusterMachineItem.vue
@@ -7,16 +7,11 @@ included in the LICENSE file.
 <script setup lang="ts">
 import yaml from 'js-yaml'
 import pluralize from 'pluralize'
-import type { Ref } from 'vue'
-import { computed, ref, toRefs, watch } from 'vue'
+import { computed, ref, watch } from 'vue'
 import WordHighlighter from 'vue-word-highlighter'
 
 import type { Resource } from '@/api/grpc'
-import type {
-  MachineConfigGenOptionsSpec,
-  MachineStatusSpec,
-  MachineStatusSpecHardwareStatusBlockDevice,
-} from '@/api/omni/specs/omni.pb'
+import type { MachineConfigGenOptionsSpec, MachineStatusSpec } from '@/api/omni/specs/omni.pb'
 import type { SiderolinkSpec } from '@/api/omni/specs/siderolink.pb'
 import {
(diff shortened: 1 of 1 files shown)


frontend/src/views/omni/Clusters/Management/ClusterMachineItem.vue
```diff
@@ -39,116 +34,86 @@ import CreateExtensions from '../../Modals/CreateExtensions.vue'
...
+const blockdevices = computed(() => item.spec.hardware?.blockdevices || [])
+const systemDiskPath = computed(
+  () => blockdevices.value.find((device) => device.system_disk)?.linux_name,
+)
+const disks = computed(() =>
+  blockdevices.value
+    .filter((device) => !device.readonly && device.type !== 'CD')
+    .map((device) => device.linux_name!),
+)
+const defaultInstallDisk = computed(() => disks.value[0])
```
Slessi: would like some confirmation if _really_ the default disk is always `disks[0]` - on the select its `item.spec.install_disk ?? disks[0]`

Unix4ever (approved):

frontend/src/views/omni/Clusters/Management/ClusterMachineItem.vue
Unix4ever: it should be `item.spec.install_disk ?? disks[0]` actually

Unix4ever: the item is loaded async if I remember correctly.

Slessi: Ok changed, wasn't sure

Unix4ever: or maybe not

Unix4ever: It comes from the props, so all good I guess

Unix4ever:
I'd propose to test it manually too.

PR #1800, opened 2025-11-05
chore: implement `InstallationMediaConfig` resource

### LINKED ISSUE #1382: Implement the API/resources to save installation media configs
The Resource should hold configured extensions, kernel args, secure boot config, image flavor, tunnel settings, labels.

Talos version and join token will be picked at the moment user downloads the image.
By default it will pick the default token version and default Talos version.

Unix4ever:
This resource is going to be used to store the saved installation media presets generated by the UI wizard.

Closes #1382


client/api/omni/specs/omni.proto
```diff
@@ -1440,3 +1440,21 @@ message InfraProviderCombinedStatusSpec {
 message MachineConfigDiffSpec {
   string diff = 1;
 }
+
+message InstallationMediaConfigSpec {
```
Unix4ever, line 1459: This might need some more thoughs, that's why the PR is still a draft.

client/api/omni/specs/omni.proto
```diff
@@ -1454,3 +1454,35 @@ message InfraProviderCombinedStatusSpec {
...
+  enum Bootloader {
+    BOOT_AUTO = 0;
```
Unix4ever: I'd suggest to unify that with https://github.com/siderolabs/omni/blob/e3835f0e3682db7d9951b4aa08aa52dcc89a4c73/client/api/omni/management/management.proto#L140-L145

You can define it in `client/api/common/omni.proto` on the top level.

Then do `import "common/omni.proto"` here and in the managment.proto

I think same goes for other message common message types.
I'd maybe only prefix it with some `PresetConfig` prefix like:
```
enum PresetConfigBootloader {}

message PresetConfigCloud {}
```

Slessi: it didnt feel like common enough to be a "common" file thing, I made it in `management.proto` as `SchematicBootloader` as it is mainly a schematic configuration item, and then i imported management into `omni.proto` as `management.SchematicBootloader` - WDYT?

Unix4ever: Yeah, that also works.

Unix4ever:
You can approve and merge it I guess :)

PR #1804, opened 2025-11-05
feat: support plain keys in the request signatures

Unix4ever:
Omni now supports ECDSA P-256 keys for signing the requests. The plain key should be encoded as PEM when it is submitted to `RegisterPublicKey` method.

Signature should be encoded using RFC4754 method (`r||s`).


Slessi:
@Unix4ever when submitting to `RegisterPublicKey` is it still on the `pgp_data` field ...? even when PEM encoding it, im told it couldnt find the armored data, which is a PGP thing

Unix4ever:
> @Unix4ever when submitting to RegisterPublicKey is it still on the pgp_data field ...? even when PEM encoding it, im told it couldnt find the armored data, which is a PGP thing

No, it will be a separate field, which is not there yet...
...because I forgot to update the version of proto files used for JS. Will do that on Monday.

Unix4ever:
So it will be necessary to set `plain_key` field instead of `pgp_data` when you call `RegisterPublicKey` in the frontend.

You will need to encode the public ECDSA key using `PEM` encoding and submit it in the following struct. Along with the not before and not after.

```ts
export type PublicKeyPlain = {
  key_pem?: string
  not_before?: GoogleProtobufTimestamp.Timestamp
  not_after?: GoogleProtobufTimestamp.Timestamp
}
```

PR #1607, opened 2025-09-21
Update the helm chart to be more robust.

dahui:
# Major Helm Chart Enhancement: External etcd Support and Production Features

## Overview

This pull request significantly enhances the Omni Helm chart with production-ready features while maintaining **full backwards compatibility** with existing deployments. The chart now supports both embedded and external etcd configurations, comprehensive ingress management, and advanced service configuration options.

## 🚨 Breaking Changes

**None** - This release maintains full backwards compatibility. Existing deployments will continue to work without any changes.

### 1. External etcd Support
- **New Configuration**: `etcd.external: true` enables external etcd cluster usage
- **Horizontal Scaling**: Multiple replicas supported when using external etcd
- **Authentication**: Both direct credentials and Kubernetes secret-based authentication
- **TLS Support**: Complete TLS configuration with file paths or secret-based certificates
- **Automatic Detection**: Chart automatically detects existing deployments and maintains compatibility

### 2. Intelligent Resource Management
- **StatefulSet for New Deployments**: Automatic PVC provisioning for embedded etcd
- **Deployment for External etcd**: Stateless deployment for external etcd configurations
- **Backwards Compatibility**: Existing Deployment-based installations continue unchanged
- **Resource Detection**: Uses Helm's `lookup` function to detect existing resources

### 3. Comprehensive Ingress Templates
- **Four Ingress Types**: API, UI, Siderolink, and Kubernetes Proxy
- **Cert-Manager Integration**: Automatic certificate provisioning
- **Wildcard Support**: Kubernetes proxy includes wildcard DNS for ArgoCD compatibility
- **Flexible Configuration**: Per-ingress customization with annotations and TLS

### 4. Enhanced Service Configuration
- **WireGuard Service Types**: NodePort, LoadBalancer, or ClusterIP support
- **Traffic Policy Control**: `externalTrafficPolicy` configuration for load balancing
- **Per-Service Annotations**: Global and service-specific annotation support
- **Automatic DNS Resolution**: WireGuard address auto-resolves to cluster DNS when not specified

### 5. Production-Ready Features
- **Pod Disruption Budgets**: Configurable availability guarantees
- **Advanced Security**: Proper capability management and device plugin support
- **Flexible Storage**: Support for storage classes and automatic PVC provisioning
- **Comprehensive Documentation**: Detailed configuration examples and troubleshooting

### Resource Selection Logic
```
1. Existing Deployment detected → Continue using Deployment (backwards compatibility)
[...]

PR #1809, opened 2025-11-05
feat(helm): add oidc auth flags support in omni chart

andrewiankidd:
Hi,

I saw OIDC support was recently implemented for 1.2.0 via b70560c1661387c5c34224140789f4e6df2304f9

I am able to use it via extraArgs but thought it would be nice to have it exposed in the Helm chart values like auth0 and saml

### DIFF
--- a/deploy/helm/omni/templates/deployment.yaml
+++ b/deploy/helm/omni/templates/deployment.yaml
@@ -91,6 +91,15 @@ spec:
           - --auth-saml-url={{ .Values.auth.saml.url }}
         {{- end }}
         {{- end }}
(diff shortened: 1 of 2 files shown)


utkuozdemir:
Hi @andrewiankidd,

Please see this comment: https://github.com/siderolabs/omni/pull/1607#issuecomment-3563066921.

I'd like to ask you to keep using `extraArgs` until the v2 chart is there - we will not keep maintaining the existing chart.
In the chart v2, all config options will be exposed in a nice way via helm values, including the Auth0 ones.

PR #1813, opened 2025-11-06
feat(frontend): add a split button component

### LINKED ISSUE #1470: [UI] Cater for a way to override tailwind classes
Tailwind classes are not overridable by default, having something like `p-0 px-1 py-2 py-1` it is unclear which set of padding you will be left with. This can happen when trying to override classes of a child in a parent, or with variants.

Something like [tailwind-merge](https://www.npmjs.com/package/tailwind-merge) can solve this

Example issue: https://github.com/siderolabs/omni/pull/1469 (can test the fix with this as a use case)
Bug that happened:

Image

Slessi:
Adds a split button component to button which is a button with a default action but also a drop-down of extra actions.

Also fixes #1470 with the inclusion of `tailwind-merge`

image


frontend/src/methods/utils.ts
Slessi, line 1: This utility func is just to make overriding tailwind classes possible

frontend/eslint.config.ts
```diff
@@ -70,6 +70,7 @@ export default defineConfigWithVueTs(
       'vue/no-boolean-default': 'error',
       'vue/no-template-target-blank': 'error',
       'vue/no-useless-mustaches': 'error',
+      'vue/require-default-prop': 'off',
```
Slessi, line 73: Particular lint rule doesn't seem helpful for vue3 SFC

frontend/src/modal.ts
```diff
@@ -2,20 +2,11 @@
...
 // included in the LICENSE file.
-
-import type { AllowedComponentProps, Component, Ref, VNodeProps } from 'vue'
+import type { Component } from 'vue'
 import { shallowRef } from 'vue'
+import type { ComponentProps } from 'vue-component-type-helpers'
 
-type ComponentProps<C extends Component> = C extends new (...args: any) => any
-  ? Omit<InstanceType<C>['$props'], keyof VNodeProps | keyof AllowedComponentProps>
-  : never
```
Slessi, line 11: saw this in passing, just replaced with `vue-component-type-helpers`

Unix4ever:
I'd probably center the dropdown using the full button width. And would make it the same width too

Slessi:
@Unix4ever done

PR #1814, opened 2025-11-06
feat(frontend): add a link generator to installation media

### LINKED ISSUE #1790: [feature] File server for ISOs
### Problem Description

When I need to upload a talos ISO to a server (read VPS) provider, often a URL is required for ISO upload. This means I need to provide a URL like https://myserver.org/my-talos-iso.iso. It is rather rare, that I can upload the ISO directly from my computer.

This requires me to spin up a machine and install a webserver (i.e. nginx) and scp the ISO there.

This is rather cumbersome in my experience: I download the ISO, I deploy a server, I scp the ISO there...

### Solution

It would be great if I had the possibility to click a button in the "Download Installation Media" dialog that is called "serve ISO for x hours" or something similar. Then, in the right sidebar of the home screen a new button will appear that will show me the URL for the ISO (for example https://myserver.org/my-talos-iso.iso) also with a button to copy the text.

Like that, after creation of the ISO, I could just head over to my VPS provider and paste the URL there.

### Alternative Solutions

All the alternatives would require me to download the ISO first.

### Notes

_No response_

Slessi:
Add a link generator to the installation media modal which provides a direct link to download the selected image from image factory.

Fixes #1790

https://github.com/user-attachments/assets/bda67874-2a27-4390-9430-c5f2e9ce4c49

### DIFF
--- a/frontend/src/methods/features.ts
+++ b/frontend/src/methods/features.ts
@@ -88,16 +88,6 @@ export const auditLogEnabled = async (): Promise<boolean> => {
   return featuresConfig.spec?.audit_log_enabled ?? false
 }
 
-export const getImageFactoryBaseURL = async (): Promise<string> => {
-  const featuresConfig = await getFeaturesConfig()
-
-  if (!featuresConfig.spec?.image_factory_base_url) {
-    throw new Error('image_factory_base_url is not set in features config')
-  }
-
-  return featuresConfig.spec?.image_factory_base_url
-}
-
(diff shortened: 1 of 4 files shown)


Unix4ever:
Maybe we can show a toast that it was copied?

Slessi:
@Unix4ever yea it crossed my mind, wasn't sure just because the PXE url copy one worked different, but I can add the toast.

Should I leave PXE URL copying the same or also a toast?

Unix4ever:
Lets align it and use toast everywhere

PR #1841, opened 2025-11-11
refactor(auth): extract interceptor from key generation logic

Slessi:
Extract interceptor code from the key generation logic in the frontend, simplified the remaining key logic, and wrote tests for all the things I touched.

### DIFF
--- a/frontend/src/App.vue
+++ b/frontend/src/App.vue
@@ -15,6 +15,11 @@ import UserConsent from '@/components/common/UserConsent/UserConsent.vue'
 import THeader from '@/components/THeader/THeader.vue'
 import TModal from '@/components/TModal.vue'
 import { suspended } from '@/methods'
+import { useRegisterAPIInterceptor } from '@/methods/interceptor'
+import { useWatchKeyExpiry } from '@/methods/key'
+
+useRegisterAPIInterceptor()
+useWatchKeyExpiry()
 
 const isSidebarOpen = ref(false)
 const router = useRouter()
--- a/frontend/src/components/common/OngoingTasks/OngoingTasks.vue
+++ b/frontend/src/components/common/OngoingTasks/OngoingTasks.vue
@@ -6,7 +6,7 @@ included in the LICENSE file.
 -->
 <script setup lang="ts">
 import { vOnClickOutside } from '@vueuse/components'
-import { computed, ref } from 'vue'
+import { ref } from 'vue'
 
 import { Runtime } from '@/api/common/omni.pb'
 import { type Resource } from '@/api/grpc'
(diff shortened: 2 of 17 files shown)


frontend/src/components/common/OngoingTasks/OngoingTasks.vue
```diff
@@ -15,26 +15,20 @@ import { EphemeralNamespace, OngoingTaskType } from '@/api/resources'
...
-import Watch from '@/components/common/Watch/Watch.vue'
+import { useWatch } from '@/components/common/Watch/useWatch'
 import IconHeaderDropdownLoading from '@/components/icons/IconHeaderDropdownLoading.vue'
-import { authorized } from '@/methods/key'
 import { formatISO } from '@/methods/time'
 
-const watchOpts = computed(() => {
-  if (authorized.value) {
-    return { resource: resource, runtime: Runtime.Omni }
-  }
```
Slessi, line 26: did a `v-if` on the parent instead

frontend/src/App.vue
```diff
@@ -15,6 +15,11 @@ import UserConsent from '@/components/common/UserConsent/UserConsent.vue'
 import THeader from '@/components/THeader/THeader.vue'
 import TModal from '@/components/TModal.vue'
 import { suspended } from '@/methods'
+import { useRegisterAPIInterceptor } from '@/methods/interceptor'
+import { useWatchKeyExpiry } from '@/methods/key'
+
+useRegisterAPIInterceptor()
+useWatchKeyExpiry()
```
Slessi, line 22: instead of the convoluted logic to load keys or reload the page, I register the hooks in `App.vue` once, and as they are reactive now they will fix themselves automatically in response to state changes, without having to reload the entire page

frontend/src/methods/interceptor.spec.ts
```diff
@@ -0,0 +1,146 @@
...
+    vi.useFakeTimers()
+    vi.setSystemTime(secondsToMilliseconds(now))
+  })
+
+  afterEach(() => {
+    vi.useRealTimers()
+  })
+
+  test.each(['/api', '/api/auth.AuthService/RevokePublicKey'])(
+    'intercepts api route: %s',
```
Slessi, line 58: @Unix4ever `/api/auth.AuthService/RevokePublicKey` is the only exception for `/api/auth` right? for interception

I test that others are not intercepted below

Unix4ever: correct

frontend/src/components/common/OngoingTasks/OngoingTasks.vue
```diff
@@ -6,7 +6,7 @@ included in the LICENSE file.
 -->
 <script setup lang="ts">
```
Unix4ever, line 7: I'd suggest to add storybook for this component as it was changed.

Slessi: done

PR #1844, opened 2025-11-12
fix: use deterministic order for machine extensions

utkuozdemir:
When machine extensions are defined on multiple levels for a single machine, e.g., at cluster, machine set and cluster machine level, the one that was picked up as the actual configuration was determined randomly.

Add a unit test to reproduce this and fix it by calculating them using a deterministic order.

Co-authored-by: Oguz Kilcan <oguz.kilcan@siderolabs.com>

### DIFF
--- a/internal/backend/runtime/omni/controllers/omni/machine_extensions.go
+++ b/internal/backend/runtime/omni/controllers/omni/machine_extensions.go
@@ -8,14 +8,17 @@ package omni
 import (
 	"context"
 	"fmt"
+	"slices"
 	"time"
 
 	"github.com/cosi-project/runtime/pkg/controller"
 	"github.com/cosi-project/runtime/pkg/controller/generic"
+	"github.com/cosi-project/runtime/pkg/controller/generic/qtransform"
 	"github.com/cosi-project/runtime/pkg/resource"
 	"github.com/cosi-project/runtime/pkg/safe"
 	"github.com/cosi-project/runtime/pkg/state"
 	"github.com/siderolabs/gen/optional"
+	"github.com/siderolabs/gen/xerrors"
 	"go.uber.org/zap"
 
 	"github.com/siderolabs/omni/client/pkg/omni/resources"
@@ -156,9 +159,7 @@ func (ctrl *MachineExtensionsController) MapInput(ctx context.Context, _ *zap.Lo
 }
 
 // Reconcile implements controller.QController interface.
-func (ctrl *MachineExtensionsController) Reconcile(ctx context.Context,
-	_ *zap.Logger, r controller.QRuntime, ptr resource.Pointer,
-) error {
+func (ctrl *MachineExtensionsController) Reconcile(ctx context.Context, logger *zap.Logger, r controller.QRuntime, ptr resource.Pointer) error {
 	configuration, err := safe.ReaderGetByID[*omni.ExtensionsConfiguration](ctx, r, ptr.ID())
 	if err != nil {
 		if state.IsNotFoundError(err) {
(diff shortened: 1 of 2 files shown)


internal/backend/runtime/omni/controllers/omni/machine_extensions.go
```diff
@@ -213,9 +226,65 @@ func (ctrl *MachineExtensionsController) Reconcile(ctx context.Context,
...
+	cmID := cm.Metadata().ID()
+	msID, _ := cm.Metadata().Labels().Get(omni.LabelMachineSet)
+	clusterID, _ := cm.Metadata().Labels().Get(omni.LabelCluster)
+
+	var clusterMachineLevel, machineSetLevel, clusterLevel *omni.ExtensionsConfiguration
+
+	for config := range configs {
+		clusterMachine, ok := config.Metadata().Labels().Get(omni.LabelClusterMachine)
+		if ok && clusterMachine == cmID {
+			clusterMachineLevel = config
```
smira, line 275: should we just `return config` here to avoid going into complexity of `continue` below?

internal/backend/runtime/omni/controllers/omni/machine_extensions.go
```diff
@@ -213,9 +226,65 @@ func (ctrl *MachineExtensionsController) Reconcile(ctx context.Context,
...
-	r controller.QRuntime, configuration *omni.ExtensionsConfiguration,
-) ([]*omni.ClusterMachine, error) {
+// determineExtensions determines the extensions for the given cluster machine.
+// The extensions are determined in the following order:
+// 1. Extensions defined for the cluster machine itself.
+// 2. Extensions defined for the machine set the cluster machine belongs to.
+// 3. Extensions defined for the cluster machine belongs to.
+//
+// If there are multiple extensions defined for the same level, the one with the lexicographically highest ID is used.
+func (ctrl *MachineExtensionsController) determineExtensions(cm *omni.ClusterMachine, configs iter.Seq[*omni.ExtensionsConfiguration]) []string {
```
smira: also as this has something I think is counter-intuitive - I would expect probably extensions to be merged, not replaced (I guess this matches config patches more closely)

utkuozdemir: yes, but it never worked that way. I'd keep this as-is, at least in this fix, unless we decide otherwise.

internal/backend/runtime/omni/controllers/omni/machine_extensions.go
utkuozdemir: it'd pick the first matching one, not the last. I could flip the logic (iterate in reverse) but don't see much win in it.

internal/backend/runtime/omni/controllers/omni/machine_extensions.go
```diff
@@ -213,9 +232,88 @@ func (ctrl *MachineExtensionsController) Reconcile(ctx context.Context,
...
+	if isCreate {
+		return true
+	}
+
+	if versionsWereSet && versionsAreUpdated {
+		return true
+	}
+
+	return false
+}
```
utkuozdemir, line 256: @Unix4ever @oguzkilcan @smira this is the idea to avoid unwanted upgrades, can you pls have a look? it also has a test.

internal/backend/runtime/omni/controllers/omni/machine_extensions.go
smira: I think the code right now prefers ClusterMachine match always, so I don't get your point. You can just quit once there's a match on ClusterMachine, no?

utkuozdemir: If there are 2 clustermachine level configs, called a and b, b should win. With your suggestion, it would return a.

internal/backend/runtime/omni/controllers/omni/machine_extensions.go
smira: I feel keeping backwards compatibility for buggy behavior makes our future life even harder. I understand the idea to avoid unwanted upgrades, but for how would we keep it?

utkuozdemir: Good question, I guess it'd be until we rework the extensions completely to simplify it. Hopefully rather soon.

Need to convince @TimJones to drop it :)

internal/backend/runtime/omni/controllers/omni/machine_extensions.go
smira: omg, is it even possible to have two? how do you decide which one is better? just by name?

internal/backend/runtime/omni/controllers/omni/machine_extensions.go
Unix4ever: I think we can drop this workaround in 1.4 or 1.5.

internal/backend/runtime/omni/controllers/omni/machine_extensions.go
Unix4ever: As that's a label someone can create a couple of them if they mess up with `omnictl`.
All high level tools don't do that, so I think it shouldn't be the case.

We can actually introduce a validation check on create/update to fail when multiple configs defined for the same resource.

Unix4ever: As of the order, I think if you have multiple defined it was also undetermined. Leaning towards alphabetically last one being higher priority.

utkuozdemir: It was undetermined, but with this change, it is, as COSI returns them in alphabetical order in the List call.
I did the "last one wins" because it is more consistent with how config patches work (priority-wise only - we don't do "merge" like config patches)

PR #1848, opened 2025-11-13
refactor(frontend): switch from openpgp to webcrypto

### LINKED ISSUE #1713: [bug] pgp keys should be issued using WebCrypto instead of using openpgp library for JS
Because we generate them we can only save them in the local storage.
Investigate if we can use WebCrypto instead and store the keys more securely so that it won't be possible to read them in the JS code, but only use them to sign the requests.

Slessi:
Switch from using openpgp to webcrypto to sign requests to the API.

Closes #1713


frontend/src/methods/key.spec.ts
```diff
@@ -291,76 +225,69 @@ describe('useSignDetached', () => {
...
+    expect(privateKey).toBeInstanceOf(CryptoKey)
+    expect(publicKey).toBeInstanceOf(CryptoKey)
 
-    expect(generateKey).toHaveBeenCalledExactlyOnceWith(
-      expect.objectContaining({
-        userIDs: [{ email: 'myfancy@email.com' }],
-      }),
+    await expect(crypto.subtle.exportKey('jwk', privateKey)).rejects.toThrowError(
+      'key is not extractable',
     )
```
Slessi, line 286: with this we can be sure it is not exportable

frontend/src/views/omni/Auth/Authenticate.vue
```diff
@@ -199,8 +199,9 @@ const confirmPublicKey = async (privateKey?: string) => {
 
     const metadata: Record<string, string> = {}
 
-    if (privateKey) {
-      const publicKeyIdSignatureBase64 = await signDetached(publicKeyId, privateKey)
+    if (keyPair) {
+      const array = new Uint8Array(await signDetached(publicKeyId, keyPair))
+      const publicKeyIdSignatureBase64 = b64Encode(array, 0, array.length)
```
Unix4ever, line 204: This should also use the public key ID from the `RegisterPublicKey` response I think.

Slessi: The code isn't the best here because it depends on function call order and a shared outside variable being updated, but it does use the response from `RegisterPublicKey` already

frontend/e2e/auth.setup.ts
```diff
@@ -19,8 +19,7 @@ setup('auth setup', async ({ page }) => {
   await page.getByRole('textbox', { name: 'Password' }).fill(process.env.AUTH_PASSWORD)
   await page.getByRole('button', { name: 'Continue', exact: true }).click()
 
-  await page.getByRole('button', { name: 'Log In' }).click()
-  await page.getByRole('heading', { name: 'Home' }).waitFor()
+  await page.getByRole('button', { name: 'Log In' }).waitFor()
```
Slessi: For E2E tests, whilst playwright supports saving the state of indexedDB, it can only save things that are serialisable to JSON, and un-exportable keys are not, so I have to move the "Log in" click to every individual test

not sure of a nicer way to automate this, the manual solution is acceptable for now

Unix4ever (approved):

PR #1857, opened 2025-11-14
chore: rekres

Slessi:
Rekres to get latest changes from kres to allow multiple commits per PR.


.github/workflows/ci.yaml
```diff
@@ -1,20 +1,20 @@
...
     branches:
-      - main
+      - frontend-webcrypto
       - release-*
     tags:
       - v*
   pull_request:
     branches:
-      - main
+      - frontend-webcrypto
```
smira: This feels wrong

PR #1880, opened 2025-11-18
feat(installation-media): implement cloud provider + sbc steps

Slessi:
- Implement cloud provider + SBC steps for installation media wizard
- Fix a scrolling issue on the installation media container page.
- Replace `useDocsLink` composable with a simpler `getDocsLink` function to allow calling in locations outside of the vue lifecycle. Don't need the complexity of a composable for this.

### DIFF
--- a/frontend/src/views/omni/InstallationMedia/Steps/CloudProvider.vue
+++ b/frontend/src/views/omni/InstallationMedia/Steps/CloudProvider.vue
@@ -5,9 +5,138 @@ Use of this software is governed by the Business Source License
 included in the LICENSE file.
 -->
 <script setup lang="ts">
+import { computed } from 'vue'
+
+import RadioGroup from '@/components/common/Radio/RadioGroup.vue'
+import RadioGroupOption from '@/components/common/Radio/RadioGroupOption.vue'
+import { getDocsLink } from '@/methods'
 import type { FormState } from '@/views/omni/InstallationMedia/InstallationMediaCreate.vue'
 
-defineModel<FormState>({ required: true })
+const formState = defineModel<FormState>({ required: true })
+
+const platforms = computed(() =>
+  [
+    {
+      value: 'aws',
+      label: 'Amazon Web Services (AWS)',
+      description: 'Runs on AWS VMs booted from an AMI',
+      documentation: '/platform-specific-installations/cloud-platforms/aws',
+    },
+    {
+      value: 'gcp',
+      label: 'Google Cloud (GCP)',
+      description: 'Runs on Google Cloud VMs booted from a disk image',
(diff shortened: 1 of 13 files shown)


frontend/src/views/omni/InstallationMedia/Steps/CloudProvider.vue
```diff
@@ -5,9 +5,138 @@ Use of this software is governed by the Business Source License
...
+
+import RadioGroup from '@/components/common/Radio/RadioGroup.vue'
+import RadioGroupOption from '@/components/common/Radio/RadioGroupOption.vue'
+import { getDocsLink } from '@/methods'
 import type { FormState } from '@/views/omni/InstallationMedia/InstallationMediaCreate.vue'
 
-defineModel<FormState>({ required: true })
+const formState = defineModel<FormState>({ required: true })
+
+const platforms = computed(() =>
```
Slessi, line 17: I plucked the platform list off of https://github.com/siderolabs/talos/blob/80ab7a0643fc8057283a8ba3eb912d0ee453c143/pkg/machinery/platforms/platforms.go#L146

I see the URLs there are the old style ones, I updated them to the new style ones here, for example:
- Before: https://www.talos.dev/v1.11/talos-guides/install/cloud-platforms/aws/
- After: https://docs.siderolabs.com/talos/v1.11/platform-specific-installations/cloud-platforms/aws

Talos probably should be updated with the new links, but then image factory also would need to be updated to point to the new domain I think or the redirects might not work, e.g. https://docs.siderolabs.com/v1.11/talos-guides/install/cloud-platforms/aws/ is a 404

frontend/src/views/omni/InstallationMedia/Steps/CloudProvider.vue
```diff
@@ -5,9 +5,138 @@ Use of this software is governed by the Business Source License
...
 
-<template>Cloud Provider</template>
+<template>
+  <RadioGroup v-mode="formState.cloudPlatform" label="Cloud">
+    <RadioGroupOption v-for="platform in platforms" :key="platform.value" :value="platform.value">
+      {{ platform.label }}
+
+      <!-- prettier-ignore -->
+      <template #description>
+        {{ platform.description }} (<a class="text-primary-p3 underline" :href="platform.documentation">documentation</a>).
```
Unix4ever: I'd suggest to add `click.prevent` or something here to avoid selecting the radio button, when you just want to see the docs.

Slessi: Yea I just noticed it too, it's fixed now. The correct one is `@click.stop` so that we stop propagating/bubbling the event to parent handlers. `@click.prevent` would prevent the default action of opening the link

PR #1885, opened 2025-11-19
feat(installation-media): implement system extensions step

Slessi:
Implement the system extension step for the installation media wizard.

### DIFF
--- a/frontend/src/views/omni/InstallationMedia/Steps/SystemExtensions.vue
+++ b/frontend/src/views/omni/InstallationMedia/Steps/SystemExtensions.vue
@@ -5,9 +5,84 @@ Use of this software is governed by the Business Source License
 included in the LICENSE file.
 -->
 <script setup lang="ts">
+import { computed, ref } from 'vue'
+
+import { Runtime } from '@/api/common/omni.pb'
+import type { TalosExtensionsSpec } from '@/api/omni/specs/omni.pb'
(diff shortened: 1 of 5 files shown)


frontend/src/views/omni/InstallationMedia/Steps/SystemExtensions.vue
```diff
@@ -5,9 +5,86 @@ Use of this software is governed by the Business Source License
...
+    </p>
+
+    <TInput v-model="filterExtensions" placeholder="Search" icon="search" />
+
+    <div class="flex flex-col gap-4 overflow-auto">
+      <TCheckbox
+        v-for="item in filteredExtensions"
+        :key="item.name"
+        :model-value="formState.systemExtensions?.includes(item.name!)"
+        @update:model-value="(value) => toggleExtension(item.name!, value)"
```
Slessi, line 77: not clear to me if we need to send to the API a list of `name` or a list of `ref`

i.e. `siderolabs/amazon-ena` or  `ghcr.io/siderolabs/amazon-ena:2.15.0-v1.11.5`

frezbo: the former to the factory api

Unix4ever: It's a list of `name` actually.
Schematic expects the list of those:
```
- siderolabs/qemu-guest-agent
```

PR #1892, opened 2025-11-20
fix: do not clear schematic meta values for non-UKI machines

utkuozdemir:
META section updates are no-op for non-UKI machines, but still, the recent changes in the kernel args PR started clearing them (since now we compute schematic ID always), causing the schematic ID to be updated, which caused cluster machines to be upgraded and restarted.

Remove the UKI check and keep meta valus always as-is.

Update the integration tests to:
- Also include META values.
- Make Omni upgrade test pick both UKI and non-UKI machines.

### DIFF
--- a/hack/test/integration.sh
+++ b/hack/test/integration.sh
@@ -99,6 +99,7 @@ fi
 KERNEL_ARGS_SCHEMATIC=$(
   cat <<EOF
 customization:
+  meta: [{ key: 42, value: test-1 }, { key: 41, value: test-2 }]
   extraKernelArgs:
     - siderolink.api=grpc://${WIREGUARD_IP}:8090?jointoken=${JOIN_TOKEN}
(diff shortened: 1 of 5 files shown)


utkuozdemir:
Not added the actual change yet, first want to see the Omni upgrade test failing.

Unix4ever:
I think the old integration test might pick only uki machines and the test will pass.

utkuozdemir:
> I think the old integration test might pick only uki machines and the test will pass.

ah right, that must be what's happening. anyways, I now reproduced it locally, and gonna test my fix.

utkuozdemir:
@oguzkilcan @Unix4ever tested locally: first reproduced the exact behavior between 1.2.1 and main. Then verified the same flow between 1.2.1 and this branch, verified that no upgrades occur.

After merging this, I'll go ahead with backporting it.

PR #1896, opened 2025-11-20
feat(frontend): style all regular links with primary

Slessi:
Adds a `link-primary` class to style regular links with our branding. Also includes some hover/active/visited feedback.

This is to standardise links to this format:
image

Stories were added for places where links were updated.

### DIFF
--- a/frontend/src/components/common/UserConsent/UserConsent.vue
+++ b/frontend/src/components/common/UserConsent/UserConsent.vue
@@ -5,12 +5,11 @@ Use of this software is governed by the Business Source License
 included in the LICENSE file.
 -->
 <script setup lang="ts">
-import { ref, watch } from 'vue'
+import { computedAsync } from '@vueuse/core'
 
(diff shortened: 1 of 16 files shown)


frontend/src/components/common/UserConsent/UserConsent.vue
```diff
@@ -22,17 +21,13 @@ const enableTracking = async () => {
...
   if (!currentUser.value) {
     return
   }
 
-  trackerToken.value = await getUserPilotToken()
-}
-
-watch(() => currentUser.value, fetchTrackingToken)
+  return await getUserPilotToken()
+})
```
Slessi, line 30: This was changed to `computedAsync` as when I wrote the stories I ran into an issue were if `currentUser.value` never changes, then `fetchTrackingToken` is never called. Could also have been fixed with `{ immediate: true }` but `computedAsync` is closer to the expected intent here.

Unix4ever (approved):

PR #1897, opened 2025-11-20
feat: check computed schematic ID to avoid unnecessary machine upgrades

utkuozdemir:
If we didn't change anything in a schematic, i.e., picked everything from the `MachineStatus`, but ended up calculating a different schematic ID than what the machine already has, just log it as a warning, but preserve the existing schematic ID.


utkuozdemir:
was a POC but we won't do it this way, closing.

PR #1900, opened 2025-11-20
feat: pull platforms and SBC information from Talos machinery

Unix4ever:
Present all of that as 3 kinds of virtual resources:

- `MetalPlatformConfig`.
- `CloudPlatformConfig`.
- `SBCConfig`

Virtual resource supports `Get` and `List` operations.


frontend/src/api/talos/machine/machine.proto
```diff
@@ -592,7 +592,7 @@ message ReadRequest {
   string path = 1;
 }
 
-// LogsContainer describes all available registered log containers.
+// LogsContainer desribes all avalaible registered log containers.
```
Slessi, line 595: :(

Unix4ever: this typo is in Talos repo

Unix4ever: it was fixed there, we'll need to update the machinery version though. Won't do it here yet

utkuozdemir: it'll eventually get to the right state, just like rest of our systems :)

internal/backend/runtime/omni/controllers/omni/talos_upgrade_status_test.go
```diff
@@ -287,6 +287,10 @@ func (suite *TalosUpgradeStatusSuite) TestReconcileLocked() {
 
 		cmtv.TypedSpec().Value.TalosVersion = cluster.TypedSpec().Value.TalosVersion
 
+		if _, ok := res.Metadata().Labels().Get(omni.LabelWorkerRole); !ok {
```
Unix4ever, line 290: Fix flaky unit test

oguzkilcan (approved):

PR #1904, opened 2025-11-20
fix(frontend): sort talosctl versions correctly and select correct default

Slessi:
Correctly sort the talosctl versions inside the download talosctl modal by version instead of alphabetically, and select the actual latest version as the default.

### DIFF
--- a/frontend/src/views/omni/Modals/DownloadTalosctl.vue
+++ b/frontend/src/views/omni/Modals/DownloadTalosctl.vue
@@ -5,7 +5,9 @@ Use of this software is governed by the Business Source License
 included in the LICENSE file.
 -->
 <script setup lang="ts">
-import { computed, onMounted, ref } from 'vue'
+import { computedAsync } from '@vueuse/core'
+import { compare } from 'semver'
+import { computed, ref } from 'vue'
 import { useRouter } from 'vue-router'
 
 import TButton from '@/components/common/Button/TButton.vue'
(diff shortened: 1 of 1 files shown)


frontend/src/views/omni/Modals/DownloadTalosctl.vue
```diff
@@ -27,46 +29,44 @@ const close = () => {
...
-  selectedVersion.value = defaultVersion.value
-
-  defaultValue.value =
+const availableVersions = computed(
+  () =>
+    new Map(
+      Object.entries(talosctlRelease.value?.release_data.available_versions ?? {}).sort(
+        ([a], [b]) => compare(a.replace('v', ''), b.replace('v', '')),
+      ),
+    ),
```
Slessi, line 47: using a map because it has guaranteed sort order and objects only kinda have order

frontend/src/views/omni/Modals/DownloadTalosctl.vue
```diff
@@ -27,46 +29,44 @@ const close = () => {
...
     defaultVersion.value &&
-    talosctlRelease.value?.release_data.available_versions[defaultVersion.value].find((item) =>
-      item.url.endsWith('linux-amd64'),
-    )?.name
-
-  selectedOption.value = defaultValue.value
-})
+    availableVersions.value
+      .get(defaultVersion.value)
+      ?.find((item) => item.url.endsWith('linux-amd64'))?.name,
```
Unix4ever, line 56: It could be cool if we could determine the OS and pick the right one automatically :)

Unix4ever (approved):

frontend/src/views/omni/Modals/DownloadTalosctl.vue
Slessi: It won't be 100% reliable but I can do something yea

https://github.com/siderolabs/omni/issues/1907

PR #1906, opened 2025-11-21
Remove vue3-popper from frontend

Slessi:
- Refactor `TActionsBox` to use `DropdownMenu` from `reka-ui`
- Replace `vue3-popper` with `Tooltip` in `TButtonGroup`
- Replace `vue3-popper` with `Tooltip` + `Popover` in `MachineSetPicker`
- Replace `vue3-popper` with `Tooltip` + `Popover` in `PatchEdit`
- Refactor `Tooltip` to use `Tooltip` from `reka-ui`

Motivation for this is that `vue3-popper` is an abandoned library which breaks our CSP if we stop allowing `'unsafe-inline'`.

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -38,7 +38,6 @@
         "vue-virtual-scroller": "^2.0.0-beta.8",
         "vue-word-highlighter": "^1.2.6",
         "vue3-apexcharts": "^1.10.0",
-        "vue3-popper": "^1.5.0",
         "whatwg-fetch": "^3.6.20"
       },
       "devDependencies": {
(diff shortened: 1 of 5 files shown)


Unix4ever:
It looks like sidebar now shows empty tooltips
https://github.com/user-attachments/assets/e8a6107c-1b79-4c59-8787-ae2ecfa5dfe4

Slessi:
@Unix4ever fixed that

Unix4ever:
Looks good to me except this unexpected change.

Before:
image

After:
image

Unix4ever (approved):

Issue #1450, opened 2025-08-20
make the automatically created `MachineSetNodes` writeable by the user


Unix4ever:
Currently, if the machine set is managed by a machine class it is not possible to edit the machine set node resource, which breaks the cases when the user:
- wants to remove the particular machine from the machine set.
- wants to lock a machine and pause all the updates.

## Solution

Create the `MachineSetNode` resources in the `MachineSetNode` controller without the owner.
This way they will be modifiable by the users.

PR #1908, opened 2025-11-21
feat: support locks,node delete and restore when using machine classes

Unix4ever:
Make `MachineSetNode` created without an owner by the `MachineSetNodeController`.

Fixes: https://github.com/siderolabs/omni/issues/1450


client/pkg/omni/resources/omni/labels.go
```diff
@@ -86,6 +86,12 @@ const (
 	// tsgen:LabelNoManualAllocation
 	LabelNoManualAllocation = SystemLabelPrefix + "no-manual-allocation"
 
+	// LabelManagedByController is set on the MachineSetNode resources which were dynamically created
+	// for a MachineSet which is using machine classes.
+	// It means that the MachineSetNode can be updated/removed by the controller too.
+	// tsgen:LabelManagedByController
+	LabelManagedByController = SystemLabelPrefix + "managed-by-controller"
```
oguzkilcan: So either label comments are too restrictive or label definition is too generic. We have other resources that are managed by user/controller as well. IMHO either this label should include them or it should only be used for `MachineSetNodes`.

Unix4ever: yep, makes sense. Will update the name of the label

internal/backend/runtime/omni/controllers/omni/machine_request_set_status_test.go
```diff
@@ -367,18 +367,10 @@ func TestInfraProviderDeletion(t *testing.T) {
...
-
-			return nil
-		})
-		require.NoError(t, err)
+		for _, id := range ids {
+			rtestutils.AssertNoResource[*infra.MachineRequest](ctx, t, st, id)
+		}
 
-		require.NoError(t, st.Destroy(ctx, machineRequestSet.Metadata()))
+		require.NoError(t, st.TeardownAndDestroy(ctx, machineRequestSet.Metadata()))
```
Unix4ever, line 374: Fix flaky test.
Error was:
```
#44 47.06     machine_request_set_status_test.go:382: 
#44 47.06         	Error Trace:	/src/internal/backend/runtime/omni/controllers/omni/machine_request_set_status_test.go:382
#44 47.06         	            				/src/internal/backend/runtime/omni/controllers/testutils/testutils.go:132
#44 47.06         	            				/src/internal/backend/runtime/omni/controllers/omni/machine_request_set_status_test.go:296
#44 47.06         	Error:      	Received unexpected error:
#44 47.06         	            	resource MachineRequestSets.omni.sidero.dev(default/test-machine-request-set@3) has pending finalizers &[MachineRequestSetStatusController]
#44 47.06         	Test:       	TestInfraProviderDeletion
```

@oguzkilcan it's better to always call `TeardownAndDestroy` or `rtestutils.Destroy`, as it removes the resource even if it has finalizers.

PR #1909, opened 2025-11-21
feat: implement helm chart v2

### LINKED ISSUE #1884: [feature] v2 Helm chart
Take some inspiration from the community PRs and create a proper Helm chart for Omni.

utkuozdemir:
**Helm Chart v2:**
- Add new Helm chart with comprehensive configuration via values.yaml
- Support for both Kubernetes Ingress and Gateway API
- Built-in validation for required fields and URL consistency
- Prometheus metrics and ServiceMonitor support
- Detailed documentation with examples for Traefik
- Workload proxy setup guide

**Deploy directory reorganization:**
- Move Docker Compose files to `deploy/compose/`
- Move existing Helm chart to `deploy/helm/omni/`
- Add top-level `deploy/README.md` pointing to deployment options
- Add deprecation warning to v1 Helm chart

**Documentation:**
- Add link to Helm chart in root README

Closes #1884.

Co-authored-by: Kevin Tijssen <kevin.tijssen@siderolabs.com>


deploy/helm/omni-v2/templates/configmap.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ include "omni-v2.fullname" . }}
+  labels:
+    {{- include "omni-v2.labels" . | nindent 4 }}
+data:
+  config.yaml: |
+    {{- .Values.config | toYamlPretty | nindent 4 }}
```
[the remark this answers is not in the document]
utkuozdemir: wrong, it exists.

deploy/helm/v2/omni/values.yaml
```diff
@@ -0,0 +1,788 @@
...
+      # Boltdb contains BoltDB storage backend configuration.
+      boltdb:
+        # Path to the primary BoltDB database.
+        # Is **NOT USED by default**: only used if the storage.default.kind is set to "boltdb".
+        # This path is mounted from the persistent volume by default.
+        path: /data/omni-boltdb.db
+      # Etcd contains etcd storage backend configuration.
+      etcd:
+        # Embedded controls whether to use embedded etcd server as the storage backend.
+        #embedded: true
```
kevintijssen: I would uncomment `embedded: true`. So by default they will use the embedded version.
If they want to use a external etcd they should set it to false

deploy/helm/v2/omni/values.yaml
```diff
@@ -0,0 +1,788 @@
...
+  # Name of an existing Secret containing the GPG private key.
+  # IMPORTANT: The existing secret MUST contain a data key named exactly "omni.asc".
+  # If your secret uses a different key, Omni will not find the file.
+  #
+  # After generating your GPG private key file (omni.asc), create the Kubernetes Secret using:
+  #
+  # kubectl create secret generic omni-key --from-file=omni.asc=./private.key
+  existingSecret: ""
+  # If 'existingSecret' is empty, a new Secret will be created using this data.
+  data:
```
kevintijssen: Remove `data:` since its obvious and set `omniAsc:` right underneath  existingSecret.

Also update the secret-etcdkey.yaml and remove `data.` there as well

deploy/helm/v2/omni/values.yaml
utkuozdemir: it's true by default (https://github.com/siderolabs/omni/blob/main/internal/pkg/config/config.go#L169), that's why I commented it out, but for this specific value, maybe it is better to have it uncommented so it is explicit and stands out.

deploy/helm/v2/omni/values.yaml
utkuozdemir: done.

deploy/helm/v2/omni/values.yaml
utkuozdemir: done.

kevintijssen: When this was commented it tried to connect to a localhost etcd address

utkuozdemir: It is because the config merging was broken below 1.5 :) You should really test with a recent commit and not do a skipVersionCheck. See the image tags here: https://github.com/siderolabs/omni/pkgs/container/omni

For example, just use `v1.5.0-beta.0-9-g883fadfe`.
In the CI, we can do something similar to:
https://github.com/siderolabs/omni-infra-provider-bare-metal/blob/main/hack/test/integration.sh#L27-L31

So that we have an Omni image with all current changes to use in the chart.

deploy/helm/v2/omni/README.md
```diff
@@ -0,0 +1,228 @@
...
+| URL                            | Purpose                |
+| ------------------------------ | ---------------------- |
+| `omni.example.com`             | UI and API             |
+| `omni-k8s.example.com`         | Kubernetes proxy       |
+| `omni-siderolink.example.com`  | SideroLink Machine API |
+
+```yaml
+# Paste the contents of your omni.asc file here
+etcdEncryptionKey:
+  data:
```
kevintijssen: `data:` was removed if I'm correct

deploy/compose/README.md
```diff
@@ -0,0 +1,20 @@
+## Omni On-Prem Compose File
+
+Follow the full documentation in the [Omni docs](https://omni.siderolabs.com/how-to-guides/self_hosted/index).
+
+The template requires the following environment variables to be set (change these for your environment):
+
+```bash
+OMNI_VERSION=v0.42.0
```
Unix4ever: We could also generate this version as part of `make generate`.

deploy/helm/v2/omni/README.md
```diff
@@ -0,0 +1,228 @@
...
+
+- `auth0` - Auth0 authentication
+- `saml` - SAML authentication
+- `oidc` - OpenID Connect authentication
+
+Additionally, at least one initial admin user must be specified in `config.auth.initialUsers`.
+
+## Example
+
+For full configuration reference, see the [config schema](https://github.com/siderolabs/omni/blob/main/internal/pkg/config/schema.json).
```
Unix4ever: I think the end goal would be to automatically generate docs page from the JSON schema and put it to the docs website.
If we don't have an issue for that, lets create it.

deploy/helm/v2/omni/Chart.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: v2
+name: omni
+description: A helm chart to deploy Omni on a Kubernetes cluster
+type: application
+version: 2.0.0
+appVersion: "v1.5.0"
```
Unix4ever, line 6: we need to generate this version in `make generate`.

deploy/helm/v2/omni/values.yaml
```diff
@@ -0,0 +1,787 @@
...
+  # Path to the configuration file (mounted via ConfigMap).
+  # Do not remove this unless you are using args only.
+  - --config-path=/config/config.yaml
+  # Enable debug logging.
+  # Uncomment this line to increase log verbosity for troubleshooting.
+  #- --debug
+
+# Omni Configuration.
+# For full configuration reference, see the schema at:
+# https://github.com/siderolabs/omni/blob/main/internal/pkg/config/schema.json
```
Unix4ever, line 55: It'd be cool to generate the config from the schema in `make generate`.
I guess we could have a template for `values.yaml`.

utkuozdemir:
@kevintijssen @Unix4ever we just did a live review with @oguzkilcan and he brought up some good points, especially regarding secrets in config. I have some ideas, I'll change all of those, then you can have another look.

deploy/helm/v2/omni/values.yaml
utkuozdemir: I tried that (also to bring the descriptions from the schema and so on, in `helm-docs` compatible), but it is too tedious. Needs lots of tricky glue code to generate this block from the schema. So, for now, I did this manually.

deploy/helm/v2/omni/README.md
utkuozdemir: Yes, I'm gonna do it eventually with `helm-docs` (as @shanduur did in image factory), but it needs some work, so will leave it for a separate PR. Gonna create the issue for that.

deploy/helm/v2/omni/README.md
utkuozdemir: yes, you are right, going to fix it.

deploy/helm/v2/omni/values.yaml
Unix4ever: we will 100% forget to keep this thing up to date :)

utkuozdemir: for sure, but I don't think it's too big of a deal. I'll look into what we can do about it, it's just not that easy.

utkuozdemir:
@kevintijssen please feel free to have a look to the new `additionalConfigSources` config. The idea is to be able to extract different parts of the config (for example, the secret values) into their own secret.

deploy/helm/v2/omni/templates/secret-config.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
```
oguzkilcan, line 1: This file name should be secret-config to match the other files. in the chart

oguzkilcan (approved):

PR #1918, opened 2025-11-25
fix(frontend): update csp for userpilot and refactor init logic

Slessi:
Update CSP again for userpilot and refactor the initialisation logic in the frontend to fix some load issues.

### DIFF
--- a/frontend/src/methods/userpilot.ts
+++ b/frontend/src/methods/userpilot.ts
@@ -1,0 +1,50 @@
+// Copyright (c) 2025 Sidero Labs, Inc.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+import { useLocalStorage } from '@vueuse/core'
+import { Userpilot } from 'userpilot'
+import { computed, ref, watch, watchEffect } from 'vue'
+
+import { getNonce } from '@/methods'
+import { currentUser } from '@/methods/auth'
+
+import { useFeatures } from './features'
+
+export function useUserpilot() {
+  const { data } = useFeatures()
+  const trackingState = useLocalStorage<boolean | null>('tracking', null)
+  const userpilotInitialised = ref(false)
+
+  const token = computed(() => data.value?.spec.user_pilot_settings?.app_token)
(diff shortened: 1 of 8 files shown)


internal/frontend/handler.go
```diff
@@ -129,12 +129,14 @@ func (handler *StaticHandler) serveFile(w http.ResponseWriter, r *http.Request,
...
+					";default-src 'self'"+
+					fmt.Sprintf(";script-src 'self' 'nonce-%s' https://*.userpilot.io", nonce)+
+					";media-src 'self' https://js.userpilot.io"+
 					";img-src * data:"+
 					";connect-src 'self' https://*.auth0.com https://*.userpilot.io wss://*.userpilot.io"+
-					";font-src 'self' data:"+
+					";font-src 'self' data: https://fonts.googleapis.com https://fonts.gstatic.com https://fonts.userpilot.io"+
 					// We are forced to use unsafe-inline for style-src due to monaco-editor https://github.com/microsoft/monaco-editor/issues/271
-					";style-src 'self' 'unsafe-inline' data: https://fonts.googleapis.com https://fonts.gstatic.com"+
+					";style-src 'self' 'unsafe-inline' data: https://fonts.googleapis.com"+
```
Slessi, line 139: I added a `script-src` policy to load remote userpilot scripts and fixed fonts loading. I double checked this with the original CSP pre-userpilot and it is as permissive as it was before, with extra looser exceptions for userpilot, so nothing should break

frontend/src/methods/userpilot.ts
```diff
@@ -0,0 +1,50 @@
...
+import { useLocalStorage } from '@vueuse/core'
+import { Userpilot } from 'userpilot'
+import { computed, ref, watch, watchEffect } from 'vue'
+
+import { getNonce } from '@/methods'
+import { currentUser } from '@/methods/auth'
+
+import { useFeatures } from './features'
+
+export function useUserpilot() {
```
Slessi, line 14: all userpilot code (except the `.reload()` on navigation) is encapsulated in here now

Unix4ever (approved):

PR #1922, opened 2025-11-26
Update image tag format in compose.yaml

danielr1996:
Otherwise `docker compose up ` fails with

```shell
 ✘ omni Error failed to resolve reference "ghcr.io/siderolabs/omni:1.3.4": ghcr.io/siderolabs/omni:1.3.4: not found                                                                   1.3s
```

### DIFF
--- a/deploy/compose.yaml
+++ b/deploy/compose.yaml
@@ -2,7 +2,7 @@ name: omni-on-prem
 services:
   omni:
     container_name: omni
-    image: "ghcr.io/siderolabs/omni:${OMNI_IMG_TAG}"
+    image: "ghcr.io/siderolabs/omni:v${OMNI_IMG_TAG}"
     devices:
       - /dev/net/tun
     volumes:


Unix4ever:
I don't think it's a good idea to change it like that now.
All docs are adding `v` prefix to the `OMNI_IMG_TAG` when you set the env var.
Changing that to prepend `v` inside the compose file will render all of them unusable.

I guess our official documentation needs to be fixed as it's using incorrect format there.

danielr1996:
Thats if course also possible, i just wanted to the sidero team to be aware that the current documentation contains issues, maybe I should have created an issue instead.

Unix4ever:
Fixed the docs here https://github.com/siderolabs/docs/pull/184

Unix4ever:
The documentation fix was merged.

danielr1996:
Thanks for fixing it, but now it must also be updated in the curl https://raw.githubusercontent.com/siderolabs/omni/v${OMNI_VERSION}/deploy/compose.yaml -o compose.yaml command, I created a pull request: https://github.com/siderolabs/docs/pull/187

PR #1923, opened 2025-11-26
refactor(frontend): wait for signing keys instead of throwing

Slessi:
Instead of throwing an error when we try to make requests before signing keys are ready, we enter a loop to wait for keys to exist. This prevents errors and the need to handle them, as well as making requests fire immediately when keys are eventually ready.

### DIFF
--- a/frontend/src/components/THeader/THeader.vue
+++ b/frontend/src/components/THeader/THeader.vue
@@ -5,13 +5,10 @@ Use of this software is governed by the Business Source License
 included in the LICENSE file.
 -->
 <script setup lang="ts">
-import { computedAsync } from '@vueuse/core'
-
 import TButton from '@/components/common/Button/TButton.vue'
 import TIcon from '@/components/common/Icon/TIcon.vue'
 import OngoingTasks from '@/components/common/OngoingTasks/OngoingTasks.vue'
 import { getDocsLink } from '@/methods'
-import { hasValidKeys } from '@/methods/key'
 
 interface Props {
   sidebarOpen?: boolean
@@ -21,9 +18,6 @@ interface Props {
 const props = defineProps<Props>()
 
 const docsLink = getDocsLink('omni')
-
-// Used to prevent sub-queries being fired before keys are created
-const authorized = computedAsync(async () => await hasValidKeys())
 </script>
 
 <template>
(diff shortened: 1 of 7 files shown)


frontend/src/components/THeader/THeader.vue
```diff
@@ -81,7 +75,7 @@ const authorized = computedAsync(async () => await hasValidKeys())
         <span class="truncate text-xs max-sm:hidden">Report an issue</span>
       </a>
 
-      <OngoingTasks v-if="authorized" />
+      <OngoingTasks />
```
Slessi, line 78: Minor side effect, now when waiting to create keys, you will see the ongoing tasks menu item, where as before you didn't. I think this was originally only being handled to not throw console errors, so unless we specifically do not want to show it at this moment, this is one less case to handle.

Image

frontend/src/methods/interceptor.ts
```diff
@@ -69,11 +69,25 @@ export function useRegisterAPIInterceptor() {
...
+      const signature = b64Encode(array, 0, array.length)
+      const fingerprint = keys.publicKeyID.value
+
+      return `${SignatureVersionV1} ${identity.value} ${fingerprint} ${signature}`
+    } catch (e) {
+      if (e instanceof KeysError && e.code === KeysErrorCode.NO_KEYS) {
+        // Enter a loop to wait for keys to be ready.
+        // Request will fire when we have keys.
+        return await new Promise<string>((resolve) => {
+          setTimeout(() => {
```
smira: this might be a dumb question, but why can't we make a function not fire an exception on no keys, and just wait on its own? 

or have two versions, one which waits and another one which throws an error?

smira: I'm trying to avoid the `setTimeout` loop here :)

Slessi: The `setTimeout` here is my version of waiting, it will essentially poll for keys every 100ms.

There are other ways to do it, could be with events & listeners for example, but I just went with the most straight forward option in front of me.

WDYT?

smira: I don't want to block the PR, I was just more curious if the Promise should not fail with "no keys" but itself wait for the keys to be establishes, avoiding the need to for polling.

Unix4ever: yeah, I'd also prefer some event handler instead

Slessi: Changed it to have no throwing and use `watch` to wait instead of `setTimeout`

PR #1933, opened 2025-11-28
feat: store machine logs in sqlite

### LINKED ISSUE #1771: machine logs in sqlite
Implement machine logs storage, retrieval and streaming from sqlite.

Provide migration from circbuf to sqlite.

utkuozdemir:
Use SQLite storage to store machine logs.
Use the same SQLite database used for the metrics state.

Refactor machine log storage to define a common interface for circular and sqlite based logs.

Closes #1771.

### DIFF
--- a/internal/pkg/config/config.go
+++ b/internal/pkg/config/config.go
@@ -159,11 +159,17 @@ func Default() *Params {
 				BufferMaxCapacity:     131072,
 				BufferSafetyGap:       256,
 				Storage: LogsMachineStorage{
-					Enabled:             true,
+					Enabled:             true, // todo: Keeping this enabled to get the logs migrated to sqlite, drop this after a while, when all logs are supposedly in SQLite.
 					Path:                "_out/logs",
 					FlushPeriod:         10 * time.Minute,
 					FlushJitter:         0.1,
 					NumCompressedChunks: 5,
+				},
+				SQLite: LogsMachineSQLite{
+					Enabled:          true,
+					Timeout:          10 * time.Second,
+					CleanupInterval:  30 * time.Minute,
+					CleanupOlderThan: 24 * 30 * time.Hour,
 				},
 			},
 		},
--- a/internal/pkg/siderolink/logstore/sqlitelog/manager.go
+++ b/internal/pkg/siderolink/logstore/sqlitelog/manager.go
@@ -1,0 +1,194 @@
+// Copyright (c) 2025 Sidero Labs, Inc.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+
+package sqlitelog
+
+import (
+	"context"
+	"database/sql"
+	"errors"
+	"fmt"
+	"strings"
+	"text/template"
+	"time"
+
+	"go.uber.org/zap"
+
+	"github.com/siderolabs/omni/internal/pkg/config"
+	"github.com/siderolabs/omni/internal/pkg/siderolink/logstore"
+)
+
+const (
+	tableName       = "machine_logs"
+	idColumn        = "id"
+	machineIDColumn = "machine_id"
+	createdAtColumn = "created_at"
+	messageColumn   = "message"
+)
+
+// StoreManager manages log stores for machines.
+type StoreManager struct {
+	db     *sql.DB
+	logger *zap.Logger
+	config config.LogsMachineSQLite
+}
+
+// Run implements the LogStoreManager interface.
+func (m *StoreManager) Run(ctx context.Context) error {
+	ticker := time.NewTicker(m.config.CleanupInterval)
+	defer ticker.Stop()
+
+	// Do the initial cleanup immediately
+	if err := m.doCleanup(ctx); err != nil {
+		return err
+	}
+
+	for {
+		select {
+		case <-ctx.Done():
+			if errors.Is(ctx.Err(), context.Canceled) {
+				return nil
+			}
+
+			return ctx.Err()
+		case <-ticker.C:
+		}
+
+		if err := m.doCleanup(ctx); err != nil {
+			m.logger.Error("failed to cleanup old logs", zap.Error(err))
+		}
+	}
+}
+
+func (m *StoreManager) doCleanup(ctx context.Context) error {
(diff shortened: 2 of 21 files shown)


internal/pkg/siderolink/logstore/sqlitelog/manager.go
```diff
@@ -0,0 +1,197 @@
...
+	} else {
+		m.logger.Info("removed logs for machine", zap.String("machine_id", id), zap.Int64("rows_affected", numRowsDeleted))
+	}
+
+	return nil
+}
+
+const schemaTmpl = `
+    CREATE TABLE IF NOT EXISTS {{.TableName}} (
+      {{.IDColumn}}        INTEGER PRIMARY KEY AUTOINCREMENT,
```
smira: https://sqlite.org/autoinc.html SQLite says 'don't use it', and instead use ROWID columns

internal/pkg/siderolink/logstore/sqlitelog/manager.go
```diff
@@ -0,0 +1,197 @@
...
+      {{.MachineIDColumn}} TEXT    NOT NULL,
+      {{.MessageColumn}}   BLOB    NOT NULL,
+      {{.CreatedAtColumn}} INTEGER NOT NULL
+    ) STRICT;
+
+    CREATE INDEX IF NOT EXISTS idx_{{.TableName}}_{{.MachineIDColumn}} 
+    ON {{.TableName}}({{.MachineIDColumn}}, {{.IDColumn}});
+
+    CREATE INDEX IF NOT EXISTS idx_{{.TableName}}_{{.CreatedAtColumn}}
```
smira: if that index is only for DELETEs, I would rather not have it

utkuozdemir: It is for the cleanup, yes.

internal/pkg/siderolink/logstore/sqlitelog/sqlitelog.go
```diff
@@ -0,0 +1,332 @@
...
+	s.subscribers = nil
+
+	return nil
+}
+
+func (s *Store) subscribe() chan logEntry {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	ch := make(chan logEntry, 128)
```
smira: we could have made this channel pure notification channel without any data, so no length needed, and the consumer can simply read new data from SQLite.

utkuozdemir: I just wanted the follow to not query sqlite when it is not needed. Do you think it's a problem?

internal/pkg/siderolink/logstore/sqlitelog/manager.go
utkuozdemir: good point, done.

internal/pkg/siderolink/logstore/sqlitelog/manager.go
utkuozdemir: dropped.

internal/pkg/siderolink/logstore/sqlitelog/sqlitelog.go
smira: I mean it adds some kind of limit of entries, a potential to drop them, and more complicated logic. 

If you just say "there's something new for you in the log, go check", the logic always reads from SQLite.

utkuozdemir: done. It made the querying logic/readline a bit more complex, though.

internal/pkg/config/config.go
```diff
@@ -159,12 +159,18 @@ func Default() *Params {
 				BufferMaxCapacity:     131072,
 				BufferSafetyGap:       256,
 				Storage: LogsMachineStorage{
-					Enabled:             true,
+					Enabled:             true, // todo: Keeping this enabled to get the logs migrated to sqlite, drop this after a while, when all logs are supposedly in SQLite.
```
smira, line 162: should we just migrate & drop the files? so it's only used as long as they exist?

internal/pkg/siderolink/logstore/sqlitelog/sqlitelog.go
```diff
@@ -0,0 +1,385 @@
...
+
+// Close implements the logstore.LogStore interface.
+func (s *Store) Close() error {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	s.closed = true
+
+	for _, ch := range s.subscribers {
+		close(ch)
```
smira, line 107: fyi, this will make all consumers wake up in a loop if they haven't had context cancel yet

I don't know if it is needed in any way (?).

utkuozdemir: A store is closed when in any of the following cases:
- Omni is shutdown
- The machine is removed

I think it makes sense in this case to tell the readers "go away, game over".

Because I check if data was actually received on the channel and return EOF if not, readers would gracefully terminate:
```go
	case _, ok := <-r.followCh:
		if !ok {
			return io.EOF
		}
```

So I think it's ok this way.

internal/pkg/config/config.go
utkuozdemir: Didn't get what you mean exactly - we already migrate and drop the old files: https://github.com/siderolabs/omni/pull/1933/files#diff-056aa9362c2bc5f1c6c022f109f18b6b071e41f339cdf809d6ca3a657d9404dcR35-R37

- Only old storage is enabled: things work as they used to do
- Old and new storage are enabled (**default from now on**): migrate old to new if only if old exists (there's a persisted circular log) and new doesn't (no logs in sqlite for that machine), drop old, use new.
- Old is disabled, new is enabled: safe to do after the migration is done.

Happening here: https://github.com/siderolabs/omni/pull/1933/files#diff-f771d4fd0d2f357e33c6da1927a017fa92172dfe2fc269507e09445befa7cffaR192-R227

utkuozdemir: (just merged this to not hold it anymore, but I can include your further suggestions in the audit logs PR or in a separate smaller PR)

Issue #1816, opened 2025-11-06
[feature] support switching tunnel mode on the machine level in omni


zgomez301:
### Problem Description

Applying a config patch with attempting to set grpc_tunnel as true on an existing cluster does not seem to work

Applying as a config patch with the following
```
machine:
  install:
    extraKernelArgs:
```
prepends the arg before the existing siderolink config 

apply a machine patch with 
```
apiVersion: v1alpha1
kind: SideroLinkConfig
apiUrl:
```

doesn't seem to apply 

### Solution

a button in the cluster home page in the feature box that allows selective enabling of the grpc_tunnel 

### Notes

It doesn't currently seem possible to enable grpc tunneling on an existing omni managed cluster without reimaging with an image that has the tunnel enabled when generated

PR #1935, opened 2025-11-28
feat: add a way to switch gRPC tunnel mode for the connected machines

Unix4ever:
Fixes: https://github.com/siderolabs/omni/issues/1816

Introduce a new command:

```
omnictl configure machine <id> --siderolink-connection=[udp|http-tunnel|auto]
```

### DIFF
--- a/internal/backend/runtime/omni/controllers/omni/machine_join_config.go
+++ b/internal/backend/runtime/omni/controllers/omni/machine_join_config.go
@@ -11,6 +11,7 @@ import (
 	"github.com/cosi-project/runtime/pkg/controller"
 	"github.com/cosi-project/runtime/pkg/controller/generic/qtransform"
 	"github.com/cosi-project/runtime/pkg/safe"
+	"github.com/cosi-project/runtime/pkg/state"
 	"go.uber.org/zap"
 
 	"github.com/siderolabs/omni/client/api/omni/specs"
@@ -45,13 +46,24 @@ func NewMachineJoinConfigController() *MachineJoinConfigController {
 					return err
 				}
 
-				joinOptions, err := siderolink.NewJoinOptions(
+				grpcTunnelConfig, err := safe.ReaderGetByID[*siderolinkres.GRPCTunnelConfig](ctx, r, machine.Metadata().ID())
+				if err != nil && !state.IsNotFoundError(err) {
+					return err
+				}
+
+				opts := []siderolink.JoinConfigOption{
 					siderolink.WithJoinToken(tokenUsage.TypedSpec().Value.TokenId),
 					siderolink.WithEventSinkPort(int(siderolinkAPIConfig.TypedSpec().Value.EventsPort)),
 					siderolink.WithLogServerPort(int(siderolinkAPIConfig.TypedSpec().Value.LogsPort)),
 					siderolink.WithMachineAPIURL(siderolinkAPIConfig.TypedSpec().Value.MachineApiAdvertisedUrl),
 					siderolink.WithMachine(machine),
-				)
+				}
+
+				if grpcTunnelConfig != nil {
+					opts = append(opts, siderolink.WithGRPCTunnel(grpcTunnelConfig.TypedSpec().Value.Enabled))
+				}
+
+				joinOptions, err := siderolink.NewJoinOptions(opts...)
 				if err != nil {
 					return err
 				}
@@ -73,6 +85,7 @@ func NewMachineJoinConfigController() *MachineJoinConfigController {
 		},
 		qtransform.WithExtraMappedInput[*siderolinkres.APIConfig](qtransform.MapperNone()),
 		qtransform.WithExtraMappedInput[*siderolinkres.JoinTokenUsage](qtransform.MapperSameID[*omni.Machine]()),
+		qtransform.WithExtraMappedInput[*siderolinkres.GRPCTunnelConfig](qtransform.MapperSameID[*omni.Machine]()),
 		qtransform.WithConcurrency(4),
 	)
 }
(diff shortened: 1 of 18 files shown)


utkuozdemir:
I think there shouldn't be a required `--id` flag, instead, we should use `args[0]` - flags kinda imply optionality, and `omnictl machine configure <id>` and `omnictl machine logs <id>` read more natural to me. I mean, `machine` commands already always operate on a single machine.

Unix4ever:
> I think there shouldn't be a required --id flag, instead, we should use args[0] - flags kinda imply optionality, and omnictl machine configure <id> and omnictl machine logs <id> read more natural to me. I mean, machine commands already always operate on a single machine.

I think cobra doesn't support positional args in the middle of command. So this kind of command is not going to work `omnictl machine configure <id> grpc-tunnel-mode <mode>`

rothgar:
Do we envision other settings for `configure`? I would think those would be "patches" to the configuration. Something like
```
omnictl patch machine $ID -p '{"spec": {"grpc-tunnel-mode": "http"}}'
```
This is the structure for kubectl patches and can work generically for objects and config

Unix4ever:
> This is the structure for kubectl patches and can work generically for objects and config

can't say I like this. Tunnel mode is not a part of the machine config, it's a separate resource. So this kind of syntax will be confusing.

rothgar:
I don't really like the syntax either but I was thinking the command order is more familiar for k8s users and the json is more flexible in case we want to patch more things in the future.

Maybe we can make the configurable patches flags instead. Something like
```
omnictl patch machine $ID --patch grpc-tunnel-mode=enabled
```
Now that I think about it, is `grpc-tunnel-mode` clear what people are doing? Should this be `--siderolink-connection=[udp,http-tunnel]` because grpc is used in multiple parts of talos it could be confusing when this is being used.

Unix4ever:
As we have config patches all over the place this `patch` in the subcommand will give wrong impression to the users.

client/pkg/omnictl/machine/configure.go
```diff
@@ -0,0 +1,106 @@
...
+	"github.com/siderolabs/omni/client/pkg/omnictl/internal/access"
+)
+
+const (
+	tunnelModeDisabled = "disabled"
+	tunnelModeEnabled  = "enabled"
+	tunnelModeAuto     = "auto"
+)
+
+// getCmd represents the get logs command.
```
oguzkilcan: Copy paste issue.

client/pkg/omnictl/machine/configure.go
```diff
@@ -0,0 +1,106 @@
...
+				if err != nil && !state.IsNotFoundError(err) {
+					return err
+				}
+
+				fmt.Printf("deleted grpc tunnel config for the machine %s\n%s", machineCmdFlags.machine, delayNotice)
+			default:
+				return fmt.Errorf("unknown mode %s, should be one of %q", mode, []string{tunnelModeDisabled, tunnelModeEnabled, tunnelModeAuto})
+			}
+
+			fmt.Printf("WARNING: the changes won't be applied until the machine is restarted")
```
oguzkilcan: Should we offer here a `--restart` flag so it's all handled in one go?

client/pkg/omnictl/machine/logs.go
```diff
@@ -0,0 +1,68 @@
...
+	"github.com/siderolabs/omni/client/pkg/omnictl/logformat"
+)
+
+var logsCmdFlags struct {
+	logFormat string
+	follow    bool
+	tailLines int32
+}
+
+// getCmd represents the get logs command.
```
oguzkilcan: Copy paste issue.

internal/backend/runtime/omni/controllers/omni/machine_join_config.go
```diff
@@ -45,13 +46,24 @@ func NewMachineJoinConfigController() *MachineJoinConfigController {
...
 					siderolink.WithMachineAPIURL(siderolinkAPIConfig.TypedSpec().Value.MachineApiAdvertisedUrl),
 					siderolink.WithMachine(machine),
-				)
+				}
+
+				if grpcTunnelConfig != nil {
+					opts = append(opts, siderolink.WithGRPCTunnel(grpcTunnelConfig.TypedSpec().Value.Enabled))
+				}
+
+				joinOptions, err := siderolink.NewJoinOptions(opts...)
```
oguzkilcan, line 66: Cluster import is also constructing joinOptions in a very similar fashion to build join config documents. 
Should we add gRPC tunnel option to there as well?

client/pkg/omnictl/user.go
```diff
@@ -4,8 +4,12 @@
...
 
-import "github.com/siderolabs/omni/client/pkg/omnictl/user"
+import (
+	"github.com/siderolabs/omni/client/pkg/omnictl/machine"
+	"github.com/siderolabs/omni/client/pkg/omnictl/user"
+)
 
 func init() {
 	RootCmd.AddCommand(user.RootCmd())
+	RootCmd.AddCommand(machine.RootCmd())
```
oguzkilcan: Why is this added here? Shouldn't be under `machine.go`?

client/pkg/omnictl/machine/configure.go
Unix4ever: restart is not always possible. For maintenance mode we can't do it

internal/backend/runtime/omni/controllers/omni/machine_join_config.go
Unix4ever: I'd rather have a separate command for that.
Now as I've changed the code you can run:

`omnictl configure machine <id1> <id2> --siderolink-connection=http-tunnel` for the imported cluster.

Unix4ever:
I've liked this syntax:

```
omnictl patch machine $ID --patch grpc-tunnel-mode=enabled
```

So I've changed the command structure again.
But made it:

```
omnictl configure machine $ID --siderolink-connection=udp
```

To avoid confusion with the config patches. And not to make the impression that you patch some `machine` resource.

PR #1938, opened 2025-12-01
chore: test omni with talos v1.12.0-beta.1

oguzkilcan:
Test omni for upcoming talos version 1.12.0-beta.1.

For #1944


smira:
> Test omni for upcoming talos version 1.12.0-beta.1.

We need not only to test, but also update everything to be ready for v1.12 :)

oguzkilcan:
> > Test omni for upcoming talos version 1.12.0-beta.1.
> 
> We need not only to test, but also update everything to be ready for v1.12 :)

That PR will be the same as this one except the changing defaults. I'll open it after confirming everything is working correctly.

PR #1947, opened 2025-12-02
feat(frontend): change default config patch for talos 1.12

### LINKED ISSUE #1943: Use new config patch format when creating a config patch for talos 1.12 machines
The current yaml default for config patches is as below:
```yml
machine:
  network:
    hostname: my-special-hostname
```

But this is not compatible with talos 1.12. We should use a default which is compatible, like:
```yml
apiVersion: v1alpha1
kind: HostnameConfig
hostname: my-special-hostname
auto: off
```

Slessi:
Change the default config patch for talos 1.12 to match the new format.

Closes #1943

### DIFF
--- a/frontend/vite.config.ts
+++ b/frontend/vite.config.ts
@@ -67,7 +67,7 @@ export default defineConfig(({ command }) => {
       "font-src 'self' data: https://fonts.googleapis.com https://fonts.gstatic.com https://fonts.userpilot.io",
       "style-src 'self' 'unsafe-inline' data: https://fonts.googleapis.com",
       'frame-src https://*.auth0.com',
-      'worker-src blob:', // Only for vite dev server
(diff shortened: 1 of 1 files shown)


frontend/src/views/omni/Clusters/Management/ClusterMachineItem.vue
```diff
@@ -194,6 +195,19 @@ const openExtensionConfig = () => {
 }
 
 const openPatchConfig = () => {
+  const isTalos112 = semver.satisfies(state.value.cluster.talosVersion!, '>=1.12', {
```
Slessi, line 198: is there a better name for this other than `isTalos112`? like, is this change "config patch v2" or something?

Unix4ever: I won't call it v2. I think what you have is good enough

PR #1950, opened 2025-12-02
fix(installation-media): correct doc links for sbc & cloud steps

Slessi:
Correct the docs links for SBC & Cloud provider steps on installation media. Also update some legacy docs links to the new format. Stories updated to reflec t the change for data being obtained from API.

### DIFF
--- a/frontend/src/methods/index.ts
+++ b/frontend/src/methods/index.ts
@@ -247,6 +247,27 @@ export const downloadMachineJoinConfig = async (
 
 type DocsType = 'talos' | 'omni' | 'k8s'
 
+/**
+ * @deprecated Update links to use getDocsLink instead where possible.
+ */
+export function getLegacyDocsLink(path?: string, options?: { talosVersion?: string }) {
+  const parts = ['https://talos.dev']
+
+  const talosVersion = options?.talosVersion || DefaultTalosVersion
+
+  const parsed = coerce(talosVersion)
+  if (!parsed) throw new Error(`Failed to parse talos version "${talosVersion}"`)
+
+  const { major, minor } = parsed
+  parts.push(`v${major}.${minor}`)
+
+  if (path) {
+    parts.push(path.replace(/^\//, ''))
+  }
+
+  return parts.join('/')
+}
+
 export function getDocsLink(
   type: 'talos',
   path?: string,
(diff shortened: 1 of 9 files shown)


frontend/src/methods/index.ts
```diff
@@ -247,6 +247,27 @@ export const downloadMachineJoinConfig = async (
 
 type DocsType = 'talos' | 'omni' | 'k8s'
 
+/**
+ * @deprecated Update links to use getDocsLink instead where possible.
+ */
+export function getLegacyDocsLink(path?: string, options?: { talosVersion?: string }) {
+  const parts = ['https://talos.dev']
```
Slessi, line 254: This function became mandatory here because now that I load doc links from the API, I have to support the old format.

I guess they should be updated in talos, and then image factory and frontend should be updated to support them.

Are there any other locations that might be using these doc links that will require updating? 🤔

frontend/src/methods/index.ts
```diff
@@ -247,6 +247,27 @@ export const downloadMachineJoinConfig = async (
 
 type DocsType = 'talos' | 'omni' | 'k8s'
 
+/**
+ * @deprecated Update links to use getDocsLink instead where possible.
+ */
+export function getLegacyDocsLink(path?: string, options?: { talosVersion?: string }) {
```
Unix4ever, line 253: can't we just use docs.siderolabs.com everywhere?

Slessi: No, see my other comment. We get the old paths from the API now. The redirects are in place but do not translate to the exact same path name, and the redirection only works from the old domain.

https://github.com/siderolabs/omni/pull/1950#discussion_r2582376392

e.g.

- we get `/talos-guides/install/cloud-platforms/aws/`
- we create https://www.talos.dev/v1.11/talos-guides/install/cloud-platforms/aws/
- redirects to https://docs.siderolabs.com/talos/v1.11/platform-specific-installations/cloud-platforms/aws
- but this doesnt work https://docs.siderolabs.com/talos/v1.11/talos-guides/install/cloud-platforms/aws/

Unix4ever: yeah, I wonder if we need to fix the links in the source to point to the right URIs

Slessi: For this PR I think it is fine like this, because changing those will require at minimum a factory update also. The redirects currently work, so there won't be anything broken here.

If there are other sources besides factory & omni that reference these links then we have to update those too, but not sure how to determine all sources that reference them.

I made an issue in the omni repo https://github.com/siderolabs/omni/issues/1956 - if you have any information to add you could add it there.

PR #1953, opened 2025-12-03
feat(installation-media): implement extra args step

Slessi:
Implement extra args step for the installation media wizard.


Slessi:
@Unix4ever renamed

PR #1955, opened 2025-12-03
fix(installation-media): adjust secureboot support check

Slessi:
Adjust the secure boot support check in the machine arch step to match how it works in factory.

### DIFF
--- a/frontend/src/api/resources.ts
+++ b/frontend/src/api/resources.ts
@@ -234,7 +234,6 @@ export const ConfigID = "siderolink-config";
 export const ConnectionParamsType = "ConnectionParams.omni.sidero.dev";
 export const DefaultJoinTokenID = "default";
 export const DefaultJoinTokenType = "DefaultJoinTokens.omni.sidero.dev";
-export const GRPCTunnelConfigType = "GRPCTunnelConfigs.omni.sidero.dev";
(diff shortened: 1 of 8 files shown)


frontend/src/api/resources.ts
```diff
@@ -234,7 +234,6 @@ export const ConfigID = "siderolink-config";
 export const ConnectionParamsType = "ConnectionParams.omni.sidero.dev";
 export const DefaultJoinTokenID = "default";
 export const DefaultJoinTokenType = "DefaultJoinTokens.omni.sidero.dev";
-export const GRPCTunnelConfigType = "GRPCTunnelConfigs.omni.sidero.dev";
```
Slessi, line 237: idk why generate removed this 🤷

internal/backend/runtime/omni/virtual/pkg/factory/configs/configs.go
```diff
@@ -124,6 +124,7 @@ func newPlatformConfig(platform platforms.Platform) *specs.PlatformConfigSpec {
 	spec.Documentation = platform.Documentation
 	spec.Architectures = platform.Architectures
 	spec.DiskImageSuffix = platform.DiskImageSuffix
+	spec.SecureBootSupported = platform.SecureBootSupported
```
Slessi, line 127: this is the only thing i added besides the proto file change, is it enough?

PR #1957, opened 2025-12-04
fix(installation-media): only list architectures supported by providers

Slessi:
Only list the architectures supported by the providers as defined in the API.

Also:
- Changed architectures from `[]string` to `[]Arch`
- Fixed a `Get` bug on `MetalPlatformConfig`


frontend/src/views/omni/InstallationMedia/Steps/MachineArch.vue
```diff
@@ -43,18 +50,38 @@ const selectedCloudProvider = computed(() =>
...
   () =>
     isGte1_5.value &&
     (formState.value.hardwareType === 'metal' ||
       selectedCloudProvider.value?.spec.secure_boot_supported),
 )
+
+const supportedArchitectures = computed(() => {
+  switch (formState.value.hardwareType) {
+    case 'sbc':
+      return [PlatformConfigSpecArch.AMD64]
```
smira: that is certainly wrong, should be arm64

Slessi: ... that is indeed certainly a typo on my part, woopsie

Slessi: fixed, nicely caught, thanks

frontend/src/views/omni/InstallationMedia/Steps/MachineArch.vue
```diff
@@ -43,18 +50,38 @@ const selectedCloudProvider = computed(() =>
   cloudProviders.value?.find((provider) => formState.value.cloudPlatform === provider.metadata.id),
 )
 
+const isGte1_5 = computed(
+  () => formState.value.talosVersion && gte(formState.value.talosVersion, '1.5.0'),
```
Unix4ever, line 54: I think we shouldn't even show Talos < 1.6.0. So we can drop these checks

Slessi: yea, i was thinking that, for now i just copy pasta'd the logic from factory, but if we arent gonna support 1.6.0 at all i _could_ just remove this check

PR #1969, opened 2025-12-08
feat: add labels to talos version metric

### LINKED ISSUE #1967: [feature] add machine status labels to the omni_machines_version metric
### Problem Description

We would like to be able to filter the machine Talos version against some machine status, specifically if the machine is connected and/or is the machine is a member of a cluster. 

### Solution

If we can add both as labels to the existing `omni_machines_version` though would be fine. Resulting in something like `omni_machines_version{talos_version=v1.11.5", connected="true", cluster_member="true"}: 12` would be great.

### Alternative Solutions

Additional metrics for the specific state would also work, such as a new `omni_cluster_machines_version` metric.

### Notes

_No response_

TimJones:
Add labels for the assigned cluster and connection status to the `omni_machines_version` metric.

Closes #1967

### DIFF
--- a/client/api/omni/specs/omni.proto
+++ b/client/api/omni/specs/omni.proto
@@ -1282,12 +1282,13 @@ message MachineExtensionsStatusSpec {
 
 // MachineStatusMetricsSpec provides aggregated state of the number of registered and connected machines for the Omni instance.
 message MachineStatusMetricsSpec {
+  reserved 5;
+
   uint32 registered_machines_count = 1;
   uint32 connected_machines_count = 2;
   uint32 allocated_machines_count = 3;
   uint32 pending_machines_count = 4;
 
-  map<string, int32> versions_map = 5;
   map<string, uint32> platforms = 6;
   map<string, uint32> secure_boot_status = 7;
   map<string, uint32> uki_status = 8;
--- a/internal/backend/runtime/omni/controllers/omni/machine_status_metrics.go
+++ b/internal/backend/runtime/omni/controllers/omni/machine_status_metrics.go
@@ -26,12 +26,18 @@ import (
 	"github.com/siderolabs/omni/client/pkg/omni/resources/omni"
 )
 
+type nodeInfo struct {
+	talosVersion string
+	cluster      string
+	connected    bool
+}
+
 // MachineStatusMetricsController provides metrics based on ClusterStatus.
 //
 //nolint:govet
 type MachineStatusMetricsController struct {
 	versionsMu  sync.Mutex
-	versionsMap map[string]int32
+	versionsMap map[nodeInfo]int32
 
 	metricsOnce sync.Once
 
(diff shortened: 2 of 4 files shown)


internal/backend/runtime/omni/controllers/omni/machine_status_metrics.go
```diff
@@ -195,7 +196,11 @@ func (ctrl *MachineStatusMetricsController) gatherMetrics(statuses iter.Seq[*omn
 		}
 
 		if ms.TypedSpec().Value.TalosVersion != "" {
-			ctrl.versionsMap[ms.TypedSpec().Value.TalosVersion]++
+			ctrl.versionsMap[strings.Join([]string{
```
Unix4ever: instead of join/split I propose to change the `versionsMap` type to:
```
type nodeInfo struct{
  talosVersion string
  cluster string
  connected bool
}

// the type
versionsMap map[nodeInfo]uint64

ctrl.versionsMap[nodeInfo{
  talosVersion: ms.TypedSpec().Value.TalosVersion,
  cluster: ms.TypedSpec().Value.Cluster,
  connected: ms.TypedSpec().Value.Connected
}]++
```

TimJones: I tried that originally, but then it was changing the API/protobuf file and I didn't think I wanted to do that.

TimJones: Specifically this would need to change:
https://github.com/siderolabs/omni/blob/1b87e58a65569a20d80c211f3e1ec213e5f3b976/client/api/omni/specs/omni.proto#L1290

Unix4ever: ah, I see.
At the same time if you encode additional fields like that in the key whatever is using this resource will have to parse it too.
I can imagine the frontend using versions statistics in the future.
Although maybe we should create a separate field for that.

Unix4ever: I don't think this field is used anywhere right now btw. So we can just drop it from the spec and use the in-memory state instead.

TimJones: OK, I will do that. Thanks for the pointers.

client/api/omni/specs/omni.proto
```diff
@@ -1282,12 +1282,14 @@ message MachineExtensionsStatusSpec {
...
+  reserved 5;
+  reserved "versions_map";
+
   uint32 registered_machines_count = 1;
   uint32 connected_machines_count = 2;
   uint32 allocated_machines_count = 3;
   uint32 pending_machines_count = 4;
 
-  map<string, int32> versions_map = 5;
   map<string, uint32> platforms = 6;
```
TimJones, line 1292: Not sure if this is the correct way, but seemed so from reading the protobuf documentation.

client/api/omni/specs/omni.proto
```diff
@@ -1282,12 +1282,14 @@ message MachineExtensionsStatusSpec {
 
 // MachineStatusMetricsSpec provides aggregated state of the number of registered and connected machines for the Omni instance.
 message MachineStatusMetricsSpec {
+  reserved 5;
+  reserved "versions_map";
```
Unix4ever: I don't think we ever add strings there. Just `reserved 5` would be enough I guess.

PR #1972, opened 2025-12-08
feat(installation-media): implement final confirmation step for wizard

### LINKED ISSUE #2001: Implement final / confirmation step of wizard

Slessi:
Implement the final confirmation step for the installation media wizard.

Closes #2001 

image

### DIFF
--- a/frontend/src/views/omni/InstallationMedia/Steps/Confirmation.vue
+++ b/frontend/src/views/omni/InstallationMedia/Steps/Confirmation.vue
@@ -33,7 +33,9 @@ import {
 } from '@/api/resources'
 import CodeBlock from '@/components/common/CodeBlock/CodeBlock.vue'
 import CopyButton from '@/components/common/CopyButton/CopyButton.vue'
+import TIcon from '@/components/common/Icon/TIcon.vue'
 import TSpinner from '@/components/common/Spinner/TSpinner.vue'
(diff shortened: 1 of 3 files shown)


frontend/src/views/omni/InstallationMedia/Steps/Confirmation.vue
```diff
@@ -5,9 +5,688 @@ Use of this software is governed by the Business Source License
...
+  )
+
+  try {
+    const { schematic_id, schematic_yml, pxe_url } = await ManagementService.CreateSchematic({
+      extensions: formState.value.systemExtensions,
+      extra_kernel_args: formState.value.cmdline?.split(' ').filter((item) => item.trim()),
+      meta_values: {
+        [LabelsMeta]: dump({ machineLabels }),
+      },
+      media_id: media.metadata.id,
```
Unix4ever, line 177: We will need to deprecate `media_id` and replace it with:
- `overlay`, which should be set to `selectedSBC.value.metadata.id`.
- `architecture` = `arch.value`.
- `profile` = `profile.value`

We will of course need to change the backend to support that. I can help with it.

Slessi: Made an issue https://github.com/siderolabs/omni/issues/2014

Issue #1874, opened 2025-11-17
[feature] enable omnictl to handle directories


gitbluf:
### Problem Description

In order to apply multiple cluster configurations, we need to run `omnictl apply -f <file>` for each one.

### Solution

Add behaviour similar to `kubectl` that can handle multiple files if directory is provided instead of a file.

Kubectl example: `kubectl apply -f manifests/`

Suggested omnictl example: `omnictl apply -f clusters/`

PR #1976, opened 2025-12-09
allow omnictl to handle directories

pranav767:
Fixes https://github.com/siderolabs/omni/issues/1874
Allow omnictl to handle directories.

### DIFF
--- a/client/pkg/omnictl/apply.go
+++ b/client/pkg/omnictl/apply.go
@@ -10,7 +10,10 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"io/fs"
 	"os"
+	"path/filepath"
+	"strings"
 
 	"github.com/cosi-project/runtime/pkg/resource"
 	"github.com/cosi-project/runtime/pkg/resource/protobuf"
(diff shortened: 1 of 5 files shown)


client/pkg/omnictl/apply.go
```diff
@@ -32,67 +35,131 @@ var applyCmdFlags struct {
...
+		return nil
+	}
+}
 
-			if state.IsNotFoundError(err) {
-				err = createResource(ctx, st, res, applyCmdFlags.options)
-				if err != nil {
-					return err
-				}
+func applyConfigFromBytes(ctx context.Context, client *client.Client, yamlRaw []byte) error {
```
pranav767, line 124: `applyConfigFromBytes` is the same as the old `applyConfig`, just with a different function signature

Now that we process multiple files in `applyConfigFiles` , we need a direct function that can be called in a loop.

Unix4ever:
I think this PR is missing important thing: there's no way to do `omnictl cluster template sync -f /folder`.
Cluster templates are more important than applying the raw Omni resources.

We can also change `omnictl apply` code to automatically fallback to `omnictl cluster template sync` if it find the cluster template YAML file.

pranav767:
@Unix4ever 
That's a pretty neat feature, to export cluster config from UI as template files - I didn't know about it!
https://docs.siderolabs.com/omni/cluster-management/export-a-cluster-template-from-a-cluster-created-in-the-ui

I believe I understand the distinction now:

`omnictl apply` is for raw Omni resources/components (low-level resources with apiVersion: omni.sidero.dev/v1alpha1)
`omnictl cluster template sync` is for cluster templates (high-level abstractions with kind: Cluster/ControlPlane/Workers)
The typical workflow being: `UI → export → template files → sync back to Omni`

Correct me if I'm wrong, but we can have 

1. Fix directory support for sync command - Make template sync work with directories like apply does
2. Make apply smart enough to auto-detect template vs resource files and route appropriately (though this might lead to conflicts/errors if a directory contains both templates & raw resources)

Unix4ever:
> The typical workflow being: `UI → export → template files → sync back to Omni`

Can also be `template files → sync back to Omni`. Can just write the template files initially :)

> Correct me if I'm wrong, but we can have
>   1. Fix directory support for sync command - Make template sync work with directories like apply does
>   2. Make apply smart enough to auto-detect template vs resource files and route appropriately (though this might lead to conflicts/errors if a directory contains both templates & raw resources)

yep, that's correct.
If the resources clash with each other then it's an operator error. I don't know how far we want to go with this smart apply mode.

But at least improving `sync` command to work with the directories might be nice.

pranav767:
Updated both sync and apply to use a `merge-first pattern` - discovering all files and processing them as a single operation with `---` separators. This follows kubectl best practices and works well with the existing `applyResources` function designed for single-file processing.

Furthermore , I was thinking about the smarter way of making things work, 
So we could have a directory which has both templates and resources in such a case our above implementation would fail, but if be made it smart to handle both apply and sync, i.e. identify the distinguishing contents within files and let apply work magic it could work like
```
$ omnictl apply -f mixed-directory/ --auto-detect
Detected 2 template files, 2 resource files
Processing templates...
Processing resources...
```
This might lead to security issues, since apply and sync were indented in a different way initially

Or we could just let the user know they have mixed files directory
```
$ omnictl apply -f mixed-directory/
Error: Mixed file types detected:
  Templates: cluster-template.yaml, extensions-template.yaml
  Resources: custom-resource.yaml, config-map.yaml

Suggestion: 
  - Use 'omnictl cluster template sync -f mixed-directory/' for templates
  - Use 'omnictl apply -f mixed-directory/' for resources  
  - Or separate into different directories

Use separate commands or directories for different file types.
```

Unix4ever:
Thanks! I'll play around to see how it works and will get back to you.

We will also need some tests: unit and integration.

Unix4ever:
I think with the current approach it's impossible to have templates managing 2 different clusters, right?
I've tried to use it this way and it tried to merge everything into a single template.

The goal of the issue was to allow multiple clusters operations.

pranav767:
You're right, Not a good approach merge won't work for either apply/sync (resources/templates).
I'm facing some trouble setting up omni with talos vm's , might need to keep files separate, and feed each file to  applyResources which does actual apply on single file.
Will test stuff out and push changes.

pranav767:
@Unix4ever , I've removed the merge approach, now it discovers each file and process each file individually.
I've tried both resources and templates on omni, 
```
 /tmp/omnictl-from-container --omniconfig /home/jinx/siderolabs/omni/omniconfig.yaml apply -f /tmp/omnictl-test/resources/ --verbose --insecure-skip-tls-verify
Processing 3 resource file(s)
Applying resources from: /tmp/omnictl-test/resources/config1.yaml
Creating resource 'test-config-1'

metadata:
    namespace: default
    type: ConfigPatches.omni.sidero.dev
    id: test-config-1
    version: undefined
    owner:
    phase: running
    created: 0001-01-01T00:00:00Z
    updated: 0001-01-01T00:00:00Z
spec:
    data: |
        machine:
          kubelet:
            extraArgs:
              rotate-server-certificates: "true"
    compresseddata: []


Applying resources from: /tmp/omnictl-test/resources/config2.yaml
Creating resource 'test-config-2'

metadata:
    namespace: default
    type: ConfigPatches.omni.sidero.dev
    id: test-config-2
    version: undefined
    owner:
    phase: running
    created: 0001-01-01T00:00:00Z
    updated: 0001-01-01T00:00:00Z
spec:
    data: |
        machine:
          time:
            servers:
              - time.cloudflare.com
    compresseddata: []


Applying resources from: /tmp/omnictl-test/resources/subdir/config3.yaml
Creating resource 'test-config-3-subdir'

metadata:
    namespace: default
    type: ConfigPatches.omni.sidero.dev
    id: test-config-3-subdir
    version: undefined
    owner:
    phase: running
    created: 0001-01-01T00:00:00Z
    updated: 0001-01-01T00:00:00Z
spec:
    data: |
        machine:
          network:
            hostname: test-machine
    compresseddata: []
```


Result:
```
~/siderolabs/omni$ /tmp/omnictl-from-container --omniconfig /home/jinx/siderolabs/omni/omniconfig.yaml get configpatches --insecure-skip-tls-verify | grep test-config
default     ConfigPatch   test-config-1          1
default     ConfigPatch   test-config-2          1
default     ConfigPatch   test-config-3-subdir   1
```


For template:
```
~/siderolabs/omni$ /tmp/omnictl-from-container --omniconfig /home/jinx/siderolabs/omni/omniconfig.yaml cluster template sync -f /tmp/omnictl-test/templates/ --verbose --dry-run --insecure-skip-tls-verify 2>&1 | head -30
Processing 2 template file(s)
Syncing template: /tmp/omnictl-test/templates/template1.yaml
* creating (dry run) Clusters.omni.sidero.dev(test-template-1)
--- /dev/null
+++ Clusters.omni.sidero.dev(default/test-template-1)
@@ -1 +1,20 @@
+metadata:
+  namespace: default
+  type: Clusters.omni.sidero.dev
+  id: test-template-1
+  version: undefined
+  owner:
+  phase: running
+  created: 2025-12-20T11:07:02+05:30
+  updated: 2025-12-20T11:07:02+05:30
+  annotations:
+    omni.sidero.dev/managed-by-cluster-templates:
+spec:
+  installimage: ""
+  kubernetesversion: 1.30.0
+  talosversion: 1.8.0
+  features:
+    enableworkloadproxy: false
+    diskencryption: false
+    useembeddeddiscoveryservice: false
+  backupconfiguration: null
* creating (dry run) MachineSets.omni.sidero.dev(test-template-1-control-planes)
--- /dev/null
+++ MachineSets.omni.sidero.dev(default/test-template-1-control-planes)
@@ -1 +1,20 @@
```

Result for templates:
```
~/siderolabs/omni$ /tmp/omnictl-from-container --omniconfig /home/jinx/siderolabs/omni/omniconfig.yaml get clusters --insecure-skip-tls-verify 2>&1
NAMESPACE   TYPE      ID                VERSION   TALOS VERSION   KUBERNETES VERSION
default     Cluster   test-template-1   12        1.8.0           1.30.0
default     Cluster   test-template-2   12        1.8.0           1.30.0
```

Unix4ever:
Looks good to me now. Thank you very much for the contribution! :+1:

pranav767:
Thanks for the review @Unix4ever :raised_hands:

PR #1983, opened 2025-12-10
feat(installation-media): add omni settings

### LINKED ISSUE #1959: [feature] Installation media Join token use name instead of vaule
### Problem Description

The join token drop down should use the name of the join token instead of the value

Image

### Solution

_No response_

### Alternative Solutions

We could also show expiration and use count to make it easier to know how much a join token is used. 

### Notes

We also need to filter out revoked join tokens. They currently show up in the drop down as available and they should be hidden.

Slessi:
- Update installation media wizard to use a join token label when selecting a token.
- Add a checkbox to enable to the wireguard gRPC tunnel over HTTP2 to the installation media wizard. 
- Add machine user labels to installation media wizard.
- Fix a bug with machine classes not properly saving the initial labels.
- Fix a bug with `TSelectList` and `TInput` sizes not respecting height set by consumers.

Closes #1959 

image

### DIFF
--- a/frontend/src/views/omni/InstallationMedia/Steps/TalosVersion.stories.ts
+++ b/frontend/src/views/omni/InstallationMedia/Steps/TalosVersion.stories.ts
@@ -5,10 +5,15 @@
 import { faker } from '@faker-js/faker'
 import { createWatchStreamHandler } from '@msw/helpers'
 import type { Meta, StoryObj } from '@storybook/vue3-vite'
+import { http, HttpResponse } from 'msw'
 
+import type { Resource } from '@/api/grpc'
+import type { GetRequest } from '@/api/omni/resources/resources.pb'
 import type { FeaturesConfigSpec, TalosVersionSpec } from '@/api/omni/specs/omni.pb'
-import type { JoinTokenStatusSpec } from '@/api/omni/specs/siderolink.pb'
+import type { JoinTokenStatusSpec, SiderolinkAPIConfigSpec } from '@/api/omni/specs/siderolink.pb'
 import {
+  APIConfigType,
+  ConfigID,
   DefaultNamespace,
   DefaultTalosVersion,
   FeaturesConfigType,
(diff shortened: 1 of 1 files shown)


frontend/src/views/omni/MachineClasses/MachineClass.vue
```diff
@@ -395,7 +396,7 @@ const submit = async () => {
       machineClass.spec.auto_provision.kernel_args = kernelArguments.value.split(' ')
     }
 
-    if (initialLabels.value.length > 0) {
+    if (Object.keys(initialLabels.value).length > 0) {
```
Slessi, line 399: this was the bug with `initialLabels`, it was typed as `any` so this slipped by, there is no `.length` on an object, so the condition was always false

frontend/src/views/omni/InstallationMedia/Steps/TalosVersion.vue
```diff
@@ -124,12 +154,39 @@ const k8sSupportMatrixDocsLink = computed(() =>
       </ul>
     </div>
 
+    <h2 class="text-base font-medium text-naturals-n14">Omni settings</h2>
+
+    <Tooltip>
+      <TCheckbox
+        v-model="formState.useGrpcTunnel"
+        :disabled="enforceGrpcTunnel"
+        label="Tunnel Omni management traffic over HTTP2"
```
smira, line 163: it should be `HTTP/2`, there's no HTTP2. 

I would make the explanation a bit different. 

"Only enable this if the network blocks UDP packets, as HTTP tunneling adds significant overhead for communications".

Slessi: Text is from https://github.com/siderolabs/omni/pull/1024 - this is just copy pasted.

Changed it in https://github.com/siderolabs/omni/pull/1998

Issue #220, opened 2024-05-08
support Talos CA rotation


smira:
Talos >= 1.7.0 supports that, so Omni could do that as well.

Talos documentation: https://www.talos.dev/v1.10/advanced/ca-rotation/

PR #1991, opened 2025-12-11
feat: add support for Talos CA rotation

oguzkilcan:
Add support for Talos CA rotation

Closes: #220


frontend/src/components/common/OngoingTasks/OngoingTasks.vue
```diff
@@ -49,7 +49,23 @@ const getCurrentVersion = (item: Resource<OngoingTaskSpec>) => {
     return item.spec.talos_upgrade.current_upgrade_version
   }
 
-  return item.spec.machine_upgrade?.schematic_id
+  if (item.spec.machine_upgrade) {
+    return item.spec.machine_upgrade.current_schematic_id
+  }
+
+  return
```
Slessi: the format before was the more concise syntax, can just replace `schematic_id` to `current_schematic_id`
```suggestion
  return item.spec.machine_upgrade?.current_schematic_id
```

frontend/src/views/cluster/Overview/components/OverviewContent.vue
```diff
@@ -158,6 +160,27 @@ const toggleUseEmbeddedDiscoveryService = async (value: boolean) => {
...
+const secretRotationStatusWatch = new Watch(secretRotationStatus)
+
+secretRotationStatusWatch.setup({
+  runtime: Runtime.Omni,
+  resource: {
+    namespace: DefaultNamespace,
+    type: ClusterSecretsRotationStatusType,
+    id: context.cluster,
+  },
+})
```
Slessi: lets stick to the simplified version of this I created:

```suggestion
const { data: secretRotationStatus } = useResourceWatch<ClusterSecretsRotationStatusSpec>(() => ({
  runtime: Runtime.Omni,
  resource: {
    namespace: DefaultNamespace,
    type: ClusterSecretsRotationStatusType,
    id: context.cluster,
  },
}))
```

> [!NOTE]
> IDK if `context.cluster` can change here but as a potentially reactive value i also put the watch params behind a getter function

frontend/src/components/common/OngoingTasks/OngoingTasks.vue
```diff
@@ -49,7 +49,23 @@ const getCurrentVersion = (item: Resource<OngoingTaskSpec>) => {
...
+}
+
+const getCurrentComponent = (item: Resource<OngoingTaskSpec>) => {
+  if (item.spec.secret_rotation) {
+    switch (item.spec.secret_rotation.component) {
+      case 1:
```
Slessi: Using `case ClusterSecretsRotationStatusSpecComponent.TALOS_CA:` would be more clear

frontend/src/views/cluster/Overview/components/OverviewContent.vue
```diff
@@ -158,6 +160,27 @@ const toggleUseEmbeddedDiscoveryService = async (value: boolean) => {
   await setUseEmbeddedDiscoveryService(context.cluster ?? '', value)
 }
 
+const getComponentInRotation = computed(() => {
+  switch (secretRotationStatus.value?.spec.component) {
+    case 1:
```
Slessi: also here

`case ClusterSecretsRotationStatusSpecComponent.TALOS_CA:`

frontend/src/views/cluster/Overview/components/OverviewContent.vue
```diff
@@ -347,6 +374,30 @@ onMounted(async () => {
...
+              v-if="clusterLocked || machineLockedForSecretRotation"
+              icon="pause-circle"
+              class="h-6 w-6"
+            />
+            <TIcon v-else icon="loading" class="h-6 w-6 animate-spin text-yellow-y1" />
+            <div class="flex-1">
+              {{ secretRotationStatus.spec.status }}
+              <template v-if="secretRotationStatus.spec.status && !clusterLocked">
+                - {{ secretRotationStatus.spec.step }}
+              </template>
```
Slessi: is the `v-if` correct? or should it check for `spec.step`?

in either case might get simpler code if you re-order the ifs from:
```
if (A && !B) else if (B)
```
to:
```
if (B) else if (A)
```

internal/backend/grpc/management.go
```diff
@@ -298,6 +298,7 @@ func (s *managementServer) markClusterAsTainted(ctx context.Context, name string
 		omnires.NewClusterStatus(resources.DefaultNamespace, name).Metadata(),
 		func(res *omnires.ClusterStatus) error {
 			res.Metadata().Labels().Set(omnires.LabelClusterTaintedByBreakGlass, "")
+			res.Metadata().Annotations().Set(omnires.TaintedByBreakGlassTimestamp, strconv.FormatInt(time.Now().Unix(), 10))
```
utkuozdemir, line 301: what happens to the clusters which were tainted before? they won't have this timestamp.

internal/backend/runtime/omni/controllers/omni/internal/secretrotation/secretrotation.go
```diff
@@ -0,0 +1,238 @@
...
+			}
+		}
+	}
+
+	if len(cp) > 0 {
+		return cp
+	}
+
+	return w
+}
```
utkuozdemir, line 91: This is a bit unclear to me. This does something beyond filtering, there's business logic in it, but is not reflected in the function. I feel like this type can be reworked (maybe removed even) for simplicity.

internal/backend/grpc/management.go
oguzkilcan: The idea is to remove label when both Talos and Kubernetes CAs are rotated. Clusters that were tainted before won't have this yes but there is also no way to rotate their CAs. 
With CA rotation being implemented here, we mark both ClusterStatus and ClusterSecret resources with timestamps. If the timestamps on ClusterSecret(which will be set when CA is rotated) is more recent than the timestamp on ClusterStatus, it's ok to remove the label. This way timestamps on ClusterStatus having empty value is not important(can be considered unix epoch time). We'd keep the Label, if the timestamp here is more recent than the one on ClusterSecret.

internal/backend/runtime/omni/controllers/omni/secrets/secrets.go
```diff
@@ -195,3 +225,86 @@ func getBackupDataFromBootstrapSpec(ctx context.Context, r controller.Reader, et
...
+	rotationStatusUntyped, err := uncachedReader.GetUncached(ctx, omni.NewClusterSecretsRotationStatus(secrets.Metadata().ID()).Metadata())
+	if err != nil {
+		return fmt.Errorf("error getting cluster secrets rotation status: %w", err)
+	}
+
+	rotationStatus, ok := rotationStatusUntyped.(*omni.ClusterSecretsRotationStatus)
+	if !ok {
+		if state.IsNotFoundError(err) { // need to wait for the secret rotation status to be created
+			return xerrors.NewTagged[qtransform.SkipReconcileTag](err)
+		}
```
utkuozdemir, line 263: I think this probably needs to be moved up

internal/backend/runtime/omni/controllers/omni/internal/secretrotation/secretrotation.go
oguzkilcan: I'll add a comment explaining the `why`

internal/backend/runtime/omni/controllers/omni/secrets/rotation_status.go
```diff
@@ -0,0 +1,407 @@
...
+			UnmapMetadataFunc: func(rotationStatus *omni.ClusterSecretsRotationStatus) *omni.ClusterSecrets {
+				return omni.NewClusterSecrets(resources.DefaultNamespace, rotationStatus.Metadata().ID())
+			},
+			TransformExtraOutputFunc:        ctrl.reconcileRunning,
+			FinalizerRemovalExtraOutputFunc: ctrl.reconcileTearingDown,
+		},
+		qtransform.WithExtraMappedInput[*omni.ClusterMachineStatus](
+			mappers.MapByClusterLabel[*omni.ClusterSecrets](),
+		),
+		qtransform.WithExtraMappedInput[*omni.MachineStatus](
```
Unix4ever: I'd recommend using `system.ResourceLabels[*omni.MachineStatus` instead, as you don't use anything other than labels there.

Unix4ever: Ah wait, it's not used anywhere actually. I guess we don't need this input.

internal/backend/runtime/omni/controllers/omni/secrets/rotation_status.go
```diff
@@ -0,0 +1,407 @@
...
+	}
+
+	if clusterSecrets.TypedSpec().Value.RotationPhase == specs.ClusterSecretsRotationStatusSpec_OK {
+		rotationStatus.TypedSpec().Value.Status = ""
+		rotationStatus.TypedSpec().Value.Step = ""
+		rotationStatus.TypedSpec().Value.Error = ""
+
+		for _, candidate := range rotationsToUpdate.Candidates {
+			secretRotation, ok := secretRotationsMap[candidate.MachineID]
+			if !ok || (secretRotation != nil && secretRotation.TypedSpec().Value.Phase == clusterSecrets.TypedSpec().Value.RotationPhase) {
```
Unix4ever, line 143: This is a bit hard to understand, can we please add a comment here?

internal/backend/runtime/omni/controllers/omni/secrets/rotation_status.go
oguzkilcan: It's used in the internal package `secretrotation` where we are building the talos client for validation.
```	machineStatus, err := safe.ReaderGetByID[*omni.MachineStatus](ctx, r, machineID)
	if err != nil {
		if state.IsNotFoundError(err) {
			return nil, xerrors.NewTagged[qtransform.SkipReconcileTag](err)
		}

		return nil, fmt.Errorf("failed to get machine status for machine %q: %w", machineID, err)
	}

	address := machineStatus.TypedSpec().Value.ManagementAddress```

File: internal/backend/runtime/omni/controllers/omni/secrets/rotation_status.go [hunk 11]
```diff
@@ -0,0 +1,407 @@
...
+		if !destroyed {
+			allDestroyed = false
+		}
+	}
+
+	if !allDestroyed {
+		return xerrors.NewTagged[qtransform.SkipReconcileTag](fmt.Errorf("waiting for all cluster machine statuses to be deleted before destroying machine secrets rotations resources"))
+	}
+
+	return nil
```
Unix4ever, line 253: can't we just do?

```go
allDestroyed, err := helpers.TeardownAndDestroyAll(ctx, r, cmStatuses.Pointers())
if !allDestroyed {
  return xerrors.NewTagged[qtransform.SkipReconcileTag](fmt.Errorf("waiting for all cluster machine statuses to be deleted before destroying machine secrets rotations resources"))
}

for res := range cmStatuses.Pointers() {
  if err := r.RemoveFinalizer(ctx, omni.NewClusterMachineStatus(resources.DefaultNamespace, res.Metadata().ID()).Metadata(), ctrl.Name()); err != nil && !state.IsNotFoundError(err) {
    return err
  }
}

return nil
```

I know that it will remove all cluster machine status finalizers in the end, when all resources are deleted. Does it slow down something, or does something get stuck?

internal/backend/runtime/omni/controllers/omni/secrets/rotation_status.go
```diff
@@ -0,0 +1,407 @@
...
+	candidate := secretrotation.Candidate{
+		MachineID:    machineStatus.Metadata().ID(),
+		Hostname:     hostname,
+		ControlPlane: isControlPlane,
+		Blocked:      locked,
+		Ready:        machineStatus.TypedSpec().Value.Ready,
+	}
+
+	// ClusterMachineStatus is being deleted, delete the corresponding ClusterMachineSecretsRotation
+	if machineStatus.Metadata().Phase() == resource.PhaseTearingDown {
```
Unix4ever: we should rename the variable too, to avoid confusion

internal/backend/runtime/omni/controllers/omni/secrets/rotation_status.go
```diff
@@ -0,0 +1,407 @@
...
+	clusterSecrets *omni.ClusterSecrets,
+	rotationStatus *omni.ClusterSecretsRotationStatus,
+	rotationsToUpdate *secretrotation.Candidates,
+	cmStatusesMap map[resource.ID]*omni.ClusterMachineStatus,
+	cmStatuses safe.List[*omni.ClusterMachineStatus],
+) error {
+	rotationToUpdate := rotationsToUpdate.Candidates[0]
+
+	if !rotationToUpdate.Ready {
+		logger.Warn("Waiting for machine to become ready", zap.String("machine", rotationToUpdate.MachineID))
```
Unix4ever: should start with lowercase letter. I think we usually use lowercase everywhere

internal/backend/runtime/omni/controllers/omni/secrets/rotation_status.go
Unix4ever: ah, I see.

`ClusterMachineStatus` also has `ManagementAddress` IIRC.

internal/backend/runtime/omni/controllers/omni/secrets/rotation_status.go
oguzkilcan: Can you please elaborate?

oguzkilcan: This would block `ClusterMachineStatuses` from being deleted before ALL of the `ClusterMachineSecretRotations `are torn down and destroyed. It'd work with that limitation. Not sure if we should go this direction.

Unix4ever: > Not sure if we should go this direction.

I think it should be fine, but I'm not insisting. The change I proposed doesn't save too much anyways.

internal/backend/runtime/omni/controllers/omni/secrets/secrets.go
```diff
@@ -195,3 +225,94 @@ func getBackupDataFromBootstrapSpec(ctx context.Context, r controller.Reader, et
 
 	return downloadedBackupData, nil
 }
+
+func (s *SecretsController) handleCARotation(
```
Unix4ever, line 251: I've got a thought: what if we put this swapping cert logic into the `rotation_status` controller?
Secrets will be kept clean this way --will just read whatever is set as primary cert and secondary from the rotation status.

Unix4ever: Otherwise the logic gets scattered across two controllers.

internal/pkg/siderolink/trustd/server.go
```diff
@@ -89,11 +92,29 @@ func (h *handler) Certificate(ctx context.Context, in *securityapi.CertificateRe
...
+	if clusterSecrets.TypedSpec().Value.RotationPhase != specs.ClusterSecretsRotationStatusSpec_OK {
+		var rotateSecretsBundle *secrets.Bundle
+
+		rotateSecretsBundle, err = omni.ToSecretsBundle(clusterSecrets.TypedSpec().Value.GetRotateData())
+		if err != nil {
+			return nil, err
+		}
+
+		acceptedCAs = append(acceptedCAs, &x509.PEMEncodedCertificate{Crt: rotateSecretsBundle.Certs.OS.Crt})
+	}
```
Unix4ever, line 116: I'll start with easy ones. I would propose to build the list of accepted CAs in the `Secrets` controller.

```suggestion
	secretsBundle, err := getSecretsBundle(ctx, h.state, tcpAddr.IP.String())
	if err != nil {
		return err
	}
	
	acceptedCAs := clusterSecrets.TypedSpec().Value.AcceptedCAs
```

oguzkilcan: We'd need acceptedCAs for each component. In this case this is CAs for Talos.

Unix4ever: I think we can also combine `acceptedCAs` into a single `[]byte` slice to avoid doing join in all the places.

Unix4ever: well, at least we can reuse it here I guess: https://github.com/siderolabs/omni/pull/1991/changes#diff-6d353c79a6b1e8967f874ccdbdd3d8807c570e56aa4166a658d23378a5649530R88

internal/pkg/certs/certs.go
```diff
@@ -62,15 +65,44 @@ func IsPEMEncodedCertificateStale(certPEM []byte, expectedValidity time.Duration
...
+
+	return clientCert, bytes.Join(
+		xslices.Map(
+			acceptedCAs,
+			func(cert *talosx509.PEMEncodedCertificate) []byte {
+				return cert.Crt
+			},
+		),
+		nil,
+	), nil
```
Unix4ever, line 107: Here we can probably reuse the generated `AcceptedCAs`.
And also we can set the active `MachineCA` in the `Secrets` controller.

```suggestion
	clientCert, err := talossecrets.NewAdminCertificateAndKey(time.Now(), secrets.TypedSpec().Value.MachineCA, roles, certificateValidity)
	if err != nil {
		return nil, nil, fmt.Errorf("error generating Talos API certificate: %w", err)
	}

	acceptedCAs := secrets.TypedSpec().Value.AcceptedCAs

	return clientCert, acceptedCAs, nil
```

oguzkilcan:
https://github.com/siderolabs/omni/pull/2198

PR #1994, opened 2025-12-11
feat: return schematic yml when creating installation media

Slessi:
Return the final schematic YML to display in the frontend when creating installation media

### DIFF
--- a/client/api/omni/management/management.proto
+++ b/client/api/omni/management/management.proto
@@ -145,6 +145,7 @@ message CreateSchematicResponse {
   string schematic_id = 1;
   string pxe_url = 2;
   bool grpc_tunnel_enabled = 3;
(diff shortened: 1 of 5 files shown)


client/api/omni/management/management.proto
```diff
@@ -145,6 +145,7 @@ message CreateSchematicResponse {
   string schematic_id = 1;
   string pxe_url = 2;
   bool grpc_tunnel_enabled = 3;
+  string schematic_yml = 4;
```
Unix4ever, line 148: I'd probably call it `schematic_data`

Slessi: i think its a bit vague if its a `string` called data, like this it is immediatley clear what it is

Issue #1993, opened 2025-12-11
[feature] Support "minimum_host_commit" metadata from Stripe


steverfrancis:
### Problem Description

Currently, when we sell a subscription to Omni, we like to have the customer commit to a minimum usage. Often, the amount the customer plans to consume is unique (such as 28 hosts.)

Currently this requires creating a specific product in Stripe with a minimum of 28 hosts, otherwise, while they are scaling up, they may start with 15 hosts. This usage will be reported to Stripe, and if they are on a standard Omni Stripe product, they would be charged for 15 hosts, not their commit level. (They may have qualified for a discount based on their commit level - so we have to charge them at least their commit.)

So it is annoying and tedious to have to create Stripe products for every host level, for every tier.

### Solution

If we add Stripe metadata to a subscription of "minimum_host_commit", then the host usage reporting logic should retrieve this metadata value from Stripe.
If the value is set, then if the host usage is < "minimum_host_commit" , report "minimum_host_commit" as the usage. If host usage is greater, then report actual usage.

The value of "minimum_host_commit"  should be dynamically (at least daily) read from Stripe so that it can be changed (as customers grow and their commit grows.) 

Mentioning @rsmitty as we discussed this briefly.

PR #1996, opened 2025-12-11
feat: add min-commit flag for omni

rsmitty:
This PR would add a new flag for minimum committed machines. This data would come from stripe and if the user's Omni environment has less that the committed machines, we'd just report the minimum specified.


rsmitty:
closes: https://github.com/siderolabs/omni/issues/1993

PR #2012, opened 2025-12-15
chore: support resource list based filtering in the `DependencyGraph`

Unix4ever:
This will allow seing what controllers are using the defined resources.

### DIFF
--- a/client/api/omni/resources/resources.proto
+++ b/client/api/omni/resources/resources.proto
@@ -107,6 +107,7 @@ message ControllersResponse {
 message DependencyGraphRequest {
   repeated string controllers = 1;
   bool show_destroy_ready = 2;
+  repeated string resources = 3;
 }
 
 message DependencyGraphResponse {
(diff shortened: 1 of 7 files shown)


client/api/omni/resources/resources.proto
```diff
@@ -106,7 +106,8 @@ message ControllersResponse {
 
 message DependencyGraphRequest {
   repeated string controllers = 1;
-  bool show_destroy_ready = 2;
+  repeated string resources = 2;
```
oguzkilcan: Doesn't this break backwards compatibility?

frontend/src/api/omni/specs/omni.pb.ts
```diff
@@ -699,6 +699,7 @@ export type UserPilotSettings = {
 
 export type StripeSettings = {
   enabled?: boolean
+  min_commit?: number
```
oguzkilcan, line 702: We definitely need to add dirty check. Do we need to include this on 1.4 release?

Unix4ever: Would be nice

PR #2015, opened 2025-12-15
chore: add new fields to the `CreateSchematic` Omni API

### LINKED ISSUE #2014: Deprecate media_id for CreateSchematic
In `ManagementService->CreateSchematic` we expect a `media_id` which needs to be deprecated in favour of `overlay`, `architecture` and `profile`.

See https://github.com/siderolabs/omni/pull/1972#discussion_r2619824482

Unix4ever:
Now it's possible to pass the `overlay` ID directly to the request. `MediaId` is also still supported, but is there only for the backward compatibility.
`InstallationMedia` resources will be used only in the `omnictl download`.

Updated the Wizard UI to no longer use `InstallationMedia` resources. Dropped `pxe_url` from the `CreateSchematic` response, as all required arguments are now on the client side (if not using `InstallationMedia` resources).

Closes #2014

### DIFF
--- a/client/pkg/omnictl/download.go
+++ b/client/pkg/omnictl/download.go
@@ -21,6 +21,7 @@ import (
 	"github.com/siderolabs/go-api-signature/pkg/message"
 	pgpclient "github.com/siderolabs/go-api-signature/pkg/pgp/client"
 	"github.com/siderolabs/go-api-signature/pkg/serviceaccount"
+	"github.com/siderolabs/talos/pkg/machinery/imager/quirks"
 	"github.com/spf13/cobra"
 	"go.yaml.in/yaml/v4"
 
@@ -238,6 +239,7 @@ func createSchematic(ctx context.Context, client *client.Client, media *omni.Ins
 	return resp, nil
 }
 
+//nolint:gocyclo,gocyclop,cyclop
 func downloadImageTo(ctx context.Context, client *client.Client, media *omni.InstallationMedia, output string, grpcTunnelMode management.CreateSchematicRequest_SiderolinkGRPCTunnelMode) error {
 	schematicResp, err := createSchematic(ctx, client, media, grpcTunnelMode)
 	if err != nil {
@@ -256,7 +258,25 @@ func downloadImageTo(ctx context.Context, client *client.Client, media *omni.Ins
 	}
 
 	if downloadCmdFlags.pxe {
-		fmt.Println(schematicResp.PxeUrl)
+		var features *omni.FeaturesConfig
+
+		features, err = safe.ReaderGetByID[*omni.FeaturesConfig](ctx, client.Omni().State(), omni.FeaturesConfigID)
+		if err != nil {
+			return err
+		}
+
+		supportsOverlays := quirks.New(downloadCmdFlags.talosVersion).SupportsOverlay()
+
+		filename := media.TypedSpec().Value.GenerateFilename(!supportsOverlays, downloadCmdFlags.secureBoot, false)
+
+		var pxeURL *url.URL
+
+		pxeURL, err = url.Parse(features.TypedSpec().Value.ImageFactoryPxeBaseUrl)
+		if err != nil {
+			return err
+		}
+
+		fmt.Println(pxeURL.JoinPath("pxe", schematicResp.SchematicId, downloadCmdFlags.talosVersion, filename).String())
 
 		return nil
 	}
--- a/frontend/src/views/omni/InstallationMedia/Steps/Confirmation.vue
+++ b/frontend/src/views/omni/InstallationMedia/Steps/Confirmation.vue
@@ -12,10 +12,10 @@ import { computed, ref } from 'vue'
 
 import { Runtime } from '@/api/common/omni.pb'
 import {
+  type CreateSchematicRequestOverlay,
   CreateSchematicRequestSiderolinkGRPCTunnelMode,
   ManagementService,
 } from '@/api/omni/management/management.pb'
-import type { InstallationMediaSpec } from '@/api/omni/specs/omni.pb'
 import {
   type PlatformConfigSpec,
   PlatformConfigSpecArch,
@@ -24,8 +24,6 @@ import {
 } from '@/api/omni/specs/virtual.pb'
 import {
   CloudPlatformConfigType,
-  EphemeralNamespace,
-  InstallationMediaType,
   LabelsMeta,
   MetalPlatformConfigType,
   SBCConfigType,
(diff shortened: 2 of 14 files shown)


frontend/src/views/omni/Modals/DownloadInstallationMedia.vue
```diff
@@ -50,6 +50,60 @@ import { showError, showSuccess } from '@/notification'
...
+const supportsOverlay = (talosVersion?: string) =>
+  !!talosVersion && semver.gte(talosVersion, '1.7.0')
+
+const getImageFactoryPXEBaseURL = (baseURL: string): string => {
+  const url = new URL(baseURL)
+
+  url.host = `pxe.${url.host}`
+
+  return url.toString()
+}
```
Slessi: is this always true though? even if factory is configured to be something other than `factory.talos.dev`? should we return it as part of the features spec from backend like we do the factory URL?

client/pkg/omnictl/download.go
```diff
@@ -621,3 +641,14 @@ func checkPath(path string) (bool, error) {
...
+func getImageFactoryPXEBaseURL(baseURL string) (*url.URL, error) {
+	url, err := url.Parse(baseURL)
+	if err != nil {
+		return nil, fmt.Errorf("invalid URL specified for the image factory: %w", err)
+	}
+
+	url.Host = fmt.Sprintf("pxe.%s", url.Host)
+
+	return url, nil
+}
```
Slessi: cant we use `ImageFactoryPXEBaseURL` inside backend?

frontend/src/views/omni/InstallationMedia/Steps/Confirmation.vue
```diff
@@ -168,13 +135,13 @@ const schematic = computedAsync(async () => {
...
   try {
-    const { schematic_id, schematic_yml, pxe_url } = await ManagementService.CreateSchematic({
+    const { schematic_id, schematic_yml } = await ManagementService.CreateSchematic({
       extensions: formState.value.systemExtensions,
       extra_kernel_args: formState.value.cmdline?.split(' ').filter((item) => item.trim()),
       meta_values: {
         [LabelsMeta]: dump({ machineLabels }),
       },
-      media_id: media.metadata.id,
+      overlay: formState.value.overlayOptions,
```
Slessi: we spoke on slack, this one is a mistake, should be an `overlayID`

frontend/src/views/omni/InstallationMedia/Steps/Confirmation.vue
```diff
@@ -231,33 +199,36 @@ const sbcDiskImagePath = computed(() => {
...
+  const url = new URL(factoryUrl.value)
+
+  url.host = `pxe.${url.host}`
+
+  return `${url.toString()}pxe`
+})
+
+const pxeBootURL = computed(
+  () =>
+    `${pxeBootBaseUrl.value}/${schematic.value?.id}/${formState.value.talosVersion}/${selectedPlatform.value?.metadata.id}-${arch.value}${secureBootSuffix.value}`,
```
Slessi: should do a check for `selectedPlatform.value` i think like the others, if we dont have a platform, better to be `undefined` than a bad link

frontend/src/views/omni/InstallationMedia/Steps/Confirmation.vue
```diff
@@ -231,33 +199,36 @@ const sbcDiskImagePath = computed(() => {
...
+    ? `${imageBaseURL.value}/${selectedPlatform.value.metadata.id}-${arch.value}${secureBootSuffix.value}.${selectedPlatform.value.spec.disk_image_suffix}`
     : undefined,
 )
 
 const qcow2DiskImagePath = computed(() =>
   selectedPlatform.value
-    ? `${imageBaseURL.value}/${selectedPlatform.value.metadata.id}-${arch.value}.qcow2`
+    ? `${imageBaseURL.value}/${selectedPlatform.value.metadata.id}-${arch.value}${secureBootSuffix.value}.qcow2`
     : undefined,
 )
```
Slessi, line 233: as far as i saw in factory there is no secureboot qcow2 option

frontend/src/views/omni/InstallationMedia/InstallationMediaCreate.vue
```diff
@@ -32,6 +32,7 @@ export interface FormState {
   secureBoot?: boolean
   cloudPlatform?: string
   sbcType?: string
+  overlayName?: string
```
Slessi: not used?

client/pkg/omnictl/download.go
Unix4ever: Oh we have it there 🤔 Yeah, maybe it makes sense to put it everywhere

internal/pkg/features/features.go
```diff
@@ -34,6 +34,13 @@ func UpdateResources(ctx context.Context, st state.State, logger *zap.Logger) er
 
 		res.TypedSpec().Value.AuditLogEnabled = config.Config.Logs.Audit.Path != "" //nolint:staticcheck
 		res.TypedSpec().Value.ImageFactoryBaseUrl = config.Config.Registries.ImageFactoryBaseURL
+
+		imageFactoryPXEBaseURL, err := config.Config.GetImageFactoryPXEBaseURL()
+		if err != nil {
+			return err
+		}
```
Slessi, line 41: How come the PXE one needs this getter? Both URLs seem to exist on `Registries` 🤔 

```go
res.TypedSpec().Value.ImageFactoryBaseUrl = config.Config.Registries.ImageFactoryBaseURL
res.TypedSpec().Value.ImageFactoryPxeBaseUrl = config.Config.Registries.ImageFactoryPXEBaseURL
```

frontend/src/views/omni/InstallationMedia/Steps/Confirmation.vue
```diff
@@ -490,7 +462,7 @@ function shortVersion(version?: string) {
         following installer image to the machine configuration:
       </p>
 
-      <CodeBlock v-if="formState.secureBoot" :code="secureBootInstallerImage" />
+      <CodeBlock v-if="formState.secureBoot" :code="installerImage" />
       <CodeBlock v-else :code="installerImage" />
```
Slessi: no longer need the if/else if the URL is the same

Slessi (approved):
LGTM 👍 left some small comments

internal/pkg/features/features.go
Unix4ever: Because of this https://github.com/siderolabs/omni/blob/2c5979cf8395b28026e0707cbe8e8e73aaf0051b/internal/pkg/config/config.go#L402-L405

If PXE boot URL is not set explicitly it's computed from the ImageFactoryBaseURL

PR #2019, opened 2025-12-16
feat(installation-media): add download progress and omni specific filenames to images

### LINKED ISSUE #2013: Generate a customised filename for images downloaded from the omni UI
In the final step of the installation media wizard we need to change the implementation to get customised filename that includes `omni`, the account/instance name, and talos version.

e.g.: `<platform>-<board>-<arch>-omni-<account-name>-<talos-version>.<ext>`

Slessi:
When downloading images in the installation media wizard, we will send the request directly to factory instead of proxying through omni. This will require image-factory enabling CORS. The custom filename will be constructed on the frontend.

Closes #2013

> [!IMPORTANT]
> Requires:
> - siderolabs/image-factory/pull/342

image

### DIFF
--- a/frontend/src/methods/useDownloadImage.ts
+++ b/frontend/src/methods/useDownloadImage.ts
@@ -1,0 +1,85 @@
+// Copyright (c) 2026 Sidero Labs, Inc.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+import { onUnmounted, ref } from 'vue'
+
+import { useFeatures } from '@/methods/features'
+
+export function useDownloadImage() {
+  const { data: features } = useFeatures()
+
+  const abortController = ref<AbortController>()
+  const isGenerating = ref(false)
+
+  async function abort() {
+    abortController.value?.abort()
+  }
+
+  async function download(imageUrl: string) {
+    try {
+      abort()
+
+      abortController.value = new AbortController()
+      isGenerating.value = true
+
+      const headUrl = new URL(imageUrl, features.value?.spec.image_factory_base_url)
+
+      const headResponse = await fetch(headUrl, {
+        method: 'HEAD',
+        signal: abortController.value.signal,
+        headers: new Headers({
+          'Cache-Control': 'no-store',
+        }),
+      })
+
+      if (!headResponse.ok) {
+        throw new Error(`request failed: ${headResponse.status} ${headResponse.statusText}`)
+      }
+
+      // URL format is hostname/image/:schematicID/:talosVersion/:image
+      const [, , , talosVersion, imageName] = headUrl.pathname.split('/')
+
+      if (!talosVersion || !imageName) {
+        throw new Error(
+          `invalid image path "${headUrl.pathname}". Got Talos version "${talosVersion}" and image name "${imageName}".`,
+        )
+      }
+
+      const accountName = features.value?.spec.account?.name ?? 'default'
+
+      const downloadUrl = new URL(headUrl)
+      downloadUrl.searchParams.set('filename', `omni-${accountName}-${talosVersion}-${imageName}`)
+
+      downloadFile(downloadUrl)
+    } catch (e) {
+      // Ignore abort errors
+      if (abortController.value?.signal.aborted) return
+
+      throw e
+    } finally {
+      isGenerating.value = false
+    }
+  }
+
+  onUnmounted(() => {
+    abort()
+  })
+
+  return {
+    isGenerating,
+    abort,
+    download,
+  }
+}
+
+async function downloadFile(url: URL | string) {
+  const a = document.createElement('a')
+  a.href = url.toString()
+  a.target = '_blank'
+
+  document.body.appendChild(a)
+
+  a.click()
+  a.remove()
+}
--- a/frontend/src/views/omni/InstallationMedia/Steps/Confirmation.vue
+++ b/frontend/src/views/omni/InstallationMedia/Steps/Confirmation.vue
@@ -5,8 +5,9 @@ Use of this software is governed by the Business Source License
 included in the LICENSE file.
 -->
 <script setup lang="ts">
+import { useEventListener } from '@vueuse/core'
 import { gte } from 'semver'
-import { computed } from 'vue'
+import { computed, useTemplateRef } from 'vue'
 
 import { Runtime } from '@/api/common/omni.pb'
 import {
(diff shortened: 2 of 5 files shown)


frontend/src/views/omni/InstallationMedia/Steps/Confirmation.vue
```diff
@@ -72,6 +74,26 @@ const arch = computed(() => {
...
+    downloadImageDialog.value?.showModal()
+    await _downloadImage(url)
+  } finally {
+    downloadImageDialog.value?.close()
+  }
+}
+
+watch(downloadImageError, (error) => {
+  if (error) showError('Download image failed', error.message)
+})
```
Slessi: can be shown as a non-toast also, just not sure where to put it

Unix4ever (commented):
maybe we should also show `x` in the downloading modal to be able to abort the download

internal/backend/factory/factory.go
```diff
@@ -81,29 +68,29 @@ func (handler *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
-	if err != nil {
-		if errors.Is(err, errNotFound) {
-			httpNotFound(w)
+	path := r.URL.Path
 
-			return
-		}
-
-		handler.handleError("failed to parse request", w, err)
+	proxyURL, err := getProxyURL(handler.config.ImageFactoryBaseURL, path)
```
Slessi: This change is to forward all requests to image factory as-is (we have the same url format) which gives support for the non-iso URLs like `1.11.5/metal-arm64.raw.zst`

internal/backend/factory/factory.go
```diff
@@ -81,29 +64,29 @@ func (handler *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+	proxyReq, err := http.NewRequestWithContext(r.Context(), r.Method, proxyURL.String(), nil)
 	if err != nil {
 		handler.handleError("failed to call image factory", w, err)
 
 		return
 	}
 
-	setContentHeaders(w, params.ContentType, params.DestinationFilename)
+	segments := strings.Split(path, "/")
+	filename := fmt.Sprintf("omni-%s-%s", config.Config.Account.Name, segments[len(segments)-1])
```
Slessi: For easier parsing, i changed it from `<platform>-<board>-<arch>-omni-<account-name>-<talos-version>.<ext>` to `omni-<account-name>-<platform>-<board>-<arch>.<ext>` but if necessary can try make it like it was before, or at least just include talos version.

I also dropped content-type which now defaults to `application/octet-stream`.

The reason for both of these is to not depend on the installation media list

internal/backend/factory/factory.go
```diff
@@ -81,29 +64,29 @@ func (handler *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
-	if err != nil {
-		if errors.Is(err, errNotFound) {
-			httpNotFound(w)
-
-			return
-		}
+	path := r.URL.Path
 
-		handler.handleError("failed to parse request", w, err)
+	proxyURL, err := getProxyURL(handler.config.ImageFactoryBaseURL, path)
```
Slessi: This change is to forward all requests to image factory as-is (we have the same url format) which gives support for the non-iso URLs like `1.11.5/metal-arm64.raw.zst`

Unix4ever:
Padding of `x` is not the same for x and y axis.
I'd suggest to have something like p-2.

Unix4ever:
LGTM except Talos version is still not included in the filename

frontend/src/methods/useDownloadImage.ts
```diff
@@ -0,0 +1,115 @@
...
+    bytesDownloaded,
+    bytesDownloadedFormatted,
+    error,
+    abort,
+    download,
+  }
+}
+
+async function doRequest(...[url, init]: Parameters<typeof fetch>) {
+  const resp = await fetch(url, init)
```
Unix4ever: When I was testing the factory I had to add `Origin` header manually. Not sure if it's a good idea, but I'll submit my diff here anyways:

```suggestion
  const resp = await fetch(url, {
    ...init,
    headers: {
      Origin: window.location.origin,
    },
  })
```

frontend/src/methods/useDownloadImage.ts
```diff
@@ -0,0 +1,115 @@
...
+
+      await doRequest(factoryUrl, {
+        signal: abortController.value.signal,
+        method: 'HEAD',
+        headers: new Headers({ 'Cache-Control': 'no-store' }),
+      })
+
+      phase.value = Phase.Loading
+
+      const resp = await doRequest(factoryUrl, { signal: abortController.value.signal })
```
Unix4ever: Actual download should be replaced with `window.open(factoryUrl)` without blank.
So the download progress will not be shown in the UI, but the browser will display it, so it should be fine.

Also should add `?filename=${omni-${features.value?.spec.account?.name}-${talosVersion}-${imageName}}` to the download URL to override the content disposition.

Slessi: so CORS for the `HEAD` to generate, then download natively, got it

frontend/src/methods/useDownloadImage.ts
Slessi: `Origin` is passed automatically by the browser. It was required to work for you because doing this replaced the `init.headers` which had `cache-control` in it.

frontend/src/methods/useDownloadImage.ts
```diff
@@ -0,0 +1,72 @@
...
+
+      const [, , , talosVersion, imageName] = headUrl.pathname.split('/')
+
+      const downloadUrl = new URL(headUrl)
+      downloadUrl.searchParams.set(
+        'filename',
+        `omni-${features.value?.spec.account?.name}-${talosVersion}-${imageName}`,
+      )
+
+      window.open(downloadUrl)
```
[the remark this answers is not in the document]
Slessi: previous required CORS

frontend/src/methods/useDownloadImage.ts
```diff
@@ -0,0 +1,72 @@
...
+      }
+
+      const [, , , talosVersion, imageName] = headUrl.pathname.split('/')
+
+      const downloadUrl = new URL(headUrl)
+      downloadUrl.searchParams.set(
+        'filename',
+        `omni-${features.value?.spec.account?.name}-${talosVersion}-${imageName}`,
+      )
+
```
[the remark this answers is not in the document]
Slessi: add some checks

frontend/src/methods/useDownloadImage.ts
```diff
@@ -0,0 +1,72 @@
...
+        headers: new Headers({
+          'Cache-Control': 'no-store',
+        }),
+      })
+
+      if (!headResponse.ok) {
+        throw new Error(`request failed: ${headResponse.status} ${await headResponse.text()}`)
+      }
+
+      const [, , , talosVersion, imageName] = headUrl.pathname.split('/')
```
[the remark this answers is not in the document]
Slessi: add some checks

frontend/src/methods/useDownloadImage.ts
```diff
@@ -0,0 +1,72 @@
...
+      const headResponse = await fetch(headUrl, {
+        method: 'HEAD',
+        signal: abortController.value.signal,
+        headers: new Headers({
+          'Cache-Control': 'no-store',
+        }),
+      })
+
+      if (!headResponse.ok) {
+        throw new Error(`request failed: ${headResponse.status} ${await headResponse.text()}`)
```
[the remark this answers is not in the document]
Slessi: changed

frontend/src/methods/useDownloadImage.ts
```diff
@@ -0,0 +1,72 @@
...
+    }
+  }
+
+  onUnmounted(() => {
+    abort()
+  })
+
+  return {
+    isGenerating,
+    error,
```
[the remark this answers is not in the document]
Slessi: removed

frontend/src/methods/useDownloadImage.ts
```diff
@@ -0,0 +1,74 @@
...
+      const headResponse = await fetch(headUrl, {
+        method: 'HEAD',
+        signal: abortController.value.signal,
+        headers: new Headers({
+          'Cache-Control': 'no-store',
+        }),
+      })
+
+      if (!headResponse.ok) {
+        throw new Error(`request failed: ${headResponse.status} ${headResponse.statusText} `)
```
Unix4ever:
```suggestion
        throw new Error(`request failed: ${headResponse.status} ${headResponse.statusText}`)
```

PR #2020, opened 2025-12-17
feat: support kernel args management in cluster templates

### LINKED ISSUE #1940: [feature] support kernel args customization support in cluster templates
Add a new field to `Machine` object in templates.
Detect the presence of this field to find out if the args are managed by the templates.
If they are, bind the lifecycle of `ExtraArgs` to templates.
Add support to `export` command - including kernel args in the exported template should be a flag, probably.

utkuozdemir:
Implement kernel args support in cluster templates.

Managing kernel args via templates is opt-in: only and only if the `kernelArgs` YAML key is defined on a `Cluster`, `ControlPlane`, `Worker` or `Machine`, the matching `KernelArgs` resource will be created/updated.

Lower levels override higher levels (Cluster -> MachineSet -> Machine).

Unlike other cluster template managed resources, they will never be destroyed, i.e, when they are removed from a template (removed completely, as in, `kernelArgs` key doesn't exist) or when `omnictl cluster template delete` is run. They instead will get updated to have the annotation `omni.sidero.dev/managed-by-cluster-templates` removed from them.

Add the new flag `--include-kernel-args` to the `omnictl cluster template export` command to optionally include them in the exported template. Note: when this flag is set, `kernelArgs` key is always included at per-machine level, not pulled up even if they are the same for all machines in a machine set or a cluster.

Update the frontend, specifically the kernel args update screen to warn the user if kernel args for that machine is managed by templates, similar to what we do for clusters.

Closes #1940.

### DIFF
--- a/client/pkg/template/internal/models/kernelargs.go
+++ b/client/pkg/template/internal/models/kernelargs.go
@@ -1,0 +1,42 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+package models
+
+import (
+	"github.com/cosi-project/runtime/pkg/resource"
+
+	"github.com/siderolabs/omni/client/pkg/omni/resources/omni"
+)
+
+// KernelArgs is a wrapper type for the kernel args, which can be specified at a Cluster, MachineSet, or Machine level in cluster templates.
+//
+// This type needs to be included in the models as an inline YAML, e.g., via `yaml:",inline"`.
+type KernelArgs struct {
+	Value *[]string `yaml:"kernelArgs,omitempty"`
+}
+
+// Get returns the kernel args value and whether they are actually defined.
+func (ka *KernelArgs) Get() (args []string, defined bool) {
+	if ka.Value != nil {
+		return *ka.Value, true
+	}
+
+	return nil, false
+}
+
+// Set sets the kernel args value.
+func (ka *KernelArgs) Set(args []string) {
+	ka.Value = &args
+}
+
+func buildKernelArgsResource(id MachineID, args []string) *omni.KernelArgs {
+	kernelArgsRes := omni.NewKernelArgs(resource.ID(id))
+
+	kernelArgsRes.Metadata().Annotations().Set(omni.ResourceManagedByClusterTemplates, "")
+
+	kernelArgsRes.TypedSpec().Value.Args = args
+
+	return kernelArgsRes
+}
--- a/frontend/src/views/cluster/ManagedByTemplatesWarning.vue
+++ b/frontend/src/views/cluster/ManagedByTemplatesWarning.vue
@@ -5,13 +5,11 @@ Use of this software is governed by the Business Source License
 included in the LICENSE file.
 -->
 <script setup lang="ts">
-import type { Ref } from 'vue'
 import { computed, ref } from 'vue'
 import { useRoute } from 'vue-router'
 
 import { Runtime } from '@/api/common/omni.pb'
 import type { Resource } from '@/api/grpc'
-import type { ClusterSpec } from '@/api/omni/specs/omni.pb'
 import { ClusterType, DefaultNamespace, ResourceManagedByClusterTemplates } from '@/api/resources'
 import Watch from '@/api/watch'
 import TAlert from '@/components/TAlert.vue'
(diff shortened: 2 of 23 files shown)


frontend/src/views/cluster/ManagedByTemplatesWarning.vue
```diff
@@ -1,45 +1,39 @@
...
 }
 
 const props = withDefaults(defineProps<Props>(), {
   warningStyle: 'alert',
 })
 
-const cluster: Ref<Resource<ClusterSpec> | undefined> = ref(props.cluster)
+const propResource = toRef(() => props.resource)
+const routeResource = ref<Resource>()
+const activeResource = computed(() => propResource.value || routeResource.value)
```
Slessi: you don't need to use `toRef` with props in vue 3.5+, they are reactive already

https://vuejs.org/guide/components/props#reactive-props-destructure

the new way of the above stuff is:

```ts
// destructure whatever you want, optionally apply defaults
const { resource, warningStyle = 'alert' } = defineProps<Props>()

const routeResource = ref<Resource>()
const activeResource = computed(() => resource || routeResource.value)
```

you can keep `propResource` name if you want with
```ts
const { resource: propResource, warningStyle = 'alert' } = defineProps<Props>()
```

frontend/src/views/cluster/ManagedByTemplatesWarning.vue
```diff
@@ -1,45 +1,39 @@
...
 
-// If the cluster is not passed explicitly, watch it from the route, if cluster exists in the route params.
-if (!props.cluster) {
+// If the resource is not passed explicitly, default to the cluster in the route, if exists.
+if (!propResource.value) {
   const route = useRoute()
-  const clusterWatch = new Watch(cluster)
+  const clusterWatch = new Watch(routeResource)
```
Slessi, line 32: IDK what the intention is here but fyi if the prop changes, this won't run again - which is maybe fine in this case for now, but in the future might be unexpected

client/pkg/template/internal/models/kernelargs.go
```diff
@@ -0,0 +1,69 @@
...
+)
+
+// KernelArgs is a wrapper type for the kernel args, which can be specified at a Cluster, MachineSet, or Machine level in cluster templates.
+//
+// The actual type is []string, but we have to work with a raw yaml.Node here to be able to distinguish between undefined (i.e., kernelArgs key isn't in the YAML at all) and null values.
+//
+// If the value is undefined, kernel args won't be managed by the templates at all.
+//
+// This type needs to be included in the models as an inline YAML, e.g., via `yaml:",inline"`.
+type KernelArgs struct {
```
smira, line 16: I think `*[]string` would give you same result as having this custom type.

frontend/src/views/omni/Modals/UpdateKernelArgs.vue
Slessi: optional refactor, but the above is also fine:

```ts
const { data: kernelArgs } = useResourceGet<KernelArgsSpec>({
  runtime: Runtime.Omni,
  resource: md,
})

const initialArgs = computed(
  () => kernelArgs.value?.spec.args?.map((arg) => arg.trim()).join(' ') ?? '',
)

watch(initialArgs, () => (args.value = initialArgs.value))
```

frontend/src/views/cluster/ManagedByTemplatesWarning.vue
utkuozdemir: yes, that was intentionally that way before, and I preserved that. But since now I needed to pass the reactive version of the resource into a separate component, I changed this to be this way.

utkuozdemir: oh sorry, wrong file. I thought this was the other file.

frontend/src/views/omni/Modals/UpdateKernelArgs.vue
utkuozdemir: you are the boss, done

frontend/src/views/cluster/ManagedByTemplatesWarning.vue
utkuozdemir: done

client/pkg/template/internal/models/kernelargs.go
utkuozdemir: done

frontend/src/views/cluster/ManagedByTemplatesWarning.vue
utkuozdemir: not sure how to do it exactly right, so I'll leave it for now.

frontend/src/views/cluster/ManagedByTemplatesWarning.vue
```diff
@@ -53,19 +49,27 @@ if (!props.cluster) {
 }
 
 const clusterIsManagedByTemplates = computed(() => {
```
Slessi: variable needs renaming now since it could be any resource

```suggestion
const resourceIsManagedByTemplates = computed(() => {
```

PR #2023, opened 2025-12-17
fix(installation-media): parse yaml for overlay options before JSON stringifying

Slessi:
Fix errors when creating SBC images due to incorrect handling of overlay options. Send overlay options as-is from frontend (as a YML string) and convert the input to a text area. In the backend, check first if options are set and then unmarshal as YAML instead of as JSON.


Unix4ever (commented):
If it's YAML originally, we can probably just parse it as YAML on the backend?

Slessi:
@Unix4ever done, looks closer to factory like this actually

PR #2028, opened 2025-12-19
fix: prevent audit logs migration from getting stuck

utkuozdemir:
Fix the issue where file audit logs would have a very long line, causing the sqlite migration to get stuck due to the bug on reading them via a scanner.

Additionally, increase the buffer size of the scanner to be able to handle lines up to 16MB in size.

### DIFF
--- a/internal/backend/runtime/omni/audit/auditlog/auditlogfile/log_file.go
+++ b/internal/backend/runtime/omni/audit/auditlog/auditlogfile/log_file.go
@@ -169,6 +169,7 @@ func (l *logReader) Close() error {
 func (l *logReader) Read() ([]byte, error) {
 	if l.scanner == nil {
 		l.scanner = bufio.NewScanner(l.rdr)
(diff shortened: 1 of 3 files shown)


internal/backend/runtime/omni/audit/auditlog/auditlogfile/log_file.go
```diff
@@ -169,6 +169,7 @@ func (l *logReader) Close() error {
 func (l *logReader) Read() ([]byte, error) {
 	if l.scanner == nil {
 		l.scanner = bufio.NewScanner(l.rdr)
+		l.scanner.Buffer(nil, 4*1024*1024) // 4MB max line size
```
smira: should we go with `16 MB` to be on the safe side (or even more given that we know there might be some really long lines)? it doesn't cost us much

utkuozdemir: done

Issue #1133, opened 2025-04-22
making namespace name in `NewResource` implicit


smira:
This would avoid repeating ourselves, and also reduce potential bugs.

`NewResource(id string)`

Issue #1121, opened 2025-04-16
[epic] backend code refactoring/simplification


smira:
- making namespace name in `NewResource` implicit 
- split controllers/resources in `omni/` into more folders/packages 
- refactor unit-test controller suite
- COSI: WriterModify for state (not controller runtime provided state)
- COSI: TeardownAndDestroy method 
- refactor Omni startup
   - one flat errgroup
   - get rid of 20+ arguments
   - components with clear inputs (other components) and options
 - QTransform + handling extra inputs destroy with finalizers

PR #2060, opened 2025-12-24
feat: make namespace implicit for all resources in auth package

pranav767:
Resolves https://github.com/siderolabs/omni/issues/1133

There are more package resources that follow the same pattern, and similar refactoring could be applied there as well.
Example:
`NewUser(ns, id string)`
`NewPublicKey(ns, id string)`
`NewCluster(ns string, id resource.ID)`

I don’t see any open issues for the other resources, but I can include changes in this PR or raise separate issues if refactoring is required elsewhere.

### DIFF
--- a/client/pkg/omni/resources/auth/identity.go
+++ b/client/pkg/omni/resources/auth/identity.go
@@ -15,9 +15,9 @@ import (
 )
 
 // NewIdentity creates a new Identity resource.
-func NewIdentity(ns, id string) *Identity {
+func NewIdentity(id string) *Identity {
 	return typed.NewResource[IdentitySpec, IdentityExtension](
-		resource.NewMetadata(ns, IdentityType, id, resource.VersionUndefined),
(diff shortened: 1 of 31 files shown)


Unix4ever:
Yeah, we actually plan to drop namespace parameter for all resources, as we don't have cases when the same resource type might be created in several different namespaces.

I guess it's fine to create several PRs for that as the diff will be huge.

pranav767:
Oh, alright. I can create a proper list of potential package resources that may need refactoring.
How should we approach this - will you be creating the issues, or is it okay if I create them? After creating the issues, we can add them to the plan/project.
https://github.com/siderolabs/omni/issues/1121

Unix4ever:
Can probably create separate PRs for resources in different folders. Though `omni` one will be huge.
I don't think it makes sense to spend time to create issues.

pranav767:
@Unix4ever, this PR is ready for review, I've refactored for all resources in auth package.
Aside looking at omni package, that might get painful to review (I'm making changes locally now), let me know if I can be of any help for that.

Issue #1929, opened 2025-11-27
[refactor] Rework the config apply rollout to work directly with `ClusterMachineConfig`


utkuozdemir:
The goal of this refactor is to simplify Omni by unifying the config rollout logic. Currently, the MachineSetStatusController adds unnecessary complexity by trying to control the rollout timing, updating ClusterMachineConfigPatches one by one. This approach is brittle and contributes to issues where the system gets stuck on invalid states.

Instead, we should create all config patches immediately and merge them into the ClusterMachineConfig right away. The actual rollout coordination logic should move down to the ClusterMachineConfig controller (or a new controller). This controller will be responsible for deciding when to apply the config based on the machine's state, such as applying immediately for broken nodes or machines not in a cluster, but coordinating safe updates for healthy nodes.

Further context if needed: https://docs.google.com/document/d/1zFXi8Vut8-qmBoWVl9ZM98nVDZE4F3dj9uamgEtKE6E/edit?tab=t.zr0wf2vhcg5

utkuozdemir:
Related to the whole topic of rollout rework - see #1344.

PR #2069, opened 2025-12-26
chore: move graceful config rollout logic to the lowest controller level

Unix4ever:
Now graceful config rollout is handled by the
`ClusterMachineConfigStatusController`.
It calculates the available update quota by adding finalizers on the `ClusterMachine` resources. By counting the resources with the finalizers it tracks the remaining quota.
It now also calculates the pending changes which are not yet applied to the machine in the `MachinePendingUpdates`.

Pending changes are not yet shown in the UI anywhere.

Fixes: https://github.com/siderolabs/omni/issues/1929


client/api/omni/specs/omni.proto
```diff
@@ -600,11 +600,36 @@ message ClusterMachineConfigStatusSpec {
...
+    string to_schematic = 2;
+
+    string from_version = 3;
+    string to_version = 4;
+  }
+
+  // Upgrade represents pending machine upgrade.
+  Upgrade upgrade = 1;
+  // ConfigDiff represents the redacted machine config diff.
+  string config_diff = 2;
```
oguzkilcan, line 632: Won't this also require compression for diffs that are too large?

internal/backend/runtime/omni/controllers/omni/cluster_machine_status_test.go
```diff
@@ -20,6 +20,8 @@ import (
 	"github.com/siderolabs/omni/client/api/omni/specs"
 	"github.com/siderolabs/omni/client/pkg/omni/resources/omni"
 	omnictrl "github.com/siderolabs/omni/internal/backend/runtime/omni/controllers/omni"
+	"github.com/siderolabs/omni/internal/backend/runtime/omni/controllers/testutils/rmock"
+	"github.com/siderolabs/omni/internal/backend/runtime/omni/controllers/testutils/rmock/options"
 )
 
 type ClusterMachineStatusSuite struct {
```
oguzkilcan: Personal opinion: Whenever I'm making changes to a unit test and it's written with `testify.Suite`, I try to migrate it to `testutils.WithRuntime`. This will help us gradually phase out `testify.Suite`.

client/api/omni/specs/omni.proto
Unix4ever: I think we can avoid compressing it. Compared to the normal configs this diff exists only if the machine is being reconfigured, so I'd assume we can afford having large uncompressed diffs there.

internal/backend/runtime/omni/controllers/omni/internal/machineset/status_handler.go
```diff
@@ -76,15 +57,32 @@ func ReconcileStatus(rc *ReconciliationContext, machineSetStatus *omni.MachineSe
 
 	_, isControlPlane := machineSet.Metadata().Labels().Get(omni.LabelControlPlaneRole)
 
+	if isControlPlane {
+		// only allow config changes if no other operations are pending for the control planes
+		spec.ConfigUpdatesAllowed = (len(rc.GetMachinesToCreate()) + len(rc.GetMachinesToTeardown()) + len(rc.GetMachinesToTeardown())) == 0
```
Unix4ever:
```suggestion
		spec.ConfigUpdatesAllowed = (len(rc.GetMachinesToCreate()) + len(rc.GetMachinesToTeardown()) + len(rc.GetMachinesToDestroy())) == 0
```

internal/backend/runtime/omni/controllers/omni/machineconfig/status.go
```diff
@@ -192,26 +240,65 @@ func (ctrl *ClusterMachineConfigStatusController) reconcileRunning(
 
 			return fmt.Errorf("failed to apply config to machine '%s': %w", machineConfig.Metadata().ID(), err)
 		}
+
+		machineConfigStatus.TypedSpec().Value.ClusterMachineConfigVersion = machineConfig.Metadata().Version().String()
+
+		if err = machineConfigStatus.TypedSpec().Value.SetUncompressedData(rc.redactedMachineConfig); err != nil {
+			return err
+		}
```
Unix4ever: drop this duplicate code

PR #2084, opened 2025-12-31
omni helm deployment improvements

xinity:
this PR adds:
- missing sqlite parameters and helm values
- additional pvc templates for etcd and sqlite volumes

### DIFF
--- a/deploy/helm/omni/templates/deployment.yaml
+++ b/deploy/helm/omni/templates/deployment.yaml
@@ -103,6 +103,7 @@ spec:
           - --private-key-source={{ .Values.privateKeySource }}
           - --siderolink-api-advertised-url={{ printf "https://%s" .Values.service.siderolink.domainName }}
           - --siderolink-wireguard-advertised-addr={{ .Values.service.siderolink.wireguard.address }}:{{ .Values.service.siderolink.wireguard.port }}
+          - --sqlite-storage-path={{ .Values.volumes.sqlite.persistentVolumeClaim.Name }}          
         {{- range $value := .Values.extraArgs }}
           - {{ $value }}
         {{- end }}


utkuozdemir:
Thanks for the PR. Please note that we are working on the V2 of the Helm chart, so we keep the changes to the existing version minimal. The ongoing work for V2 is here: https://github.com/siderolabs/omni/pull/1909

Edit: since this PR introduced breaking changes and seems it was not tested, I will have to reject this PR, since the ongoing V2 work will supersede this. Regardless of that, thank you very much for your effort.

deploy/helm/omni/values.yaml
```diff
@@ -52,7 +52,13 @@ volumeMounts:
     readOnly: true
 volumes:
   etcd:
-    persistentVolumeClaimName: omni-pvc
+    persistentVolumeClaim:
+      Name: omni-pvc
+      Size: 10G
```
utkuozdemir, line 57: These changes are not backwards compatible. We will introduce breaking changes altogether in the V2 of the chart - WIP, here: https://github.com/siderolabs/omni/pull/1909

deploy/helm/omni/templates/deployment.yaml
```diff
@@ -103,6 +103,7 @@ spec:
           - --private-key-source={{ .Values.privateKeySource }}
           - --siderolink-api-advertised-url={{ printf "https://%s" .Values.service.siderolink.domainName }}
           - --siderolink-wireguard-advertised-addr={{ .Values.service.siderolink.wireguard.address }}:{{ .Values.service.siderolink.wireguard.port }}
+          - --sqlite-storage-path={{ .Values.volumes.sqlite.persistentVolumeClaim.Name }}
```
utkuozdemir, line 106: This would not work, as `--sqlite-storage-path` expects a file path, not a directory path, like `/path/to/sqlite.db`, as documented in the changelog. This is an indication that the changes introduced were not tested.

xinity:
@utkuozdemir this PR Has been tested as i'm using it to deploy my omni instance :)
closed anyway

PR #2111, opened 2026-01-08
feat(frontend): add extra information to userpilot

### LINKED ISSUE #2036: [feature] Add company to userpilot data
### Problem Description

Need a way to filter types of users from userpilot flows. This should include a user's `sign_up_date` and company (based on omni domain).

### Solution

When a user first logs in and accepts userpilot cookie we need to add two additional fields to the user object.

The company field can be based on the Omni domain (just the bottom level).

eg, my omni jg.us-west-1.omni.siderolabs.io my company should be "jg"

The sign_up_date should be based on when the user is first seen. I'm not sure if that gets sent when the user is first seen or if that needs to happen as part of the Omni provisioning flow.

### Alternative Solutions

sign_up_date might need to be handled by our onboarding web form instead of in Omni. If that's the case we can figure out how to create a company and sign up date for users in that company.

### Notes

_No response_

Slessi:
Add extra information to userpilot as per [RFD-28](https://www.notion.so/siderolabs/RFD-28-Userpilot-Implementation-and-Omni-Usage-Analytics-6024961dd8cc43eea0ef7797841be51b?d=2dfb1211badf809fbd3f001c79056040#118b1211badf804b9de5edde0f22da96)

Closes #2036

### DIFF
--- a/frontend/src/methods/userpilot.ts
+++ b/frontend/src/methods/userpilot.ts
@@ -4,7 +4,7 @@
 // included in the LICENSE file.
 import { useLocalStorage } from '@vueuse/core'
 import { Userpilot } from 'userpilot'
-import { computed, ref, watch, watchEffect } from 'vue'
+import { computed, ref, toValue, watchEffect } from 'vue'
 
 import { getNonce } from '@/methods'
 import { currentUser } from '@/methods/auth'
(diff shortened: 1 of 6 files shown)


frontend/src/methods/userpilot.ts
```diff
@@ -4,36 +4,44 @@
...
-      role: user.spec.role!,
+    if (!user?.spec.user_id || !userpilotInitialised.value) return
+
+    Userpilot.identify(user.spec.user_id, {
+      role: user.spec.role,
+      email: user.spec.identity,
+      created_at: user.metadata.created,
+
+      company: {
+        id: features.value?.spec.account?.id,
```
Unix4ever, line 43: We now have `name` in the account, but it's not used here? Then what was the point to add it?

Slessi: Justin/Jeff seemed like they might want it later, and for https://github.com/siderolabs/omni/pull/2019 I had to add it for the filename (though we might scrap that PR)

Mostly it just looked potentially useful, so figured why not. Can remove for this particular PR if u prefer not to have it just yet.

Unix4ever: Yeah, it's useful for sure

PR #2128, opened 2026-01-13
test: fix the flake in redacted machine config test

utkuozdemir:
It was flaky because it relied on the fake clock and timings, which do not always behave deterministic.

Remove the fake clock, trigger the cleanups manually from the test (via an additional optional channel).

Change the ID format of the diffs from being millisecond precision to have nanosecond precision, as it caused updates instead of new diffs being created in cases where a config gets updated multiple times very quickly.

Move the controller and its test to a dedicated package and refactor their tests from using a suite to `testutils.WithRuntime`.

Additionally, increase the timeouts in machine config status tests, as in rare cases where tests get starved of CPU, 5 seconds were not enough for the machines to get configured.


frezbo:
> It was flaky because it relied on the fake clock and timings, which do not always behave deterministic.
> 
> Remove the fake clock, trigger the cleanups manually from the test (via an additional optional channel).
> 
> Change the ID format of the diffs from being millisecond precision to have nanosecond precision, as it caused updates instead of new diffs being created in cases where a config gets updated multiple times very quickly.
> 
> Move the controller and its test to a dedicated package and refactor their tests from using a suite to `testutils.WithRuntime`.

time can be faked with testing/synctest see https://github.com/siderolabs/talos/pull/11915

utkuozdemir:
> > It was flaky because it relied on the fake clock and timings, which do not always behave deterministic.
> > Remove the fake clock, trigger the cleanups manually from the test (via an additional optional channel).
> > Change the ID format of the diffs from being millisecond precision to have nanosecond precision, as it caused updates instead of new diffs being created in cases where a config gets updated multiple times very quickly.
> > Move the controller and its test to a dedicated package and refactor their tests from using a suite to `testutils.WithRuntime`.
> 
> time can be faked with testing/synctest see [siderolabs/talos#11915](https://github.com/siderolabs/talos/pull/11915)

Yes, I tried to make it work with that first, but:
- the cleanup and reconcile timings were still hard to get right, in rare cases it still flaked
- it has a timeout problem - if for some reason the test hangs, it times out only after the go test command times out (10s). It is not possible to define a (real) timeout via context in those tests.

So I went with this way instead.

Issue #2126, opened 2026-01-13
Split integration tests to qemu, talemu and e2e


oguzkilcan:
Split integration tests so they can run in parallel. End result should have:
* Omni integration tests running with talemu
* Omni integration tests running with qemu
* Omni e2e tests running with qemu
* Omni e2e tests running with talemu*

PR #2132, opened 2026-01-13
test: add talemu fixtures and split into talemu + qemu tests

Slessi:
- Split tests into `qemu` and `talemu` projects/folders
- Adds a setup step to the `talemu` project to create a machine class to use in fixtures
- Adds a cluster fixture to `talemu` project to create throw away clusters on demand
- Updated `View all clusters` and `View all machines` to use Talemu clusters as a POC

Closes #2129

Issue #2115, opened 2026-01-09
Setup talemu as infra provider for E2E tests


Slessi:

Slessi:
Apparently the omni instance used for frontend E2E already has talemu on it, so this is not needed.

PR #2134, opened 2026-01-13
chore: separate integration-tests

oguzkilcan:
Separate integration/e2e tests for qemu and talemu, so we can run them in parallel.

Closes: #2126


hack/test/e2e-qemu.sh
Slessi, line 1: I noticed in #2132 that talemu is present as an infra provider for E2E tests, and I actually need that for playwright tests. I was going to add it in #2115 but closed since it was present already.

Can we have that here also? Or is it here already?

oguzkilcan: It's not here anymore. But let me add `e2e-talemu` tests so the picture is complete.

PR #2139, opened 2026-01-14
fix(frontend): fix incorrect cluster query checking for disk_encryption

Slessi:
Fix the incorrect watch query in PatchEdit which checks for the `disk_encryption` feature. Fixed it with refactor to `useResourceWatch` and setting the correct query & types.

### DIFF
--- a/frontend/src/views/cluster/Config/PatchEdit.vue
+++ b/frontend/src/views/cluster/Config/PatchEdit.vue
@@ -7,7 +7,6 @@ included in the LICENSE file.
 <script setup lang="ts">
 import type monaco from 'monaco-editor/esm/vs/editor/editor.api'
 import { MarkerSeverity, MarkerTag } from 'monaco-editor/esm/vs/editor/editor.api'
-import type { Ref } from 'vue'
 import { computed, defineAsyncComponent, onMounted, ref, watch } from 'vue'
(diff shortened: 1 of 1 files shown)


frontend/src/views/cluster/Config/PatchEdit.vue
```diff
@@ -354,46 +366,8 @@ const notes = computed(() => {
...
-    selectors: [`${LabelCluster}=${route.params.cluster}`],
-  })
-
-  clusterWatch.setup({
-    runtime: Runtime.Omni,
-    resource: {
-      namespace: DefaultNamespace,
-      type: ClusterMachineStatusType,
-      id: route.params.cluster as string,
-    },
```
Slessi, line 391: The bug is here. It is querying `ClusterMachineStatusType`, but `features.disk_encryption` only exists on `ClusterType`. It's been this way since the first commit of omni 22 months ago though so maybe worth validating if this behaviour is the desired behaviour? 🤷 

https://github.com/siderolabs/omni/blob/1483aacbd3c23e43cceae99d8c546d34d9a618cb/frontend/src/views/cluster/Config/PatchEdit.vue#L143-L147

With fix, if `ClusterSpec.features?.disk_encryption` is true:
image

Without fix, will always be this (this particular message is because the current yaml config schema doesnt have it in the spec anymore):
image

PR #2145, opened 2026-01-15
chore: bump node to 24.13

Slessi:
Bump node to 24.13 to address CVEs https://nodejs.org/en/blog/vulnerability/december-2025-security-releases


Slessi:
@smira yeah indeed. The upgrade in this case doesn't have a real cost so its very much a "might as well"

smira:
> @smira yeah indeed. The upgrade in this case doesn't have a real cost so its very much a "might as well"

my comment was more to set expectations for our users ;)

PR #2149, opened 2026-01-16
chore: rekres and bump deps

oguzkilcan:
* rekres
* bump deps
* bump go to 1.25.6
* fix linter errors


internal/backend/runtime/options.go
```diff
@@ -57,6 +57,7 @@ func WithName(name string) QueryOption {
 // WithLabelSelectors enables filtering by label.
 func WithLabelSelectors(selectors ...string) QueryOption {
 	return func(o *QueryOptions) {
+		//nolint:prealloc
 		var labelSelectors []string
```
smira: I guess this can be pre-allocated to the length of `selectors` ?

smira (approved):

PR #2150, opened 2026-01-16
feat: generate omni config from schema

### LINKED ISSUE #2157: generate/validate Omni config from a JSON schema
This will address multiple issues:
- Unexpected behavior on config merge with zero values (particularly booleans), blocking the work on Helm chart v2 in #1884 
- Automatically generating documentation for Omni config
- Overall consistency of config code and its parsing/validation

utkuozdemir:
Make all leaf fields nillable, so that we can distinguish unset from explicit empty, and merging of CLI args and YAML configs work correctly.

Generate nil-safe accessors (getter/setters) for these nillable fields and use them in the code.

Wrap the cobra command line parser to support nillable flags.

Move all validations into the JSON schema and drop go-validator usage and its annotations.

Closes #2157.


internal/pkg/config/config.go
```diff
@@ -6,21 +6,25 @@
 // Package config contains the application config loading functions.
 package config
 
+//go:generate go tool go-jsonschema --only-models --struct-name-from-title --tags=json,yaml --package=config --extra-imports -o=types.generated.go schema.json
+
+//go:generate go tool go-accessor-gen --source=types.generated.go --output=accessors.generated.go
```
utkuozdemir: Note to reviewer:
This is a tool I wrote to generate nil-safe accessors for the config fields, so that we would not pollute the codebase with many `pointer.To`/`pointer.Safederef` calls, as now they all are nillable.

I am open to moving this project under `./hack` or to a repo under our organization.

go.mod
```diff
@@ -3,13 +3,20 @@ module github.com/siderolabs/omni
...
 	// forked saml library that has the fix for Fusion Auth ACS parsing
 	github.com/crewjam/saml => github.com/unix4ever/saml v0.0.0-20250630213700-66b137182abe
 
 	// use nested module
 	github.com/siderolabs/omni/client => ./client
 )
 
+tool (
+	github.com/atombender/go-jsonschema
+	github.com/utkuozdemir/go-accessor-gen
```
utkuozdemir: Note to reviewer:
Please see: https://github.com/siderolabs/omni/pull/2150/changes#r2704503636

cmd/omni/cmd/flag.go
utkuozdemir, line 1: Note to reviewer:
I iterated a few times on this - with generics, reflection, custom value type and so on. Ended up with this, as I found it to be the cleanest solution among them.

utkuozdemir:
Locally tested:
- Only file based config
- Only flags based config
- File + flag overrides

internal/pkg/config/schema.json
```diff
@@ -0,0 +1,1117 @@
...
+        "snapshotsEnabled": {
+          "type": "boolean"
+        },
+        "snapshotsPath": {
+          "type": "string",
+          "deprecated": true
+        },
+        "snapshotsInterval": {
+          "type": "string",
+          "pattern": "^([0-9]+(\\.[0-9]+)?(ns|us|µs|ms|s|m|h))+$",
```
utkuozdemir, line 307: This is Go's time pattern. If we instead used `format: duration`, it would enforce JS's pattern, which is not correct in our case.

internal/pkg/config/schema.json
```diff
@@ -0,0 +1,1117 @@
...
+          "goJSONSchema": {
+            "type": "time.Duration"
```
utkuozdemir, line 309: Hint the generator on what Go type to use for durations.

internal/pkg/config/schema.json
```diff
@@ -0,0 +1,1117 @@
...
+        "id",
+        "name"
+      ],
+      "properties": {
+        "id": {
+          "type": "string",
+          "minLength": 1,
+          "description": "Stable identifier of the instance.",
+          "goJSONSchema": {
+            "type": "*string"
```
utkuozdemir, line 64: This field is marked as required, so without this hint to the generator, it would generate the type `string`. We want the schema to have it as required (so I can validate my config files against them correctly), but we want the Go type to still be `*string`, as we want to still be able to correctly distinguish empty vs unset.

cmd/omni/cmd/flag.go
```diff
@@ -0,0 +1,123 @@
+// Copyright (c) 2025 Sidero Labs, Inc.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+
+package cmd
```
Unix4ever, line 6: I think it could be nice to have the tests for these functions

utkuozdemir:
> Can we use a single library for all json schema operations? Maybe we can switch the provider validation code to use the same library which is used for the config.

I had a look into that, and I would leave it as-is: the new library I used (omissis/go-jsonschema) is focused on schema/code generation, while the previous lib (santhosh-tekuri/jsonschema) is focused on being full-fledged validator and nothing else.

The new library has limited validation, but far from being feature complete: it doesn't validate for `oneOf`, `allOf`, `if-else` and so on, which we use in our schema.

Also, the old library supports taking YAML as input directly, which is beneficial for us, because when we marshal our final config to YAML, the durations marshal as strings (like `1h30s`), while when marshaled to JSON, they marshal to numbers (millis or nanos or something). So YAML fits to us better there, to pass it as JSON for validation, we would need to do more handling.

cmd/omni/cmd/flag.go
utkuozdemir: Good point, added a unit test which tests all of those flag functions.

Unix4ever:
We talked with @smira about this PR a bit. He had an idea of using https://github.com/siderolabs/gen/tree/main/optional instead of raw pointers here.
It kinda makes sense, as this way it will be harder to cause a nil reference error. Optional forces you to use getters and it distinguishes reference from the pointer more reliably.
But I guess it will require more changes in the JSON schema generation and such...

utkuozdemir:
> We talked with @smira about this PR a bit. He had an idea of using https://github.com/siderolabs/gen/tree/main/optional instead of raw pointers here.
> 
> It kinda makes sense, as this way it will be harder to cause a nil reference error. Optional forces you to use getters and it distinguishes reference from the pointer more reliably.
> 
> But I guess it will require more changes in the JSON schema generation and such...

I'll give this a try, shouldn't be a big effort unless something unexpected comes up.

utkuozdemir:
@Unix4ever updated it after our discussion - brought `go-accessor-gen` into the project.

PR #2153, opened 2026-01-16
Update CONTRIBUTING.md

steverfrancis:


smira:
@steverfrancis you might need to make your membership in the org public to make CLA assistant pick this up

smira:
Forked to #2167

PR #2154, opened 2026-01-16
Create .license-header.community.txt

steverfrancis:


smira:
I'm not sure we need this with CLA?

PR #2169, opened 2026-01-21
feat: utilize kubernetes server side apply for kubernetes bootstrap manifests

Orzelius:
* utilize server side apply for dry-run apply and diffs
* implement pruning of deleted resources

part of https://github.com/issues/assigned?issue=siderolabs%7Comni%7C2064

This is my first time actually touching grpc so please let me know if there's room for improvements!


Unix4ever:
I think [this controller](https://github.com/siderolabs/omni/blob/0f8a3d6c6f06110769d3b2a5f92209a78c6698c9/internal/backend/runtime/omni/controllers/omni/kubernetes_upgrade_manifest_status.go#L212) should also be updated.

Orzelius:
> I think [this controller](https://github.com/siderolabs/omni/blob/0f8a3d6c6f06110769d3b2a5f92209a78c6698c9/internal/backend/runtime/omni/controllers/omni/kubernetes_upgrade_manifest_status.go#L212) should also be updated.

Ah I completely missed it! I've updated the controller to use the new SSA logic.

Unix4ever:
As the library changed so much it's easier to do from scratch. New PR https://github.com/siderolabs/omni/pull/2453

PR #2173, opened 2026-01-22
fix: open OIDC plugin link in a new tab

utkuozdemir:
Update the OIDC plugin link to open in a new tab by adding target="_blank" and rel="noopener noreferrer" attributes.

Additionally: if workload proxying feature is disabled in Omni, make its feature checkbox on cluster overview page disabled instead of completely hidden - it makes it consistent with other features and makes the feature more discoverable.

### DIFF
--- a/frontend/src/views/cluster/Overview/components/OverviewContent.vue
+++ b/frontend/src/views/cluster/Overview/components/OverviewContent.vue
@@ -44,7 +44,7 @@ import {
   setClusterWorkloadProxy,
   setUseEmbeddedDiscoveryService,
 } from '@/methods/cluster'
-import { embeddedDiscoveryServiceFeatureAvailable } from '@/methods/features'
+import { embeddedDiscoveryServiceFeatureAvailable, useFeatures } from '@/methods/features'
 import { useResourceWatch } from '@/methods/useResourceWatch'
 import ClusterMachines from '@/views/cluster/ClusterMachines/ClusterMachines.vue'
 import OverviewRightPanel from '@/views/cluster/Overview/components/OverviewRightPanel/OverviewRightPanel.vue'
@@ -123,6 +123,8 @@ const { data: talosUpgradeStatus } = useResourceWatch<TalosUpgradeStatusSpec>(()
 }))
 
 const { canManageClusterFeatures } = setupClusterPermissions(clusterId)
+
+const { data: features } = useFeatures()
 
 const isEmbeddedDiscoveryServiceAvailable = ref(false)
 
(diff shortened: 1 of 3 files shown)


frontend/src/views/cluster/Overview/components/OverviewContent.vue
```diff
@@ -327,7 +329,7 @@ onMounted(async () => {
             <div class="flex flex-col gap-2">
               <ClusterWorkloadProxyingCheckbox
                 :model-value="enableWorkloadProxy"
-                :disabled="!canManageClusterFeatures"
+                :disabled="!canManageClusterFeatures || !features?.spec.enable_workload_proxying"
```
utkuozdemir, line 332: @Slessi brought it here to this PR

frontend/src/views/omni/Clusters/ClusterWorkloadProxyingCheckbox.vue
```diff
@@ -21,14 +20,14 @@ type Props = {
 defineProps<Props>()
 
 const checked = defineModel<boolean>({ default: false })
-const { data: features } = useFeatures()
 </script>
 
 <template>
-  <Tooltip v-if="features?.spec.enable_workload_proxying" placement="bottom">
```
utkuozdemir, line 28: @Slessi and removed this if - the checkbox was completely not visible

Slessi: I think that the intent was to do exactly that, to hide this checkbox if the feature is disabled. Hiding the other stuff was a mistake though.

We can start showing it but that might be an unexpected functionality change, to show a checkbox for a feature disabled on ur instance. IDK.

@Unix4ever WDYT?

utkuozdemir: I think it was me back then who put that intent there. I built this feature.
The other checkboxes, for example, the embedded discovery service one, is disabled when feature is disabled in omni:
image

This change makes it consistent with it.
IMO, disabled is better than non-visible, as it is easier for the users to find out what feature is available, and what they need to do to unlock it.

OTOH, setting up workload proxying in on-prem setups need additional infra work. So, on prem users might want to have this checkbox out of their sight, idk.

Slessi: Makes sense to me for consistency. Not against it, just wanted to check that its a conscious decision to change this behaviour 👍

PR #2174, opened 2026-01-22
fix: always show features section in cluster overview

utkuozdemir:
It was not displayed when cluster workload proxying feature was disabled instance-wide, but that was wrong, as the features section contains toggles for other features as well.

### DIFF
--- a/frontend/src/views/cluster/Overview/components/OverviewContent.vue
+++ b/frontend/src/views/cluster/Overview/components/OverviewContent.vue
@@ -44,7 +44,7 @@ import {
   setClusterWorkloadProxy,
   setUseEmbeddedDiscoveryService,
 } from '@/methods/cluster'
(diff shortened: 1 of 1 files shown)


frontend/src/views/cluster/Overview/components/OverviewContent.vue
Slessi, line 332: @utkuozdemir should we hide this checkbox then? or disable it when feature is disabled, if you cant ever enable it?

Slessi: actually, it does that internally inside the checkbox already, so nvm

utkuozdemir: oh, sorry, saw the comment after merging. good point, I can add that to the other fix PR.

Slessi: @utkuozdemir there is no need, its already working this way

PR #2179, opened 2026-01-22
chore: remove unused vite-plugin-node-polyfills

Slessi:
Remove the unused vite-plugin-node-polyfills. We use native browser stream APIs, not the node ones.


Unix4ever:
I added that for Firefox browser, but it was long time ago. Please make sure that it won't be broken

PR #2193, opened 2026-01-23
fix: fix config patches encryption when encryption is disabled

utkuozdemir:
When the resource compression was disabled in the Omni config, we were not generating the ClusterMachineConfigPatches correctly.

The issue was: it was attempting to "force-compress" the ClusterMachineConfigPatches when any of the patches' size was above the threshold. But when it was trying to do that, it did not override the global setting of false.

The default setting for resource compression is `true`, but when a config file is used to configure Omni, and it was not specified in the config YAML, it was getting overwritten to be `false` due to the boolean merging behavior, which was fixed in https://github.com/siderolabs/omni/pull/2150.

Also: fix the compression kicking in even in cases when it is disabled in config but above the threshold.

### DIFF
--- a/client/api/omni/specs/cluster_machine_config_patches.go
+++ b/client/api/omni/specs/cluster_machine_config_patches.go
@@ -12,8 +12,6 @@ import (
 
 	"github.com/siderolabs/gen/xslices"
 	"go.yaml.in/yaml/v4"
-
-	"github.com/siderolabs/omni/client/pkg/constants"
 )
 
 // MarshalJSON implements json.Marshaler interface.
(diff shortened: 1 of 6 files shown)


client/api/omni/specs/cluster_machine_config_patches.go
```diff
@@ -161,25 +161,22 @@ func (x *ClusterMachineConfigPatchesSpec) SetUncompressedPatches(patches []strin
...
-) error {
+func (x *ClusterMachineConfigPatchesSpec) FromConfigPatches(patches iter.Seq[*ConfigPatchSpec], compressionConfig CompressionConfig) error {
 	aboveThreshold, err := anyAboveThreshold(patches)
 	if err != nil {
 		return fmt.Errorf("failed to check if one patch is above threshold: %w", err)
 	}
 
-	if compressionEnabled || aboveThreshold {
-		return x.fromConfigPatchesCompress(patches)
+	if compressionConfig.Enabled || aboveThreshold {
```
utkuozdemir: Even though I did a fix, I found this part of the code (which was already like this) confusing: it decides to compress the `ClusterMachineConfigPatches` if compression is enabled OR if any of the patches is above the threshold.

This means, even when I disable compression instance-wide, I cannot get rid of future compressions completely.

This is unlike the other places in the code, for example, in `RedactedClusterMachineConfigSpec`, we compress the data only if the compression is enabled AND if data is above the min threshold.

@Unix4ever do you remember if we did this this intentionally this way back then?

Unix4ever: It looks like a bug to me to be honest. It should be enabled if " thecompression is enabled AND if data is above the min threshold".

utkuozdemir: Yep, I also thought so, therefore kept the PR open. I'll fix that.

utkuozdemir: fixed.

PR #2194, opened 2026-01-26
refactor(frontend): rekres and use request error from fetch.pb.ts

Slessi:
Rekres and use the RequestError class generated in fetch.pb.ts and remove the grpc.ts one.


frontend/src/schemas/config_1_12.schema.json
```diff
@@ -1,6 +1,6 @@
 {
   "$schema": "https://json-schema.org/draft/2020-12/schema",
-  "$id": "https://talos.dev/v1.13/schemas/config.schema.json",
+  "$id": "https://talos.dev/v1.12/schemas/config.schema.json",
```
Slessi, line 3: not sure what happened here, the 1.12 schema got regenerated as 1.13 by a previous PR

PR #2195, opened 2026-01-26
chore: update packages

Orzelius:
* update client image-factory import
* update machinery package and fix test error assertions


client/pkg/clusterimport/clusterimport_test.go
```diff
@@ -410,10 +409,10 @@ func (data *testData) generateMachineConfig(t *testing.T, machineType machinetyp
 	conf, err := input.Config(machineType)
 	require.NoError(t, err)
 
-	machineConfig, ok := conf.Machine().(*v1alpha1.MachineConfig)
```
smira, line 413: this is certainly wrong, and it should use proper `//nolint` for now (and keep using legacy machine config)

Orzelius: Is there no way of doing this with the new environment document?

smira: this using a bit hacky way here, it's possible, but new machine config is Talos 1.13 only, so I would keep old code at the cost of `//nolint:staticcheck`

PR #2209, opened 2026-01-28
fix: bind wireguard to configured address instead of all interfaces

utkuozdemir:
Having issues with wireguard connectivity from QEMU machines to Omni running on macOS revealed an issue: If macOS has multiple interfaces with IPs in the same subnet (for example connected both via ethernet and Wi-Fi), it could respond to WireGuard packets not from the interface they are received from, but from the other one, even when the wg endpoint was explicitly set to be a specific IP:PORT in Omni config. And this was breaking wg handshakes.

The core issue seems to be the wireguard-go library not implementing sticky sockets (`IP_PKTINFO`) on macOS.

While investigating, we found that the standard wireguard-go `StdNetBind` always binds to all interfaces (`0.0.0.0`), ignoring any specific host in the endpoint configuration. Add a custom bind implementation that respects the configured host.

This fixes the macOS issue as a side benefit.

### DIFF
--- a/hack/release.toml
+++ b/hack/release.toml
@@ -62,3 +62,10 @@ title = "gRPC Tunnel Management"
 description = """\
 Added the ability to switch gRPC tunnel modes for connected machines.
 """
+
+[notes.wireguard-bind-address]
+title = "WireGuard Bind Address"
+description = """\
+The WireGuard endpoint (`services.siderolink.wireGuard.endpoint` / `--siderolink-wireguard-bind-addr`) is now respected. \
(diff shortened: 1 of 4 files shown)


oguzkilcan (approved):

utkuozdemir:
@Unix4ever would you like to have a look before I merge this?

smira (approved):

Unix4ever:
> ould you like to have a look before I merge this?

yep, going to take a look right now

Issue #1003, opened 2025-03-11
[bug] Cache collision on `kubectl oidc-login` cache for different users


utkuozdemir:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

The kubectl auth plugin we use, https://github.com/int128/kubelogin, keeps a cache under `~/.kube/cache/oidc-login`.

The file names there are the cache keys, and those keys are the hashes of the args of the `user` in the KUBECONFIG file.

For example, in kubeconfig, let's say we have two users for cluster `mycluster`, `user1@example.org` and `user2@example.org`:

```yaml
- name: myomnicontext-mycluster-user1@example.org
  user:
    exec:
      apiVersion: client.authentication.k8s.io/v1beta1
      args:
        - oidc-login
        - get-token
        - --oidc-issuer-url=https://omni.example.org/oidc
        - --oidc-client-id=native
        - --oidc-extra-scope=cluster:mycluster
      command: kubectl
      env: null
      provideClusterInfo: false

- name: myomnicontext-mycluster-user2@example.org
  user:
    exec:
      apiVersion: client.authentication.k8s.io/v1beta1
      args:
        - oidc-login
        - get-token
        - --oidc-issuer-url=https://omni.example.org/oidc
        - --oidc-client-id=native
        - --oidc-extra-scope=cluster:mycluster
      command: kubectl
      env: null
      provideClusterInfo: false
```

These users will end up in the same cache key, which means changing the context used by kubectl, assuming that it will authenticate as a different user will not work as expected.
This causes confusion, and the way to get around it atm is to run `kubectl oidc-login clean` or `rm -rf ~/.kube/cache/oidc-login`.
This is not ideal, as it is still confusing to authenticate as the unexpected user, and clearing the cache means we need to re-authenticate.

In other words, the `user`s we define in the KUBECONFIGs we generate are unique by `instance+cluster` (`--oidc-issuer-url + --oidc-extra-scope=cluster:`), not `instance+cluster+user`.

We should look into whether we want to include user information in the oidc-login input one way or another - if we do, we also need to verify it on the server side during authentication.

### Expected Behavior

Multiple user definitions against the same omni instance and the same cluster live happily and separately in my kubeconfig.

### Steps To Reproduce

1. Create a cluster.
2. Do `omnictl kubeconfig --cluster=***`
3. Either switch Omni context to use another Omni user OR simply edit KUBECONFIG file to create a duplicate user & context with a different name.
[...]

PR #2213, opened 2026-01-28
Use one cache directory for kubectl oidc-login per cluster+identity combination

filleokus:
This PR sets a custom cache directory for the kubectl oidc-login / https://github.com/int128/kubelogin based on the cluster+identity combination.

I ran in to https://github.com/siderolabs/omni/issues/1003 when testing out multiple accounts, and used the proposed solution from the issue https://github.com/int128/kubelogin/issues/29#issuecomment-1657057575 locally / manually and it worked great.

Thought it could be a simple fix do add it in the generated kubeconfig file.

### DIFF
--- a/internal/backend/runtime/kubernetes/kubernetes.go
+++ b/internal/backend/runtime/kubernetes/kubernetes.go
@@ -240,6 +240,7 @@ users:
         - get-token
         - --oidc-issuer-url={{ .EndpointOIDC }}
         - --oidc-client-id={{ .ClientID }}
+        - --cache-dir=~/.kube/cache/oidc-login/{{ .InstanceName }}-{{ .ClusterName }}{{- if .Identity }}-{{ .Identity }}{{ end }}
         - --oidc-extra-scope=cluster:{{ .ClusterName }}
         {{- range $option := .ExtraOptions }}
         - --{{ $option }}


oguzkilcan:
I think it would be better to add `cache-dir` as a cli flag with default value empty and set is on the template if its explicitly set. This way we would ensure backwards compatibility while giving users the option to set a custom cache directory whenever they want.

filleokus:
Hmm, yes it might be better. But should the user then specify a template string in the cli flag? Since the goal is to have dynamic cache dirs per identity to avoid them conflicting, setting a "static" cache dir won't help.

oguzkilcan:
It's a cli flag, a different setting can be provided with each execution.
Having custom cache dir for `kubelogin` is an extra feature. For example; users that has a single cluster do not actually need this change. 
Also having the ability to pass a custom cache dir would help with environments where writable directories are limited. This would remove the limit on dependency to have `HOME/.kube/cache/oidc-login` writable.

Having said these, I see your point on not putting extra effort to change the exact cache location for each `omnictl kubeconfig` execution. We can have two cli flags :
 1. `cache-dir`(string): changes the base location of the kubelogin cache. Default: `$HOME/.kube/cache/oidc-login`
 2. `use-dynamic-cache-dir`(bool): adds `cluster` and `user` info to `cache-dir`. Default: `false`.
 
This way, we can preserve existing functionality, have a way to change the cache-dir altogether, and use dynamic naming for the cache with a single flag.

filleokus:
Thanks for the feedback and taking the time! My apologies, I did not consider the `omnictl kubeconfig` case - I was focusing on when the user downloads the kubeconfig via the Omni Web UI, and assumed you with the cli flag ment a flag/argument to Omni itself to configure the behavior. 

For the  `omnictl kubeconfig` case it would of course work with just one flag as you say, but the two might be preferable? Do you think it would make sense to match those two flags in Omni as well? If so I would be happy to take a stab at it!

oguzkilcan:
Yes, it makes sense to also make them configurable for the Omni instance for the UI use-case. 

Feel free to tackle it.

oguzkilcan:
Thanks for the contribution. Functionality of this PR has been implemented on another PR.

PR #2215, opened 2026-01-29
fix(installation-media): fix bug when setting arch to amd64

Slessi:
Fix a bug when setting arch to AMD64 which was enum value 0 and was being omitted in responses to frontend.

### DIFF
--- a/client/api/omni/specs/virtual.proto
+++ b/client/api/omni/specs/virtual.proto
@@ -53,15 +53,16 @@ message AdvertisedEndpointsSpec {
 
 message PlatformConfigSpec {
   enum BootMethod {
-    UNKNOWN = 0;
+    UNKNOWN_BOOT = 0;
     DISK_IMAGE = 1;
     ISO = 2;
     PXE = 3;
   }
 
   enum Arch {
-    AMD64 = 0;
-    ARM64 = 1;
+    UNKNOWN_ARCH = 0;
+    AMD64 = 1;
+    ARM64 = 2;
   }
 
   string label = 1;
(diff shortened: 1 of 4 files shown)


client/api/omni/specs/virtual.proto
```diff
@@ -53,15 +53,16 @@ message AdvertisedEndpointsSpec {
...
+    UNKNOWN_BOOT = 0;
     DISK_IMAGE = 1;
     ISO = 2;
     PXE = 3;
   }
 
   enum Arch {
-    AMD64 = 0;
-    ARM64 = 1;
+    UNKNOWN_ARCH = 0;
```
smira, line 63: I _believe_ enum names are namespaced, so you can do simply `UNKNOWN` (it won't conflict)

Slessi: Maybe if they are defined outside of a `message` (?) but I tried that first and got a conflict.

Slessi: Nah even outside of a message, it fails:

```
0.241 /client/api/omni/specs/virtual.proto:63:3: "UNKNOWN" is already defined in "specs".
0.241 /client/api/omni/specs/virtual.proto:63:3: Note that enum values use C++ scoping rules, meaning that enum values are siblings of their type, not children of it.  Therefore, "UNKNOWN" must be unique within "specs", not just within "Arch".
```

client/api/omni/specs/virtual.proto
```diff
@@ -53,15 +53,16 @@ message AdvertisedEndpointsSpec {
...
+    AMD64 = 1;
+    ARM64 = 2;
   }
```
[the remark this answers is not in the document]
Slessi: Aware of this, but its an unreleased feature.

PR #2218, opened 2026-01-29
fix: stop generating node unique token in `NodeUniqueTokenStatus`

Unix4ever:
Instead, create the `PendingMachine` resource there and let `PendingMachineStatus` controller handle node unique token creation instead.

### DIFF
--- a/internal/backend/runtime/omni/controllers/omni/node_unique_token_status.go
+++ b/internal/backend/runtime/omni/controllers/omni/node_unique_token_status.go
@@ -7,28 +7,19 @@ package omni
 
 import (
 	"context"
-	"strings"
-	"time"
 
 	"github.com/cosi-project/runtime/pkg/controller"
 	"github.com/cosi-project/runtime/pkg/controller/generic/qtransform"
 	"github.com/cosi-project/runtime/pkg/safe"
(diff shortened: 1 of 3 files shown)


internal/backend/runtime/omni/omni.go
```diff
@@ -265,7 +265,9 @@ func NewRuntime(talosClientFactory *talos.ClientFactory, dnsService *dns.Service
 
 	if config.Config.Services.Siderolink.GetJoinTokensMode() != config.SiderolinkServiceJoinTokensModeLegacy {
 		qcontrollers = append(qcontrollers,
-			omnictrl.NewNodeUniqueTokenStatusController(),
+			omnictrl.NewNodeUniqueTokenStatusController(
+				st.defaultState,
```
utkuozdemir: any chance to avoid passing state into controller? why do we need this?

Unix4ever: Actually we can avoid it. Will just need to create it without owner, which we can actually do now

Unix4ever: Normally `PendingMachines` are created in the `provision.go`, so they have no owner.
And also `provision.go` has the cleanup runner for them.

PR #2220, opened 2026-01-29
feat(installation-media): replace edit naming with clone for installation media

Slessi:
Since in the edit flow for installation media we are actually saving new copies rather than editing, we can change the labelling from "Edit" to "Clone". Also changed the icon.


frontend/src/components/icons/IconArrowUpOnSquareStack.vue
Slessi, line 1: Open to ideas for different icons, got this one from https://heroicons.com/

Image

Unix4ever (approved):

PR #2231, opened 2026-01-30
feat: add a way to sync Kubernetes manifests in Omni

Unix4ever:
Manifests support two modes:
- `FULL` - Omni will keep the manifest in sync always.
- `ONE_TIME` - Omni will apply the manifest only if it doesn't exist. If the manifest is removed by hand and then changed in Omni it will be applied too.

Manifests are applied using service side apply, Omni now has two inventories: `omni-internal-inventory` and `omni-user-inventory`:

- User inventory will be used for user managed manifests.
- Internal one will be used for the manifests which are created by Omni controllers (workloadproxy, advanced healtcheck service).

Manifests also support setting namespace to all namespaced resources. It might be useful for the huge manifest files which are supplied without the namespace (similar to `kubectl apply -n namespace -f manifest.yaml`).


client/api/omni/specs/omni.proto
```diff
@@ -1580,3 +1580,35 @@ message ClusterMachineSecretsSpec {
...
+  enum Mode {
+    UNKNOWN = 0;
+    FULL = 1;
+    ONE_TIME = 2;
+  }
+
+  // Deprecated: use accessor methods GetUncompressedData/SetUncompressedData to manage this field.
+  bytes compressed_data = 1;
+  // Deprecated: use accessor methods GetUncompressedData/SetUncompressedData to manage this field.
+  string data = 2;
```
Orzelius, line 1594: I'm quite new to proto files. Is this the common pattern for avoiding direct property access?

client/api/omni/specs/omni.proto
```diff
@@ -1580,3 +1580,35 @@ message ClusterMachineSecretsSpec {
   // Rotation holds the status of the ongoing secrets rotation.
   Rotation rotation = 2;
 }
+
+message KubernetesManifestSpec {
+  enum Mode {
+    UNKNOWN = 0;
```
Orzelius, line 1586: Why have this unknown option?

client/pkg/omni/resources/omni/cluster_kubernetes_manifest_status.go
```diff
@@ -0,0 +1,48 @@
...
+	"github.com/cosi-project/runtime/pkg/resource"
+	"github.com/cosi-project/runtime/pkg/resource/meta"
+	"github.com/cosi-project/runtime/pkg/resource/protobuf"
+	"github.com/cosi-project/runtime/pkg/resource/typed"
+
+	"github.com/siderolabs/omni/client/api/omni/specs"
+	"github.com/siderolabs/omni/client/pkg/omni/resources"
+)
+
+// NewClusterKubernetesManifestsStatus creates new cluster machine diagnostics resource.
```
Orzelius:
```suggestion
// NewClusterKubernetesManifestsStatus creates new cluster kubernetes manifest status resource.
```

client/pkg/omni/resources/omni/kubernetes_manifest.go
```diff
@@ -0,0 +1,48 @@
...
+	"github.com/cosi-project/runtime/pkg/resource"
+	"github.com/cosi-project/runtime/pkg/resource/meta"
+	"github.com/cosi-project/runtime/pkg/resource/protobuf"
+	"github.com/cosi-project/runtime/pkg/resource/typed"
+
+	"github.com/siderolabs/omni/client/api/omni/specs"
+	"github.com/siderolabs/omni/client/pkg/omni/resources"
+)
+
+// NewKubernetesManifest creates new cluster machine diagnostics resource.
```
Orzelius:
```suggestion
// NewKubernetesManifest creates new cluster kubernetes manifest resource.
```

internal/backend/runtime/omni/controllers/omni/kubernetes/cluster_manifest_status.go
```diff
@@ -0,0 +1,444 @@
...
+
+// NewClusterManifestsStatusController initializes ClusterKubernetesManifestsStatusController.
+func NewClusterManifestsStatusController() *ClusterManifestsStatusController {
+	ctrl := &ClusterManifestsStatusController{}
+
+	ctrl.QController = qtransform.NewQController(
+		qtransform.Settings[*omni.Cluster, *omni.ClusterKubernetesManifestsStatus]{
+			Name: "ClusterKubernetesManifestsStatusController",
+			MapMetadataFunc: func(manifest *omni.Cluster) *omni.ClusterKubernetesManifestsStatus {
+				return omni.NewClusterKubernetesManifestsStatus(manifest.Metadata().ID())
```
Orzelius:
```suggestion
			MapMetadataFunc: func(cluster *omni.Cluster) *omni.ClusterKubernetesManifestsStatus {
				return omni.NewClusterKubernetesManifestsStatus(cluster.Metadata().ID())
```

internal/backend/runtime/omni/controllers/omni/kubernetes/cluster_manifest_status.go
```diff
@@ -0,0 +1,444 @@
...
+	discoveryClient, err := discovery.NewDiscoveryClientForConfig(client.Config)
+	if err != nil {
+		return fmt.Errorf("error building discovery client: %w", err)
+	}
+
+	dc := memory.NewMemCacheClient(discoveryClient)
+
+	mapper := restmapper.NewDeferredDiscoveryRESTMapper(memory.NewMemCacheClient(dc))
+
+	inventories := make(map[string]inventory.Inventory, 2)
```
Orzelius: I think I'm missing something but this inventory part seems to be never used

internal/backend/runtime/omni/controllers/omni/kubernetes/cluster_manifest_status.go
```diff
@@ -0,0 +1,444 @@
...
+			mappers.MapClusterResourceToLabeledResources[*omni.Cluster](),
+		),
+	)
+
+	return ctrl
+}
+
+// nolint:gocyclo,cyclop
+func (ctrl *ClusterManifestsStatusController) transform(ctx context.Context, r controller.Reader, logger *zap.Logger,
+	cluster *omni.Cluster, kubernetesManifestStatus *omni.ClusterKubernetesManifestsStatus,
```
Orzelius:
```suggestion
	cluster *omni.Cluster, kubernetesManifestsStatus *omni.ClusterKubernetesManifestsStatus,
```

internal/backend/runtime/omni/controllers/omni/kubernetes/cluster_manifest_status.go
```diff
@@ -0,0 +1,444 @@
...
+
+	inventories := make(map[string]inventory.Inventory, 2)
+
+	for _, id := range []string{
+		omniconsts.SSAOmniInternalInventory,
+		omniconsts.SSAOmniUserInventory,
+	} {
+		var inventory inventory.Inventory
+
+		inventory, err = getInventory(ctx, client.Config, mapper, dc, id)
```
Orzelius: This will error if the inventory doesn't exist yet

internal/backend/runtime/omni/controllers/omni/kubernetes/cluster_manifest_status.go
```diff
@@ -0,0 +1,444 @@
...
+	logger.Info("sync manifests full", zap.String("inventory", inventory), zap.Int("count", len(objects)))
+
+	panichandler.Go(func() {
+		errCh <- manifests.SyncSSA(ctx, objects, client.Config, resultCh, ssaOpts)
+	}, logger)
+
+	for {
+		select {
+		case result := <-resultCh:
+			id := result.ObjectID.Namespace + "/" + result.ObjectID.Name
```
Orzelius: This should also include the Kind otherwise you risk conflicting with Objects that have the same name and ns

internal/backend/runtime/omni/controllers/omni/kubernetes/cluster_manifest_status.go
```diff
@@ -0,0 +1,444 @@
...
+				clusterKubernetesManifestsStatus.TypedSpec().Value.Manifests[id] = &specs.ClusterKubernetesManifestsStatusSpec_ManifestStatus{
+					LastError: result.Error.Error(),
+					Phase:     specs.ClusterKubernetesManifestsStatusSpec_ManifestStatus_FAILED,
+					Mode:      specs.KubernetesManifestSpec_FULL,
+				}
+
+				continue
+			}
+
+			if result.Type == event.PruneType || result.Type == event.RolloutType {
```
Orzelius: The API will change with the rewrite of the go-kubernetes helpers, but currently this is missing the `WaitType` check. I think a switch statement would be better here

internal/backend/runtime/omni/controllers/omni/kubernetes/cluster_manifest_status.go
```diff
@@ -0,0 +1,444 @@
...
+
+		updated++
+
+		clusterKubernetesManifestsStatus.TypedSpec().Value.Manifests[id] = &specs.ClusterKubernetesManifestsStatusSpec_ManifestStatus{
+			Phase: specs.ClusterKubernetesManifestsStatusSpec_ManifestStatus_APPLIED,
+			Mode:  specs.KubernetesManifestSpec_ONE_TIME,
+		}
+	}
+
+	logger.Info("sync manifests one time", zap.Int("count", len(objects)), zap.Int("updated", updated))
```
Orzelius: perhaps this is a bit more clear
```suggestion
	logger.Info("completed one time manifest sync", zap.Int("count", len(objects)), zap.Int("updated", updated))
```

client/api/omni/specs/omni.proto
Unix4ever: We use this kind of approach only for compressed resources.

client/api/omni/specs/omni.proto
Unix4ever: It is better to use 0 as unknown, because when protobuf is encoded to JSON `0` will make the field to be omitted.

internal/backend/runtime/omni/controllers/omni/kubernetes/cluster_manifest_status.go
Unix4ever: I think I'll just change it to:

```golang
			if result.Type != event.ApplyType {
				continue
			}
```

internal/backend/runtime/omni/controllers/omni/kubernetes/cluster_manifest_status.go
Unix4ever: right, I forgot to drop it

internal/backend/runtime/omni/controllers/omni/kubernetes/cluster_manifest_status.go
Unix4ever: I think this is handled by:

```golang
		if apierrors.IsNotFound(err) {
			return &inventory.SingleObjectInventory{}, nil
		}
```

internal/backend/runtime/omni/controllers/omni/kubernetes/cluster_manifest_status.go
Unix4ever: Yeah, updated to use `ObjectID.String()`

client/api/omni/specs/omni.proto
```diff
@@ -1580,3 +1580,35 @@ message ClusterMachineSecretsSpec {
...
+    FULL = 1;
```
oguzkilcan, line 1587: Should we have a limit on how many attempts will be made to sync, when sync fails for some reason? To To protect Omni from attempting to sync a broken manifest indefinitely.

client/pkg/omni/resources/omni/cluster_kubernetes_manifest_status.go
```diff
@@ -0,0 +1,48 @@
...
+// ClusterKubernetesManifestsStatusExtension provides auxiliary methods for ClusterKubernetesManifestsStatus resource.
+type ClusterKubernetesManifestsStatusExtension struct{}
+
+// ResourceDefinition implements [typed.Extension] interface.
+func (ClusterKubernetesManifestsStatusExtension) ResourceDefinition() meta.ResourceDefinitionSpec {
+	return meta.ResourceDefinitionSpec{
+		Type:             ClusterKubernetesManifestsStatusType,
+		Aliases:          []resource.Type{},
+		DefaultNamespace: resources.DefaultNamespace,
+		PrintColumns:     []meta.PrintColumn{},
```
oguzkilcan: I'd be nice to have sync status as extra column.

client/api/omni/specs/omni.proto
Unix4ever: the current limit for broken manifests is 1 :) 
Invalid manifests errors are saved to the status and `nil` is returned from the controller.

client/pkg/omni/resources/omni/cluster_kubernetes_manifest_status.go
Unix4ever: I think currently we don't have a good option for the global status. But can figure something out.

rothgar:
How is someone intended to use this initially? Is this only part of cluster templates or can it be added to any cluster in Omni?

Unix4ever:
> How is someone intended to use this initially? Is this only part of cluster templates or can it be added to any cluster in Omni?

I'm going to make it part of the cluster templates in the next PR.
Also will move workloadproxy manifests to this system.

client/pkg/omni/resources/omni/cluster_kubernetes_manifest_status.go
Unix4ever: added out of sync and total counts there

Unix4ever:
Pretty much rewrote the controller. Made the status updates more frequent, fixed more issues there:
- mappers for the LoadbalancerStatus and ClusterStatus were incorrect.
- reduced the prune/rollout timeouts to make the controller retry and update the status.

The status now contains more phases as well as the information about the resource being created.

Example:

```yaml
metadata:
    namespace: default
    type: ClusterKubernetesManifestsStatuses.omni.sidero.dev
    id: talos-default
    version: 34
    owner: ClusterKubernetesManifestsStatusController
    phase: running
    created: 2026-02-06T12:38:46Z
    updated: 2026-02-06T13:20:00Z
spec:
    manifests:
        default_nginx-deployment_apps_Deployment:
            phase: 3
            lasterror: ""
            mode: 1
            kind: Deployment
            name: nginx-deployment
            namespace: default
    errors: {}
    outofsync: 0
    total: 1
```

client/api/omni/specs/kubernetes_manifest.go
```diff
@@ -0,0 +1,161 @@
...
+	if !compress || len(data) < config.MinThreshold {
+		x.Data = string(data)
+		x.CompressedData = nil
+
+		return nil
+	}
+
+	compressedData := doCompress(data, config)
+
+	x.Data = ""
```
utkuozdemir, line 104: What if we compress them unconditionally? Our old compressing resources were done kinda that way for kind of backwards compatibility. Maybe we can just do this simpler.

client/api/omni/specs/omni.proto
```diff
@@ -1580,3 +1580,43 @@ message ClusterMachineSecretsSpec {
...
+
+    Phase phase = 1;
+    string last_error = 2;
+    KubernetesManifestSpec.Mode mode = 3;
+    string kind = 4;
+    string name = 5;
+    string namespace = 6;
+  }
+
+  map<string, ManifestStatus> manifests = 1;
```
utkuozdemir, line 1618: Curious, does this mean there is no order of apply between manifests? I guess it'd probably work anyway, our code would keep trying to apply them, but normally there can be an order in k8s manifests - an example is creating a CRD and a CR. If so, we could maybe turn these into slices instead?

Unix4ever: This is the status resource.

While the manifests are ordered: separate KubernetesManifest resources are ordered by COSI using the ID, and it's possible to put many manifests to a single resource delimiting by `---`, then it's a slice.

client/api/omni/specs/kubernetes_manifest.go
Unix4ever: It has more than that. It also doesn't compress if the manifest size is small. So I thought it kinda makes sense to use the same approach here.

client/pkg/omni/resources/omni/labels.go
```diff
@@ -49,6 +49,10 @@ const (
 	// tsgen:LabelSystemPatch
 	LabelSystemPatch = SystemLabelPrefix + "system-patch"
 
+	// LabelSystemManifest marks the manifest as the system manifest, so it shouldn't be editable by the user.
+	// tsgen:LabelSystemManifest
+	LabelSystemManifest = SystemLabelPrefix + "system-manifest"
```
Unix4ever, line 54: Show the system manifest flag inside the `ClusterKubernetesManifestsStatus`.

internal/backend/runtime/omni/controllers/omni/kubernetes/cluster_manifest_status.go
```diff
@@ -0,0 +1,621 @@
...
+
+	for inventory, objects := range manifestsByInventory {
+		if err = ctrl.sync(ctx, logger, objects, inventory, client, clusterKubernetesManifestsStatus); err != nil {
+			errs = errors.Join(errs, err)
+		}
+	}
+
+	if len(oneTimeManifests) > 0 {
+		if err = ctrl.syncOneTime(ctx, logger, oneTimeManifests, client, clusterKubernetesManifestsStatus); err != nil {
+			errs = errors.Join(errs, err)
```
Unix4ever, line 287: wrap the error to add more context

internal/backend/runtime/omni/controllers/omni/kubernetes/cluster_manifest_status.go
```diff
@@ -0,0 +1,621 @@
...
+	var errs error
+
+	// update the status early to display pending manifests, as SSA is a long running operation
+	if err = ctrl.updateStatus(ctx, r, clusterKubernetesManifestsStatus); err != nil {
+		return err
+	}
+
+	for inventory, objects := range manifestsByInventory {
+		if err = ctrl.sync(ctx, logger, objects, inventory, client, clusterKubernetesManifestsStatus); err != nil {
+			errs = errors.Join(errs, err)
```
Unix4ever, line 281: wrap the error to add more context

client/api/omni/specs/omni.proto
```diff
@@ -1580,3 +1580,43 @@ message ClusterMachineSecretsSpec {
...
+      DELETING = 4;
+      FAILED = 5;
+    }
+
+    Phase phase = 1;
+    string last_error = 2;
+    KubernetesManifestSpec.Mode mode = 3;
+    string kind = 4;
+    string name = 5;
+    string namespace = 6;
```
Unix4ever, line 1615: add last modified date.

internal/backend/runtime/omni/controllers/omni/kubernetes/cluster_manifest_status.go
```diff
@@ -0,0 +1,621 @@
...
+
+func validationError(err error) bool {
+	var validationError *validation.Error
+
+	return errors.As(err, &validationError)
+}
+
+func isFatalError(err error) bool {
+	return apierrors.IsInvalid(err) || apierrors.IsBadRequest(err) || apierrors.IsForbidden(err) ||
+		apierrors.IsRequestEntityTooLargeError(err) || webhookError(err) || validationError(err) || strings.Contains(err.Error(), "reconcile failed")
```
Unix4ever, line 620: Consider making it less restricted

internal/backend/runtime/omni/controllers/omni/kubernetes/cluster_manifest_status.go
```diff
@@ -0,0 +1,621 @@
...
+				}
+
+				errs = errors.Join(result.Error)
+
+				continue
+			}
+
+			//nolint:exhaustive
+			switch result.Type {
+			case event.PruneType:
```
Unix4ever, line 465: Try to use inventory to track prune/delete events.

internal/backend/runtime/omni/controllers/omni/kubernetes/cluster_manifest_status.go
```diff
@@ -0,0 +1,621 @@
...
+		if o != nil {
+			setManifestStatus(
+				specs.ClusterKubernetesManifestsStatusSpec_ManifestStatus_APPLIED,
+				"",
+			)
+
+			continue
+		}
+
+		_, err = dr.Create(ctx, manifest, v1.CreateOptions{FieldManager: constants.KubernetesFieldManagerName})
```
Unix4ever, line 571:
```suggestion
		_, err = dr.Apply(ctx, manifest, v1.CreateOptions{FieldManager: constants.KubernetesFieldManagerName})
```

PR #2233, opened 2026-01-30
test: add unit and e2e tests to the helm chart

kevintijssen:
# 1
fix: unify helm chart services and ingresses, remove JSON schema

Merge the separate gRPC service and ingress into the main service using h2c (HTTP/2 cleartext), since the backend already demuxes HTTP and gRPC on the same port. Split out the Kubernetes proxy into its own service because it doesn't support h2c. Remove the dedicated gRPC GatewayAPI route. Add hostNetwork support for environments where NET_ADMIN is insufficient for WireGuard. Remove the JSON schema in favor of values file comments and validation template.

# 2
test: add unit and e2e tests to the helm chart

Add helm unit tests (via helm-unittest) covering services, ingresses, HTTPRoutes, secrets, PrometheusRules and ServiceAccounts. Add a helm-based e2e test workflow that deploys Omni on a Talos cluster with Traefik and etcd, runs integration tests including workload proxy, and verifies the full stack end-to-end. Add a configurable TestOptions struct to the workload proxy test to allow running with smaller scale in helm e2e.


smira:
@kevintijssen please make your org membership public

kevintijssen:
> @kevintijssen please make your org membership public

Done

deploy/helm/v2/e2e/testdata/omni.asc
utkuozdemir, line 1: @kevintijssen This file is too large. Is this intentional? If not, I'm gonna replace it with a smaller one.

utkuozdemir:
Closes #2278.

deploy/helm/v2/e2e/testdata/omni.asc
kevintijssen: We can generated every time we're running the pipeline or make it smaller. What do you prefer?

utkuozdemir: No need for generating it each time, having a static one is fine. I'm already on the PR, gonna make it smaller. Thanks.

deploy/helm/v2/e2e/tests/01-omni/06-test-omni-ingress.yaml
```diff
@@ -0,0 +1,98 @@
...
+
+              echo ""
+              echo "Create MachineClass"
+              cat <<EOF > /tmp/mc.yaml
+              metadata:
+                namespace: default
+                type: MachineClasses.omni.sidero.dev
+                id: test-mc
+              spec:
+                matchlabels:
```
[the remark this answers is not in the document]
utkuozdemir: it's correct as-is - it's not k8s, it's an omni resource.

hack/test/integration-qemu.sh
```diff
@@ -81,7 +81,7 @@ fi
 
 # Create a Talos cluster to be imported.
 if [ "${CREATE_TALOS_CLUSTER:-false}" == "true" ]; then
-  create_talos_cluster name=test-cluster-import cp_count=3 wk_count=3 cidr=172.28.0.0/24 talos_version="${QEMU_TALOS_VERSION}"
+  create_talos_cluster name=test-cluster-import cp_count=3 wk_count=3 cidr=172.28.0.0/24 talos_version="${QEMU_TALOS_VERSION}" true false
```
oguzkilcan: create_talos_cluster takes named arguments. You are passing here them without names.

PR #2246, opened 2026-02-02
feat(installation-media): replace modal with wizard by default

### LINKED ISSUE #2009: Remove feature flag + add E2E test
Remove the feature flag to enable installation media for all. This will require updating the current E2E test which downloads media using the modal.

Slessi:
- Replace the Installation Media modal with the wizard by default
- Add an E2E test to cover the wizard instead of the modal
- Update some components for better a11y, to help with testing

Closes #2009

E2E test in action:

https://github.com/user-attachments/assets/9a88da75-ade3-4c39-aa08-5f74a76c1316

### DIFF
--- a/frontend/src/components/common/Checkbox/TCheckbox.vue
+++ b/frontend/src/components/common/Checkbox/TCheckbox.vue
@@ -5,7 +5,8 @@ Use of this software is governed by the Business Source License
 included in the LICENSE file.
 -->
 <script setup lang="ts">
-import TAnimation from '@/components/common/Animation/TAnimation.vue'
+import { CheckboxIndicator, CheckboxRoot } from 'reka-ui'
+
 import TIcon from '@/components/common/Icon/TIcon.vue'
 
 type Props = {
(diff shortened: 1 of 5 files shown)


frontend/src/views/omni/Settings/JoinTokens.vue
```diff
@@ -191,7 +189,7 @@ const openDeleteToken = (token: string) => {
                 </TActionsBoxItem>
                 <TActionsBoxItem
                   icon="long-arrow-down"
-                  @select="() => openDownloadInstallationMedia(item.metadata.id!)"
+                  @select="() => openDownloadInstallationMedia()"
                 >
                   Download Installation Media
                 </TActionsBoxItem>
```
Slessi: not sure it makes sense to keep this one, its a dropdown item on the join tokens page, before it would open the modal with i assume the specific join token pre-selected, but what would i do here? always open the wizard to create a new media and preselect a token? Might distract you from being able to re-use saved media.

Image

rothgar: I'd be fine removing that link in the drop down.

Unix4ever: Yeah, can drop it

frontend/src/components/common/Checkbox/TCheckbox.vue
```diff
@@ -21,19 +22,18 @@ const checked = defineModel<boolean>({ default: false })
...
   <label class="inline-flex cursor-pointer items-center gap-2 has-disabled:cursor-not-allowed">
-    <input v-model="checked" type="checkbox" :disabled="disabled" class="peer sr-only fixed" />
-
-    <div
-      class="flex size-3.5 items-center justify-center rounded-xs border border-naturals-n7 peer-checked:border-primary-p6 peer-checked:bg-primary-p6 peer-disabled:border-naturals-n5 peer-disabled:bg-naturals-n4"
+    <CheckboxRoot
+      v-model="checked"
+      :disabled
+      class="flex size-3.5 items-center justify-center rounded-xs border border-naturals-n7 transition-colors data-disabled:border-naturals-n5 data-disabled:bg-naturals-n4 data-[state=checked]:border-primary-p6 data-[state=checked]:bg-primary-p6"
     >
```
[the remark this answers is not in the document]
Slessi: Untrue, `disabled` is no value, `:disabled` is equivalent to assigning the variable disabled to it, i.e. `:disabled="disabled"`, which is from props.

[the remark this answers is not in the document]
Slessi: Untrue, `:disabled` is sugar for `:disabled="disabled"` which itself is sugar for `v-bind:disabled="disabled"` which means it is being assigned the `disabled` variable, in this case from the props. If there was no semicolon (`:`) then this would be correct.

Unix4ever (approved):

Unix4ever:
:100:

PR #2258, opened 2026-02-04
feat: move omni schematic cache to ephemeral

### LINKED ISSUE #2240: [bug] (Re)create schematic ID by Omni if not exist

### Current Behavior

If you are running a self-hosted Omni instance you can also use a self-hosted Image-Factory. When you create a "Installation Media" it will create a schematics id and stores it in your registry. Omni will from that point never recreate the schematic id. So if you lose the registry data Omni won't recreate the schematic id anymore

### Expected Behavior

Recreate the schematic when not found.

### Steps To Reproduce

* Setup self-hosted Omni, Image-factory, registry
* Create a "Install Media" without extensions
* Delete registry data
* Restart Image-factory
* Try to download the same "Install Media"

Chrome

### Anything else?

_No response_

Slessi:
Move the factory schematic cache in omni from the default namespace to the ephemeral one, so that it is only stored in memory. Also includes a migration to clean-up the cache from the default namespace.

Resolves #2240

### DIFF
--- a/internal/backend/runtime/omni/migration/migrations.go
+++ b/internal/backend/runtime/omni/migration/migrations.go
@@ -231,3 +231,28 @@ func changeClusterMachineConfigPatchesOwner(ctx context.Context, st state.State,
 
 	return nil
 }
+
+func moveSchematicCacheToEphemeral(ctx context.Context, st state.State, logger *zap.Logger, _ migrationContext) error {
+	kind := resource.NewMetadata(
+		resources.DefaultNamespace,
+		omni.SchematicType,
+		"",
+		resource.VersionUndefined,
+	)
+
+	list, err := st.List(ctx, kind)
+	if err != nil {
+		return err
+	}
+
+	logger.Info("cleaning up schematics from default namespace", zap.Int("count", len(list.Items)))
+
+	for _, item := range list.Items {
+		err = st.Destroy(ctx, item.Metadata(), state.WithDestroyOwner(item.Metadata().Owner()))
+		if err != nil && !state.IsNotFoundError(err) {
+			return err
+		}
+	}
+
+	return nil
+}
--- a/client/pkg/omni/resources/omni/schematic.go
+++ b/client/pkg/omni/resources/omni/schematic.go
@@ -17,7 +17,7 @@ import (
 // NewSchematic creates new schematic resource.
 func NewSchematic(id resource.ID) *Schematic {
 	return typed.NewResource[SchematicSpec, SchematicExtension](
-		resource.NewMetadata(resources.DefaultNamespace, SchematicType, id, resource.VersionUndefined),
+		resource.NewMetadata(resources.EphemeralNamespace, SchematicType, id, resource.VersionUndefined),
 		protobuf.NewResourceSpec(&specs.SchematicSpec{}),
 	)
 }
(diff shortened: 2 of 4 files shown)


Unix4ever:
We can't simply drop the cache I think. For example `EnsureSchematic` is called for each registered machines in the machine status controller [see this](https://github.com/siderolabs/omni/blob/48fecf007e96fe97ca5ddb68867ae3f12e828ab1/internal/backend/runtime/omni/controllers/omni/machine_status.go#L511).
If we drop cache it might put burden on the factory. It will call factory for each machine on every Omni restart. Not sure if that's what we want.

If we drop the cache, we'll need to delete the resources in the migration.

We can actually keep the cache and make it include the factory URL in the cache key, then it will work properly.

Slessi:
@Unix4ever so like, we will cache schematics as currently, but they will be related to specific factory URL?

That will fix swapping out factories, but in kevin's case he would have run into the same problem because he was losing factory state on the same factory URL.

Unix4ever:
> @Unix4ever so like, we will cache schematics as currently, but they will be related to specific factory URL?

Yeah, something like that.

> That will fix swapping out factories, but in kevin's case he would have run into the same problem because he was losing factory state on the same factory URL.

Like I said, if we're fine with Omni doing lots of requests to the IF, we can drop the cache.

smira:
Does it make sense if we cache in memory? So on restart the cache is gone?

Unix4ever:
> Does it make sense if we cache in memory? So on restart the cache is gone?

Yep, that can be an option. @Slessi to make it cached in memory we can change the namespace of the `Schematic` resource to be `ephemeral`.

utkuozdemir:
> > Does it make sense if we cache in memory? So on restart the cache is gone?
> 
> Yep, that can be an option. @Slessi to make it cached in memory we can change the namespace of the `Schematic` resource to be `ephemeral`.

Should we cleanup the existing schematics? Via a migration or something?

Slessi:
Updated PR to just switch namespace to ephemeral and includes a migration to wipe it from default 🚀

internal/backend/runtime/omni/migration/migrations.go
```diff
@@ -231,3 +231,27 @@ func changeClusterMachineConfigPatchesOwner(ctx context.Context, st state.State,
 
 	return nil
 }
+
+func moveSchematicCacheToEphemeral(ctx context.Context, st state.State, logger *zap.Logger, _ migrationContext) error {
```
utkuozdemir, line 235: can we write a test for the migration? we do it for all of them.

internal/backend/runtime/omni/migration/migrations.go
```diff
@@ -231,3 +231,27 @@ func changeClusterMachineConfigPatchesOwner(ctx context.Context, st state.State,
...
+
+	list, err := st.List(ctx, kind)
+	if err != nil {
+		return err
+	}
+
+	logger.Info("cleaning up schematics from default namespace", zap.Int("count", len(list.Items)))
+
+	for _, item := range list.Items {
+		if err = st.Destroy(ctx, item.Metadata()); err != nil {
```
utkuozdemir: Here, to be on the safe side, we can:
- do the delete with the owner (should always be empty, but to be extra safe `state.WithDestroyOwner(item.Metadata().Owner())`)
- Do not fail if the returned error is `state.IsNotFoundError`. 

I know it's just defensive code, but a failing migration on upgrade is not fun, so at least I write my migrations like that.

PR #2284, opened 2026-02-06
feat: allow unrejecting machines from the ui

### LINKED ISSUE #2243: Rejected machines should be shown in the UI to be un-rejected
### Current Behavior

When a machine has been rejected from pending machines there's no way to unreject them from the UI

### Expected Behavior

There should be a list or button of previously rejected nodes that you can delete so they can go back through a provisioning flow

Slessi:
Allow unrejecting previously rejected machines from the UI.

Closes #2243

### DIFF
--- a/client/pkg/omni/resources/infra/machine.go
+++ b/client/pkg/omni/resources/infra/machine.go
@@ -84,3 +84,28 @@ func (MachineExtension) ResourceDefinition() meta.ResourceDefinitionSpec {
 		},
 	}
 }
+
+// Make implements [typed.Maker] interface.
+func (MachineExtension) Make(_ *resource.Metadata, spec *MachineSpec) any {
+	return (*machineAux)(spec)
+}
+
+type machineAux MachineSpec
+
+func (m *machineAux) Match(searchFor string) bool {
+	val := m.Value
+
+	if searchFor == "pending" {
+		return val.GetAcceptanceStatus() == specs.InfraMachineConfigSpec_PENDING
+	}
+
+	if searchFor == "accepted" {
+		return val.GetAcceptanceStatus() == specs.InfraMachineConfigSpec_ACCEPTED
+	}
+
+	if searchFor == "rejected" {
+		return val.GetAcceptanceStatus() == specs.InfraMachineConfigSpec_REJECTED
+	}
+
+	return false
+}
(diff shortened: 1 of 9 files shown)


client/pkg/omni/resources/infra/machine.go
Slessi, line 1: @Unix4ever WDYT about this implementation?

On frontend, we list pending machines by finding `InfraMachine` with `LabelMachinePendingAccept` label. Then we accept/reject by updating `acceptance_status` on `InfraMachineConfig`.

I noticed both of those resources have an `acceptance_status` though, so I added this `searchFor` here which should let me get a list of rejected or pending. IDK if for sure those statuses will be in sync though.

frontend/src/methods/machine.ts
```diff
@@ -175,6 +175,12 @@ export const rejectMachine = async (machine: string) => {
   })
 }
 
+export const unrejectMachine = async (machine: string) => {
+  await updateInfraMachineConfig(machine, (r: Resource<InfraMachineConfigSpec>) => {
+    r.spec.acceptance_status = InfraMachineConfigSpecAcceptanceStatus.PENDING
+  })
+}
+
```
Slessi: Do I need to manually re-add the `LabelMachinePendingAccept` in this case? Or will it automatically return to the machine? 👀

client/pkg/omni/resources/infra/machine.go
Unix4ever: I think that's a correct solution 👍

frontend/src/methods/machine.ts
Unix4ever: I think `LabelMachinePendingAccept` appears on the `infra.Machine`, while you're changing `InfraMachineConfig` here. So you don't need to add the label there.

client/pkg/omni/resources/infra/machine.go
Slessi: Not sure if I missed some critical detail in the implementation for this, but it doesn't work great with watches. When `acceptance_status` changes, it doesn't send any events to the watch, so the machine stays on screen. 🤷 

Also, updating `acceptance_status` from accepted/rejected -> pending is not allowed by omni validation rules. I could change it, to allow at least rejected -> pending, but maybe that is not desirable.

I tried doing it by deleting the link like when machines are destroyed, hoping they would be reconciled by the provider back into a pending state, but they come back as rejected.

PR #2285, opened 2026-02-06
refactor(frontend): reimplement tabs using reka-ui

Slessi:
Re-implement tabs using reka-ui for a11y.

https://github.com/user-attachments/assets/e4761523-e733-4656-9eb9-56639562a76f


smira (approved):
🆒

PR #2286, opened 2026-02-06
refactor: remove global runtime registry, inject runtimes to services

utkuozdemir:
As an effort to clean up the codebase, remove the global runtime registry and lookups, which was an antipattern: hiding the service dependencies, making the testing and reasoning harder.


smira (approved):

internal/backend/k8sproxy/multiplex.go
```diff
@@ -22,15 +22,18 @@ import (
...
 
 	"github.com/siderolabs/omni/client/api/common"
-	"github.com/siderolabs/omni/internal/backend/runtime"
-	"github.com/siderolabs/omni/internal/backend/runtime/kubernetes"
 	"github.com/siderolabs/omni/internal/pkg/ctxstore"
 )
 
+type kubeconfigGetter interface {
+	GetKubeconfig(ctx context.Context, cluster *common.Context) (*rest.Config, error)
+}
```
Unix4ever, line 30: We could probably define these interfaces in a single common place instead of redefining them everywhere. But it's not a big deal anyways.

Unix4ever (approved):

internal/backend/k8sproxy/multiplex.go
utkuozdemir: Actually I thought about it, but kinda aimed to follow the principle of "smaller interfaces, declared at the consumer side". So, in a single package, I didn't create multiple interfaces for different functions, but left it at that level, didn't want to pull them more up.

PR #2296, opened 2026-02-10
feat: update machinery doc links

### LINKED ISSUE #1956: Update docs links for SBCs & cloud providers.
The documentation links for SBCs and cloud providers are using the paths from the old docs page. We should update these to the new links. This will also require an update in at least omni and image factory. Unsure if they are used elsewhere.

See: https://github.com/siderolabs/talos/blob/d4309d7b1aec9d2852173fd704b09dfabe2cf217/pkg/machinery/platforms/platforms.go#L157

e.g.

- we get `/talos-guides/install/cloud-platforms/aws/`
- we create https://www.talos.dev/v1.11/talos-guides/install/cloud-platforms/aws/
- we redirect to https://docs.siderolabs.com/talos/v1.11/platform-specific-installations/cloud-platforms/aws

N.B. doing it blindly like this would not work as link paths are not 1:1
https://docs.siderolabs.com/talos/v1.11/talos-guides/install/cloud-platforms/aws/

Slessi:
Bump Talos version and update machinery doc links to point to docs.siderolabs.com.

Closes #1956

### DIFF
--- a/client/go.mod
+++ b/client/go.mod
@@ -28,8 +28,8 @@ require (
 	github.com/siderolabs/image-factory v1.0.0
 	github.com/siderolabs/proto-codec v0.1.3
 	github.com/siderolabs/siderolink v0.3.15
-	github.com/siderolabs/talos v1.13.0-alpha.0.0.20260126150111-859d3f03c444
-	github.com/siderolabs/talos/pkg/machinery v1.13.0-alpha.0.0.20260126150111-859d3f03c444
+	github.com/siderolabs/talos v1.13.0-alpha.1.0.20260210134859-406b8c83c9b3
+	github.com/siderolabs/talos/pkg/machinery v1.13.0-alpha.1.0.20260210134859-406b8c83c9b3
 	github.com/spf13/cobra v1.10.2
 	github.com/stretchr/testify v1.11.1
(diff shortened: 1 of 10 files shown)


client/go.mod
```diff
@@ -28,8 +28,8 @@ require (
 	github.com/siderolabs/image-factory v1.0.0
 	github.com/siderolabs/proto-codec v0.1.3
 	github.com/siderolabs/siderolink v0.3.15
-	github.com/siderolabs/talos v1.13.0-alpha.0.0.20260126150111-859d3f03c444
-	github.com/siderolabs/talos/pkg/machinery v1.13.0-alpha.0.0.20260126150111-859d3f03c444
+	github.com/siderolabs/talos v1.13.0-alpha.1.0.20260210134859-406b8c83c9b3
```
smira, line 31: I guess you might need `go mod tidy`, there should be more deps updated transitively

smira:
For 

```
talosCli.ImageList(client.WithNode(ctx, nodeAddress), common.ContainerdNamespace_NS_CRI)
```

Add `//nolint:staticcheck`, Omni needs to support old Talos versions

client/go.mod
Slessi: Yea I did that, didn't get any other changes

PR #2298, opened 2026-02-10
fix: fix memory leak in the config diff compute code

Unix4ever:
There was a missing `buffer.Free` method call.
Also add shortcut to skip generating the diff if both configs are equal.
And if the previousDiff is empty, use simpler implementation for calculating the diff as the whole config will be added.


client/api/omni/specs/cluster_machine_config.go
```diff
@@ -21,7 +21,7 @@ func (x *ClusterMachineConfigSpec) MarshalJSON() ([]byte, error) {
 		return nil, err
 	}
 
-	defer buffer.Free()
+	defer buffer.Close()
```
utkuozdemir: It was my idea but I just checked my IDE and golangci-lint, but it seems my IDE doesn't warn in either, and there's no generic linter for checking `Close()` calls - there are only specific ones for SQL rows and HTTP bodies. So actually this doesn't benefit us in any way (btw, this doesn't implement `io.Closer` either, but even if it did, it wouldn't help).

I'd say, it's to you to keep it this way or revert back to Free.

internal/backend/runtime/omni/controllers/omni/machineconfig/reconciliation_context.go
```diff
@@ -292,3 +301,24 @@ func computeDiff(previousData, newData []byte) (string, error) {
 
 	return diffStr, nil
 }
+
+// Helper to format a string as a diff addition block
+func formatAsAdditions(content string) string {
```
utkuozdemir: Didn't the diff code do this efficiently by itself?

utkuozdemir: Also, should we add a unit test for this, in which we compare the string this produces with the one the normal diff code would produce to be exactly equal?

client/api/omni/specs/cluster_machine_config.go
smira: I think .Close() is somewhat confusing here (as it's an I/O) operation. 

.Free() is better (or .Release()) as I guess this all about a buffer pool of some sort?

internal/backend/runtime/omni/controllers/omni/machineconfig/reconciliation_context.go
smira: yep, I would like to see a benchmark that proves that we need to special-case this.

Also in general, in terms of code organization, I would rather extract `computeDiff` into a new small internal package, make the function public, provide unit-tests and benchmarks.

Also it feels it does excessive computation all the way, and it could be optimized to allocate significantly less I guess. 

I don't know if we can drop removing `.machine.install` section from the config (how much important that is?), but it certainly does lots of allocations/marshaling/etc. which can be avoided in the hot path.

Unix4ever: I've changed it to drop `.machine.install` section because when you upgrade, the config also changes.
So `Reconfiguring` and `Upgrading` phases start to fight with each other.

I think the other way to fix it might be to change `cluster_machine_config` controller to stop updating `.machine.install` section.

smira: so the diff is actually used to trigger the controller work? if it does, we need to make a decision if we update the install section or not, if we do, we should show the diff (and apply it), if we don't, we should skip it.

from Talos point of view, `.machine.install` is useless after the initial install (so patching it exactly once is fine, or using the initial schematic/version is also good).

Unix4ever: Dropped the config parsing

client/api/omni/specs/cluster_machine_config.go
Unix4ever: Reverted to Free.

PR #2300, opened 2026-02-11
test: use automation sa directly in integration tests

utkuozdemir:
Instead of doing the fake user auth flow in the integration tests via the `clientconfig` package, use the automation service account directly. Remove all other usages of that package as well, and drop it completely.

The package predates the initial service account token feature of Omni, its purpose was to authenticate to the Omni API in the integration tests. We have the automation key now, so we don't need that anymore.


utkuozdemir:
> I give 👍 for the cleanup idea!

This bit me pretty badly in helm chart tests, wasted quite a bit of my and @oguzkilcan 's time till we figured it out, so I wanted to fix it for good.

The followup idea getting rid of the special env var in Omni code which allows us to confirm public keys without the auth flow in debug builds. It is used by talemu atm, but doesn't need to be that way.

Also, the actual user auth flow (register public key/confirm) is tested in FE e2e tests, so we don't sacrifice test coverage here.

PR #2308, opened 2026-02-11
test: reach maintenance mode machines' Talos API through Omni in tests

utkuozdemir:
In the integration tests, we were accessing the API of the Talos machines which are in maintenance mode by directly hitting their SideroLink mgmt endpoint.

This worked only because the test was running on the same host as Omni itself (as we spawned Omni as process). This approach breaks when we install Omni via its helm chart on a Kubernetes cluster.

Fix this by going to them through Omni as well.
Additionally, centralize the talos client creation in the tests.

Additionally: bump Talos machinery, and pass the service account key explicitly to the Talos client when creating it, instead of relying on it to pick it from env vars.


client/pkg/omni/resources/omni/cluster_machine_config_patches_test.go
```diff
@@ -35,10 +35,23 @@ func TestClusterMachineConfigPatchesSpecW_marshal(t *testing.T) {
 
 	out := must.Value(yaml.Marshal(must.Value(resource.MarshalYAML(original))(t)))(t)
 
+	// TODO(workaround): yaml/v4 rc.4 passes DocumentNode to UnmarshalYAML, but COSI runtime
+	//                   doesn't handle it yet. Replace this block with:
+	//                   var dest protobuf.YAMLResource; err = yaml.Unmarshal(out, &dest)
+	//                   require.NoError(t, err)`
```
utkuozdemir, line 41: After https://github.com/cosi-project/runtime/pull/659 is merged, we can remove this workaround again.

client/pkg/template/template_test.go
```diff
@@ -101,7 +101,7 @@ func TestLoad(t *testing.T) {
 		{
 			name:          "clusterBadYAML1",
 			data:          clusterBadYAML1,
-			expectedError: "error decoding document at line 1:1: yaml: unmarshal errors:\n  line 7: field containerd not found in type models.Cluster",
+			expectedError: "error decoding document at line 1:1: yaml: construct errors:\n  line 7: field containerd not found in type models.Cluster",
```
utkuozdemir, line 104: New yaml library (brought transitively by bumped machinery) also changed the errors it returns.

internal/integration/integration_test.go
```diff
@@ -201,18 +201,15 @@ func TestIntegration(t *testing.T) {
 		}
 	}
 
-	// Talos API calls try to use user auth if the service account var is not set
-	os.Setenv(serviceaccount.OmniServiceAccountKeyEnvVar, serviceAccount)
```
utkuozdemir, line 205: This "workaround" is now removed, and we pass the whole TestOptions to many tests now, so they can access the serviceAccountKey where they construct a Talos client and pass it as a functional option.

internal/integration/common_test.go
```diff
@@ -131,74 +126,31 @@ func nodes(ctx context.Context, cli *client.Client, clusterName string, labels .
...
-		return nil, err
-	}
+func getTalosClient(ctx context.Context, t *testing.T, options *TestOptions) *talosclient.Client {
+	return getTalosClientForCluster(ctx, t, options, "")
+}
 
-	if len(data) == 0 {
-		return nil, errors.New("empty talosconfig")
-	}
+func getTalosClientForCluster(ctx context.Context, t *testing.T, options *TestOptions, clusterName string) *talosclient.Client {
```
utkuozdemir, line 133: This function is now the central place where we create Talos clients in integration tests.

internal/integration/common_test.go
```diff
@@ -131,74 +126,31 @@ func nodes(ctx context.Context, cli *client.Client, clusterName string, labels .
...
 		talosclient.WithConfig(config),
-		talosclient.WithCluster(cluster),
-	)
-}
-
-func talosClientForCluster(ctx context.Context, cli *client.Client, clusterName string) (*talosclient.Client, error) {
-	data, err := cli.Management().Talosconfig(ctx)
-	if err != nil {
-		return nil, err
+		talosclient.WithServiceAccount(options.serviceAccountKey),
```
utkuozdemir, line 143: Because of this, we pass whole TestOptions to many tests and propagate it downwards - so we can access `options.serviceAccountKey` here. This functional option was added very recently: https://github.com/siderolabs/talos/pull/12769

internal/integration/extensions_test.go
```diff
@@ -42,16 +43,19 @@ func AssertExtensionsArePresent(ctx context.Context, cli *client.Client, cluster
...
+		checkExtensionsWithRetries(ctx, t, options, extensions, machineIDs)
 	}
 }
 
-func checkExtensionsWithRetries(ctx context.Context, t *testing.T, cli *client.Client, extensions []string, machineIDs []resource.ID) {
+func checkExtensionsWithRetries(ctx context.Context, t *testing.T, options *TestOptions, extensions []string, machineIDs []resource.ID) {
 	for _, machineID := range machineIDs {
 		numErrs := 0
 
+		nodeCtx := talosclient.WithNode(ctx, machineID)
```
utkuozdemir, line 54: Brought `WithNode` back, as we don't use the trick of hitting the machine directly via its mgmt endpoint anymore. We go to Omni, and it needs to know which node to route to.

internal/backend/runtime/omni/controllers/omni/cluster_machine_config_test.go
```diff
@@ -251,7 +251,7 @@ func (suite *ClusterMachineConfigSuite) TestGenerationError() {
 		&suite.OmniSuite,
 		*omni.NewClusterMachineConfig(machines[0].Metadata().ID()).Metadata(),
 		func(cfg *omni.ClusterMachineConfig, assert *assert.Assertions) {
-			expectedError := "yaml: unmarshal errors"
+			expectedError := "yaml: construct errors"
```
utkuozdemir, line 254: New yaml library (brought transitively by bumped machinery) also changed the errors it returns.

PR #2317, opened 2026-02-12
feat: support custom Vault Kubernetes auth mount path

utkuozdemir:
Add a new `k8sAuthMountPath` config field under `storage.vault` and a `--vault-k8s-auth-mount-path` CLI flag. When set, it is passed as `auth.WithMountPath` to the Vault Kubernetes auth method, allowing Omni to authenticate against a Vault instance that mounts K8s auth at a non-default path (e.g., when Vault runs on a different cluster).

### DIFF
--- a/deploy/helm/v2/omni/values.yaml
+++ b/deploy/helm/v2/omni/values.yaml
@@ -421,6 +421,10 @@ config:
       # Tip: Use additionalConfigSources to load this from an existing Secret,
       # or set the VAULT_TOKEN environment variable via env/envFrom.
       token: ""
+      # K8sAuthMountPath is the mount path of the Kubernetes auth method in Vault.
+      # When not set, it defaults to "kubernetes".
+      # This is useful when Vault is running on a different cluster and has multiple Kubernetes auth mounts.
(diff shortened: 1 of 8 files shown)


deploy/helm/v2/omni/values.yaml
```diff
@@ -421,6 +421,10 @@ config:
       # Tip: Use additionalConfigSources to load this from an existing Secret,
       # or set the VAULT_TOKEN environment variable via env/envFrom.
       token: ""
+      # K8sAuthMountPath is the mount path of the Kubernetes auth method in Vault.
+      # When not set, it defaults to "kubernetes".
+      # This is useful when Vault is running on a different cluster and has multiple Kubernetes auth mounts.
+      #k8sAuthMountPath: ""
```
TimJones, line 427: Where is this actually used in the chart?

TimJones: Ahh, this whole `config` field is just slurped up as one big yaml block 👍

smira (approved):

oguzkilcan (approved):

PR #2322, opened 2026-02-13
release(v1.5.3): prepare release

utkuozdemir:
This is the official v1.5.3 release.


hack/test/common.sh
```diff
@@ -28,7 +28,7 @@ export KUBERNETES_VERSION=1.35.0
 # - Omni upgrade tests, to prevent Talos changes interfering with Omni changes
 # - Talos minor upgrade tests
 export STABLE_TALOS_VERSION=1.11.6
-export ANOTHER_OMNI_VERSION="${ANOTHER_OMNI_VERSION:-$LATEST_STABLE_OMNI}"
+export ANOTHER_OMNI_VERSION=v1.5.2
```
utkuozdemir, line 31: @oguzkilcan did this by your updated instructions. does it look right?

oguzkilcan: Yes it looks alright. This mainly matters when we are doing a release on an older minor version.

PR #2323, opened 2026-02-14
feat(frontend): show more detailed node disk information

### LINKED ISSUE #1116: [feature] provide a view of disks and partitions
### Problem Description

Talos since 1.8 provides a view of both block devices (`talosctl get disks`) containing low-level information about the disk itself, and also information about discovered volumes (information about disks and partitions, `talosctl get discoveredvolumes`). 

We can combine the both for a per-machine view of block devices similar to the screenshot above (screenshot just to give an idea, actual view might be anything).

We can also pull in `MountStatus` resource to show where each volume is mounted.

We can show sizes, file system information, calculated unallocated space, etc.

### Solution

_No response_

### Alternative Solutions

_No response_

### Notes

_No response_

Slessi:
Show more detailed node disk information in the node overview disks tab (renamed from mounts).

Closes #1116 

image


Unix4ever:
Do we also show that for user disks?

Slessi:
@Unix4ever I think so, in E2E I get these extra disks that I don't get locally, which may be user disks.

Not sure what we should or shouldn't show, I just got the Talos data from `Disk`, `DiscoveredVolume`, and `MountStatus`, and tried to show everything that looked useful.

For review the part I am interested the most is definitely if I am using the data properly. Maybe I'm showing things that are not helpful, or missing things that could be helpful. If nothing is bad in the current version also possible to merge as is and iterate later.

image

Unix4ever:
I'd probably rename Mounts to Disks there.
And will also check how it looks like on various data tomorrow.

rothgar:
I love this view and want to figure out if we can show this type of simplified view in Omni when provisioning a node. During provisioning is the only time you can create the partitions so we'll need a way to represent than data (maybe a modal like patches and extensions).

One problem you're going to run into is large disks (eg 1TB+) will only show ephemeral because the other partitions are going to be such a small percentage of the disk it'll be hard to notice they're there.

If you're going to show encryption status you should also show when it's `unavailable` Some partitions can't be encrypted (eg `STATE`) and showing `unknown` might be confusing.

Unix4ever:
image
image

user disks look fine I guess.

What might be also really nice to show is the rest of the data from the resource.

```yaml
    dev_path: /dev/sda
    size: 25165824000
    pretty_size: 25 GB
    io_size: 512
    sector_size: 512
    readonly: false
    cdrom: false
    model: QEMU HARDDISK
    modalias: scsi:t-0x00
    wwid: t10.ATA     QEMU HARDDISK                           QM00001
    bus_path: /pci0000:00/0000:00:1f.2/ata1/host0/target0:0:0/0:0:0:0
    sub_system: /sys/class/block
    transport: sata
    rotational: true
    symlinks:
        - /dev/disk/by-diskseq/13
        - /dev/disk/by-id/ata-QEMU_HARDDISK_QM00001
        - /dev/disk/by-path/pci-0000:00:1f.2-ata-1
        - /dev/disk/by-path/pci-0000:00:1f.2-ata-1.0
```

If we show it in a popup when clicking some (i) button on the disk item it could be enough.

Or maybe we can just skip that step and allow creating assisted config patches for disk partitioning.

For example, to allocated space on the /dev/sda, I've used the following config:

```yaml
apiVersion: v1alpha1
kind: UserVolumeConfig
name: rook-2
provisioning:
    diskSelector:
        match: "disk.bus_path == '/pci0000:00/0000:00:1f.2/ata1/host0/target0:0:0/0:0:0:0'"
    minSize: 10GB
    maxSize: 12GB
```

Also we need to show the disks list for Machines running in maintenance mode as well.

Slessi:
@Unix4ever I don't think I can show the maintenance mode information, because the resource service proxy thing doesn't seem to work for unclustered machines. At least not currently, I just get this

```
{
    "error": {
        "code": 5,
        "message": "talos API client is not available: failed to get: resource ClusterEndpoints.omni.sidero.dev(default/@undefined) doesn't exist"
    }
}
```

Unix4ever:
Did you use machine uuid as the node name there?

PR #2324, opened 2026-02-14
test: use envsubst in tests and do small improvements

utkuozdemir:
Now that we have envsubst in the build container, we can simplify our scripts a bit.

Also do other cosmetic improvements in the test scripts.


utkuozdemir:
> Another solution would have been to use yq templates, same way we use jq templates in talos repo, but we just need something that works, 🆒

Yes, I just needed templating, didn't need to be fancy. For now envsubst is getting the job done.

PR #2326, opened 2026-02-16
fix: break the dep loop in the cluster machine config status controller

Unix4ever:
Extract the fields required by the `MachineConfigStatusController` to a separate resource.
Otherwise there's circular loop: `MachinePendingUpdates` ->
`MachineSetStatus` -> `MachineConfigStatus` -> `MachinePendingUpdates`...

Also change the way machine pending is calculated: do not delete the pending machine updates resource if the Talos version/schematic is not in sync.


Unix4ever:
Tests found the bug in my implementation.
Not all errors should release the config change lock.

Should be good if we do this:
https://github.com/siderolabs/omni/compare/9ddba76ee9f1bd23e3032d384bd567719abe47ee..69c2759b8bc5fe39f731c126921cf9ced4e7d468

PR #2330, opened 2026-02-16
refactor: replace the old helm chart with the new one

utkuozdemir:
Since we don't want to support/maintain the old chart anymore, we simply replace it with the new chart.

Added a validation which fails on upgrades from the old one to the new one.


frontend/src/components/common/DiffRenderer/sample_diff.diff
utkuozdemir, line 1: @Slessi this is a victim of find&replace, but I guess it's fine to include.

Slessi: @utkuozdemir yea, this is just dummy data to use in storybook for the ConfigDiffs page. Doesn't really matter what is in it as long as it is valid diff syntax, which it still is

oguzkilcan (approved):

PR #2336, opened 2026-02-17
fix: track dependendants for searchFor in watch

Slessi:
Keep track of IDs of resources that were sent for watches using searchFor, so as to correctly update them even if changes to the resource cause them to no longer match the filter.

(Videos are 3 tabs of cluster view with filters -> [`Not Ready`][`Ready`][`All`])

Before:
> Destroyed clusters linger in Ready tab, created ones linger in Not ready

https://github.com/user-attachments/assets/d55855dc-6352-4766-90a7-fb74b486f430

After:

https://github.com/user-attachments/assets/76c343d2-ed56-407a-929b-99172575c954

### DIFF
--- a/internal/backend/runtime/proxy_runtime.go
+++ b/internal/backend/runtime/proxy_runtime.go
@@ -77,8 +77,17 @@ func watchResponseProducer(
 	offsetLimiter := MakeStreamOffsetLimiter(opts.Offset, opts.Limit, safeCmp(cmp, cmpNamespaceID[WatchResponse]))
 	total := int32(0)
 
+	var sent map[string]struct{} // tracks resources sent to the client for searchFor filtering
+	if len(opts.SearchFor) > 0 {
+		sent = map[string]struct{}{}
+	}
+
 	return func(ctx context.Context, wr WatchResponse) (bool, error) {
-		if !match(wr, opts.SearchFor) {
+		if len(opts.SearchFor) > 0 && wr.ID() != "" {
+			if !reconcileSearchEvent(wr, opts.SearchFor, sent) {
+				return true, nil
+			}
+		} else if !match(wr, opts.SearchFor) {
 			return true, nil
 		}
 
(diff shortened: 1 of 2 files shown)


internal/backend/runtime/proxy_runtime.go
```diff
@@ -76,9 +76,17 @@ func watchResponseProducer(
 ) func(ctx context.Context, wr WatchResponse) (bool, error) {
 	offsetLimiter := MakeStreamOffsetLimiter(opts.Offset, opts.Limit, safeCmp(cmp, cmpNamespaceID[WatchResponse]))
 	total := int32(0)
+	var sent map[string]struct{} // tracks resources sent to the client for searchFor filtering
+	if len(opts.SearchFor) > 0 {
+		sent = map[string]struct{}{}
```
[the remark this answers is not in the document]
Slessi: I don't think this applies because you watch one resource at a time.

Unix4ever:
I guess I'd prefer to move it to COSI level.

Slessi:
@Unix4ever if this is ok for now can we get this in and leave cosi migration as a task for later? As its a bug already, and I wanna get un-rejecting machines in

internal/backend/runtime/proxy_runtime.go
```diff
@@ -108,6 +117,47 @@ func match(ev runtime.Matcher, searchFor []string) bool {
...
+
+		return true
+	case matches:
+		return true
+	case wasSent:
+		delete(sent, wr.ID())
+		wr.Unwrap().Event.EventType = resources.EventType_DESTROYED
+		wr.Unwrap().Event.Old = ""
+
+		return true
```
Unix4ever, line 155: This case block doesn't make sense to me.

Switch can get only:
```
	EventType_CREATED
	EventType_UPDATED
	EventType_BOOTSTRAPPED
```

We don't want to map any of those to DESTROYED.

I think even BOOTSTRAPPED won't make it there, as it's `wr.ID()` is empty and it will be skipped.
So this case block won't be ever reached.

Unix4ever:
Well, actually there's no such thing as COSI layer for `searchFor` :upside_down_face:
So I guess that's the only fix we can have.
Lets merge it after we fix the thing I found.

Slessi:
@Unix4ever yea there isnt but maybe we can do something like this https://github.com/cosi-project/runtime/pull/660 and make it work with the `searchFor` in omni

Then omni only does searchFor -> func, but none of the tracking (that should work in COSI like it does for selectors). Would only work for local resources since you cant serialise a function but I think for Omni thats ok?

also possible _i guess_ could be creating some string syntax that COSI could parse instead of a filter function but meh

internal/backend/runtime/proxy_runtime.go
Slessi: the switch is switching on `true`, not event types

this case is:

- resource is NOT currently a match for the filter
- resource previously matched the filter and was sent to the listener

then -> replace whatever event we have here with a destroyed event so listener knows to remove it, since it is no longer a match for the filter

Unix4ever: ah I see, then it's all good I guess

Issue #2256, opened 2026-02-04
talk directly to Talos node over SideroLink when possible


smira:
> Short version: stop doing proxying for Talos API when this is not required, use SideroLink tunnel to talk directly to Talos API of a specific machine.

## Why?

Relying on Talos API multiplexing with Omni seems not reliable:

1. It relies on connectivity between the machines, while Omni has a direct p2p tunnel to each machine.
2. Proxying wouldn't work if the pod and node CIDRs overlap, direct connection should work.
3. If the machine is multi-homed, `--nodes` should be picked up so that the address allows node-to-node communication, while if Omni picks it up (by resolving node name or UUID), Omni doesn't know which of the multiple addresses would work.

## How
 
1. If we still use `nodes` (`client.WithNodes`) anywhere in Omni, let's drop it.
2. For any access to Talos API, if it doesn't have `nodes` set (or `nodes` is a singular item) and when `node` is set, set the `endpoint` of the client to be a single machine we're talking to, and drop `nodes`/`node` completely.
3. There should be no change for maintenance mode API.

## Caveats

Today Omni does caching and re-use of Talos API clients as they all are "same" (talk to the same set of endpoints which is controlplane endpoints).  This change might make cache way less efficient.  We can either drop the cache, make the TTL very short (to offset the connection setup time), or pick something else.

smira:
Paths we need to check/update:

* Talos API gRPC proxy (used by `talosctl` against Omni as proxy)
* `talos` backend in Omni (used by the frontend)

We can make all these paths use one2one proxy over SideroLink point-to-point if the path allows it.

Audit if we have any calls done by Omni backend/frontend with `nodes` and move them to use `node`. (client.WithNodes vs. client.WithNode).

smira:
Idea: remove Talos API proxying in `talemu`, and integration tests should still pass.

pranav767:
I understand that the current implementation expects reliable cluster-wide networking. When that breaks, or when machines are multi-homed, proxying via the control-plane becomes unreliable. So instead of building multi-endpoint clients, we can use a single endpoint per request and rely on SideroLink, which provides a direct end-to-end connection between Omni and each machine.

This would avoid the guesswork around which IP to use and remove the dependency on node-to-node routing.

However, I agree this could impact caching, since we would be caching per-node clients instead of a single cluster-level client. Maybe introducing a per-node cache with a shorter lifetime could help balance this.

Let me know if my understanding is correct & I can proceed with the above plan.

PR #2337, opened 2026-02-17
feat: allow communicating with unclustered machines

Slessi:
Currently Omni can only communicate to clustered machines through the resource service. With this PR we can communicate with unclustered machines.

### DIFF
--- a/internal/backend/runtime/talos/clients.go
+++ b/internal/backend/runtime/talos/clients.go
@@ -7,9 +7,12 @@ package talos
 
 import (
 	"context"
+	"crypto/tls"
 	"errors"
 	"fmt"
+	"net"
 	"runtime"
+	"strconv"
 	"strings"
 	"time"
 
@@ -21,6 +24,7 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/siderolabs/talos/pkg/machinery/client"
 	clientconfig "github.com/siderolabs/talos/pkg/machinery/client/config"
+	talosconstants "github.com/siderolabs/talos/pkg/machinery/constants"
 	"go.uber.org/zap"
 	"golang.org/x/sync/singleflight"
 	"google.golang.org/grpc"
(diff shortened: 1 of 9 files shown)


internal/backend/runtime/talos/clients.go
```diff
@@ -260,6 +264,73 @@ func (factory *ClientFactory) Get(ctx context.Context, clusterName string) (*Cli
 	}
 }
 
+// GetDirect creates an insecure Talos client connected directly to a node address.
+// This is used for unclustered/maintenance machines that are not part of any cluster.
+func (factory *ClientFactory) GetDirect(ctx context.Context, nodeAddress string) (*Client, error) {
```
Unix4ever: It's almost identical to `Get`.

I think we should extract common parts there.

Factory build can be replaced with a callback.

Then we can have two factories: 1st for the old clients for the clustered clients, 2nd implementation for direct clients.

Something like:

```go
// Like the current Get function, with a couple of changes here and there. Also use the logger from the params
func (factory *ClientFactory) get(ctx context.Context, logger *zap.Logger, prefix, id string, build func(ctx context.Context, id string) (*Client, error) {
   key := prefix + id
   ...

	ch := factory.sf.DoChan(key, func() (any, error) {
		...
		client, err := build(ctx, id)
		...
	}
}

// I would also suggest to rename Get to GetForCluster.
func (factory *ClientFactory) GetForCluster(ctx context.Context, clusterName string) (*Client, error) {
	logger := factory.logger.With(logging.String("cluster", clusterName))

	return get(ctx, logger,  "", clusterName, factory.buildForCluster)
}

func (factory *ClientFactory) GetForMachine(ctx context.Context, address string) (*Client, error) {
	logger := factory.logger.With(logging.String("node", address))

	return get(ctx, logger, "machine-", address, factory.buildDirect)
}

func (factory *ClientFactory) buildDirect(ctx context.Context, nodeAddress string) (*Client, error) {
		endpoint := net.JoinHostPort(nodeAddress, strconv.FormatInt(talosconstants.ApidPort, 10))

		c, err := client.New(ctx,
			client.WithTLSConfig(&tls.Config{
				InsecureSkipVerify: true,
			}),
			client.WithEndpoints(endpoint),
			client.WithGRPCDialOptions(
				grpc.WithDefaultCallOptions(
					grpc.MaxCallRecvMsgSize(constants.GRPCMaxMessageSize),
				),
				grpc.WithSharedWriteBuffer(true),
			),
		)
		if err != nil {
			return nil, err
		}

		return NewClient(c, ""), nil
}
```

Slessi: done!

internal/backend/runtime/talos/clients.go
```diff
@@ -298,6 +317,28 @@ func (factory *ClientFactory) build(ctx context.Context, clusterName string) (*C
...
+			InsecureSkipVerify: true,
+		}),
+		client.WithEndpoints(endpoint),
+		client.WithGRPCDialOptions(
+			grpc.WithDefaultCallOptions(
+				grpc.MaxCallRecvMsgSize(constants.GRPCMaxMessageSize),
+			),
+			grpc.WithSharedWriteBuffer(true),
+		),
+	)
```
Slessi, line 334: From copilot:

> consider restricting direct dialing to addresses that resolve to a known Machine in Omni state to avoid SSRF-style outbound connections.

I looked also at code and I can see how I guess u could pass some IP yourself instead of a UUID and the code in [`resolve.go:92-98`](https://github.com/siderolabs/omni/blob/d262e03bad9f94763aeff53dbb3a9f0456c6ca10/internal/backend/grpc/router/resolve.go#L92-L99) will just forward it along, and eventually this new stuff in `buildForMachine` will just start talking to it. You still need auth to get to this point as far as I know though so maybe not a big deal.

Slessi: There is also the problem that after the machine joins a cluster, this wont work anymore (complains about no TLS) and we will have to pass the cluster in, so e.g. on a page like `/machine/:machineId/disks` i'll have to first check if the machine is in a cluster, then adjust the `context` accordingly, which is non-obvious

Unix4ever: Yeah. I think we can address that here though.
The trick is to have a different cache key for secure and insecure clients.

Then when creating the client check if the machine belongs to a cluster, get the cluster name and use the certs for it.

It will be a good fix btw. If we do it, then it will actually be a fix for https://github.com/siderolabs/omni/issues/2256

Then the UI won't need to pass the cluster name at all, it should always rely on the machine UUID as the node name.

Slessi:
Closing in favor of the solution for #2256

PR #2340, opened 2026-02-18
test: fix flaky unit tests

oguzkilcan:
Fix several flaky unit tests:
  * Fix infra provider cleanup test using wrong service account identity ID (prefix+id instead of id+suffix).
  * Fix cluster bootstrap status test using separate time.Now() calls for backup upload and snapshot lookup, causing mismatched snapshot names when even 1 second elapsed.
  * Increase context timeouts for machineconfig status and talosconfig tests to prevent deadline exceeded under load.
  * Fix siderolink tests by increasing timeouts and waiting for JoinTokenStatus to be ACTIVE before provisioning.
  * Replace time.Sleep with EventuallyWithT polling in stripe metrics reporter test to avoid flaky timing.
  * Make kubernetes_node_audit mock thread-safe with mutex, deduplicate deletions, remove incorrect assertNoResource (qtransform creates output even with SkipReconcileTag).
  * Fix encrypt (crypt) blocked upload mock to return ctx.Err() immediately, preventing race between context cancellation and io.Copy on small payloads.
  * Fix cluster_status break glass test to trigger reconcile via Cluster input update after modifying ClusterStatus output, since qtransform only watches inputs.
  * Fix machine_set_status cluster locks test by moving sleep between lock and MachineSetNode destruction so pending reconciles drain before state changes.
  * Fix infra_provider_cleanup test to create Link and InfraMachineStatus before controller registration, avoiding race where teardown fires before watch events reach the controller cache.
  * Use inmem state with history options in NewTestState to prevent watch event loss.
  * Use inmem state with history options in testutils.WithRuntime to prevent watch event loss.
  * Refactor machine_set_etcd_audit from testify.Suite to testutils.WithRuntime + rmock pattern with per-test isolated runtimes.

Fixes: #2328

### DIFF
--- a/internal/backend/runtime/omni/controllers/omni/machine_set_status_test.go
+++ b/internal/backend/runtime/omni/controllers/omni/machine_set_status_test.go
@@ -7,8 +7,6 @@ package omni_test
 
 import (
 	"context"
-	"encoding/json"
-	"errors"
 	"fmt"
 	"testing"
 	"time"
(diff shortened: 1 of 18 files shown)


internal/backend/runtime/omni/controllers/omni/machine_set_status_test.go
```diff
@@ -537,10 +537,14 @@ func (suite *MachineSetStatusSuite) TestClusterLocks() {
 
 	suite.Require().NoError(err)
 
+	// Wait for the controller to drain any pending reconciles from the setup phase
+	// and process the lock event before we modify the machine set.
+	time.Sleep(time.Second * 2)
```
Unix4ever: I'd rather suggest to do:

```go
rtestutils.AssertResources(suite.ctx, suite.T(), suite.state, []string{machines[0]}, func(*omni.ClusterMachine, *assert.Assertions) {})
```

We only need to wait for `ClusterMachine` to be created.

Unix4ever (approved):

internal/backend/runtime/omni/controllers/omni/machine_set_status_test.go
utkuozdemir: @oguzkilcan you could take care of other sleeps while you are at it, if it applies. There are some others.

PR #2351, opened 2026-02-18
feat(frontend): add ability to delete pending machines

### LINKED ISSUE #1655: Allow deleting pending machines from the UI
Currently Decline keeps the link and it's fine. But if you want to drop the provider completely, and remove all related machines there's no way to drop the links of not accepted machines in the UI.

Slessi:
Allow deleting pending or rejected machines from the omni UI.

Closes #1655


frontend/src/views/omni/Machines/components/MachineDelete.vue
```diff
@@ -0,0 +1,65 @@
...
+    @confirm="confirm"
+  >
+    <ul class="list-inside list-disc">
+      <li v-for="machine in machines" :key="machine">
+        <code>{{ machine }}</code>
+      </li>
+    </ul>
+
+    <p class="py-2 text-xs">Please confirm the action.</p>
+    <p class="py-2 text-xs text-primary-p2">Deleted machines will be removed from omni.</p>
```
Slessi: not sure if any additional text is required for this?

Unix4ever:
```suggestion
    <p class="py-2 text-xs text-primary-p2">Deleted machines will be removed from Omni.</p>
```

Unix4ever: In case of bare metal provider there's no need to reset them I guess.
So maybe it's fine as it is.

PR #2360, opened 2026-02-18
fix(frontend): fix multi-doc parsing when creating single node clusters

### LINKED ISSUE #2333: Bug: Can't use multi-doc patch during cluster creation
Trying to set a multidoc patch during cluster creation and the frontend throws an error

```
 {name: 'YAMLException', reason: 'expected a single document in the stream, but found more', mark: undefined, message: 'expected a single document in the stream, but found more', stack: 'YAMLException: expected a single document in the
```

The cluster fails to create and nothing happens on the Omni backend.

Slessi:
When creating clusters with only a single node, we do a check to see if we should ask the user whether we should allow scheduling user workloads on that node. As part of that we parse config patches to see if its already enabled there, and that parsing did not support multi-doc. Now it does.

Fixes #2333

### DIFF
--- a/frontend/src/states/cluster-management.ts
+++ b/frontend/src/states/cluster-management.ts
@@ -5,7 +5,7 @@
 
 // MachineSet is the state used in cluster creation flow.
 
-import yaml from 'js-yaml'
+import { loadAll } from 'js-yaml'
 import { ref } from 'vue'
 
 import { Runtime } from '@/api/common/omni.pb'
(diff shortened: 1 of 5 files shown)


frontend/src/states/cluster-management.ts
```diff
@@ -608,12 +608,12 @@ export class State {
...
   private checkSchedulingEnabled(data: string) {
-    const loaded = yaml.load(data) as
-      | { cluster?: { allowSchedulingOnControlPlanes?: boolean } }
+    const loaded = loadAll(data) as
+      | { cluster?: { allowSchedulingOnControlPlanes?: boolean } }[]
       | undefined
       | null
 
-    return loaded?.cluster?.allowSchedulingOnControlPlanes
+    return loaded?.some((y) => y.cluster?.allowSchedulingOnControlPlanes)
```
Slessi, line 616: bug here, rest is just so we dont import the entire js-yaml lib

rothgar (approved):

PR #2380, opened 2026-02-20
chore: bump Talos machinery to the latest main and use 1.12.4 schema

Unix4ever:
There was a feature backport to 1.12.4, we want to support that in the config validators.


deploy/helm/omni/Chart.yaml
```diff
@@ -3,4 +3,4 @@ name: omni
 description: A helm chart to deploy Omni on a Kubernetes cluster
 type: application
 version: 0.0.3
-appVersion: "v1.5.4"
+appVersion: "v1.5.3"
```
Slessi: is this one right? 👀

Unix4ever: nope

Issue #2342, opened 2026-02-18
[UX] rename node `Force Destroy` and `omnictl cluster machine delete` `--force` flag to something that better hints the idea of what it's doing


Unix4ever:
Force Destroy and `--force` flag actual idea is to unblock control plane machine set scale down when something went wrong with a node joining etcd cluster.
This use case is very narrow. Overusing force destroy might also lead to breaking etcd quorum with no way to repair it.

People often confuse this force destroy with the flow when a machine should be deleted from Omni machines page if it's not reachable.

So the steps to fix it can be:
1. We should have a confirmation dialog in the UI, explaining the use case.
2. Rename `Force Destroy` context menu item to something like `Force Etcd Leave`, make it red.
3. Rename `--force` to `--force-etcd-leave` in the CLI.
4. Have the confirmation (y/N), with the description what it does and consequences it might cause in CLI. Should have `-y` flag.

PR #2386, opened 2026-02-23
fix: rename `--force` flag to `--force-etcd-leave`, same in the UI

Unix4ever:
Also add the confirmation in the CLI, explaining what this thing is doing.

Fixes: https://github.com/siderolabs/omni/issues/2342

### DIFF
--- a/frontend/src/views/omni/Modals/NodeDestroy.vue
+++ b/frontend/src/views/omni/Modals/NodeDestroy.vue
@@ -56,9 +56,9 @@ const close = (goBack?: boolean) => {
   router.go(-1)
 }
 
-const forceDestroy = ref(false)
+const forceEtcdLeave = ref(false)
 
-const initRequiresForceDestroy = async () => {
+const initRequiresForceEtcdLeave = async () => {
   try {
     await ResourceService.Get(
       {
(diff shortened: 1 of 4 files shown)


frontend/src/views/omni/Modals/NodeDestroy.vue
```diff
@@ -182,9 +182,9 @@ const destroyNode = async () => {
 
   close()
 
-  if (forceDestroy.value) {
+  if (forceEtcdLeave.value) {
     showSuccess(
-      `The Machine ${node.value} is Attempted to be Forcefully Removed from the Cluster`,
+      `The Machine ${node.value} is Marked for Destruction with Force Etcd Leave`,
```
Slessi: Capitalisation is a bit weird, and maybe more understandable if u just mention the flag? For me at least (who doesnt know what this is) seeing it as a flag would make it more clear


```suggestion
      `The machine ${node.value} is marked for destruction with --force-etcd-leave`,
```

Unix4ever: I was trying to follow header capitalisation rules. I can change it to have `--force-etcd-leave` flag.

Issue #2375, opened 2026-02-20
limit number of users and service accounts


oguzkilcan:
Add two cli flags/configs to limit maximum number allowed users and service accounts

PR #2391, opened 2026-02-24
feat: enforce configurable limits on user and service account creation

oguzkilcan:
Add state validation that rejects identity creation when the configured maximum number of users or service accounts is reached. The gRPC resource and management servers now use the validated state so these limits are enforced for all creation paths (CLI, UI, API). Identity is created before the user resource so the validation fires before any side effects.

Also adds create validation for join token name, e2e Playwright tests covering UI and AccountLimits integration test covering API and CLI for limit enforcement.

Closes: #2375


internal/backend/runtime/omni/state_validation_test.go
```diff
@@ -2130,3 +2139,139 @@ func (m *mockEtcdBackupStore) Download(_ context.Context, _ []byte, clusterUUID,
...
+		t.Parallel()
+
+		ctx, cancel := context.WithTimeout(t.Context(), 3*time.Second)
+		t.Cleanup(cancel)
+
+		maxUsers := 2
+
+		innerSt := state.WrapCore(namespaced.NewState(inmem.Build))
+		st := state.WrapCore(validated.NewState(innerSt, omni.AccountLimitsValidationOptions(innerSt, config.AuthLimits{
+			MaxUsers: new(uint32(maxUsers)),
```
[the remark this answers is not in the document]
oguzkilcan: This is valid with Go 1.26

client/pkg/omnictl/user/create.go
```diff
@@ -55,11 +55,17 @@ func createUser(email string) func(ctx context.Context, client *client.Client) e
 			return fmt.Errorf("identity with email %q already exists", email)
 		}
 
+		if err := client.Omni().State().Create(ctx, identity); err != nil {
```
Unix4ever: I think the best option would be to define a method to create a user.
We can also do best effort revert there, but it won't depend on the CLI/UI interruptions in the middle of the operation.

internal/backend/runtime/omni/state_validation.go
```diff
@@ -954,6 +954,50 @@ func hasUppercaseLetters(s string) bool {
...
+				if err != nil {
+					return fmt.Errorf("failed to list users: %w", err)
+				}
+
+				if uint32(existing.Len()) >= maxUsers {
+					return fmt.Errorf("maximum number of users (%d) reached", maxUsers)
+				}
+			}
+
+			return nil
```
Unix4ever, line 997: More "go" way is to use early returns:

```suggestion
			if isServiceAccount {
				maxServiceAccounts := limits.GetMaxServiceAccounts()
				if maxServiceAccounts == 0 {
					return nil
				}

				existing, err := safe.StateListAll[*authres.Identity](ctx, st,
					state.WithLabelQuery(resource.LabelExists(authres.LabelIdentityTypeServiceAccount)),
				)
				if err != nil {
					return fmt.Errorf("failed to list service accounts: %w", err)
				}

				if uint32(existing.Len()) >= maxServiceAccounts {
					return fmt.Errorf("maximum number of service accounts (%d) reached", maxServiceAccounts)
				}
				
				return nil
			}
			
			maxUsers := limits.GetMaxUsers()
			if maxUsers == 0 {
				return nil
			}

			existing, err := safe.StateListAll[*authres.Identity](ctx, st,
				state.WithLabelQuery(resource.LabelExists(authres.LabelIdentityTypeServiceAccount, resource.NotMatches)),
			)
			if err != nil {
				return fmt.Errorf("failed to list users: %w", err)
			}

			if uint32(existing.Len()) >= maxUsers {
				return fmt.Errorf("maximum number of users (%d) reached", maxUsers)
			}

			return nil
```

client/pkg/omnictl/user/create.go
Unix4ever: Also swapping the order of creation here will break the UI.
I don't think we show identites without the users. But we show the users without identities, so you can clean them up manually.

PR #2392, opened 2026-02-24
feat: enforce auth_time in auth0 token validation

Slessi:
In the backend enfore auth_time checks to validate that a re-authentication has taken place, and that the token meets our minimum age policy.

### DIFF
--- a/internal/pkg/auth/auth0/token.go
+++ b/internal/pkg/auth/auth0/token.go
@@ -21,6 +21,7 @@ import (
 const (
 	tokenValidationCacheDuration    = 5 * time.Minute
 	tokenValidationAllowedClockSkew = 5 * time.Minute
+	tokenValidationMaxAge           = 2 * time.Minute
 )
 
 // IDTokenVerifier is an Auth0 ID token verifier.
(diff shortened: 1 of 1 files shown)


internal/pkg/auth/auth0/token.go
```diff
@@ -73,6 +74,14 @@ func (v *IDTokenVerifier) Verify(ctx context.Context, token string) (*jwt.Claims
 		return nil, errors.New("unexpected custom claims type")
 	}
 
+	if customClaims.AuthTime == 0 {
+		return nil, errors.New("auth_time claim is missing")
+	}
+
+	if time.Since(time.Unix(customClaims.AuthTime, 0)) > tokenValidationMaxAge+tokenValidationAllowedClockSkew {
```
Slessi: as an aside, the `tokenValidationAllowedClockSkew` is 5 minutes which seems a bit excessive, i think a server with a 5minute wrong clock should just fix its clock, but maybe this is a real usecase i dont know of

smira: might be on-prem Omni on a server with a clock skew (in theory)

PR #2394, opened 2026-02-24
fix(frontend): handle invalid jwt response from backend

### LINKED ISSUE #2349: [bug] stale jwt key not handled

### Current Behavior

After having a hosted Omni tab open for some time, it redirected to (domain)/authenticate?flow=frontend&redirect=/. I chose the login in the Authenticate UI Access dialog. A popup of "Failed to confirm public key" pops up.

### Expected Behavior

It should redirect to the authentication source (Auth0 for a Google account in the hosted instance) to renew that authentication information.

### Steps To Reproduce

1. Leave the Omni instance idle for some time on the dashboard.

Microsoft Edge

### Anything else?

Repeating the experience seemed like similar behavior occured but sourced in a JSON response getting mishandled and put into the JWT popup. The way screens/dialogs are titled, it seems like it's trying to redirect but it's not doing so consistently or reliably.

Hosted Omni is currently 1.4.9

Slessi:
If the backend returns an invalid jwt error, the frontend must re-trigger the authentication flow

Closes #2349

### DIFF
--- a/frontend/src/views/omni/Auth/Authenticate.vue
+++ b/frontend/src/views/omni/Auth/Authenticate.vue
@@ -12,7 +12,7 @@ import { jwtDecode } from 'jwt-decode'
 import { computed, onMounted, ref, watch } from 'vue'
 import { useRoute, useRouter } from 'vue-router'
 
-import { b64Encode, type fetchOption } from '@/api/fetch.pb'
+import { b64Encode, type fetchOption, RequestError } from '@/api/fetch.pb'
 import { Code } from '@/api/google/rpc/code.pb'
 import { AuthService } from '@/api/omni/auth/auth.pb'
 import { withMetadata } from '@/api/options'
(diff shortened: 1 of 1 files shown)


frontend/src/views/omni/Auth/Authenticate.vue
```diff
@@ -182,23 +182,8 @@ const generatePublicKey = async (identity: string) => {
   await router.replace({ path: !redirect.startsWith('/') ? '/' : redirect })
 }
 
-let renewIdToken = false
-
 const confirmPublicKey = async (publicKeyId: string, keyPair?: CryptoKeyPair) => {
   try {
-    // If the JWT validation has failed on the backend at the moment of clicking "Login", get a new ID token from Auth0 on the next click.
-    // This way, the user will not have to reload the page after validating their email - they can simply click "Login" again to get in.
```
Slessi, line 190: User was stuck in an "invalid JWT" loop because of this.

Im not sure about the original thinking behind this, but pretty sure we cant support this flow. If there is something wrong with the token, we have to re-auth, to meet our security policies.

Even if not, this isn't a nice flow:

- Click login
- Login error
- Click login
- Login accepted

Should've just worked the first time, if user interaction wasn't required to fix the problem

Issue #2070, opened 2025-12-26
[bug] Worker systemExtensions are being overridden and applied as "inherited from cluster" level config


ryjogo:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

I have two worker definitions in my template, and a cluster. The cluster has the ` siderolabs/iscsi-tools` and `siderolabs/util-linux-tools` and a selection of worker template have `siderolabs/qemu-guest-agent` and some without (metal). 

Yet it seems all machinesets get `siderolabs/qemu-guest-agent` 

This is Omni SaaS v1.4.3

### Expected Behavior

I expected that metal-workers-1 will only have ` siderolabs/iscsi-tools` and `siderolabs/util-linux-tools` yet they seem to be "inheriting" config from the cluster level, and reporting that it should have the `siderolabs/qemu-guest-agent` 

### Steps To Reproduce

```
kind: Cluster
name: qe1
labels:
  template: true
  cni: none
  kubespan: enabled
kubernetes:
  version: v1.34.2
talos:
  version: v1.11.5
....
systemExtensions:
  - siderolabs/iscsi-tools
  - siderolabs/util-linux-tools
```
and the workers

```
---
kind: Workers
name: metal-workers-1
labels:
  type: worker
  qe1.dk/location: local
machines:
  - 16afd6f4-7bfd-fd59-4de0-27524344c90b
  - 9493718b-61f6-fcd0-2799-9c4c19e74255
```
```
---
kind: Workers
name: cloud-workers-1
labels:
  type: worker
  qe1.dk/location: cloud
machines:
  - a7af7ba9-54a9-46d8-a54c-dccc8e72e38b
systemExtensions:
  - siderolabs/qemu-guest-agent
```

I have included my template.yaml file for clarity.

[template.yaml](https://github.com/user-attachments/files/24350266/template.yaml)

### Anything else?

Had a chat with @Unix4ever regarding this, the work around for now is to explicitly define the `systemExtensions` in each worker definition. This seems to work. 

```
---
kind: Workers
name: metal-workers-1
labels:
  type: worker
  qe1.dk/location: local
machines:
  - 16afd6f4-7bfd-fd59-4de0-27524344c90b
  - 9493718b-61f6-fcd0-2799-9c4c19e74255
systemExtension:
  - siderolabs/iscsi-tools
  - siderolabs/util-linux-tools
```

PR #2395, opened 2026-02-24
fix: properly select extensions when they're defined for cluster/ms lvl

Unix4ever:
Fixes: https://github.com/siderolabs/omni/issues/2070

### DIFF
--- a/internal/backend/runtime/omni/controllers/omni/machine_extensions.go
+++ b/internal/backend/runtime/omni/controllers/omni/machine_extensions.go
@@ -18,6 +18,7 @@ import (
 	"github.com/cosi-project/runtime/pkg/state"
 	"github.com/siderolabs/gen/optional"
 	"github.com/siderolabs/gen/xerrors"
+	"github.com/siderolabs/gen/xslices"
 	"go.uber.org/zap"
 
 	"github.com/siderolabs/omni/client/pkg/omni/resources"
@@ -208,7 +209,7 @@ func (ctrl *MachineExtensionsController) Reconcile(ctx context.Context, logger *
 
 		tracker.keep(status)
 
-		if err = safe.WriterModify[*omni.MachineExtensions](ctx, r, status, func(r *omni.MachineExtensions) error {
+		if err = safe.WriterModify(ctx, r, status, func(r *omni.MachineExtensions) error {
 			if !ctrl.shouldRecalculateExtensions(r, configs) {
 				return nil
 			}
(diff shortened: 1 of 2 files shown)


internal/backend/runtime/omni/controllers/omni/machine_extensions.go
```diff
@@ -277,7 +277,7 @@ func (ctrl *MachineExtensionsController) determineExtensions(cm *omni.ClusterMac
 		}
 
 		machineSet, ok := config.Metadata().Labels().Get(omni.LabelMachineSet)
-		if ok && machineSet == msID {
```
smira, line 280: I'm confused with `if clusterMachineLevel != nil` check above... we're in a loop, so I guess it can be removed or changed to `break` statement above... but I think both of these are a bit confusing

smira: or probably in other words, we should continue if we find e.g. a cluster machine label on the extensions config? as it doesn't matter what other labels are, if the most specific label didn't match?

smira: in other words, I like the fix, but I think the code was confusing, and now it's even more confusing to me ;)

Unix4ever: The logic is to pick the last ExtensionsConfiguration which is matching the particular machine ID.
If someone is using it this way we want to make sure we don't break their cluster.

internal/backend/runtime/omni/controllers/omni/machine_extensions.go
```diff
@@ -277,7 +277,7 @@ func (ctrl *MachineExtensionsController) determineExtensions(cm *omni.ClusterMac
...
+		if ok && machineSet == msID && clusterMachine == "" {
```
smira: ok, if we do this, why do we proceed with the loop on line 268 is `clusterMachine != ""` ? I'm trying to follow the logic.

If the ExtensionConfig belongs to a specific ClusterMachine, we have two cases:

1. It's our ClusterMachine so we found it, and we can return it immediately? 
2. It's not our ClusterMachine, so we skip `ExtensionConfig` completely?

Same flow doesn't work for MachineSets - it might be _our_ MachineSet, but we don't know if we find a more specific ExtensionConfig for a cluster machine, but still if we find the extension config being labled by a machine set label, we should not test it for cluster labels, and move on to the next one?

Unix4ever: This check should be dropped, yeah. ClusterMachine and MachineSet labels shouldn't be set at the set time, compared to the Cluster label.

smira: it's not clear to me if there might be some invalid label combinations.

all I'm trying to say is to write this code in a way it's easy to understand what it is doing exactly. the current loop has issues, it has too many exit points, and many conditions.

can we do it something like:

```go

// return nil if no match
func matchExtensionConfigByLabel(labelName, labelValue string, list []*ExtensionConfig) *ExtensionConfig


...

if matched := matchExtensionConfigByLabel(ClusterMachine, cmID, list); matched != nil { 
   return matched
}

if matched := matchExtensionConfigByLabel(MatchineSet, msID, list); matched != nil { 
   return matched
}

if matched := matchExtensionConfigByLabel(Cluster, clusterID, list); matched != nil { 
   return matched
}

return nil

```

I'm basically unrolling into three loops instead of one, but this time (I think) it's way more clear and harder to make a bug

Unix4ever: ok, gotcha. Should be better now

smira:
label system still looks confusing to me 🤦

Issue #2393, opened 2026-02-24
[chore] move Talos graceful upgrade to the lowest level


Unix4ever:
Do the same trick as for the config patches. Make ClusterMachineConfig status handle locking mechanism.
- the locks should happen on the Cluster scope.
- quota should always be 1.

Changes in `talos_upgrade_status.go` controller:
- drop graceful rollout on per-machine basis, keep graceful rollout per machineset, as we need to roll out control planes first.

PR #2412, opened 2026-02-26
chore: move graceful upgrades to the lowest level

Unix4ever:
Rewrite `TalosUpgradeStatus` controller to use the completely different flow:
- update all `ClusterMachineTalosVersion` resources immediately.
- to control quotas and rollout sequence use `UpgradeRollout` resource, it has a single field which is a map of MachineSetName -> Current Quota:
  - if control plane is updating it sets quota 0 on all other machine sets.
  - the number of not running/unhealthy machines is subtracted from the quota.
  - quota is now copied from the new `UpgradeStrategy`, so it's possible to have more than one machine updated in parallel.
- `ClusterMachineConfigStatus` controller now adds a new finalizer for upgrades on all `ClusterMachines` which are currently being updated to acquire/release locks and reads quotas from the `UpgradeRollout`.

Fixes: https://github.com/siderolabs/omni/issues/2393


internal/backend/runtime/omni/controllers/omni/machineconfig/status.go
```diff
@@ -844,6 +862,67 @@ func (ctrl *ClusterMachineConfigStatusController) getClient(
...
+		if clusterMachine.Metadata().Finalizers().Has(UpgradeFinalizer) {
+			quota--
+		}
+
+		if quota <= 0 {
+			return xerrors.NewTaggedf[qtransform.SkipReconcileTag]("quota %d for upgrades reached, waiting for the locks to be released", quota)
+		}
+	}
+
+	return r.AddFinalizer(ctx, rc.clusterMachine.Metadata(), UpgradeFinalizer)
```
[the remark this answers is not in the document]
Unix4ever: that's expected

internal/backend/runtime/omni/controllers/omni/talosupgrade/status.go
```diff
@@ -0,0 +1,478 @@
...
+			continue
+		}
+
+		if clusterMachine.Metadata().Finalizers().Has(machineconfig.UpgradeFinalizer) {
+			outdatedMachines[clusterMachine.Metadata().ID()] = clusterMachine
+
+			continue
+		}
+
+		if schematicConfiguration.TypedSpec().Value.TalosVersion != cluster.TypedSpec().Value.TalosVersion {
```
Unix4ever, line 480: Add a comment why we have this check here

internal/backend/runtime/omni/controllers/omni/talosupgrade/status.go
```diff
@@ -0,0 +1,478 @@
...
+	machineSetNodes, err := safe.ReaderListAll[*omni.MachineSetNode](ctx, r, state.WithLabelQuery(resource.LabelEqual(omni.LabelCluster, cluster.Metadata().ID())))
+	if err != nil {
+		return err
+	}
+
+	machineSetNodeMap := xslices.ToMap(slices.Collect(machineSetNodes.All()), func(r *omni.MachineSetNode) (string, *omni.MachineSetNode) {
+		return r.Metadata().ID(), r
+	})
+
+	for clusterMachine := range clusterMachines.All() {
```
Unix4ever: we can move that to a common place

internal/backend/runtime/omni/controllers/omni/talosupgrade/status.go
```diff
@@ -0,0 +1,478 @@
...
+		if pendingUpdate.TypedSpec().Value.Upgrade.FromSchematic == "" ||
+			pendingUpdate.TypedSpec().Value.Upgrade.FromVersion == "" {
+			continue
+		}
+
+		msn, ok := machineSetNodeMap[pendingUpdate.Metadata().ID()]
+		if ok {
+			_, locked := msn.Metadata().Annotations().Get(omni.MachineLocked)
+			if locked {
+				lockedMachines++
```
Unix4ever: rename this var to `pendingLockedMachines`

internal/backend/runtime/omni/controllers/omni/talosupgrade/status.go
```diff
@@ -0,0 +1,478 @@
...
+	clusterMachineTalosVersions, err := safe.ReaderListAll[*omni.ClusterMachineTalosVersion](ctx, r,
+		state.WithLabelQuery(resource.LabelEqual(omni.LabelCluster, cluster.Metadata().ID())),
+	)
+	if err != nil {
+		return err
+	}
+
+	visited := make(map[string]struct{}, schematicConfigurations.Len())
+
+	for schematicConfiguration := range schematicConfigurations.All() {
```
Unix4ever: Make it depend on the clusterMachine resources.
If clusterMachine is gone/tearingDown skip it in the loop.

internal/backend/runtime/omni/controllers/omni/machineconfig/status.go
```diff
@@ -844,6 +860,69 @@ func (ctrl *ClusterMachineConfigStatusController) getClient(
 	return result, nil
 }
 
+func (ctrl *ClusterMachineConfigStatusController) acquireUpgradeLock(ctx context.Context, r controller.ReaderWriter,
+	rc *ReconciliationContext,
+) error {
+	ctrl.acquireLockMu.Lock()
```
oguzkilcan, line 866: It's safer to use mutex for the whole method but you are actually calling etcd (using uncached list) to get all cluster machines, this may leave this lock acquired for a long time. Should we acquire the lock after we fetch all the cluster machines and before we calculate the quota?

internal/backend/runtime/omni/controllers/omni/talosupgrade/status.go
```diff
@@ -0,0 +1,492 @@
...
+
+				return ctrl.reconcileUpgradeVersions(ctx, r, cluster, upgradeStatus)
+			},
+			FinalizerRemovalExtraOutputFunc: func(ctx context.Context, r controller.ReaderWriter, _ *zap.Logger, cluster *omni.Cluster) error {
+				clusterMachineTalosVersions, err := safe.ReaderListAll[*omni.ClusterMachineTalosVersion](ctx, r, state.WithLabelQuery(resource.LabelEqual(omni.LabelCluster, cluster.Metadata().ID())))
+				if err != nil {
+					return err
+				}
+
+				if _, err = helpers.TeardownAndDestroyAll(ctx, r, clusterMachineTalosVersions.Pointers()); err != nil {
```
oguzkilcan: What if sth went wrong and not everything was destroyed, should we really ignore the confirmation?

internal/backend/runtime/omni/controllers/omni/talosupgrade/status.go
```diff
@@ -0,0 +1,492 @@
...
+				if !ready {
+					return xerrors.NewTaggedf[qtransform.SkipReconcileTag]("upgrade rollout %q tearing down", omni.NewUpgradeRollout(cluster.Metadata().ID()).Metadata().ID())
+				}
+
+				return nil
+			},
+		},
+		qtransform.WithExtraMappedInput[*omni.TalosVersion](
+			func(ctx context.Context, _ *zap.Logger, r controller.QRuntime, _ controller.ReducedResourceMetadata) ([]resource.Pointer, error) {
+				// reconcile all cluster TalosUpgradeStatus on TalosVersion changes
```
oguzkilcan, line 116: Why?

internal/backend/runtime/omni/controllers/omni/talosupgrade/status.go
```diff
@@ -0,0 +1,492 @@
...
+		qtransform.WithExtraMappedInput[*omni.ClusterConfigVersion](
+			qtransform.MapperNone(),
+		),
+		qtransform.WithExtraMappedInput[*omni.MachineSet](
+			mappers.MapByClusterLabel[*omni.Cluster](),
+		),
+		qtransform.WithExtraOutputs(
+			controller.Output{
+				Type: omni.ClusterMachineTalosVersionType,
+				Kind: controller.OutputShared,
```
oguzkilcan: Is there a different place that create this resource?

internal/backend/runtime/omni/controllers/omni/talosupgrade/status.go
```diff
@@ -0,0 +1,492 @@
...
+func (ctrl *TalosUpgradeStatusController) reconcileStatus(ctx context.Context, r controller.ReaderWriter,
+	clusterMachines safe.List[*omni.ClusterMachine], outdatedMachines *outdatedMachines,
+	cluster *omni.Cluster, upgradeStatus *omni.TalosUpgradeStatus,
+) error {
+	machineSetNodes, err := safe.ReaderListAll[*omni.MachineSetNode](ctx, r, state.WithLabelQuery(resource.LabelEqual(omni.LabelCluster, cluster.Metadata().ID())))
+	if err != nil {
+		return err
+	}
+
+	machineSetNodeMap := xslices.ToMap(slices.Collect(machineSetNodes.All()), func(r *omni.MachineSetNode) (string, *omni.MachineSetNode) {
```
oguzkilcan, line 176: I think we need a version of ToMap that accepts iterators.

internal/backend/runtime/omni/controllers/omni/talosupgrade/status.go
```diff
@@ -0,0 +1,492 @@
...
+}
+
+func (ctrl *TalosUpgradeStatusController) updateRolloutState(ctx context.Context, r controller.ReaderWriter,
+	outdatedMachines *outdatedMachines, cluster *omni.Cluster, upgradeStatus *omni.TalosUpgradeStatus,
+) error {
+	clusterMachineStatuses, err := safe.ReaderListAll[*omni.ClusterMachineStatus](ctx, r, state.WithLabelQuery(
+		resource.LabelEqual(omni.LabelCluster, cluster.Metadata().ID()),
+	))
+	if err != nil {
+		if state.IsNotFoundError(err) {
```
oguzkilcan: Do we still get `IsNotFoundError` when we are using `ReaderListAll`?

internal/backend/runtime/omni/controllers/omni/talosupgrade/status.go
Unix4ever: because we compute available upgrade versions in each status.

internal/backend/runtime/omni/controllers/omni/talosupgrade/status.go
Unix4ever: maybe not actually. Lemme check

internal/backend/runtime/omni/controllers/omni/talosupgrade/status.go
Unix4ever: no, good catch

oguzkilcan (approved):

internal/backend/runtime/omni/controllers/omni/machineconfig/status.go
Unix4ever: that's intentional. We want to hold the lock during the whole operation, including the list uncached.
I saw issues in the config changes without the lock.

The trick is to make list + update finalizer atomic. Otherwise:
- machine X lists cluster machines - no locks
- machine Y lists cluster machines - no locks
- machine X adds finalizer to a machine
- `somewhere here we will have machine Y to check that there are no locks again`
- machine Y adds finalizer to a machine
- wrong calculation - 2 locks

With locks:
- machine X lists cluster machines - no locks
- machine Y lists cluster machines - paused on the mutex
- machine X adds the finalizer
- machine Y unblocked - 1 lock
- machine Y skips

internal/backend/runtime/omni/controllers/omni/talosupgrade/status.go
Unix4ever: Yeah, I don't think there's one

Issue #2242, opened 2026-02-02
[feature] Add Metric for Monthly Active Users (MAU) Tracking


KarmaPoliceT2:
### Problem Description

Expose a prometheus metric that provides aggregate Monthly Active User counts for the tenant, enabling operators to track usage patterns for capacity planning and adoption insights. This is a standard operational metric exposed by most enterprise platforms and would bring Omni in line with expectations.

### Solution

Expose a metric (e.g. omni_monthly_active_users) representing the count of unique authenticated users with activity in the trailing 30-day window. 

- Metric should be an aggregate count, not individual user identifiers
- Appropriate labels for tenant should be applied
- Activeness should include authentication to omni via UI, API, or omnictl

While this is technically feasible to do via counts from audit/access logs, that represents opening up a level of data access that feels far broader than should be needed to gain such a basic operational metric.

smira:
I think we can make a useful feature and a metric on top of it:

* track the last time user/service account has been used using the metrics namespace (don't update on every interaction, update e.g. if the last used time is > 5 mins since now)
* show the "last activity" time in the Users view
* from this data, calculate the MAU metric above, probably split it by "users" and "service accounts" domains

smira:
Update Omni UI (Users) and CLI (`omnictl user list`?) to show the new last activity.

KarmaPoliceT2:
> * from this data, calculate the MAU metric above, probably split it by "users" and "service accounts" domains

Definitely agree that we should split users and service accounts in this... in fact I'm not sure service accounts should be included at all in the MAU tally, though I definitely see the value in having a "last activity" time for the service accounts logged in the UI and such.

PR #2424, opened 2026-03-02
feat: track user and service account last activity

oguzkilcan:
* Add `IdentityLastActive` resource to record the last time each identity(`User`/`ServiceAccount`) made a gRPC call.
* Add `IdentityStatusController` to aggregate identity, user role, and last-active data into an ephemeral `IdentityStatus` resource.
* Expose last_active in ListUsers/ListServiceAccounts gRPC responses, omnictl CLI output, and the frontend Users/ServiceAccounts views.
* Add `UserMetricsController` exposing `omni_users` (total) and `omni_active_users` (7d/30d windows) Prometheus gauges.

Closes: #2242


internal/backend/runtime/omni/controllers/omni/metrics/users.go
```diff
@@ -0,0 +1,183 @@
...
+
+		if err := ctrl.updateMetrics(ctx, r); err != nil {
+			return err
+		}
+
+		select {
+		case <-ctx.Done():
+			return nil
+		case <-time.After(reconcileInterval):
+		}
```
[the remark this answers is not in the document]
oguzkilcan: This is intentional, ticket is there to keep metrics up to date even if there are no new events.

client/pkg/omni/resources/auth/identity_last_active.go
```diff
@@ -0,0 +1,49 @@
...
+
+// IdentityLastActiveExtension provides auxiliary methods for IdentityLastActive resource.
+type IdentityLastActiveExtension struct{}
+
+// ResourceDefinition implements [typed.Extension] interface.
+func (IdentityLastActiveExtension) ResourceDefinition() meta.ResourceDefinitionSpec {
+	return meta.ResourceDefinitionSpec{
+		Type:             IdentityLastActiveType,
+		Aliases:          []resource.Type{},
+		DefaultNamespace: resources.DefaultNamespace,
```
smira: I wonder if we should push this out of etcd to "sqlite" state, as this is going to be updated frequently

oguzkilcan: Actually it won't be updated that much. There is a 1 minute debounce on the writes for each user activity.

smira: yep, I see that it's limited, but I wonder if it should be in sqlite along with other "metric"-like resources... up to you to decide on that

Unix4ever: I'd also vote for sqlite state.

internal/pkg/auth/interceptor/activity.go
```diff
@@ -0,0 +1,125 @@
...
+func (a *Activity) trackActivity(ctx context.Context) {
+	identity := identityFromContext(ctx)
+	if identity == "" {
+		return
+	}
+
+	now := time.Now()
+
+	if last, ok := a.lastUpdate.Load(identity); ok {
+		if lastTime, lastTimeOK := last.(time.Time); lastTimeOK && now.Sub(lastTime) < activityDebounceInterval {
```
Unix4ever: You can use https://github.com/siderolabs/gen/blob/main/containers/syncmap.go which is a typed wrapper around sync.Map. Then there will be no need in type casting.

internal/pkg/auth/interceptor/activity.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+
+	// Write asynchronously with a detached context to avoid blocking the RPC and to prevent client disconnects from canceling the write.
+	panichandler.Go(func() { //nolint:contextcheck
+		writeCtx, cancel := context.WithTimeout(context.Background(), activityWriteTimeout)
+		defer cancel()
+
+		writeCtx = actor.MarkContextAsInternalActor(writeCtx)
+
+		if err := safe.StateModify(writeCtx, a.state, authres.NewIdentityLastActive(identity),
```
Unix4ever: I think we better use StateUpdateWithConflicts and ignore not found errors.
Imagine identity is deleted at the very moment when activity is tracked, `StateModify` will re-create deleted identity.

oguzkilcan: Oh, good catch.

Issue #2064, opened 2025-12-26
support SSA manifest sync


smira:
See the change https://github.com/siderolabs/talos/pull/12362: use same logic to pick up the syncer for bootstrap manifest based on Talos version (current Talos version, not the initial version).

PR #2453, opened 2026-03-05
chore: force SSA manifests sync mode for Talos >= 1.13

Unix4ever:
Backend now automatically switches between legacy and SSA modes for different Talos versions.
Fixes: https://github.com/siderolabs/omni/issues/2064


internal/backend/grpc/kubernetes_manifests.go
```diff
@@ -0,0 +1,406 @@
...
+	req *management.KubernetesSyncManifestRequest,
+	sync *kubernetesSyncHelpers,
+	logger *zap.Logger,
+	srv management.ManagementService_KubernetesSyncManifestsServer,
+) error {
+	if req.Ssa == nil {
+		// fallback to defaults if SSA options are not provided
+		req.Ssa = &management.KubernetesSSAOptions{
+			InventoryPolicy: management.KubernetesSSAOptions_ADOPT_IF_NO_INVENTORY,
+			NoPrune:         true,
```
Orzelius, line 100: why disable pruning?

internal/backend/grpc/kubernetes_manifests.go
```diff
@@ -0,0 +1,406 @@
...
+
+	if req.Ssa.ReconcileTimeout == nil {
+		req.Ssa.ReconcileTimeout = durationpb.New(time.Second * 30)
+	}
+
+	if req.DryRun {
+		return s.diffSSA(ctx, req, sync, srv)
+	}
+
+	manager, err := ssa.NewManager(ctx, sync.cfg,
```
Orzelius, line 112: The manager creation bit is duplicated in the diff and apply. I think I'd be better to pass it to the methods.

internal/backend/runtime/omni/controllers/omni/kubernetes_upgrade_manifest_status.go
```diff
@@ -190,40 +194,30 @@ func NewKubernetesUpgradeManifestStatusController(talosClientGetter TalosClientG
...
-								logger.Error("failed to sync bootstrap manifests", zap.String("cluster", clusterID), zap.Error(err))
-								manifestStatus.TypedSpec().Value.LastFatalError = err.Error()
-
-								return nil
-							}
+				if talosVersion.SupportsSSAManifestSync() {
+					err = reconcileSSA(ctx, logger, bootstrapManifests, kubeconfigGetter, clusterID, manifestStatus)
+				} else {
+					err = reconcileLegacy(ctx, logger, bootstrapManifests, kubeconfigGetter, clusterID, manifestStatus)
+				}
```
Orzelius, line 208: I think there's no point in keeping the legacy here as we only the diff is performed which has no side effects.

Orzelius (commented):
my late review

PR #2454, opened 2026-03-05
refactor(frontend): migrate to file based routing

### LINKED ISSUE #1428: [UI] Move to file based routing

### LINKED ISSUE #1924: Provide/Inject pattern for permissions & currentUser
Currently `currentUser` and `permissions` are handled imperatively and in a coupled way with their consumers. This makes it hard to test & write stories.

Instead we can refactor them to a provider/injector pattern which can make them easily mocked with mock providers, and allows us to hook into the vue life cycle.

Slessi:
Migrate to file based routing (now included in vue-router v5 by default) which allows for type-safe routing and helps with file organisation by putting pages in a dedicated src/pages folder.

Due to `beforeEnter` route guards not being a thing, I had to also move route guard logic into the components themselves, and did some refactoring of the `currentUser` and `permissions` ref exports to cater for this.

Closes #1428 
Closes #1924 - (solves it in a different way, in 2nd commit)

### DIFF
--- a/frontend/src/components/SideBar/TSideBar.vue
+++ b/frontend/src/components/SideBar/TSideBar.vue
@@ -36,13 +36,7 @@ import type { SideBarItem } from '@/components/SideBar/TSideBarList.vue'
 import TSidebarList from '@/components/SideBar/TSideBarList.vue'
 import { getContext } from '@/context'
 import { setupBackupStatus } from '@/methods'
-import {
-  canManageBackupStore,
-  canManageUsers,
-  canReadClusters,
-  canReadMachines,
-  setupClusterPermissions,
-} from '@/methods/auth'
+import { useClusterPermissions, usePermissions } from '@/methods/auth'
 import { useFeatures } from '@/methods/features'
 import { useIdentity } from '@/methods/identity'
 import { useResourceWatch } from '@/methods/useResourceWatch'
@@ -53,6 +47,7 @@ const context = getContext()
 const { avatar, fullname, identity } = useIdentity()
 
 const { data: featuresConfig } = useFeatures()
+const { canManageBackupStore, canManageUsers, canReadClusters, canReadMachines } = usePermissions()
 
 const currentCluster = computed(() =>
   'cluster' in route.params ? route.params.cluster : undefined,
@@ -63,7 +58,7 @@ const currentMachine = computed(() =>
 
 const { status: backupStatus } = setupBackupStatus()
 const { canSyncKubernetesManifests, canManageClusterFeatures } =
-  setupClusterPermissions(currentCluster)
+  useClusterPermissions(currentCluster)
 
 const { data: machineMetrics } = useResourceWatch<MachineStatusMetricsSpec>({
   resource: {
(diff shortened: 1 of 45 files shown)


frontend/src/components/SideBar/TSideBar.vue
```diff
@@ -53,12 +47,19 @@ const context = getContext()
 const { avatar, fullname, identity } = useIdentity()
 
 const { data: featuresConfig } = useFeatures()
+const { canManageBackupStore, canManageUsers, canReadClusters, canReadMachines } = usePermissions()
 
-const { status: backupStatus } = setupBackupStatus()
-const { canSyncKubernetesManifests, canManageClusterFeatures } = setupClusterPermissions(
-  computed(() => route.params.cluster as string),
+const currentCluster = computed(() =>
+  'cluster' in route.params ? route.params.cluster : undefined,
```
Unix4ever, line 53: shouldn't just `route.params.cluster` work?

Slessi: `useRoute` is a lot more magic now, it is typed to exactly the type of the route your are currently looking at. so if you are in a `/clusters/MY-CLUSTER/**/*` type route, `route.params.cluster` is typed as string.

But here, in `TSideBar`, the router is aware that it is _not_ inside any such component, so `route.params` is a union type of every possible param combination, so its like trying to access `.cluster` on `{} | { cluster: string } | { machine: string} | { cluster: string; machine: string' }` and TS blocks you.

`'cluster' in route.params` fixes it because it is a type guard that narrows the type.

frontend/src/pages/(authenticated)/clusters/[cluster]/machine/[machine]/logs/[service].vue
```diff
@@ -166,7 +168,7 @@ const err = computed(() => {
 </template>
 
 <style scoped>
-@reference "../../../index.css";
+@reference "../../../../../../../index.css";
```
Comment (Unix4ever) [inline frontend/src/pages/(authenticated)/clusters/[cluster]/machine/[machine]/logs/[service].vue:171]: this looks even worse now 😅 can we somehow allow @/ syntax here?

Comment (Slessi) [reply frontend/src/pages/(authenticated)/clusters/[cluster]/machine/[machine]/logs/[service].vue:171]: As far as I am aware there is not but the plan is to completely get rid of these, its only needed if you wanna `@apply <tailwindstuff>` and theres never really a need for that

Ive been progressively inlining the classes and removing these sections entirely, eventually all will be gone. Inline classes is the tailwind way!

frontend/src/components/SideBar/TSideBar.vue
Unix4ever: 🤯

frontend/src/main.ts
```diff
@@ -35,6 +36,11 @@ if (import.meta.env.PROD) {
   }
 }
 
+// This will update routes at runtime without reloading the page
+if (import.meta.hot) {
```
Unix4ever, line 40: neat!

PR #2461, opened 2026-03-06
fix: use localhost for internal kubeconfig server address

utkuozdemir:
The internally generated Kubeconfig resource used the SideroLink WireGuard address as the kube-apiserver endpoint. Since this kubeconfig is only consumed by Omni itself (k8s proxy, internal k8s client) and the load balancer runs in the same process, rewrite the server to localhost instead.

The primary reason for doing this is that leaving the host as-is does not work consistently across platforms. macOS does not route traffic to a local WireGuard interface address back to the host itself, causing Kubernetes proxy connections to hang.

Existing clusters will pick up the change on the next kubeconfig regeneration (e.g., secret rotation).

### DIFF
--- a/internal/pkg/certs/kubernetes.go
+++ b/internal/pkg/certs/kubernetes.go
@@ -10,6 +10,8 @@ import (
 	stdlibx509 "crypto/x509"
 	"encoding/base64"
 	"fmt"
+	"net"
+	"net/url"
 	"time"
 
 	talosx509 "github.com/siderolabs/crypto/x509"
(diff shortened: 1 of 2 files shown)


internal/pkg/certs/kubernetes.go
```diff
@@ -161,3 +177,24 @@ func NewKubernetesCertificateAndKey(ca *talosx509.PEMEncodedCertificateAndKey, c
...
+	if err != nil {
+		return "", "", fmt.Errorf("failed to parse endpoint %q: %w", endpoint, err)
+	}
+
+	_, port, err := net.SplitHostPort(u.Host)
+	if err != nil {
+		return "", "", fmt.Errorf("failed to split host:port %q: %w", u.Host, err)
+	}
+
+	tlsServerName = u.Hostname()
```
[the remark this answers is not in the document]
utkuozdemir: defensive, no need, we know the input.

smira:
I'm confused on the OS X statement here... if it doesn't route back to Wireguard, it should break many other things? Wireguard link should be a tunnel device with the address assigned to it?

smira:
Or in other words, what is "OS X" here? Do you run Omni outside of a docker container?

utkuozdemir:
> Or in other words, what is "OS X" here? Do you run Omni outside of a docker container?

Yes, I run Omni directly, on macOS.

When Omni makes a request to the siderolink ip, which is held on the same machine as itself (`fdae:41e4:649b:9303::1`), it works on Linux - the IP is recognized as a local IP, and the packet gets delivered locally to itself.

This doesn't work on mac though - it gets routed into the utun device instead it getting delivered locally. WireGuard then has no peer with that address, so it goes nowhere.

I wanted to solve this to make my development env better (I also solved a few other things related to macOS in the past).

BTW, only Kubernetes breaks, because it is the only place (I think) where we make an external request to ourselves.

internal/pkg/certs/kubernetes.go
smira: kube-apiserver should have a SAN for localhost always, so I think using 127.0.0.1 directly is fine here, and it will reduce the size of the change

smira: correction: it looks like Talos started adding 127.0.0.1 two years ago, previously "localhost" was there, it should work always

utkuozdemir: so, use `localhost`, not `127.0.0.1`? I did the `tlsServerName` to be on the safe side.

smira: `localhost` 100% 🤣

utkuozdemir: tested locally with localhost, all good.

PR #2477, opened 2026-03-10
feat: add configurable bandwidth rate limiting for SideroLink tunnel

oguzkilcan:
Introduce token-bucket based bandwidth rate limiting for the SideroLink WireGuard tunnel, configurable via services.siderolink.bandwidthLimitMbps and services.siderolink.bandwidthLimitBurstBytes config fields (with corresponding CLI flag fallbacks).
Rate limiting is applied in both directions: outbound via a wrapped conn.Bind and inbound via a TUN input packet filter. A shared limiter drops packets exceeding the budget, relying on TCP congestion control to throttle senders. Disabled by default (0 = unlimited).

Also adds a Grafana service to docker-compose with pre-built Omni dashboards for local development 

When all values are zero (default), rate limiting is disabled. Dropped packets are tracked via a Prometheus counter (`omni_siderolink_ratelimit_dropped_packets_total`).

Resolves: #2441

### DIFF
--- a/cmd/omni/cmd/cmd.go
+++ b/cmd/omni/cmd/cmd.go
@@ -207,6 +207,12 @@ func defineServiceFlags(rootCmd *cobra.Command, rootCmdFlagBinder *FlagBinder, f
 	rootCmdFlagBinder.IntVar("event-sink-port", flagDescription("services.siderolink.eventSinkPort", schema), &flagConfig.Services.Siderolink.EventSinkPort)
 	rootCmdFlagBinder.IntVar("log-server-port", flagDescription("services.siderolink.logServerPort", schema), &flagConfig.Services.Siderolink.LogServerPort)
 	EnumVar(rootCmdFlagBinder, "join-tokens-mode", flagDescription("services.siderolink.joinTokensMode", schema), &flagConfig.Services.Siderolink.JoinTokensMode)
+	rootCmdFlagBinder.Uint64Var("siderolink-bandwidth-limit-mbps",
+		flagDescription("services.siderolink.bandwidthLimitMbps", schema),
+		&flagConfig.Services.Siderolink.BandwidthLimitMbps)
+	rootCmdFlagBinder.Uint64Var("siderolink-bandwidth-limit-burst-bytes",
+		flagDescription("services.siderolink.bandwidthLimitBurstBytes", schema),
+		&flagConfig.Services.Siderolink.BandwidthLimitBurstBytes)
 
 	// MachineAPI
 	for _, prefix := range []string{"siderolink", "machine"} {
--- a/hack/compose/grafana/provisioning/dashboards/dashboards.yml
+++ b/hack/compose/grafana/provisioning/dashboards/dashboards.yml
@@ -1,0 +1,10 @@
+apiVersion: 1
+
+providers:
+  - name: Omni
+    type: file
+    disableDeletion: false
+    editable: true
+    options:
+      path: /var/lib/grafana/dashboards
+      foldersFromFilesStructure: false
--- a/internal/backend/server.go
+++ b/internal/backend/server.go
@@ -100,6 +100,7 @@ import (
 	"github.com/siderolabs/omni/internal/pkg/kms"
 	"github.com/siderolabs/omni/internal/pkg/machineevent"
 	"github.com/siderolabs/omni/internal/pkg/siderolink"
+	"github.com/siderolabs/omni/internal/pkg/siderolink/ratelimit"
 	"github.com/siderolabs/omni/internal/pkg/xcontext"
 )
 
@@ -613,6 +614,25 @@ func (s *Server) runMachineAPI(ctx context.Context) error {
 	omniState := s.state.Default()
 	machineEventHandler := machineevent.NewHandler(omniState, s.logger, s.siderolinkEventsCh, s.installEventCh)
 
+	var rateLimiter *ratelimit.Limiter
+
+	if s.cfg.Services.Siderolink.GetBandwidthLimitMbps() > 0 {
+		rateLimiter = ratelimit.NewLimiter(
+			s.cfg.Services.Siderolink.GetBandwidthLimitMbps(),
+			s.cfg.Services.Siderolink.GetBandwidthLimitBurstBytes(),
+			s.logger,
+		)
+	}
+
+	if rateLimiter != nil {
+		prometheus.MustRegister(rateLimiter)
+
+		s.logger.Info("bandwidth rate limiting enabled",
+			zap.Uint64("mbps", s.cfg.Services.Siderolink.GetBandwidthLimitMbps()),
+			zap.Uint64("burst_bytes", s.cfg.Services.Siderolink.GetBandwidthLimitBurstBytes()),
+		)
+	}
+
 	slink, err := siderolink.NewManager(
 		ctx,
 		omniState,
@@ -624,6 +644,7 @@ func (s *Server) runMachineAPI(ctx context.Context) error {
 		s.logHandler,
 		machineEventHandler,
 		s.linkCounterDeltaCh,
+		rateLimiter,
 	)
 	if err != nil {
 		return err
--- a/internal/pkg/siderolink/ratelimit/bind.go
+++ b/internal/pkg/siderolink/ratelimit/bind.go
@@ -1,0 +1,38 @@
+// Copyright (c) 2026 Sidero Labs, Inc.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+
+package ratelimit
+
+import (
+	"time"
+
+	"golang.zx2c4.com/wireguard/conn"
+)
+
+// rateLimitedBind wraps a conn.Bind to drop outbound packets that exceed
+// the bandwidth budget.
+type rateLimitedBind struct {
+	conn.Bind
+	limiter *Limiter
+}
+
+// Send implements conn.Bind. It drops packets that exceed the rate limit.
+func (b *rateLimitedBind) Send(bufs [][]byte, ep conn.Endpoint) error {
+	now := time.Now()
+
+	filtered := make([][]byte, 0, len(bufs))
+
+	for _, buf := range bufs {
+		if b.limiter.allow(now, len(buf), MetricLabelDirOutbound) {
+			filtered = append(filtered, buf)
+		}
+	}
+
+	if len(filtered) == 0 {
+		return nil
+	}
+
+	return b.Bind.Send(filtered, ep)
+}
--- a/internal/pkg/siderolink/ratelimit/limiter.go
+++ b/internal/pkg/siderolink/ratelimit/limiter.go
@@ -1,0 +1,136 @@
+// Copyright (c) 2026 Sidero Labs, Inc.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+
+// Package ratelimit provides bandwidth rate limiting for the SideroLink WireGuard tunnel.
+//
+// Rate limiting is applied at two points:
+//   - Outbound: WireGuard Bind Send (packets from Omni to peers) via WrapBind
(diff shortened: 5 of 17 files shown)


internal/pkg/siderolink/ratelimit/limiter.go
```diff
@@ -0,0 +1,158 @@
...
+
+	if globalMbps > 0 {
+		bytesPerSec := globalMbps * 1_000_000 / 8
+		burst := int(globalBurstBytes)
+
+		if burst == 0 {
+			burst = int(bytesPerSec)
+		}
+
+		l.globalLimiter = rate.NewLimiter(rate.Limit(bytesPerSec), burst)
```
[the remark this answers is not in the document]
Unix4ever: this seems like a valid issue

cmd/omni/cmd/cmd.go
```diff
@@ -207,6 +207,18 @@ func defineServiceFlags(rootCmd *cobra.Command, rootCmdFlagBinder *FlagBinder, f
...
+	rootCmdFlagBinder.Uint64Var("siderolink-bandwidth-limit-per-node-mbps",
+		flagDescription("services.siderolink.bandwidthRateLimit.perNodeMbps", schema),
+		&flagConfig.Services.Siderolink.BandwidthRateLimit.PerNodeMbps)
+	rootCmdFlagBinder.Uint64Var("siderolink-bandwidth-limit-per-node-burst-bytes",
+		flagDescription("services.siderolink.bandwidthRateLimit.perNodeBurstBytes", schema),
+		&flagConfig.Services.Siderolink.BandwidthRateLimit.PerNodeBurstBytes)
+	rootCmdFlagBinder.Uint64Var("siderolink-bandwidth-limit-global-mbps",
+		flagDescription("services.siderolink.bandwidthRateLimit.globalMbps", schema),
+		&flagConfig.Services.Siderolink.BandwidthRateLimit.GlobalMbps)
+	rootCmdFlagBinder.Uint64Var("siderolink-bandwidth-limit-global-burst-bytes",
```
Orzelius: I think I'd be useful to document how the burst bandwith is measured somewhere. Currently it's unclear to developer and user alike without looking at the code

Orzelius: unless it's a standard thing I'm ignorant about

Unix4ever (approved):

internal/pkg/siderolink/ratelimit/bind.go
```diff
@@ -0,0 +1,47 @@
...
+// Send implements conn.Bind. It drops packets that exceed the rate limit.
+func (b *rateLimitedBind) Send(bufs [][]byte, ep conn.Endpoint) error {
+	addrPort, err := netip.ParseAddrPort(ep.DstToString())
+	if err != nil {
+		// If the endpoint address is unparseable, skip rate limiting and let the
+		// underlying bind handle (or reject) the packet.
+		return b.Bind.Send(bufs, ep)
+	}
+
+	addr := addrPort.Addr()
```
smira: there's already `ep.DstIP()`, I don't think we need to parse once again

internal/pkg/siderolink/ratelimit/limiter.go
```diff
@@ -0,0 +1,158 @@
...
+	dirInbound  = "inbound"
+	dirOutbound = "outbound"
+)
+
+// Limiter applies bandwidth rate limiting using token bucket algorithm.
+// It supports both per-peer and global rate limits.
+type Limiter struct {
+	globalLimiter  *rate.Limiter
+	droppedPackets *prometheus.CounterVec
+	peers          containers.ConcurrentMap[netip.Addr, *rate.Limiter]
```
smira: I _think_ this might cause more complexity than needed, as this map potentially needs to be cleaned up, etc.

What is the usecase for limiting by address? This might prevent starvation in general, but besides that it doesn't get any benefits imho over the global limit (which is way easier to reason about)

internal/pkg/siderolink/ratelimit/limiter.go
```diff
@@ -0,0 +1,158 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/siderolabs/gen/containers"
+	"github.com/siderolabs/siderolink/pkg/tun"
+	"golang.org/x/net/ipv6"
+	"golang.org/x/time/rate"
+	"golang.zx2c4.com/wireguard/conn"
+)
+
+const (
+	dirInbound  = "inbound"
```
smira: Looks like this is not used for actual limits (we limit on in+out), but rather it is used for metrics only. Probably it makes sense to be explicit?

internal/pkg/siderolink/ratelimit/limiter.go
oguzkilcan: It's adding a lot more complexity. Should I drop peer rate limiting(per talos node) and only do global(instance wide) rate limiting?

oguzkilcan: The idea of having that was to solve a potential noisy neighbor problem.

smira: What is the noisy neighbor in this case? Like you have Omni going crazy querying a single machine? Or Talos machine going crazy and sending e.g. too many logs?

oguzkilcan: I don't think those would create excessive traffic. My main worry was misuse of workload proxy or k8s proxy starving the global bandwidth and affecting Omni's internal workings.

cmd/omni/cmd/cmd.go
oguzkilcan: I'll add documentation for how it works.

internal/pkg/siderolink/ratelimit/limiter.go
smira: I mean to document it more explicitly (or drop it?)

internal/pkg/siderolink/ratelimit/limiter.go
smira: If we throttle by machine, k8s API traffic goes to controlplanes, so it will throttle the most important thing. Workload proxy goes to any random node, so still sounds weird that it will be throttled (from user point of view).

We could expose the throttling in the UI as a warning I guess if we want to.

oguzkilcan: I'll drop peer rate limiting then. We'll only do instance wide rate limiting.

jeffbehl-beep:
Is there any sort of notification or metric one can view to see if one is being rate limited?  Without that it's a black box as to why things are slower than expected

oguzkilcan:
> Is there any sort of notification or metric one can view to see if one is being rate limited? Without that it's a black box as to why things are slower than expected

There will be a metric counting dropped packages, however this metrics is not being utilized anywhere to show that siderolink bandwith is being rate limited ATM.

Unix4ever:
> Is there any sort of notification or metric one can view to see if one is being rate limited? Without that it's a black box as to why things are slower than expected

I think it's a good idea to show that data somewhere in Omni. Having metrics is nice, but users can't see our metrics in SaaS

oguzkilcan:
I tried finding a place for this but nothing jumped to me. I was thinking in talos dashboard we are showing siderolink status so this info could go next to it but in Omni we do not show need to show siderolink status if it's not working machine is disconnected. So I need some UX guidance for it. However, this setting requires explicit configuration changes, it's not enabled by default. We can tackle the UI part on a different PR.

jeffbehl-beep:
> I tried finding a place for this but nothing jumped to me.

UI would be great, but I was actually just seeing if it'd be a possibility to log it in some manner.  Obviously not for every dropped packed, but something that would mention "X packets dropped over the last X seconds".  It's something I've seen in other rate-limiting implementations and is more likely to be seen by an operator than via metrics

oguzkilcan:
> "X packets dropped over the last X seconds"
I'm not really sure if this will be really helpful, because dropped package number doesn't actually tell much information by itself. I works in conjunction with siderolink_traffic for received/sent bytes. If we have set ratelimit to be 20 Mbps and total traffic is averaging around 20 Mbps, then the dropped package counter increasing tells us we are actively rate limiting. 

I feel like, I'd like to know as an Operator whether I'm being actively rate limited and how many seconds I was rate limited in a certain timeframe.

cmd/omni/cmd/cmd.go
```diff
@@ -207,6 +207,12 @@ func defineServiceFlags(rootCmd *cobra.Command, rootCmdFlagBinder *FlagBinder, f
 	rootCmdFlagBinder.IntVar("event-sink-port", flagDescription("services.siderolink.eventSinkPort", schema), &flagConfig.Services.Siderolink.EventSinkPort)
 	rootCmdFlagBinder.IntVar("log-server-port", flagDescription("services.siderolink.logServerPort", schema), &flagConfig.Services.Siderolink.LogServerPort)
 	EnumVar(rootCmdFlagBinder, "join-tokens-mode", flagDescription("services.siderolink.joinTokensMode", schema), &flagConfig.Services.Siderolink.JoinTokensMode)
+	rootCmdFlagBinder.Uint64Var("siderolink-bandwidth-limit-mbps",
+		flagDescription("services.siderolink.bandwidthLimitMbps", schema),
+		&flagConfig.Services.Siderolink.BandwidthLimitMbps)
+	rootCmdFlagBinder.Uint64Var("siderolink-bandwidth-limit-burst-bytes",
+		flagDescription("services.siderolink.bandwidthLimitBurstBytes", schema),
+		&flagConfig.Services.Siderolink.BandwidthLimitBurstBytes)
```
utkuozdemir, line 215: might wanna add these also into the chart values

hack/compose/grafana/provisioning/dashboards/dashboards.yml
utkuozdemir, line 1: ❤️ 
By the way, later we might consider getting the dashboards ops guys are using in (might need to add some relabeling rules to satisfy them).

internal/backend/server.go
```diff
@@ -100,6 +100,7 @@ import (
 	"github.com/siderolabs/omni/internal/pkg/kms"
 	"github.com/siderolabs/omni/internal/pkg/machineevent"
 	"github.com/siderolabs/omni/internal/pkg/siderolink"
+	sidelinkratelimit "github.com/siderolabs/omni/internal/pkg/siderolink/ratelimit"
```
utkuozdemir: Why custom package name?

internal/backend/server.go
```diff
@@ -613,6 +614,20 @@ func (s *Server) runMachineAPI(ctx context.Context) error {
 	omniState := s.state.Default()
 	machineEventHandler := machineevent.NewHandler(omniState, s.logger, s.siderolinkEventsCh, s.installEventCh)
 
+	rateLimiter := sidelinkratelimit.NewLimiter(
+		s.cfg.Services.Siderolink.GetBandwidthLimitMbps(),
+		s.cfg.Services.Siderolink.GetBandwidthLimitBurstBytes(),
+	)
+
+	if rateLimiter != nil {
```
utkuozdemir, line 627: You construct it above unconditionally, but here you check for nil. I guess the intention was something else here. Maybe you need to do a nil check on `BandwidthLimitMbps`, and construct it inside the if?

utkuozdemir: Also, maybe it should go inside the manager? Instead of constructing it here and passing it in?

hack/compose/grafana/provisioning/dashboards/dashboards.yml
oguzkilcan: These are those dashboards without all the selectors.

internal/backend/server.go
oguzkilcan: Ratelimit is enabled if BandwidthLimitMbps > 0, otherwise it should be nil. I forgot to add that check.

cmd/omni/cmd/cmd.go
oguzkilcan: Should we always update chart values when we make changes to Omni config?

PR #2480, opened 2026-03-10
chore: bump default talos version

Slessi:
Bump default talos version to 1.12.5


hack/test/common.sh
```diff
@@ -22,14 +22,14 @@ echo "127.0.0.1 omni.localhost" | tee -a /etc/hosts
...
-export KUBERNETES_VERSION=1.35.1
+export TALOS_VERSION=1.12.5
+export KUBERNETES_VERSION=1.35.2
 # To use in:
 # - Omni upgrade tests, to prevent Talos changes interfering with Omni changes
 # - Talos minor upgrade tests
 export STABLE_TALOS_VERSION=1.11.6
 export ANOTHER_OMNI_VERSION="${ANOTHER_OMNI_VERSION:-$LATEST_STABLE_OMNI}"
-export ANOTHER_KUBERNETES_VERSION=1.34.4
+export ANOTHER_KUBERNETES_VERSION=1.35.1
```
Slessi: @utkuozdemir is this correct for "another k8s version" ?

utkuozdemir: I guess better to keep it to point to previous minor, 1.34.5 in this case.

PR #2486, opened 2026-03-10
fix: do not release lock on apply config fails

Unix4ever:
The code there was also incorrect: it was skipping setting the `LastError` on the `ClusterMachineConfigStatus` resource. Also add an integration test to verify that invalid config errors are properly reported.


Unix4ever:
> does this fix also pending config diff? or is this separate?

Doing a separate fix for that

PR #2504, opened 2026-03-12
IGNORE: chore: fix chart 2.5.10 - not to be merged

utkuozdemir:
Run a workflow to re-push the accidentally overwritten 2.5.10 chart in OCI registry.


utkuozdemir:
This pushed the correct version of chart 2.5.10 to OCI - thanks @kevintijssen for pointing this out.

PR #2507, opened 2026-03-12
feat: access all Talos nodes directly via their SideroLink endpoint

utkuozdemir:
Eliminate Talos node-to-node proxying from Omni. Instead of connecting to a control plane node and using `node`/`nodes` gRPC headers to proxy to the target, Omni now connects directly to each node's SideroLink management address.

Two client types replace the old single cluster-wide client:
- `GetForCluster`: connects to all CP management addresses, gRPC load-balances across them. For operations targeting any healthy CP (e.g., etcd audit, k8s upgrade manifest status).
- `GetForMachine`: connects directly to a specific node's SideroLink address. The cache supports both regular (secure) and maintenance clients.

There are three paths Omni uses to talk to Talos nodes:
- Path 1: Controllers build ad-hoc Talos clients directly to nodes. These already connected directly, no proxying was involved.
- Path 2: talosctl / Talos client library / frontend direct Talos API calls → Omni gRPC router → Talos. The router now creates per-node backends connecting directly to each node's SideroLink address, and strips `node`/`nodes` headers before forwarding.
- Path 3: Frontend resource service (Get/List/Watch) → Omni resolves machine IDs → Talos runtime dispatches to `GetForMachine` or `GetForCluster`. `WithNode` is no longer used.

Both the Talos client factory and the gRPC router use LRU caches with larger capacity but shorter TTLs, and actively invalidate entries when relevant resources change.

The support bundle collector now uses per-node clients instead of connecting to a CP node and proxying. The frontend sends a single machine ID instead of a list.

Fixes: https://github.com/siderolabs/omni/issues/2256


internal/backend/grpc/router/router.go
```diff
@@ -164,58 +176,82 @@ func (r *Router) Director(ctx context.Context, fullMethodName string) (proxy.Mod
 			return proxy.One2One, nil, err
 		}
 
+		if len(backends) > 1 {
```
smira: there is one slight issue here, and I'm not sure how to fix it exactly.

One2Many proxying relies on a specific structure of the API, which most of the "legacy" Talos API adhere to, but newer APIs don't support One2Many proxying at all.

`talosctl` does things correctly and doesn't allow calling new APIs with multiple nodes (or handles multiplexing client-side). 

But if we imagine a random Talos API client calling through Omni, if it specifies `nodes: a,b` and uses API which is not there for One2Many, it will receive garbage response back.

Talos's `apid` knows which APIs are not eligible for one2many, and explicitly rejects such calls. In the previous implementation this flow worked, as Talos apid saw the `nodes` and could reject, this new way might have some consequences.

internal/backend/grpc/router/omni_backend.go
```diff
@@ -8,15 +8,12 @@ package router
...
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/metadata"
-
-	"github.com/siderolabs/omni/internal/backend/dns"
 )
 
-// ResolvedNodesHeaderKey is used to propagate the node IP information from the node/nodes headers to the backend.
-const ResolvedNodesHeaderKey = "resolved-nodes"
+// ResolvedMachinesHeaderKey is used to propagate machine IDs from the node/nodes headers to the backend.
+const ResolvedMachinesHeaderKey = "resolved-nodes"
```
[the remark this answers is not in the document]
utkuozdemir: rather keep this to be less invasive.

go.mod
```diff
@@ -6,6 +6,10 @@ replace (
 	// forked saml library that has the fix for Fusion Auth ACS parsing
 	github.com/crewjam/saml => github.com/unix4ever/saml v0.0.0-20250630213700-66b137182abe
 
+	// TODO(Utku): remove and use upstream after merged
+	github.com/siderolabs/go-kubernetes => github.com/utkuozdemir/sidero-go-kubernetes v0.0.0-20260313013548-611725496bc7
+	github.com/siderolabs/go-talos-support => github.com/utkuozdemir/sidero-go-talos-support v0.0.0-20260306111941-6c9c09e9bb9d
+
```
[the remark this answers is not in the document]
utkuozdemir: Yes, we know, no worries.

internal/backend/runtime/omni/loader_test.go
```diff
@@ -189,7 +189,7 @@ func errString(expectedErr string) func(*testing.T, omni.Loader, error) {
 func IgnoreUnexported(vals ...any) cmp.Option {
 	return cmpopts.IgnoreUnexported(xslices.Map(vals, func(v any) any {
 		val := reflect.ValueOf(v)
-		if val.Kind() == reflect.Ptr {
+		if val.Kind() == reflect.Pointer {
```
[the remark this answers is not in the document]
utkuozdemir: done

internal/backend/runtime/talos/clients.go
```diff
@@ -314,29 +493,57 @@ func (factory *ClientFactory) StartCacheManager(ctx context.Context) error {
 		return fmt.Errorf("failed to watch TalosConfigs: %w", err)
 	}
 
+	err = factory.omniState.WatchKind(ctx, clusterMachineMd, eventCh)
+	if err != nil {
+		return fmt.Errorf("failed to watch ClusterMachines: %w", err)
+	}
```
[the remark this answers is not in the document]
utkuozdemir: that doesn't change.

hack/test/e2e-talemu.sh
```diff
@@ -11,6 +11,7 @@ set -eoux pipefail
 source ./hack/test/common.sh
 
 TALEMU_INFRA_PROVIDER_IMAGE=ghcr.io/siderolabs/talemu-infra-provider:latest
+TALEMU_INFRA_PROVIDER_IMAGE=ghcr.io/utkuozdemir/talemu-infra-provider:drop-node-proxying # TODO(Utku): remove this after testing / when it is merged to talemu main
```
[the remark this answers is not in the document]
utkuozdemir: gonna be removed.

hack/test/integration-talemu.sh
```diff
@@ -11,6 +11,7 @@ set -eoux pipefail
 source ./hack/test/common.sh
 
 TALEMU_INFRA_PROVIDER_IMAGE=ghcr.io/siderolabs/talemu-infra-provider:latest
+TALEMU_INFRA_PROVIDER_IMAGE=ghcr.io/utkuozdemir/talemu-infra-provider:drop-node-proxying # TODO(Utku): remove this after testing / when it is merged to talemu main
```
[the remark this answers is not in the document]
utkuozdemir: that's gonna be removed anyway.

internal/backend/dns/service.go
```diff
@@ -90,92 +71,79 @@ type Service struct {
...
+		case *omni.MachineStatus:
+			d.deleteMachineMappings(r.Metadata().ID())
+		}
+	case state.Created, state.Updated:
+		switch r := ev.Resource.(type) {
+		case *omni.ClusterMachineIdentity:
+			d.updateEntryByIdentity(r)
+		case *omni.MachineStatus:
+			d.updateEntryByMachineStatus(r)
+		}
```
Unix4ever, line 151: why did we drop default case with the warn? I think it's good to keep it to prevent coding errors

internal/integration/talos_test.go
```diff
@@ -180,25 +180,11 @@ func AssertTalosAPIAccessViaOmni(testCtx context.Context, options *TestOptions,
 			assert.NoError(t, err)
 			assert.Len(t, version.Messages, len(machineNames))
 
-			assert.Equal(t,
```
Unix4ever, line 177: If we don't get the hostname, how would we map the responses? Then why do we need to support multiple nodes?

Just in case one2many case is only useful when the request is load balanced to any of the nodes.

internal/backend/dns/service.go
utkuozdemir: I just thought it was defensive code, but can bring it back certainly.

internal/backend/grpc/router/router.go
utkuozdemir: changed it to preserve when "nodes" is present and >1.

utkuozdemir:
> I see you changed the frontend `context` from `nodes[]` to `node`
> 
> I knew already that in fact the API only worked if we submit exactly one node frontend side, but I got the impression we accepted multiple nodes from other sources (e.g. omnictl/talosctl ?)
> 
> Should we keep it `nodes[]` and eventually support actually passing multiple? Or is this exactly what the PR is addressing, to remove this `nodes[]` array behaviour in general?

The frontend always submitted one node actually, even when using "nodes". In such cases, better to specify "node". 
Our goal is to eliminate proxying through other nodes for single-node calls.

With these changes, Omni always makes a direct connection to the target machine when a single node is targeted (be it via "node" or "nodes" header from frontend).
Only when nodes>=2, we rely on the Talos's proxying behavior (request hits one node, it proxies to others). We eliminate all other cases of proxying.

But from the API consumers standpoint (talosctl etc.), nothing has changed - both "node" and "nodes" work.

The long term goal is to get rid of all proxying - for example, when multiple nodes are specified, do the multiplexing/merging responses at client side. Newer Talos APIs are already designed with this in mind.

internal/integration/talos_test.go
utkuozdemir: You're right, these assertions are required. I removed them when I was attempting to remove the 'nodes' proxying completely. Even then removing these assertions (and .metadata.hostname) were wrong. Then Andrey pointed out it's not really possible, so I brought it back, and now I brought back these assertions.

About one2many, I didn't exactly get what you mean. We right now have load-balancing (grpc's round robin) for cluster-wide cluents, the nodes header when >=2 is preserved, and Talos's apid will do its one2many proxying.

utkuozdemir: Ah, we probably need to even preserve the header (and hostname on response) on single node calls via "nodes" header. Looking into it.

utkuozdemir: brought back the nodes header logic, and the assertions.

internal/backend/grpc/router/router.go
```diff
@@ -343,74 +443,95 @@ func (r *Router) getClusterCredentials(ctx context.Context, clusterName string)
...
 
-	if err := s.WatchKind(ctx, resource.NewMetadata(resources.DefaultNamespace, omni.ClusterType, "", resource.VersionUndefined), events); err != nil {
-		return err
+	if err := s.WatchKind(ctx, omni.NewCluster("").Metadata(), events); err != nil {
+		return fmt.Errorf("failed to watch Clusters: %w", err)
 	}
 
-	if err := s.WatchKind(ctx, resource.NewMetadata(resources.DefaultNamespace, omni.ClusterSecretsType, "", resource.VersionUndefined), events); err != nil {
-		return err
+	if err := s.WatchKind(ctx, omni.NewClusterSecrets("").Metadata(), events); err != nil {
```
utkuozdemir, line 479: Watch also ClusterMachineSecrets

internal/backend/grpc/router/router.go
utkuozdemir: add metrics breakdown by type of clients

PR #2508, opened 2026-03-13
feat: add a way to sync Kubernetes manifests in Omni

Unix4ever:
Manifests support two modes:
- `FULL` - Omni will keep the manifest in sync always.
- `ONE_TIME` - Omni will apply the manifest only if it doesn't exist. If the manifest is removed by hand and then changed in Omni it will be applied too.

Manifests are applied using service side apply, Omni now has three inventories: `omni-internal-inventory`, `omni-user-inventory` and `omny-sync-one-time`:

- User inventory will be used for user managed manifests.
- Internal one will be used for the manifests which are created by Omni controllers (workloadproxy, advanced healtcheck service).
- One time inventory is used with NoPrune enabled. If the manifest is applied it's just removed from the list of applied manifests: that ensures that manifests changes are not going happen.

Manifests also support setting namespace to all namespaced resources. It might be useful for the huge manifest files which are supplied without the namespace (similar to `kubectl apply -n namespace -f manifest.yaml`).

Status is now computed from the actual inventory and resource state: before running the sync and after it Omni updates the status of manifests for the whole cluster.
It gets all resources from the inventory, queries the Kubernetes resources and checks if they're running.
Wait is not called in sync anymore. This makes controller loop faster.


internal/integration/kubernetes_test.go
```diff
@@ -574,3 +580,235 @@ func AssertKubernetesDeploymentHasRunningPods(testCtx context.Context, managemen
...
+
+		// wait a bit to ensure that the manifest group controller processes the update of the manifest group
+		// we don't have a reliable signal of the config being processed unfortunately
+		time.Sleep(5 * time.Second)
+
+		// the service should not be created because group mode is one time
+		_, err = kubeClient.CoreV1().Services("default").Get(ctx, "nginx-full", metav1.GetOptions{})
+		require.Error(t, err)
+		require.True(t, kubeerrors.IsNotFound(err))
+
```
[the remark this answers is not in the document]
Unix4ever: this is not a solution for this particular issue.

internal/backend/runtime/omni/controllers/omni/kubernetes/cluster_manifest_status.go
```diff
@@ -0,0 +1,651 @@
...
+		groups[res.Metadata().ID()] = res.TypedSpec().Value
+
+		inventory := omniconsts.SSAOmniUserInventory
+
+		_, systemManifest := res.Metadata().Labels().Get(omni.LabelSystemManifest)
+		if systemManifest {
+			inventory = omniconsts.SSAOmniInternalInventory
+		}
+
+		if res.TypedSpec().Value.Mode == specs.KubernetesManifestGroupSpec_ONE_TIME {
```
Unix4ever, line 239: might have issues with ONE_TIME + internal manifest types. But we might not need to create such manifest.

client/api/omni/specs/cluster_kubernetes_manifests_status.go
```diff
@@ -0,0 +1,112 @@
...
+		Name:      manifest.Name,
+		Kind:      manifest.GroupKind.Kind,
+		Group:     manifest.GroupKind.Group,
+		Namespace: manifest.Namespace,
+	}
+
+	gr.Manifests[id] = status
+}
+
+func (x *ClusterKubernetesManifestsStatusSpec) Reconcile(groups map[string]*KubernetesManifestGroupSpec) {
```
Unix4ever: rename to UpdateGroups

frontend/typed-router.d.ts
Slessi, line 1: @Unix4ever same as a comment i left for oguz https://github.com/siderolabs/omni/pull/2560#discussion_r2966108299

This shouldn't be committed, the `vue-router` bump slightly changed how this file was generated 😬 I think this old version of the file got comitted here due to the dev server running the old `vue-router` version, a bit annoying...

Unix4ever: okay

PR #2512, opened 2026-03-13
feat: add hsts header for omni frontend

Slessi:
Add HSTS (Strict-Transport-Security) response header to the frontend

### DIFF
--- a/internal/frontend/handler.go
+++ b/internal/frontend/handler.go
@@ -110,6 +110,9 @@ func (handler *StaticHandler) serveFile(w http.ResponseWriter, r *http.Request, 
 
 		defer file.Close() //nolint:errcheck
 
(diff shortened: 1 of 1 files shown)


internal/frontend/handler.go
```diff
@@ -110,6 +110,9 @@ func (handler *StaticHandler) serveFile(w http.ResponseWriter, r *http.Request,
 
 		defer file.Close() //nolint:errcheck
 
+		w.Header().Set("Strict-Transport-Security", "max-age=31536000; includeSubDomains")
+		w.Header().Set("X-Content-Type-Options", "nosniff")
```
Slessi, line 114: `Strict-Transport-Security` / `X-Content-Type-Options` are relevant for all requests so moved them out of the if/else

Issue #2464, opened 2026-03-07
Infra provider disconnect is not detected, leaving stale "Healthy" status and silently blocking cluster operations


jeffbehl-beep:
## Summary

When an infra provider process stops, Omni does not detect the disconnect. The provider continues to show as "Healthy" in the UI indefinitely, and cluster operations that depend on the provider (e.g., deletion) hang with no indication of what is blocking them.

## Environment

- **Omni**: v1.5.8
- **Talos**: v1.12.4
- **Kubernetes**: v1.35.2
- **Infra Provider**: Proxmox (`omni-infra-provider-proxmox`)

## Steps to Reproduce

1. Set up an infra provider (e.g., Proxmox) and create a cluster using its machine classes (tested with 1 CP + 5 workers)
2. Wait for the cluster to reach Ready state
3. Stop the provider process (e.g., `docker stop <container>`)
4. Wait >30 seconds (past the heartbeat interval + buffer)
5. Observe the provider still shows "Healthy" in Settings > Infra Providers
6. Delete the cluster
7. Observe:
   - The cluster hangs in "Destroying" state indefinitely
   - Worker nodes disappear from the cluster view but their VMs continue running in the infrastructure layer
   - No indication is provided as to why the deletion is blocked
8. Start the provider process again
9. The provider decommissions VMs and the cluster deletion completes

### 1. Provider health status is stale

After stopping the provider, the Infra Providers page continues to show the provider as "Healthy" indefinitely. It should show "Disconnected" within a reasonable time after the last heartbeat.

### 2. Cluster deletion hangs silently

Deleting a cluster that uses machines from the disconnected provider hangs in the "Destroying" state with no explanation. The user has no way to determine why the deletion is blocked without deep knowledge of Omni internals.

### 3. Misleading UI during blocked teardown

During the stuck deletion, worker nodes disappear from the cluster view, giving the impression they were fully removed. In reality, the underlying VMs continue running on the infrastructure side (e.g., in Proxmox). The UI provides no indication that these VMs still exist or that resources associated with them are stuck waiting on the disconnected provider.

## Suggested Improvements

- Provider health should transition to "Disconnected" within ~30-60 seconds of the last heartbeat
- When a cluster operation is blocked because a provider is disconnected, this should be clearly surfaced to the user (e.g., in cluster status, as an event, or as a warning on the teardown)
- During a blocked teardown, the UI should not give the impression that machines have been fully removed when their infrastructure-side VMs still exist
[...]

PR #2514, opened 2026-03-13
git fix: properly update infra provider connection status

Unix4ever:
The controller was missing requeue interval. Without it nothing will poke it co trigger reconciliation when the provider is gone.

Fixes: https://github.com/siderolabs/omni/issues/2464

### DIFF
--- a/internal/backend/runtime/omni/controllers/omni/infraprovider/combined_status.go
+++ b/internal/backend/runtime/omni/controllers/omni/infraprovider/combined_status.go
@@ -3,7 +3,7 @@
 // Use of this software is governed by the Business Source License
 // included in the LICENSE file.
 
-package omni
+package infraprovider
 
 import (
 	"context"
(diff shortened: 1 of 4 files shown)


internal/backend/runtime/omni/controllers/omni/infraprovider/combined_status.go
```diff
@@ -48,10 +47,16 @@ func NewInfraProviderCombinedStatusController() *InfraProviderCombinedStatusCont
 
 				infraProviderStatus.TypedSpec().Value.Health = &specs.InfraProviderCombinedStatusSpec_Health{}
 
+				var requeueInterval time.Duration
+
 				if providerHealthStatus != nil {
 					lastHeartbeatTime := providerHealthStatus.TypedSpec().Value.LastHeartbeatTimestamp.AsTime()
 
-					infraProviderStatus.TypedSpec().Value.Health.Connected = time.Since(lastHeartbeatTime) < constants.InfraProviderHealthCheckInterval+time.Second
+					infraProviderStatus.TypedSpec().Value.Health.Connected = time.Since(lastHeartbeatTime) < healthCheckInterval+time.Second
```
smira, line 55: another solution (which is simpler from my PoV) is to do something like https://github.com/siderolabs/omni/blob/7ed46ba968613fcc6f82c24ac6758a9cdb8d34bf/client/pkg/omni/resources/system/cert_refresh_tick.go#L17-L24

We just have a resource which gets updated every X minutes, and add it as an input to this controller, so it always wakes up every now and then to recheck

Unix4ever: Do you mean some global ticker that is used by multiple controllers?
Then it might be simpler... Kinda. At the same time the current controller will have to iterate over all infra providers when it triggers.
Code wise my current solution feels like less efforts.

smira: Up to you, mine was just a suggestion 👍

PR #2525, opened 2026-03-16
fix: load balancer COSI cache races

oguzkilcan:
- Fix health status diff: persist `lastHealthy` back into the running map after updating it — wrapper is a value type, so mutations were silently discarded, causing the status resource to be written on every health poll tick
- Fix stopped status race: move "set stopped" writes before the `ReaderListAll` cleanup loop so they no longer depend on a potentially stale COSI reader cache
- Fix duplicate port allocations: track allocated ports in an in-memory cache instead of `ReaderListAll`, which may lag behind recent writes when multiple clusters are created in quick succession


smira:
On unrelated note, this is a perfect example where we should have used LB port (endpoint) as the resource ID, and use this uniqueness constraint to avoid conflicts.

Issue #2535, opened 2026-03-17
[bug] backwards compatibility for omnictl v1.6 is broken


oguzkilcan:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

omnictl v1.6.0 doesn't support interacting with Omni instances older than v1.6. 

### Expected Behavior

Newer omnictl versions should support interacting with old Omni instances.

### Steps To Reproduce

1. Use omnictl v1.6.0 against an Omni instance older than v1.6.0
2. See error:
```[WARN] omnictl version differs from the backend version: "1.6.0" vs "1.5.10".
Error: failed to list notifications: rpc error: code = PermissionDenied desc = failed to validate: 1 error occurred:
	* rpc error: code = PermissionDenied desc = no access is permitted
```

PR #2537, opened 2026-03-17
fix: add omnictl backward compatibility with older Omni servers

oguzkilcan:
v1.6 omnictl uses new gRPC management endpoints for user operations (CreateUser, ListUsers, UpdateUser, DestroyUser) that don't exist on older Omni versions, and queries the Notification resource which was introduced in v1.6. This makes every omnictl command fail against pre-v1.6 Omni.

Detect the server version and:
 - Skip Notification listing on Omni instances older than v1.6
 - Fall back to direct COSI resource manipulation for user commands create, delete, list, set-role) on older Omni instances
 - Return clear version errors for v1.6-only features (secret rotation)
 - Pass ServerInfo through WithClient so commands can version-gate

Fixes: #2535


client/pkg/omnictl/user/create.go
```diff
@@ -29,14 +34,46 @@ var createCmd = &cobra.Command{
...
+		}
+
 		_, err := client.Management().CreateUser(ctx, email, createCmdFlags.role)
 
 		return err
 	}
 }
 
+// createUserLegacy creates a user by directly manipulating COSI resources (for servers older than v1.6).
+func createUserLegacy(ctx context.Context, client *client.Client, email string) error {
```
utkuozdemir, line 50: Do we need to bring back these legacy methods? I assumed we would only take care of the issue with `Notification` resources. Other things are arguably just maintenance burden - we anyway recommend users to keep their omnictl in sync with their Omni. The fix with `Notification`s would make sure that "not everything breaks", but we don't need full backward compatibility, IMO.

oguzkilcan: It's not new code and it will only work for Omni's older than v1.6. I don't see a problem bringing them back.

utkuozdemir: I know it's not new, and it would work, certainly. This is more of a policy question. If we commit to keeping omnictl always backward-compatible, more and more compatibility code will accumulate over time. And could shift expectations of users.

If we decide with having them though, I'm ok with that.

@Unix4ever wdyt?

oguzkilcan: I think we should aim to support at least 1 or 2 minor versions.

utkuozdemir: Btw, otherwise LGTM, so approving.

Unix4ever: I would drop it completely. But I guess it's already 2 approvals, so I guess it's fine

utkuozdemir: I can also live with it.

Issue #2280, opened 2026-02-06
[feature] Allow configuring the full domain for the workload service proxy


utkuozdemir:
Currently we only allow `--workload-proxying-subdomain`, which makes the workload proxy services to be served at the same level as Omni, from a different domain (it is not really a subdomain). This was done this way for our SaaS, but we can add another config flag to specify the full domain to provide more flexibility to on-prem setups. (For example, real subdomains to Omni could be used, or completely another domain).

logaritmisk:
This would be amazing! I currently have two instances of Omni running, and right now I have to do some less nice things to get that to work properly.

If it would be possible, it would be nice as well to have a toggle to remove the instance name suffix in the workload proxy url. As an example the address I get now is h2ct44-omni1.omni1.infra.my-domain.org, and it would be nice to remove that duplicate -omni1.

Also (sorry for spamming) I'm having issues to access the workloads with this setup. I can only access workloads from the instance that I most recently logged into. Even if I have logged in on both, one will always return a 403. My guess is due to the cookies publicKeyIdSignatureBase64 and publicKeyId both are on the parent domain. So in my case the domains for omni is omni1.infra.my-domain.org and omni2.infra.my-domain.org, but the cookies are on .infra.my-domain.org. So when I login on one instance they are overwritten. Could maybe this be related to how domain is handled for workload proxy? If not, I will create another ticket for that :)

PR #2538, opened 2026-03-17
feat: add useOmniSubdomain option for workload proxy

utkuozdemir:
Add a new boolean config field `services.workloadProxy.useOmniSubdomain` that places the workload proxy subdomain under Omni's own domain rather than as a sibling. When enabled, service URLs become `<alias>.<subdomain>.<omni-domain>` without the instance name suffix, and dashes are allowed in service aliases. This is recommended for on-prem and single-instance deployments.

Also add a regex pattern validation on the `subdomain` config field, and clear the stale URL on ExposedService resources when alias validation fails.

Closes: https://github.com/siderolabs/omni/issues/2280

### DIFF
--- a/internal/backend/runtime/omni/controllers/omni/internal/exposedservice/reconciler.go
+++ b/internal/backend/runtime/omni/controllers/omni/internal/exposedservice/reconciler.go
@@ -14,6 +14,7 @@ import (
 	"fmt"
 	"io"
 	"net/url"
+	"regexp"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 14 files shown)


internal/backend/runtime/omni/controllers/omni/internal/exposedservice/reconciler.go
```diff
@@ -27,24 +28,25 @@ import (
 	"github.com/siderolabs/omni/client/pkg/omni/resources/omni"
 )
 
+// dnsLabelRegexp validates an RFC 1123 DNS label: lowercase alphanumeric with optional dashes, no leading/trailing dash, max 63 chars.
+var dnsLabelRegexp = regexp.MustCompile(`^[a-z0-9]([a-z0-9-]{0,61}[a-z0-9])?$`)
```
utkuozdemir, line 32: Also added this validation - not only for the "new format", but for all service aliases. It is good to have IMO.

PR #2542, opened 2026-03-17
fix: use dynamic SQLite pool

smira:
Configure by default number of connections: 4 cached and 64 max.

This uses https://github.com/cosi-project/state-sqlite/pull/10

The goal is to lower overall memory usage by SQLite.

New metric:

```
 # HELP omni_sqlite_pool_connections Number of connections in the sqlite pool.
 # TYPE omni_sqlite_pool_connections untyped
 omni_sqlite_pool_connections 3
```

PR #2543, opened 2026-03-18
fix: correct SQLite size metrics to include indexes and freelist

oguzkilcan:
Table size queries (`omni_sqlite_subsystem_size_bytes`) filtered by dbstat name, missing index sizes. Join with sqlite_master to attribute index pages to their parent table.

DB size (`omni_sqlite_db_size_bytes`) used dbstat sum which excludes freelist pages. Use page_count * page_size to match actual file size.

Add `omni_sqlite_db_freelist_size_bytes` metric to track wasted space.

### DIFF
--- a/hack/compose/grafana/dashboards/omni-details.json
+++ b/hack/compose/grafana/dashboards/omni-details.json
@@ -1723,6 +1723,17 @@
           "legendFormat": "Total DB size",
           "range": true,
           "refId": "A"
+        },
+        {
+          "datasource": {
+            "type": "prometheus",
+            "uid": "Prometheus"
+          },
+          "editorMode": "code",
+          "expr": "omni_sqlite_db_freelist_size_bytes",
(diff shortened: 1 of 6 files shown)


hack/compose/grafana/dashboards/omni-details.json
```diff
@@ -1723,6 +1723,17 @@
...
+        {
+          "datasource": {
+            "type": "prometheus",
+            "uid": "Prometheus"
+          },
+          "editorMode": "code",
+          "expr": "omni_sqlite_db_freelist_size_bytes",
+          "legendFormat": "Freelist size",
+          "range": true,
+          "refId": "B"
```
smira, line 1736: could you please 🙏  add the metrics as well I put in https://github.com/siderolabs/omni/pull/2542 - I forgot that we have Grafana now!

PR #2551, opened 2026-03-18
release(v1.6.1): prepare release

Unix4ever:
This is the official v1.6.1 release.


Unix4ever:
> is this missing some files like `internal/version/data/tag` and helm ?

should be good now

Issue #2457, opened 2026-03-06
[bug] Omni's embedded discovery should not report public IP back to Talos


smira:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

It reports SideroLink address (as it sees from Talos connecting to it).

### Expected Behavior

No public IP discovery with embedded discovery service.

### Steps To Reproduce

See https://github.com/siderolabs/talos/issues/12897

The fix is to never set `cluster_ip` field here: https://github.com/siderolabs/discovery-api/blob/9c06846e6f9f4f5765d5e431f8e25dc44a7ff337/api/v1alpha1/server/cluster.proto#L52-L57

utkuozdemir:
Good catch

PR #2564, opened 2026-03-20
fix: disable client IP reporting in embedded discovery service

utkuozdemir:
When using Omni's embedded discovery service, nodes connect over SideroLink, so the discovery service sees the SideroLink tunnel address as the peer IP. It then reports this back to Talos as the node's "public IP", which Talos adds as a KubeSpan endpoint candidate. This causes KubeSpan to try connecting to peers via SideroLink addresses, which breaks connectivity.

Disable client IP reporting in the embedded discovery service so that SideroLink addresses are never reported back as public IPs.

Also fix the embedded discovery service checkbox on the cluster create page not properly binding to the cluster state.

Fixes: siderolabs/omni#2457

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -66,7 +66,7 @@ require (
 	github.com/siderolabs/crypto v0.6.4
 	github.com/siderolabs/discovery-api v0.1.8
 	github.com/siderolabs/discovery-client v0.1.15
-	github.com/siderolabs/discovery-service v1.0.15
+	github.com/siderolabs/discovery-service v1.0.16
 	github.com/siderolabs/gen v0.8.6
 	github.com/siderolabs/go-api-signature v0.3.12
 	github.com/siderolabs/go-debug v0.6.2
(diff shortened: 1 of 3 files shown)


go.mod
```diff
@@ -9,6 +9,9 @@ replace (
 	// fix for multiple panics, see https://github.com/neticdk/go-stdlib/pull/44
 	github.com/neticdk/go-stdlib => github.com/smira/go-stdlib v0.0.0-20260318082201-9d387eb2130d
 
+	// TODO(utku): switch to upstream after siderolabs/discovery-service#98 is merged and released
+	github.com/siderolabs/discovery-service => github.com/utkuozdemir/sidero-discovery-service v0.0.0-20260320142709-379016ae4d6f
```
utkuozdemir: Not going to merge this, ofc.

frontend/src/pages/(authenticated)/clusters/create.vue
```diff
@@ -382,10 +367,9 @@ const list = useTemplateRef('list')
         </Tooltip>
         <ClusterWorkloadProxyingCheckbox v-model="state.cluster.features.enableWorkloadProxy" />
         <EmbeddedDiscoveryServiceCheckbox
-          :checked="state.cluster.features.useEmbeddedDiscoveryService"
+          v-model="state.cluster.features.useEmbeddedDiscoveryService"
```
utkuozdemir: @Slessi This checkbox didn't work before - selection didn't persist to the cluster.

Slessi: I see the issue, but don't we need to check if `embeddedDiscoveryServiceAvailable` is true though? It was part of the toggle logic you removed.

Slessi: I made a "full" fix here https://github.com/siderolabs/omni/pull/2565

Issue #2172, opened 2026-01-21
MVP: Kubernetes Manifest Sync


smira:
The least common denominator language is the Kubernetes raw manifest. On top of it, we can support other ways like `kustomize` or `Helm`. 

The first MVP is to provide an option to sync the list of Kubernetes manifest from the cluster template to Omni, and Omni will handle the sync of these manifest to Kubrnetes using `go-kubernetes` library and SSA (Server-Side Apply), which includes support for creating, updating and pruning manifests. 

Put initial limit on the size of the manifest(s), might reconsider after MVP.

Ensure that manifests (as resources) when stored in Omni are compressed.

Support two sync methods:

* `full` (do the sync always, prune, etc.)
* `bootstrap` / `one-shot` (create once, never touch later, e.g. bootstrapping ArgoCD once, and ArgoCD manages itself later)

Consider manifest deduplication across clusters (content-addressable, user-managed, reference counting, back references, ...)

```yaml
cluster:
   name: foo
   kubernetes:
     manifests:
      - mode: full # default
        name: ArgoCD
        inline: -|
             # apiVersion: ...
             meta:
                 name: pod-1
        file: ../../my-argo-cd.yaml # get naming in sync with config patches
```

bernardgut:
All right so I've been running a production Talos cluster managed entirely through Omni for a while now, and I wanted to share the workaround I've built for deploying and updating Helm charts since native manifest sync is exactly what I've been working around.

The core idea is a pre-render pipeline: Helm charts are rendered locally into raw Kubernetes manifests and injected into Talos machine config patches as cluster.inlineManifests, which Omni then syncs via omnictl cluster template sync.

The pipeline per chart looks like this:

```bash
helm template -n <namespace> --include-crds \
  -f apps/<chart>.helm.yaml \
  <release> <repo>/<chart> \
  | python3 helm-template-fix.py <namespace> \
  | yq -i 'with(.cluster.inlineManifests.[] | select(.name=="<name>"); \
      .contents=load_str("/dev/stdin"))' patches/<name>.yaml
```

Where:

- `apps/<chart>.helm.yaml` : Helm values files, one per chart (versioned in git)
- `helm-template-fix.py` : a small Python script that fixes the well-known [Helm namespace issue](https://github.com/helm/helm/issues/3553) by injecting the namespace into resources that helm template leaves without one (I might get rid of that one soon depending on when it is in GA)
[...]

PR #2574, opened 2026-03-23
feat: support Kubernetes manifests in the cluster templates

Unix4ever:
Fixes: https://github.com/siderolabs/omni/issues/2172

Leverage kubernetes manifest resources and expose them through cluster templates.

### DIFF
--- a/client/pkg/template/internal/models/kubernetes_manifest.go
+++ b/client/pkg/template/internal/models/kubernetes_manifest.go
@@ -1,0 +1,212 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+package models
+
+import (
+	"bytes"
+	"errors"
+	"fmt"
+	"os"
+
+	"github.com/cosi-project/runtime/pkg/resource"
+	"github.com/cosi-project/runtime/pkg/resource/kvutils"
+	"github.com/siderolabs/gen/pair"
+	"github.com/siderolabs/gen/xslices"
+	"go.yaml.in/yaml/v4"
+
+	"github.com/siderolabs/omni/client/api/omni/specs"
+	"github.com/siderolabs/omni/client/pkg/omni/resources/omni"
+)
+
+type KubernetesManifestMode specs.KubernetesManifestGroupSpec_Mode
+
+// UnmarshalYAML implements yaml.Unmarshaler.
+func (c *KubernetesManifestMode) UnmarshalYAML(unmarshal func(any) error) error {
+	var value string
+
+	if err := unmarshal(&value); err != nil {
+		return err
+	}
+
+	switch value {
+	case "one-time":
+		*c = KubernetesManifestMode(specs.KubernetesManifestGroupSpec_ONE_TIME)
+	case "full":
+		*c = KubernetesManifestMode(specs.KubernetesManifestGroupSpec_FULL)
+	default:
+		return fmt.Errorf("unknown mode name %q", value)
+	}
+
+	return nil
+}
+
+// MarshalYAML implements yaml.Marshaler.
+func (c KubernetesManifestMode) MarshalYAML() (any, error) {
+	switch c {
+	case KubernetesManifestMode(specs.KubernetesManifestGroupSpec_FULL):
+		return "full", nil
+	case KubernetesManifestMode(specs.KubernetesManifestGroupSpec_ONE_TIME):
+		return "one-time", nil
+	default:
+		return "", fmt.Errorf("unknown mode %q", c)
+	}
+}
+
+// KubernetesManifest represents the kubernetes manifest to be applied in the cluster.
+// The manifests are applied on the server side.
+type KubernetesManifest struct {
+	Descriptors Descriptors            `yaml:",inline"`
+	Name        string                 `yaml:"name"`
+	File        string                 `yaml:"file,omitempty"`
+	Inline      []map[string]any       `yaml:"inline,omitempty"`
+	Mode        KubernetesManifestMode `yaml:"mode"`
+}
+
+// Validate a kubernetes manifest.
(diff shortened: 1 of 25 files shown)


client/pkg/template/operations/export.go
```diff
@@ -504,6 +527,73 @@ func collectClusterResources(ctx context.Context, st state.State, clusterID stri
...
+
+	dec := yaml.NewDecoder(bytes.NewReader(buffer.Data()))
+
+	for {
+		var doc map[string]any
+
+		if err := dec.Decode(&doc); err != nil {
+			break
+		}
+
```
Orzelius, line 580: Is the break here intentional? won't it swallow errors?

client/pkg/template/internal/models/kubernetes_manifest.go
```diff
@@ -0,0 +1,214 @@
...
+	id := fmt.Sprintf("%03d-%s-%s", weight, prefix, name)
+
+	var (
+		raw []byte
+		err error
+	)
+
+	switch {
+	case km.File != "":
+		raw, err = os.ReadFile(km.File)
```
Orzelius, line 128: Potential security risk? I think this will go and read any file, right?

client/pkg/template/internal/models/kubernetes_manifest.go
```diff
@@ -0,0 +1,214 @@
...
+	Descriptors Descriptors            `yaml:",inline"`
+	Name        string                 `yaml:"name"`
+	File        string                 `yaml:"file,omitempty"`
+	Namespace   string                 `yaml:"namespace"`
+	Inline      []map[string]any       `yaml:"inline,omitempty"`
+	Mode        KubernetesManifestMode `yaml:"mode"`
+}
+
+// Validate a kubernetes manifest.
+func (km *KubernetesManifest) Validate() error {
```
Orzelius, line 68: perhaps also validate the namespace to spare the user from a headache in case it's invalid

client/pkg/template/internal/models/kubernetes_manifest.go
```diff
@@ -0,0 +1,214 @@
...
+			return fmt.Errorf("duplicate manifest name %q", name)
+		}
+
+		names[name] = struct{}{}
+
+		return nil
+	})...)
+}
+
+// Translate the list of patches into a list of resources.
```
Orzelius: I don't understand. What patches? Did you mean manifests?

Unix4ever: yeah, that's a copy/paste issue, thanks

client/pkg/template/operations/export.go
Unix4ever: not intentional. Good catch!

client/pkg/template/internal/models/kubernetes_manifest.go
Unix4ever: Manifests are validated on the server side.

client/pkg/template/internal/models/kubernetes_manifest.go
Unix4ever: We do the same for patches. Do you suggest to allow only local directories? 
User are already got used to the way patches work, I guess it might be confusing if we use a different approach here. Also might break someone's setup if lock it out for patches.

Orzelius: Currently it seems like a bad actor could potentially specify a file path and expose some secrets if the contents are shown (for example in an error message). I hope that everybody runs Omni dockerized so I am not sure how big of an issue this is

smira: This is client-side code (it runs on a customer's machine), so if it reads a sensitive file, it's a file they put into a cluster template (not sure what the risk is here).

They can run it containerized or whatever else, but the risk is same as having `cat` available.

client/pkg/template/internal/models/kubernetes_manifest.go
```diff
@@ -0,0 +1,214 @@
...
+		return "", fmt.Errorf("unknown mode %q", c)
+	}
+}
+
+// KubernetesManifest represents the kubernetes manifest to be applied in the cluster.
+// The manifests are applied on the server side.
+type KubernetesManifest struct {
+	Descriptors Descriptors            `yaml:",inline"`
+	Name        string                 `yaml:"name"`
+	File        string                 `yaml:"file,omitempty"`
```
utkuozdemir, line 62: Just occurred to me: kustomize has something called `LoadRestrictions`. By default, it's on and it doesn't let you point to `..` and so on - basically it only allows cwd and subdirs.
It can be overridden, but is there for security reasons (don't know exact reasons).
We might want to consider doing something similar. Just as a note.

client/pkg/template/internal/models/kubernetes_manifest.go
utkuozdemir: btw - wrote https://github.com/siderolabs/omni/pull/2574/changes#r2982441664 before seeing this discussion.

client/pkg/template/internal/models/kubernetes_manifest.go
smira: I wonder if this is important if you download `kustomize` stuff that _someone_ _else_ authored, so you're at risk running this on your machine.

So applied to Omni, what is someone publishes a malicious cluster template which e.g. reads in `/etc/passwd`. I think this is valid, but we need to introduce the check gracefully.

Unix4ever: Do you think we can postpone that until we have `kustomize` support, or we should use sandbox right now?

frontend/typed-router.d.ts
Slessi, line 1: @Unix4ever i think this again is from another branch?

Unix4ever: ugh, yeah

Unix4ever: I'm getting this feeling when I deal with that router thingy 


image

PR #2575, opened 2026-03-23
chore: rewrite cluster workload proxy controller to use manifests

Unix4ever:
Also change it to use `QTransform`.


frontend/e2e/qemu/clusters.spec.ts
```diff
@@ -258,8 +258,8 @@ test('exposed services', async ({ page }, testInfo) => {
       await servicePage?.close()
     }
   }, 'Wait for service to be running').toPass({
-    intervals: [milliseconds({ seconds: 5 })],
-    timeout: milliseconds({ minutes: 1 }),
+    intervals: [milliseconds({ seconds: 15 })],
+    timeout: milliseconds({ minutes: 2 }),
```
Slessi: @Unix4ever default playwright timeout for the entire test is 30 seconds, so we need to also bump the timeout for the entire test to cater for this. It was actually missing already, just worked fast enough to never need the full minute.

Add this as first line of the test (line 194?):
(2minutes + an extra minute for the rest of the test)
```ts
  test.setTimeout(milliseconds({ minutes: 3 }))
```

Unix4ever: ah, I see

PR #2582, opened 2026-03-24
fix(frontend): prevent invalid auth state

Slessi:
- Fix some frontend login race conditions related to vue state not flushing before redirects happen, resulting in stale identity data being loaded into local storage
- Extract local storage hooks out of the useIdentity hook so as to have a single shared ref for all of them, preventing accidental overwrites from race conditions when waiting for storage events to update other listeners
- Update public key confirmation logic to cater for a situation where an auth0 login was required, but keys were still saved before being confirmed.


Slessi:
@claude review

PR #2583, opened 2026-03-25
fix(frontend): prevent invalid auth states in frontend

Slessi:
- Update public key confirmation logic to cater for a situation where an auth0 login was required, but keys were saved before being confirmed.
- Update interceptor to invalidate keys if a signed request is rejected with a 401 whilst having keys stored in the frontend.
- Update some k8s proxy response codes from 401 to more correct ones 500/403 for the relevant cases

### DIFF
--- a/frontend/src/methods/interceptor.ts
+++ b/frontend/src/methods/interceptor.ts
@@ -23,17 +23,14 @@ import { signDetached, useKeys } from '@/methods/key'
  * This will add the necessary authorization headers for Omni gRPC calls.
  */
 export function useRegisterAPIInterceptor() {
-  const { keyPair, publicKeyID } = useKeys()
+  const { keyPair, publicKeyID, invalidate: invalidateKeys } = useKeys()
   const { identity } = useIdentity()
 
   const unregister = fetchIntercept.register({
     async request(url, config?: { headers?: Headers; method?: string }) {
       url = encodeURI(url)
 
-      if (
-        !/^\/(api|image)/.test(url) ||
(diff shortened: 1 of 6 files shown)


internal/backend/k8sproxy/middleware.go
```diff
@@ -80,30 +80,22 @@ func AuthorizeRequest(next http.Handler, keyFunc KeyProvider, clusterUUIDResolve
 			Set("impersonate.user", claims.Subject).
 			Set("impersonate.groups", claims.Groups)
 
-		if clusterName == "" {
-			ctxzap.Error(ctx, "cluster name is empty")
-
-			w.WriteHeader(http.StatusUnauthorized)
-
-			return
-		}
```
Slessi, line 89: Tried to change this to a `400` bad request, but then when updating test realised this is actually dead code. It will fail earlier with a `401` for having invalid claims due to:

https://github.com/siderolabs/omni/blob/2a863fcf5f446794e4b80b8152a150a3dea756a2/internal/backend/k8sproxy/jwt.go#L39-L41

frontend/src/methods/interceptor.ts
```diff
@@ -62,6 +59,19 @@ export function useRegisterAPIInterceptor() {
...
+      if (
+        response.status === 401 &&
+        keyPair.value &&
+        isSignedRequest(new URL(response.url).pathname)
+      ) {
+        clearKeys()
+      }
+
+      return response
+    },
```
Slessi, line 73: this will tell the frontend to just clear whatever keys it has, if it tries a signed request with keys it thinks are valid, but the backend throws a `401`

PR #2587, opened 2026-03-25
fix(frontend): hide machine tutorial card if we have machines

Slessi:
Currently we erroneously show the machine tutorial card if we have machines in an instance but they get filtered out for some reason. This is not necessary as the user has already been able to have machines join his instance.


frontend/src/views/omni/Machines/components/AddingMachinesTutorial.vue
```diff
@@ -26,7 +44,7 @@ const pxeBootCode =
 
 <template>
   <section
-    v-if="!isDismissed"
+    v-if="!isDismissed && !loading && !data?.spec.connected_machines_count"
```
Unix4ever:
```suggestion
    v-if="!isDismissed && !loading && !data?.spec.registered_machines_count"
```

Issue #905, opened 2025-02-05
[feature] implement `omnictl edit`


utkuozdemir:
Similar to `kubectl edit`.

Unix4ever:
I wanted that for a long time. Will do it as it's a quick win.

PR #2601, opened 2026-03-27
feat: support `omnictl edit` command

Unix4ever:
Works same way as `talosctl edit`, `kubectl edit`.

Fixes: https://github.com/siderolabs/omni/issues/905


client/pkg/omnictl/edit.go
```diff
@@ -0,0 +1,412 @@
...
+			var changed int
+
+			for _, res := range editedResources {
+				editedData, marshalErr := marshalResourceCanonical(res)
+				if marshalErr != nil {
+					return marshalErr
+				}
+
+				originalResource, ok := originalResources[resourceKey(res)]
+				if !ok {
```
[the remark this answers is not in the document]
Unix4ever: expected

smira:
@claude review

client/pkg/omnictl/get.go
```diff
@@ -39,95 +40,52 @@ To get a list of all available resource definitions, issue 'omnictl get rd'`,
...
+		var id string
 
 		if len(args) > 1 {
-			resourceID = args[1]
+			id = args[1]
 		}
 
-		rd, err := resources.ResolveType(ctx, st, resourceType)
+		req, err := createResourceRequest(ctx, client, getCmdFlags.namespace, args[0], id, getCmdFlags.selector, getCmdFlags.idRegexp)
 		if err != nil {
```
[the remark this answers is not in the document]
Unix4ever: that's fine

client/pkg/omnictl/get.go
```diff
@@ -239,6 +197,75 @@ func getResources(cmd *cobra.Command, args []string) func(ctx context.Context, c
...
+		}
+
+		idQuery = append(idQuery, resource.IDRegexpMatch(compiledIDRegexp))
+	}
+
+	return resourceRequest{
+		md:                resource.NewMetadata(getCmdFlags.namespace, rd.TypedSpec().Type, resourceID, resource.VersionUndefined),
+		idQueryOptions:    idQuery,
+		labelQueryOptions: labelQuery,
+		rd:                rd,
```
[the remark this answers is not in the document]
Unix4ever: fixed

client/pkg/omnictl/edit.go
```diff
@@ -0,0 +1,412 @@
...
+						message += "\n"
+					}
+
+					message += fmt.Sprintf("A copy of your changes has been stored to %q\nEdit canceled, no valid changes were saved.\n", path)
+
+					return errors.New(message)
+				}
+			}
+
+			if len(bytes.TrimSpace(stripComments(edited))) == 0 {
```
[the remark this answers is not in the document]
Unix4ever: fixed

client/pkg/omnictl/edit.go
```diff
@@ -0,0 +1,412 @@
...
+					continue
+				}
+
+				origData, ok := originalYAML[resourceKey(res)]
+				if ok && bytes.Equal(editedData, origData) {
+					continue
+				}
+
+				if err := updateResource(ctx, c.Omni().State(), originalResource, res, options{}); err != nil {
+					lastError = err.Error()
```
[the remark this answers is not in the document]
Unix4ever: fixed

client/api/omni/specs/omni_yaml.go
```diff
@@ -19,3 +21,26 @@ func (c *ClusterMachineSpec) MarshalYAML() (any, error) {
...
+		valueNode := node.Content[i+1]
+
+		switch keyNode.Value {
+		case "kubernetes_version":
+			if err := valueNode.Decode(&c.KubernetesVersion); err != nil {
+				return err
+			}
+		default:
+			return fmt.Errorf("unexpected field: %s", keyNode.Value)
+		}
```
Orzelius, line 42: Won't this break if any fields are added to the `ClusterMachineSpec` or is that a necessary evil?

Unix4ever: I'm not sure what's the history behind `kubernetes_version` initially. But I made it more proper now.

Unix4ever:
> Very cool and certainly useful for dev purposes, but I'm fairly sure that we absolutely don't want users touching it, right? One could very easily make an impressive mess with this that would require a hard reset or advanced state surgery to recover from.
> 
> Should the command be hidden and perhaps should it not even be included in production builds of omnictl?

Sometimes users need a way to edit things. Like for example machine classes.
I don't think edit adds more chaos, as you can do the same with `get` + `apply` commands.

Orzelius:
Utku explained this to me yesterday. Previously I was thinking this allows to edit any resource, but now I understand it's limited to the user owned ones. Sorry for the ignorant comment earlier!

Unix4ever:
> Utku explained this to me yesterday. Previously I was thinking this allows to edit any resource, but now I understand it's limited to the user owned ones. Sorry for the ignorant comment earlier!

all good!

client/pkg/omnictl/edit.go
```diff
@@ -0,0 +1,416 @@
...
+		fmt.Fprintf(&sb, "# %s\n", line) //nolint:errcheck
+	}
+
+	return sb.String()
+}
+
+// editCmd represents the edit command.
+var editCmd = &cobra.Command{
+	Use:   "edit <type> [<id>]",
+	Short: "Edit Omni resource with the default editor.",
```
Orzelius:
```suggestion
	Short: "Edit Omni resources with the default editor.",
```

client/pkg/omnictl/get.go
```diff
@@ -239,6 +197,75 @@ func getResources(cmd *cobra.Command, args []string) func(ctx context.Context, c
...
+	st := client.Omni().State()
+
+	rd, err := resources.ResolveType(ctx, st, resourceType)
+	if err != nil {
+		return resourceRequest{}, err
+	}
+
+	if namespace == "" {
+		namespace = rd.TypedSpec().DefaultNamespace
+	}
```
Orzelius, line 225: This is different from the [previous default behavior](https://github.com/siderolabs/omni/blob/43be52c7b1053630184a00841027391c2eb4d712/client/pkg/omnictl/get.go#L65-L67) where the resource types default namespace was used by default unless user specified the `--namespace` flag explicitly. I don't know how big of an issue this is but it might break the get command for some users.

client/pkg/omnictl/edit.go
```diff
@@ -0,0 +1,416 @@
...
+					continue
+				}
+
+				changed++
+
+				if err := updateResource(ctx, c.Omni().State(), originalResource, res, editCmdFlags.options); err != nil {
+					lastError = err.Error()
+
+					continue
+				}
```
Orzelius, line 208: I don't think the `lastError` set here is ever shown to the user?

client/pkg/omnictl/get.go
Unix4ever: it's noop as it falls back to default in both cases

client/pkg/omnictl/edit.go
Unix4ever: it's not printed, but we write it to the comments of the edited file. That the only way to present it back

PR #2619, opened 2026-04-02
chore: bump sqlite metrics collector timeout and interval

Unix4ever:
Timeout: 10s -> 60s
Interval: 60s -> 120s

### DIFF
--- a/internal/backend/runtime/omni/sqlite/metrics.go
+++ b/internal/backend/runtime/omni/sqlite/metrics.go
@@ -23,8 +23,8 @@ import (
 )
 
 const (
-	defaultRefreshInterval = 60 * time.Second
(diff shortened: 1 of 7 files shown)


internal/backend/runtime/omni/sqlite/metrics.go
```diff
@@ -23,8 +23,8 @@ import (
 )
 
 const (
-	defaultRefreshInterval = 60 * time.Second
-	queryTimeout           = 10 * time.Second
+	defaultRefreshInterval = 120 * time.Second
+	queryTimeout           = 60 * time.Second
```
utkuozdemir: wdyt about moving them to config? would never need to change them again.

PR #2641, opened 2026-04-08
feat: move config defaults into the JSON schema

utkuozdemir:
Since omissis/go-jsonschema#521 is merged and released in v0.23.0, we can now use the `goJSONSchema.pointer` field to force pointer generation on leaf fields regardless of whether they have a default value. This removes the blocker that previously prevented us from declaring defaults in the schema.

All static default values are now defined in schema.json and read at runtime by walking the schema and extracting them. The hardcoded Default() function is replaced with a schema-driven one. Only three dynamic defaults that depend on runtime state (local IP, machine API endpoint, resource logger types) remain in Go code.

Additionally, the SQLite metrics config was incorrectly generated as a nillable (pointer) intermediate struct. It is now non-nillable, matching all other intermediate structs. Missing CLI flag bindings for the SQLite metrics fields are also added.

A new unit test validates the structural invariant of the config types: all leaf fields must be nillable (to distinguish unset from zero value during config layer merging), and all intermediate struct fields must be non-nillable (for safe navigation without nil checks).


utkuozdemir:
@claude review

utkuozdemir:
> Do we have tests for the defaults?

I wrote one to make sure that we didn't regress on any of the defaults. It passed.
But then, I thought, having such test would beat the purpose of "changing only a single place to add a default" idea. We would need to add the default to the schema, and then go and add the same value into the unit test as well. 

That's why I removed that test afterwards. I can bring it back. We can keep the existing values tested at least, even if we don't add the new ones' defaults. Wdyt?

Unix4ever:
> > Do we have tests for the defaults?
> 
> I wrote one to make sure that we didn't regress on any of the defaults. It passed. But then, I thought, having such test would beat the purpose of "changing only a single place to add a default" idea. We would need to add the default to the schema, and then go and add the same value into the unit test as well.
> 
> That's why I removed that test afterwards. I can bring it back. We can keep the existing values tested at least, even if we don't add the new ones' defaults. Wdyt?

I thought it could be enough to test different cases we have for defaults: with refs, without refs. No need to cover every default value.

utkuozdemir:
> > > Do we have tests for the defaults?
> > 
> > 
> > I wrote one to make sure that we didn't regress on any of the defaults. It passed. But then, I thought, having such test would beat the purpose of "changing only a single place to add a default" idea. We would need to add the default to the schema, and then go and add the same value into the unit test as well.
> > That's why I removed that test afterwards. I can bring it back. We can keep the existing values tested at least, even if we don't add the new ones' defaults. Wdyt?
> 
> I thought it could be enough to test different cases we have for defaults: with refs, without refs. No need to cover every default value.

I actually brought back that test - it covers a lot of cases (all defaults, with and without refs), and I don't think we will ever gonna change any defaults, needing to change two places in the code is not really an issue (even if were, no big deal).

PR #2654, opened 2026-04-09
feat: comment serviceaccount create output

rothgar:
The plain text output makes it less friendly to automation and saving to .env files because it's interpreted by the shell.

### DIFF
--- a/client/pkg/omnictl/serviceaccount.go
+++ b/client/pkg/omnictl/serviceaccount.go
@@ -69,13 +69,13 @@ var (
 					return err
 				}
 
-				fmt.Printf("Created service account %q with public key ID %q\n", name, publicKeyID)
-				fmt.Printf("\n")
-				fmt.Printf("Set the following environment variables to use the service account:\n")
+				fmt.Fprintf(os.Stderr, "Created service account %q with public key ID %q\n", name, publicKeyID)
+				fmt.Fprintf(os.Stderr, "\n")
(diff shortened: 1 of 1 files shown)


oguzkilcan:
I think it'd be better to add an output flag that takes json or yaml to print machine readable output rather than changing the free text to make it easier for the machines to read.

rothgar:
The output is usually sourced in a shell. I'm fine if we want to add flags for different output handling, but I think the default output should be comment escaped for the common use case.

Ideally the output wouldn't have the extra text or the extra text would be written to stderr, but I didn't want to change the behavior, I just wanted to make it easier for people to copy/paste the output or source it without editing a file.

utkuozdemir:
I think this is a good change.
I was considering adding `export ` prefix to the env vars on that output, but that wouldn't work in `.env` files ~I guess~. So it's ~probably~ an either/or.

Unix4ever:
Actually the correct fix might be to move the human readable part to stderr. Then stdout will be clean for scripts to parse

rothgar:
> Actually the correct fix might be to move the human readable part to stderr. Then stdout will be clean for scripts to parse

I originally was going to do that but it felt like to big of a change. I agree it's the "correct" way to handle it even though I hate sending useful information to stderr.

I can change it to stderr output for the text

Unix4ever:
Yep, lets change it to use stderr. That's the standard we tried to follow all this time.

rothgar:
The current commit is stderr writes instead of # comments

oguzkilcan (approved):

PR #2657, opened 2026-04-10
feat: implement eula guard for omni

### LINKED ISSUE #2576: Require acceptance of the EULA on first login
### Problem Description

For self-hosted Omni (only - so not for SaaS) we need the first user to log in to Omni to be required to accept the EULA.

For first user only:

END USER LICENSE AGREEMENT
Before installing Sidero Omni, please review the End User License Agreement ("Agreement") at:
https://siderolabs.com/eula/
By checking the box below and clicking Accept, you agree to the terms of the Agreement on behalf of yourself and the organization you represent, confirm that you have authority to bind that organization, and agree to ensure that all users of this deployment comply with the Agreement's terms. If you do not agree, click Decline — installation will not proceed.
☐  I have read and agree to the End User License Agreement.
[Decline]    [Accept] (greyed until box checked)

If they do not accept the EULA via the checkbox, and click Decline, Omni must terminate, and re-present the same option on next log in.

If they do accept, log locally the time, hostname and machine _id, username, and acceptance method ("gui-checkbox-accepted").

### Solution

_No response_

### Alternative Solutions

_No response_

### Notes

_No response_

Slessi:
Implement a guard for Omni to prevent usage until users accept an EULA through the UI or a startup flag.

Trying to use the UI before accepting EULA, will redirect you to `/eula`:
image

All UI / CLI actions will be blocked until this is accepted, at the API level in Omni. Using CLI before accepting will notify you of this:
```
rpc error: code = FailedPrecondition desc = EULA has not been accepted; please accept the End User License Agreement before using Omni: https://dev.omni.local/eula
```

In addition to the UI flow, EULA can be accepted programmatically with Omni flags:
```sh
--eula-accept-name=John Doe
--eula-accept-email=john@doe.com
```

Or with Omni config:
```yml
eulaAccept:
  name: John Doe
  email: john@doe.com
```

Closes #2576


internal/backend/runtime/omni/state_validation.go
```diff
@@ -1697,6 +1698,41 @@ func kubernetesManifestsValidationOptions() []validated.StateOption {
...
+			}
+
+			return nil
+		})),
+		validated.WithUpdateValidations(validated.NewUpdateValidationForType(func(_ context.Context, _ *authres.EulaAcceptance, _ *authres.EulaAcceptance, _ ...state.UpdateOption) error {
+			return fmt.Errorf("EULA acceptance cannot be modified")
+		})),
+		validated.WithDestroyValidations(validated.NewDestroyValidationForType(func(_ context.Context, _ resource.Pointer, _ *authres.EulaAcceptance, _ ...state.DestroyOption) error {
+			return fmt.Errorf("EULA acceptance cannot be revoked")
+		})),
```
Unix4ever: As we forbid update/delete in state_access we don't need this validation.

frontend/src/pages/eula.vue
```diff
@@ -0,0 +1,143 @@
...
+const name = ref('')
+const email = ref('')
+const accepted = ref(false)
+const accepting = ref(false)
+
+const invalidForm = computed(() => !accepted.value || !name.value.trim() || !email.value.trim())
+
+function decline() {
+  window.alert('TODO: Implement')
+}
```
Slessi: just waiting for feedback on what this should do (and for text adjustments)

https://talossystems.slack.com/archives/C05348STLHY/p1776073872620789?thread_ts=1774342084.865779&cid=C05348STLHY

Slessi:
@claude review

internal/pkg/config/schema.json
```diff
@@ -1767,6 +1772,31 @@
           }
         }
       }
+    },
+    "EulaAccept": {
+      "type": "object",
+      "dependentRequired": {
+        "name": ["email"],
+        "email": ["name"]
+      },
```
utkuozdemir, line 1781: Does this mean I can leave both empty and it'll simply not require acceptance?
Is there a way to disable the EULA requirement completely?

utkuozdemir: Also wondering: why is it partly config (flags, this), partly data (resource in DB)? Shouldn't it be either one or the other?

Slessi: How I wrote it, you can indeed supply both or none. Just email or just name will be rejected.

There is no way to disable the EULA requirement _entirely_. One way or another it has to be accepted whether that is programatically or through the UI is up to the user.

For SaaS, we can self-accept by inputting our own information as omni flags, as SaaS users have already effectively accepted EULA through the SaaS registration process (as I understand it) but everyone else _must_ accept EULA themselves.

@steverfrancis can correct me if I am mistaken about that, but I got this from the ticket https://github.com/siderolabs/omni/issues/2576

Slessi: Ah, for 2nd question about why it is a resource, its because you can also accept through UI and because once accepted this is immutable information and shouldn't be modified. So dropping or modifying flags later does not modify your EULA acceptance information.

utkuozdemir: So when I pass both these flags, would Omni start auto-accepted?

Rolling this out as-is can be quite disruptive. I'm thinking, maybe we could instead add an `--enforce-eula` flag or so.

If it stays as-is, ops team should be informed about this when rolling it out, and to release.toml, it should definitely go as:
```
"Urgent Upgrade Notes **(No, really, you MUST read this before you upgrade)**"
```

It's because, if I upgrade my on-prem without changing anything, it will effectively stop working.

Unix4ever: > For SaaS, we can self-accept by inputting our own information as omni flags, as SaaS users have already effectively accepted EULA through the SaaS registration process (as I understand it) but everyone else _must_ accept EULA themselves.

We have emails there, not sure about the full name.  Maybe we can also get that from Stripe.
It should definitely support self-accept before we roll out the new version to prod.

Slessi: > So when I pass both these flags, would Omni start auto-accepted?

Yes.

For SaaS, ops team need to add the flags to accept EULA on omni start for users, because they have already accepted through other agreements. There is no disruption to users there, as it will always be accepted for them.

> maybe we could instead add an `--enforce-eula` flag or so.

This does not make sense. Self hosted users will then never accept EULA beacuse they never enforce EULA, which is the exact target user base we want to force to accept EULA.

> if I upgrade my on-prem without changing anything, it will effectively stop working

This is true, but we already do this for other things like adding required flags for sqlite, or forcing users to remove of deprecated flags that have been removed from Omni.

> It should definitely support self-accept before we roll out the new version to prod

It does, through omni flag/config, or is there some extra feature you expect here?

utkuozdemir: Then if auto-accept is there, and we add the urgent update notes to the changelog, we are good I guess.

cmd/omni/pkg/app/app.go
```diff
@@ -37,12 +37,13 @@ import (
...
 	"github.com/siderolabs/omni/internal/pkg/config"
 	"github.com/siderolabs/omni/internal/pkg/ctxstore"
+	"github.com/siderolabs/omni/internal/pkg/eula"
 	"github.com/siderolabs/omni/internal/pkg/features"
 	"github.com/siderolabs/omni/internal/pkg/siderolink"
 )
 
 // Run the Omni service.
-func Run(ctx context.Context, state *omni.State, cfg *config.Params, logger *zap.Logger) error {
+func Run(ctx context.Context, state *omni.State, cfg *config.Params, eulaParams eula.AcceptParams, logger *zap.Logger) error {
```
Unix4ever: I don't think we need a separate parameter for EULA: email/name is already defined in cfg.

So we can do something like:
```golang
if config.EulaAccept.GetName() != "" && config.EulaAccept.GetEmail() != "" {
	if err = eula.Accept(ctx, state.Default(), eula.AcceptParams{Name: config.EulaAccept.GetName(), Email: config.EulaAccept.GetEmail()}); err != nil {
		return fmt.Errorf("failed to accept EULA: %w", err)
	}
}
```

frontend/src/pages/eula.vue
```diff
@@ -0,0 +1,143 @@
...
+        },
+        spec: {
+          accepted_by_name: name.value.trim(),
+          accepted_by_email: email.value.trim(),
+        },
+      },
+      withRuntime(Runtime.Omni),
+      (req: OmniRequestOptions) => {
+        req.skipSignature = true
+      },
```
Unix4ever: I would suggest to write a helper func for consistency:
something like: `withSkipRequestSignature()`

internal/backend/runtime/omni/state_validation.go
```diff
@@ -1697,6 +1697,33 @@ func kubernetesManifestsValidationOptions() []validated.StateOption {
 	}
 }
 
+// eulaValidationOptions returns validation options for the EulaAcceptance resource.
+// The EULA can only be accepted once (Create), and never updated or destroyed.
+func eulaValidationOptions(st state.State) []validated.StateOption {
+	return []validated.StateOption{
+		validated.WithCreateValidations(validated.NewCreateValidationForType(func(ctx context.Context, res *authres.EulaAcceptance, _ ...state.CreateOption) error {
```
Unix4ever, line 1704: I would suggest adding the validation that eula ID is always `EulaAcceptanceID`.

client/pkg/omni/resources/auth/eula.go
```diff
@@ -0,0 +1,56 @@
...
+
+// ResourceDefinition implements [typed.Extension] interface.
+func (EulaAcceptanceExtension) ResourceDefinition() meta.ResourceDefinitionSpec {
+	return meta.ResourceDefinitionSpec{
+		Type:             EulaAcceptanceType,
+		Aliases:          []resource.Type{},
+		DefaultNamespace: resources.DefaultNamespace,
+		PrintColumns: []meta.PrintColumn{
+			{Name: "Name", JSONPath: "{.acceptedByName}"},
+			{Name: "Email", JSONPath: "{.acceptedByEmail}"},
```
Unix4ever: Should be all lowercase to work:

```suggestion
			{Name: "Name", JSONPath: "{.acceptedbyname}"},
			{Name: "Email", JSONPath: "{.acceptedbyemail}"},
```

internal/pkg/auth/interceptor/eula.go
utkuozdemir, line 1: GRPC is not the only entry point we have - there's also the HTTP API, which is used by workload proxy and other stuff. I guess we should also intercept access to those, in package: internal/pkg/auth/handler

We could maybe move the eula check to its own package (+ component/struct), containing the `accepted atomic.Bool` and so on. Then we could call it from both places.

cmd/omni/pkg/app/app.go
Slessi: done

frontend/src/pages/eula.vue
Slessi: done

internal/backend/runtime/omni/state_validation.go
Slessi: done

client/pkg/omni/resources/auth/eula.go
Slessi: done

.vscode/settings.json
```diff
@@ -1,5 +1,14 @@
...
+  },
+  "gopls": {
+    "formatting.gofumpt": true,
+    "ui.semanticTokens": true
+  },
+  "go.lintTool": "golangci-lint-v2",
+  "go.lintOnSave": "package",
+  "go.toolsEnvVars": {
+    "GOFLAGS": "-tags=integration"
+  }
```
Slessi, line 13: my vscode was not linting go properly, these are the correct settings that fix it for me (and i would assume, others)

frontend related ones, i usually open frontend as its own vscode window, and there are `frontend/.vscode/settings.json` for that

internal/pkg/auth/interceptor/eula.go
Slessi: I could, but not sure if its worth bothering. You can't authenticate without accepting EULA, which blocks new & current users really. New users for sure won't be able to set up anything you might use an HTTP API for later (e.g. workload proxy). Current users, I dunno, maybe ur workload proxy still works, but at minimum your auth is going to expire after 8 hours anyway.

What do you think? Worth doing regardless?

utkuozdemir: I was thinking more from technical standpoint, but real-world implications. I don't think it's actually a big deal. @Unix4ever wdyt?

Unix4ever: I guess as we block user creation, it should be fine as is

Slessi: I added notes to `release.toml`, is there anything else from you guys? ~im just waiting on steve's final text~

updated text also, so ready

Unix4ever (approved):

PR #2659, opened 2026-04-10
fix: always set last error on manifests sync failure

Unix4ever:
Also handle `namespace not specified` as a fatal error. All transient errors are logged as generic errors to avoid doing to many writes to the resource.


oguzkilcan (approved):

Unix4ever:
@claude review

Issue #2656, opened 2026-04-10
UX inconsistencies when workload proxying feature is disabled instance-wide


utkuozdemir:
When it is disabled instance-wide (`--workload-proxying-enabled=false`):
- [x] The checkox for enabling the feature on the cluster creation screen is not disabled, it is clickable. It should be disabled, should be consistent with the cluster overview screen. cc @Slessi. 
  - https://github.com/siderolabs/omni/pull/2660 
- [ ] No validation on the cluster.features field - the feature can be toggled when the feature is disabled instance-wide - user can, via cluster templates, toggle it (when adding validation for it, we should be careful of not breaking things on `Cluster` resource updates).
- [ ] Enabling the feature on a cluster (because there is no validation) causes `kube-service-exposer` to be deployed to the workload cluster. It'll be no-op, but it shouldn't be possible at all, probably.

Unix4ever:
> No validation on the cluster.features field - the feature can be toggled when the feature is disabled instance-wide - user can, via cluster templates, toggle it (when adding validation for it, we should be careful of not breaking things on Cluster resource updates).

This validation is already there.

PR #2660, opened 2026-04-10
fix(frontend): disable workload proxy checkbox if disabled on instance

Slessi:
If Omni is running with workload proxying disabled, then disable the checkbox to enable workload proxying on the frontend.

Related #2656


Slessi:
@claude review

Unix4ever (approved):

PR #2665, opened 2026-04-10
feat: allow force destroying machine requests

### LINKED ISSUE #2642: [bug] Infra provider teardown stuck when provisioned machine never joined Omni

### Current Behavior

When an infra provider successfully provisions a VM (all provision steps complete) but the machine never fully
   joins Omni via SideroLink, the cluster destroy flow gets permanently stuck. The SDK's reconcileTearingDown   
  never calls the provider's Deprovision method, leaving the cluster in "Destroying" state indefinitely with no 
  way to resolve from the UI.                                                                                   
                                                                                                                
  The Omni UI shows:                                                                                            
  - Cluster: "Destroying"                                                                                       
  - Control plane machine: "Provisioned / Waiting for the machine to join Omni"
  - Worker: "Destroying" (blocked waiting for control plane)                   
                                                                                                                
  Provider logs show only reconcileRunning (provision path succeeds with no-op), never reconcileTearingDown. The
   provider's Deprovision is never called. 

Image

### Expected Behavior

The provider's Deprovision method should be called even when the machine never joined Omni. The provider can  
  then clean up infrastructure-side resources (VM, storage) and the cluster destroy should complete.
                                                                                                                
  If the machine never joined, there's nothing to gracefully drain — the teardown should proceed directly to    
  calling Deprovision. 

### Steps To Reproduce

  1. Deploy an infra provider using infra.NewProvider() + provision.Step pattern                                
  2. Create a cluster — provider completes all provision steps, VM is created and running                     
  3. VM fails to reconnect via SideroLink (e.g., VM is restarted during a finalization step, network issue, or  
  VM is deleted externally from the infrastructure)                                                             
  4. Omni shows machine as "Provisioned / Waiting for the machine to join Omni"                                 
  5. Delete the cluster in Omni UI                                                                              
  6. Cluster enters "Destroying" state and stays there permanently                                            
  7. Provider logs show reconcile succeeded (provision path) — reconcileTearingDown is never entered  

Chrome

### Anything else?

 Looking at the SDK code in pkg/infra/controllers/provision.go, reconcileTearingDown (line 473) has several    
  prerequisite steps before calling Deprovision (line 516):     
                                                                                                                
  1. removePatches (line 482) — blocks if config patches can't be cleaned up                                  
  2. TeardownAndDestroyAll on MachineRequestStatus (line 495) — blocks until ready
  3. Machine state existence check (line 506) — skips Deprovision entirely if not found                         
                                                                                                                
  One of these gates appears to block indefinitely when the machine was provisioned but never joined Omni.      
                                                                                                                
  Environment: Omni client SDK v1.6.4, TrueNAS provider using standard SDK patterns. (I am building it btw)                    
                                                                                                              
  Workaround: Force-delete via omnictl delete cluster <name> --destroy or omnictl delete machinerequest <id>    
  --namespace infra-provider.                                                                                 
                                                                                                                
  Related: #2629 (provision errors not visible in UI — same controller, different symptom)

oguzkilcan:
Machine requests are now created without a controller owner, allowing operators and admins to teardown stuck or unwanted requests directly. The controller replaces destroyed requests automatically to maintain the desired machine count. Includes a migration to clear ownership on existing requests.

Fixes: #2642

### DIFF
--- a/frontend/src/views/ClusterMachines/MachineRequest.vue
+++ b/frontend/src/views/ClusterMachines/MachineRequest.vue
@@ -5,21 +5,31 @@ Use of this software is governed by the Business Source License
 included in the LICENSE file.
 -->
 <script setup lang="ts">
-import { computed, toRefs } from 'vue'
+import { computed, ref, toRefs } from 'vue'
 
+import { Runtime } from '@/api/common/omni.pb'
 import type { Resource } from '@/api/grpc'
+import { ResourceService } from '@/api/grpc'
 import type { ClusterMachineRequestStatusSpec } from '@/api/omni/specs/omni.pb'
 import { ClusterMachineRequestStatusSpecStage } from '@/api/omni/specs/omni.pb'
+import { withRuntime } from '@/api/options'
+import { InfraProviderNamespace, MachineRequestType } from '@/api/resources'
+import TActionsBox from '@/components/ActionsBox/TActionsBox.vue'
+import TActionsBoxItem from '@/components/ActionsBox/TActionsBoxItem.vue'
 import TIcon from '@/components/Icon/TIcon.vue'
 import IconHeaderDropdownLoading from '@/components/icons/IconHeaderDropdownLoading.vue'
 import TStatus from '@/components/Status/TStatus.vue'
+import Tooltip from '@/components/Tooltip/Tooltip.vue'
 import { TCommonStatuses } from '@/constants'
+import { showError, showSuccess } from '@/notification'
 
 const props = defineProps<{
   requestStatus: Resource<ClusterMachineRequestStatusSpec>
+  canDestroy: boolean
 }>()
 
 const { requestStatus } = toRefs(props)
+const destroying = ref(false)
 
 const stage = computed(() => {
   switch (requestStatus.value.spec.stage) {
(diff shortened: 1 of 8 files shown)


oguzkilcan:
@claude review

frontend/src/views/ClusterMachines/MachineRequest.vue
```diff
@@ -58,5 +98,13 @@ const stage = computed(() => {
...
       {{ requestStatus.spec.status }}
     </div>
+
+    <div class="flex items-center justify-end">
+      <TActionsBox v-if="canDestroy">
+        <TActionsBoxItem icon="delete" danger :disabled="destroying" @select="forceDestroy">
+          Destroy
+        </TActionsBoxItem>
+      </TActionsBox>
+    </div>
```
Slessi, line 124: instead of v-if, its better UX to disable the item, and have a tooltip explaining why its disabled

we are also missing the `canRemoveMachines` permission check as copilot mentioned

Texts are examples only but something like:
- if `!canDestroy`
  - "Cant destroy because wrong stage" or whatever the reason is (variable name not clear why we cant destroy)
- if `!hasDestroyPermission`
  - "Dont have right role for this"
- if `destroying`
  - This is a loading state, and should probably have a spinner near "destroying..." text or so, instead of a tooltip (but still disabled)

Slessi (approved):
frontend is gucci

Unix4ever (approved):

PR #2666, opened 2026-04-10
feat: toggle info buttons if already opened

rothgar:
Changes the behavior of info buttons to close the panel if it's already open


Slessi (commented):
Sounds gud but i thought it was already this way 🤔

Ill check on monday if not will approve and merge

Slessi (approved):

PR #2668, opened 2026-04-13
chore: prepare omni with talos v1.13.0-rc

### LINKED ISSUE #12963: test: Omni and Talos 1.13.0-beta.0
New maintenance API, apply config flow and other gotchas.

Look into new Talos APIs - do they need to be specifically exposed?

New machinery, config schemas, etc.

oguzkilcan:
Prepare omni with talos v1.13.0-beta

Fixes https://github.com/siderolabs/talos/issues/12963


smira:
@oguzkilcan we have 1.13.0-rc.0 now, 🤞 it goes well

frontend/e2e/talemu/cluster_fixtures.ts
```diff
@@ -53,6 +54,20 @@ const test = base.extend<ClusterFixtures>({
 
       await use({ name: clusterName })
 
+      // Save support bundle if the test failed
+      if (testInfo.status !== 'passed') {
```
Slessi, line 59: does this make sense for talemu tests? wont u get garbage back? or u only interested in the omni logs?

frontend/e2e/qemu/clusters.spec.ts
```diff
@@ -4,18 +4,42 @@
...
+          console.error(`failed to save support bundle for cluster ${clusterName}:`, e)
+        }
+      }
+    },
+    { auto: true },
+  ],
+})
+
+test.describe.configure({ mode: 'serial', retries: 0 })
 const cpMachineName = 'deadbeef'
```
Slessi, line 44: this is a total nit but can u move the earlier const down so the diff doesnt look weird? 😅 

if left exactly like this it wont show unchanged lines as changed

```suggestion
/*your new thingy*/

test.describe.configure({ mode: 'serial', retries: 0 })

const clusterName = 'talos-test-cluster'
const cpMachineName = 'deadbeef'
```

frontend/e2e/talemu/cluster_fixtures.ts
```diff
@@ -6,6 +6,7 @@ import { faker } from '@faker-js/faker'
 import { milliseconds } from 'date-fns'
 import fs from 'fs/promises'
 import { dump } from 'js-yaml'
+import path from 'path'
```
Slessi: probably want `node:path` to make sure u get the node core library, not some npm package that may also be called path
```suggestion
import path from 'node:path'
```

Slessi (commented):
support bundle sounds like a nice idea 👍

Unix4ever (approved):

utkuozdemir (approved):

frontend/e2e/qemu/clusters.spec.ts
Unix4ever: I think it also has Omni resources which won't be garbage and are actually useful :)

smira (approved):

PR #2693, opened 2026-04-16
feat: add Talos version end-of-support notifications and metrics

### LINKED ISSUE #2364: Create a warning when there are Omni machines older than the supported min version
It should warn users to upgrade them.

Ideas:
- For Talos versions approaching their EOL (from Omni support perspective), we can say that "upgrade soon, otherwise it'll go out of support"
- For the versions which are already unsupported, the warning message can be different (e.g., is out of support, and might be locked out of upgrades anytime)
- We can instead start with a milder warning only when **adding** machines below the supported version.

oguzkilcan:
* Track machines running Talos versions approaching or past end of support relative to MinTalosVersion.
* Replace the config-driven non-ImageFactory deprecation notification with hardcoded constants and add two new notifications (approaching end of support, end of support reached) with corresponding Prometheus metrics.
* Add startup validation hooks (currently disabled) that will refuse to start when unsupported machines are detected.
* Fix frontend notification namespace from Default to Ephemeral.

Resolves: #2364


oguzkilcan:
@claude review

Unix4ever (approved):

PR #2694, opened 2026-04-16
feat(frontend): allow re-saving omni support bundle

### LINKED ISSUE #2680: allow to restart support bundle download after fetching is complete
### Problem Description

1. pull support bundle from large cluster. Can take >10m
2. go do something else
3. download dialogue pops up
4. accidentally close it
5. clicking "download" on the (completed) support bundle pull restarts the complete process

### Solution

The support bundle should be resumable, if a user accidentally canceled the download dialogue.

This can be done in various ways, e.g. to save the support bundle to memory, if <100MB.

### Alternative Solutions

Not sure if this is possible:
Ask for a download path right when initiating the support bundle pull, then stream file there as soon as it's available upstream?

### Notes

_No response_

Slessi:
After download completes on the Omni support bundle, the user may click save again to save the bundle again without having to initiate the download again. This helps incase you accidentally click out of the first save, or deleted it, or anything like that. If you want a fresh bundle, you can still get that when you close & re-open the modal.

https://github.com/user-attachments/assets/6e158c90-8a0a-4be0-9986-6fcd949b2629

Closes #2680

### DIFF
--- a/frontend/src/views/Overview/components/OverviewRightPanel/DownloadSupportBundleModal.vue
+++ b/frontend/src/views/Overview/components/OverviewRightPanel/DownloadSupportBundleModal.vue
@@ -61,10 +61,10 @@ const { data } = useResourceWatch<ClusterMachineIdentitySpec>(() => ({
 }))
 
 const downloading = ref(false)
-const done = ref(false)
+const downloadedURL = ref<string>()
 
 const closeText = computed(() => {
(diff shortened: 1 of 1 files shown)


smira (approved):

fsgh42:
rel: https://github.com/siderolabs/omni/issues/2680

PR #2696, opened 2026-04-16
release(v1.7.0): prepare release

Unix4ever:
This is the official v1.7.0 release.


utkuozdemir (approved):

smira (approved):
so Talos 1.13 support in the next release?

Unix4ever:
1.13 support checklist we had is done. Tests are not migrated yet and we still use beta machinery.
So I think 1.13 should work without issues on 1.7.

We can backport machinery bump later. WDYT?

frezbo (approved):

smira:
> 1.13 support checklist we had is done. Tests are not migrated yet and we still use beta machinery. So I think 1.13 should work without issues on 1.7.
> 
> We can backport machinery bump later. WDYT?

I'm looking at https://github.com/siderolabs/omni/pull/2668 which seems to be failing still

Unix4ever:
> > 1.13 support checklist we had is done. Tests are not migrated yet and we still use beta machinery. So I think 1.13 should work without issues on 1.7.
> > We can backport machinery bump later. WDYT?
> 
> I'm looking at #2668 which seems to be failing still

ah dang... Ok, we can postpone it. Need to figure out what's up there

PR #2701, opened 2026-04-17
feat: introduce powered off machine state and power on support

### LINKED ISSUE #1634: [feature] Add a "Shutdown" state to Omni
### Problem Description

If you use Omni to shutdown a machine in a cluster, that machine will always be in state "Shutting Down", and because the machine is powered off, it can't have the state updated, so it will always be in the greyed out state of "Machine is unreachable - last known state"

Image

### Solution

Once the machine logs "task shutdown (1/1): starting" we should not expect more messages or SisderoLink status, so change the state to "Shutdown" in the Cluster level Overview:

Image

Should also reflect that state in the Cluster view of All Nodes:

Image
i.e. also show that the node is Shutdown there, not just Not Ready.

Also show it in the Stage field of the Cluster/ node detail view, which currently shows Shutting Down:

Image

### Alternative Solutions

_No response_

### Notes

_No response_

utkuozdemir:
Machines that were shutting down and then disconnect are now shown as "Powered Off" in the UI instead of being stuck in "Shutting Down" with a greyed-out unreachable state.

For machines managed by a static infra provider, shutting down a machine now prevents the provider from automatically powering it back on due to cluster allocation. The provider honors the shutdown request until the machine goes through a deallocation cycle, at which point the request is considered stale.

Intentionally powered-off machines are also excluded from the "disconnected machines" list on the frontend when destroying a cluster, to avoid them being force-destroyed.

The shutdown modal in the frontend now calls a new management API endpoint instead of the Talos API directly. The CLI gains \`omnictl machine shutdown\` and \`omnictl machine power-on\` commands.

Closes siderolabs/omni#1634.
Part of siderolabs/omni-infra-provider-bare-metal#103.

### DIFF
--- a/frontend/src/views/Modals/NodePowerOn.vue
+++ b/frontend/src/views/Modals/NodePowerOn.vue
@@ -1,0 +1,84 @@
+<!--
+Copyright (c) 2026 Sidero Labs, Inc.
+
+Use of this software is governed by the Business Source License
+included in the LICENSE file.
+-->
+<script setup lang="ts">
+import { computed, ref } from 'vue'
+
+import { Runtime } from '@/api/common/omni.pb'
+import { ManagementService } from '@/api/omni/management/management.pb'
+import type { ClusterMachineStatusSpec } from '@/api/omni/specs/omni.pb'
+import {
+  ClusterMachineStatusLabelNodeName,
+  ClusterMachineStatusType,
+  DefaultNamespace,
+} from '@/api/resources'
+import ConfirmModal from '@/components/Modals/ConfirmModal.vue'
+import { useClusterPermissions } from '@/methods/auth'
+import { useResourceWatch } from '@/methods/useResourceWatch'
+import { showError, showSuccess } from '@/notification'
+
+const { clusterId, machineId } = defineProps<{
+  clusterId: string
+  machineId: string
+}>()
+
+const open = defineModel<boolean>('open', { default: false })
+const loading = ref(false)
+
+const { data: clusterMachineStatus } = useResourceWatch<ClusterMachineStatusSpec>(() => ({
+  skip: !open.value,
+  resource: {
+    namespace: DefaultNamespace,
+    type: ClusterMachineStatusType,
+    id: machineId,
+  },
+  runtime: Runtime.Omni,
+}))
+
+const nodeName = computed(
+  () =>
+    clusterMachineStatus.value?.metadata.labels?.[ClusterMachineStatusLabelNodeName] ?? machineId,
+)
+
+const close = () => {
+  open.value = false
+}
+
+const { canRebootMachines } = useClusterPermissions(computed(() => clusterId))
+
+const powerOn = async () => {
+  loading.value = true
+
+  try {
+    await ManagementService.MachinePowerOn({ machine_id: machineId })
+  } catch (e) {
+    close()
+
+    showError('Failed to Issue Power On', e instanceof Error ? e.message : String(e))
+
+    return
+  } finally {
+    loading.value = false
+  }
+
+  close()
+
+  showSuccess('Machine Power On', `Power on request sent for machine ${nodeName.value}.`)
+}
+</script>
+
+<template>
+  <ConfirmModal
+    v-model:open="open"
+    :title="`Power on the Machine ${nodeName} ?`"
+    action-label="Power On"
+    :disabled="!canRebootMachines"
+    :loading="loading"
+    @confirm="powerOn"
+  >
+    <p class="text-xs">Please confirm the action.</p>
+  </ConfirmModal>
+</template>
--- a/internal/backend/grpc/router/router.go
+++ b/internal/backend/grpc/router/router.go
@@ -193,10 +193,6 @@ func (r *Router) Director(ctx context.Context, fullMethodName string) (proxy.Mod
 			return proxy.One2One, nil, err
 		}
 
-		if err = r.talosAuditor.AuditTalosAccess(ctx, fullMethodName, getClusterName(md), getNodeID(md)); err != nil {
-			return proxy.One2One, nil, err
-		}
-
 		return proxy.One2One, []proxy.Backend{backend}, nil
 	}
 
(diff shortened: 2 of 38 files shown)


frontend/src/views/Modals/NodePowerOn.vue
Slessi, line 1: Could you do the modal the "new way" like how `ClusterDestroy.vue` is?

Claude can probably figure out the migration if you give that as an example. They are both confirmation type modals so same structure. There is non-confirmation type modal as well FYI but not for this PR

frontend/src/views/Modals/ClusterDestroy.vue
```diff
@@ -35,7 +37,7 @@ const close = () => {
   open.value = false
 }
 
-const { data: disconnectedMachines, loading } = useResourceWatch<MachineStatusSpec>(() => ({
+const { data: rawDisconnectedMachines, loading } = useResourceWatch<MachineStatusSpec>(() => ({
```
Slessi, line 40: just call it `machineStatuses` imo

frontend/src/views/Modals/NodePowerOn.vue
```diff
@@ -0,0 +1,115 @@
...
+          {{ state }}
+        </TButton>
+      </div>
+    </div>
+  </div>
+</template>
+
+<style scoped>
+@reference "../../index.css";
+
```
Slessi, line 84: plz never `<style scoped>` unless u have to do something that isn't possible another way (e.g keyframe animation) - tailwind classes should be inlined

frontend/src/views/Modals/ClusterDestroy.vue
utkuozdemir: done

frontend/src/views/Modals/NodePowerOn.vue
utkuozdemir: done, it was just c/p from NodeShutdown.vue.

frontend/src/views/Modals/NodePowerOn.vue
utkuozdemir: Done, but I left those old popups as-is, only did for the new one.

Slessi: Not done, `TModal` should have zero edits, that is the old router based way of showing modals. We should be showing modals based on `v-model:open`.

So, there should be no business like `router.go(-1)` for example to close the modal, it should be setting its `open` model to false

internal/backend/grpc/router/router.go
```diff
@@ -193,10 +193,6 @@ func (r *Router) Director(ctx context.Context, fullMethodName string) (proxy.Mod
 			return proxy.One2One, nil, err
 		}
 
-		if err = r.talosAuditor.AuditTalosAccess(ctx, fullMethodName, getClusterName(md), getNodeID(md)); err != nil {
-			return proxy.One2One, nil, err
-		}
-
```
utkuozdemir, line 199: Realized this audit was happening too early, so moved it into `TalosBackend.GetConnection`. At that point the `ctx` has gone through authentication and ACL evaluation, so the audit event sees the authenticated actor and the effective role.

frontend/src/views/Modals/NodePowerOn.vue
utkuozdemir: ah I see, you can have another look now.

Slessi: Now its good ✅ thanks

I think the `disabled` on `ConfirmModal` is a bit superfluous because the button that opens it is `disabled` for the same reason, but its fine.

frontend/src/views/Modals/ClusterDestroy.vue
```diff
@@ -35,15 +37,45 @@ const close = () => {
...
+const { data: clusterMachineStatuses, loading: clusterMachineStatusesLoading } =
+  useResourceWatch<ClusterMachineStatusSpec>(() => ({
+    skip: !open.value,
+    resource: {
+      namespace: DefaultNamespace,
+      type: ClusterMachineStatusType,
+    },
+    selectors: [`${LabelCluster}=${clusterId}`],
+    runtime: Runtime.Omni,
+  }))
```
Unix4ever: I thought we already had that. Disconnected label shouldn't exist on powered off machines if they're coming from the bare-metal provider. Did it break?

Unix4ever: And actually I'd rather fix MachineStatus resource to get this powered off state from the cluster machine status, than do that in the UI.

utkuozdemir: addressed both. I also moved power and status related logic into the new `machine` package, and added unit tests for those. Here: https://github.com/siderolabs/omni/pull/2701/changes#diff-a974c94bb0c56f87a9b7412b4a7533df44f6537bc56be65ca12e68a057f35018

internal/backend/runtime/omni/controllers/omni/machine_status_snapshot.go
utkuozdemir, line 1: @Unix4ever The changes in this file are also important.
tldr: I removed the logic from `ClusterMachineStatusController`, and brought it into two places instead: `MachineStatusSnapshotController` and `MachineStatusController`: both were the places where we handled power status of BM provider machines. Now they are unified - all machines' power status are determined in these two places.

internal/backend/runtime/omni/controllers/omni/machine/status.go
```diff
@@ -0,0 +1,99 @@
...
+
+import (
+	"github.com/siderolabs/omni/client/api/omni/specs"
+	"github.com/siderolabs/omni/client/pkg/omni/resources/infra"
+	"github.com/siderolabs/omni/client/pkg/omni/resources/omni"
+)
+
+// UpdateConnectionStatus updates the connection-related labels and the Connected spec field on machineStatus
+// based on the machine's current connectivity, power state, and infra provider information.
+func UpdateConnectionStatus(machineStatus *omni.MachineStatus, machine *omni.Machine, machineStatusSnapshot *omni.MachineStatusSnapshot, infraMachineStatus *infra.MachineStatus) {
```
utkuozdemir, line 16: The logic got more complex in this function, so it was worth refactoring it, moving it to a package and unit-testing it - I made sure we don't introduce a regression here.

internal/backend/runtime/omni/controllers/omni/machine/status.go
```diff
@@ -0,0 +1,99 @@
...
+
+	// If the machine is known to be powered off OR if we can power it on
+	// (e.g., it is managed by the bare-metal infra provider,
+	// and the provider has the BMC credentials of the machine),
+	// we do not mark it as disconnected anymore, as the machine being not connected is expected.
+	if isPoweredOff || canBePoweredOn {
+		machineStatus.Metadata().Labels().Delete(omni.MachineStatusLabelDisconnected)
+	} else {
+		machineStatus.Metadata().Labels().Set(omni.MachineStatusLabelDisconnected, "")
+	}
```
utkuozdemir, line 46: This is the "new" logic of the `Disconnected` label - just extended it to check also `isPoweredOff`, actually.

Unix4ever (approved):

PR #2710, opened 2026-04-21
feat: allow reader access to join token

Slessi:
Explicitly allow readers to read join tokens

### DIFF
--- a/frontend/src/pages/(authenticated)/clusters/[cluster]/machine/[machine].vue
+++ b/frontend/src/pages/(authenticated)/clusters/[cluster]/machine/[machine].vue
@@ -14,13 +14,16 @@ import { DefaultNamespace, MachineStatusType } from '@/api/resources'
 import TabButton from '@/components/Tabs/TabButton.vue'
 import TabContent from '@/components/Tabs/TabContent.vue'
 import Tabs from '@/components/Tabs/Tabs.vue'
+import { useClusterPermissions } from '@/methods/auth'
 import { useResourceGet } from '@/methods/useResourceGet'
 import NodesHeader from '@/views/Nodes/NodesHeader.vue'
 
 definePage({ name: 'NodeDetails' })
 
 const route = useRoute()
-const machine = computed(() => route.params.machine as string)
+const machine = computed(() => route.params.machine)
+const { canReadMachineConfig, canReadConfigPatches, canReadMachinePendingUpdates } =
+  useClusterPermissions(() => route.params.cluster)
 
 const routes = computed(() => {
   return [
@@ -39,18 +42,22 @@ const routes = computed(() => {
     {
       name: 'Config',
       to: { name: 'NodeConfig', params: { machine: machine.value } },
+      disabled: !canReadMachineConfig.value,
     },
     {
       name: 'Pending Updates',
       to: { name: 'NodePendingUpdates', params: { machine: machine.value } },
+      disabled: !canReadMachinePendingUpdates.value,
     },
     {
       name: 'Config History',
       to: { name: 'NodeConfigDiffs', params: { machine: machine.value } },
+      disabled: !canReadMachineConfig.value,
     },
     {
       name: 'Patches',
       to: { name: 'NodePatches', params: { machine: machine.value } },
+      disabled: !canReadConfigPatches.value,
     },
     {
       name: 'Disks',
@@ -96,7 +103,14 @@ const nodeName = computed(
       tabs-list-class="px-4 md:px-6"
     >
       <template #triggers>
-        <TabButton v-for="{ name, to } in routes" :key="name" :as="RouterLink" :value="to.name" :to>
+        <TabButton
+          v-for="{ name, to, disabled } in routes"
+          :key="name"
+          :as="RouterLink"
+          :value="to.name"
+          :to
+          :disabled
+        >
           {{ name }}
         </TabButton>
       </template>
--- a/internal/backend/runtime/omni/state_access.go
+++ b/internal/backend/runtime/omni/state_access.go
@@ -456,6 +456,8 @@ func filterAccess(ctx context.Context, access state.Access) error {
 		siderolink.PendingMachineType,
 		siderolink.LinkStatusType,
 		siderolink.JoinTokenStatusType,
+		siderolink.JoinTokenType,
+		siderolink.DefaultJoinTokenType,
 		siderolink.NodeUniqueTokenStatusType,
 		siderolink.GRPCTunnelConfigType,
 		omni.MachineClassType,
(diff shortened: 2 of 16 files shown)


internal/backend/runtime/omni/state_access.go
```diff
@@ -505,7 +499,23 @@ func filterAccess(ctx context.Context, access state.Access) error {
 		virtual.PermissionsType:
 		// allow access with just valid signature
 		_, err = auth.CheckGRPC(ctx, auth.WithValidSignature(true))
-	case // admin-only resources
+	case
+		omni.RedactedClusterMachineConfigType,
+		omni.MachineConfigDiffType,
```
Unix4ever: I _think_ we don't need to hide this for the reader role.
Join tokens are redacted there.

Slessi: not 100% sure either but if so there is the schematic ID which is not redacted and can be used to obtain the join token indirectly, so we might have to redact schematic ID then, WDYT? does reader need access to machine config in the first place?

smira: I think MachineConfigDiff same as redacted machine config might be sensitive in general (even if the join token is redacted)

Unix4ever: Then we also need to restrict the config patches. I think they're currently available for reader

Slessi: Ye, i will add those also

Unix4ever: And pending updates are readable too.
But I guess it all will become irrelevant when we get the new RBAC in place.

Slessi: According to RFD we still want to offer the 3 current roles as static uneditable roles, so they should still be "good" roles in that case

Slessi: updated to also hide config patches!

frontend/src/pages/(authenticated)/clusters/[cluster]/machine/[machine].vue
```diff
@@ -39,6 +43,7 @@ const routes = computed(() => {
     {
       name: 'Config',
       to: { name: 'NodeConfig', params: { machine: machine.value } },
+      disabled: !canReadMachineConfig.value,
```
Comment (Unix4ever) [inline frontend/src/pages/(authenticated)/clusters/[cluster]/machine/[machine].vue:45]: Maybe we can just hide it completely?

frontend/src/pages/(authenticated)/clusters/[cluster]/machine/[machine].vue
```diff
@@ -39,6 +43,7 @@ const routes = computed(() => {
...
     },
     {
       name: 'Pending Updates',
```
Comment (Unix4ever) [inline frontend/src/pages/(authenticated)/clusters/[cluster]/machine/[machine].vue:48]: We need to disable this page too.

internal/backend/runtime/omni/state_access.go
utkuozdemir: Allowing read on config patches was intentional since the beginning, we didn't aim for Reader to not be able to read anything sensitive. This was not the function of the Reader role. Why did we want to do this change?

utkuozdemir: hiding all sensitive information, especially the join token, is I think not easily achievable. You can see machine's schematic in its logs, and from that you can also find out the token. I wouldn't try to fill all the gaps on this.

Slessi: The start issue was reading join tokens, as you can then as a reader start joining machines to clusters which is conceptually a mutation not a read.

Then, when catering for this, there was the issue that schematic IDs are shown in configs, which potentially embeds the join token in kernel args, so same problem.

Then its either redact schematic IDs from configs, or prevent showing readers configs. For now went with removing configs for readers on basis of "Do they even need to see configs?" so if you have a reason why its good to show readers configs maybe we can switch back and go for redacting schematic ID in config? Not sure if redacting that could be an issue for other people.

frontend/src/pages/(authenticated)/clusters/[cluster]/machine/[machine].vue
Comment (Slessi) [reply frontend/src/pages/(authenticated)/clusters/[cluster]/machine/[machine].vue:45]: I can hide, but usually for UX showing I sway towards disabled over hiding, as its more clear that an action _could_ be possible, just not by _you_. If it's hidden, user can be left wondering "how can I do this?" without realising that he cannot.

But if you prefer hiding I will hide.

internal/backend/runtime/omni/state_access.go
utkuozdemir: You are right about join tokens - whoever gets their hands on them can effectively join a machine, and it can be matched via a machine class and be added to a cluster and so on, so it's effectively a write.

But, join tokens are special in that sense, and the correct solution might be the secret management or so. For example, the join tokens would be redacted on all views, or something like that.

I am just not sure if restricting access of Reader, which has the original intention of "can read everything, but not write", is the right solution for the problem. It feels like a more complex problem to me than that.

utkuozdemir: @Unix4ever wdyt?

frontend/src/pages/(authenticated)/clusters/[cluster]/machine/[machine].vue
Comment (Slessi) [reply frontend/src/pages/(authenticated)/clusters/[cluster]/machine/[machine].vue:48]: Done

internal/backend/runtime/omni/state_access.go
Slessi: It boils down to what is the reader role for. ATM I only can imagine two situations

- `A`: Full read access to everything, no mutations. Use case: You only want to manage things with cluster template
- `B`: A reduced access user, who can only see some safe things. Use case: You don't trust this user with access to sensitive data

I lean towards `B`. Scenario `A` still requires you to have another user that does have access, to manage clusters through CLI, so not sure what the point of a user that can read sensitive data but can't act on it is.

Slessi: We discussed in slack, will make join tokens visible to readers

https://talossystems.slack.com/archives/C05348FENHG/p1776779027200659?thread_ts=1776754964.093119&cid=C05348FENHG

frontend/src/pages/(authenticated)/clusters/[cluster]/machine/[machine].vue
Comment (Unix4ever) [reply frontend/src/pages/(authenticated)/clusters/[cluster]/machine/[machine].vue:45]: Then we might need to make it more clear that it's disabled maybe. Right now it's just a tiny bit darker text

Comment (Slessi) [reply frontend/src/pages/(authenticated)/clusters/[cluster]/machine/[machine].vue:45]: It has also the `cursor-not-allowed` so it should be clear, but if design changes still needed can we do them in a different PR?

Comment (Unix4ever) [reply frontend/src/pages/(authenticated)/clusters/[cluster]/machine/[machine].vue:45]: yeah, I'm fine with separate PR

Unix4ever (approved):

PR #2713, opened 2026-04-22
fix(frontend): clear page state when keys are cleared

Slessi:
After clearing keys, use `location.replace` instead of `router.replace` to do a full page reload to clear any invalid key related state. This addresses an issue where user might see a blank screen if something breaks with their keys whilst still having a valid auth session. Usually an invalid auth session triggers a redirect to the auth provider, causing the same state reset. This also fixes a flakey test in `e2e-talemu` suite which was suffering from this.


oguzkilcan (approved):

frezbo (approved):
👏

smira (approved):

PR #2715, opened 2026-04-22
feat: add omnictl media command group with preset support

### LINKED ISSUE #2625: [feature] support downloading installation media using presets in CLI
This is the last gap of the installation media presets feature we need to cover.
Currently `omnictl download` only supports built-in presets.

oguzkilcan:
The UI exposes an Installation Media configurator that saves presets as InstallationMediaConfig resources, but the CLI only allowed downloading built-in images via inline flags. This adds `omnictl media preset {create,list,delete}` for managing presets and `omnictl media download <preset>` for downloading from them, with create-time validation against the server's CloudPlatformConfig (architecture and secure-boot support, min Talos version), SBCConfig (min Talos version), and TalosExtensions resources. Download flags are restricted to runtime overrides only - flags that redefine the preset (platform, overlay, arch, bootloader) are intentionally not accepted, and overrides that affect validation re-run the relevant checks.

The legacy `omnictl download` is preserved but deprecated, and its download logic is extracted into an internal/download package shared by both commands.

Resolves: #2625


oguzkilcan:
@claude review

Unix4ever:
Introducing all possible arguments as overrides to the `omnictl media download` gives us the same issue we had with the old UI for installation media download: you can get combinations which can't exist.
Like:
- can supply overlay for the cloud image.
- can set arm64 for Digital Ocean that doesn't support that.
- can set `overlay-options` without `overlay` maybe?
- can set secure boot for SBCs.

It quickly gets messy... 
While we can just drop all these args from `omnictl media download`, we can't get `create` right. It should be a wizard.

We should only have `talos-version`, `join-token`, `tunnel-mode`, `kernel-args`, `extensions` in the download command: the flags that can be applied to all possible images.

Unix4ever:
I also think that `omnictl media list` output doesn't have enough data to understand what's going on there.
While it might be fine to have it this way we might need `-owide` option or something.

I think `overlay` has higher priority than secure boot. Also it might be nice to see the extensions list.

Unix4ever:
I think we still need to add more data to the `omnictl media preset list`:
```
 → _out/omnictl-linux-amd64 media preset list
NAME    ARCH    TALOS VERSION   TYPE    SECURE BOOT
test    amd64   1.12.6          metal   false
test2   arm64   1.12.6          cloud   true
test3   amd64   1.12.6          cloud   false
```

I would suggest to add at least overlay and platform there.

oguzkilcan:
> I think we still need to add more data to the `omnictl media preset list`:
> 
> ```
>  → _out/omnictl-linux-amd64 media preset list
> NAME    ARCH    TALOS VERSION   TYPE    SECURE BOOT
> test    amd64   1.12.6          metal   false
> test2   arm64   1.12.6          cloud   true
> test3   amd64   1.12.6          cloud   false
> ```
> 
> I would suggest to add at least overlay and platform there.

Did you try `-o wide`?

Unix4ever:
ah, I didn't... Still I think we should swap PLATFORM/OVERLAY with SECURE BOOT.
Or just have it in addition to SECURE BOOT.

Unix4ever (approved):

PR #2735, opened 2026-04-24
feat(frontend): add join token/talos version placeholders in installation media

### LINKED ISSUE #2719: [feature] support join token/talos version placeholders in the presets
## `default`

Join token should support the `default` option that will make this schematic always pick the *current* default token when the image is downloaded

## `stable`

Talos version should support the `stable` option that makes it always download *current* default Talos version.

Support token + version + arch (?) picker in the download image dialog.

### Implementation

Preset resource can support empty strings for token and version. If empty it will fall back to the default/latest.
The UI will have dropdowns for that.
So for join tokens for example the dropdown will have all available tokens with option `default`.

oguzkilcan:
InstallationMediaConfig can now use empty strings for talosVersion and joinToken, which resolve to the current stable version and default token at download time.
    
The create wizard adds "Automatic" options to the version and token dropdowns, and the download modal shows version/token/arch pickers for all presets.

Resolves: #2719

image

image

### DIFF
--- a/frontend/src/pages/(authenticated)/machines/installation-media/create/talos-version.vue
+++ b/frontend/src/pages/(authenticated)/machines/installation-media/create/talos-version.vue
@@ -6,7 +6,7 @@ included in the LICENSE file.
 -->
 <script setup lang="ts">
 import { compare } from 'semver'
-import { computed, onBeforeMount, watch } from 'vue'
+import { computed, onBeforeMount } from 'vue'
 
 import { Runtime } from '@/api/common/omni.pb'
 import type { TalosVersionSpec } from '@/api/omni/specs/omni.pb'
@@ -23,7 +23,11 @@ import TSelectList from '@/components/SelectList/TSelectList.vue'
 import { getDocsLink } from '@/methods'
 import { useFeatures } from '@/methods/features'
 import { useResourceWatch } from '@/methods/useResourceWatch'
-import type { FormState } from '@/views/InstallationMedia/useFormState'
+import {
+  AUTOMATIC_VERSION,
+  type FormState,
+  resolveTalosVersion,
+} from '@/views/InstallationMedia/useFormState'
 
 definePage({ name: 'InstallationMediaCreateTalosVersion' })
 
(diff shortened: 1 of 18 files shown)


frontend/src/pages/(authenticated)/machines/installation-media/create/talos-version.vue
```diff
@@ -48,15 +48,26 @@ const { data: joinTokenList, loading: joinTokensLoading } = useResourceWatch<Joi
...
-const talosVersions = computed(() =>
-  talosVersionList.value
+const talosVersions = computed(() => {
+  const versions = talosVersionList.value
     .filter((v) => !v.spec.deprecated)
-    .map((v) => v.spec.version!)
-    .sort(compare),
+    .map((v) => ({ label: v.spec.version!, value: v.spec.version! }))
+
+  versions.push({ label: `${DefaultTalosVersion} (Stable)`, value: 'stable' })
```
Slessi: I think it would be more clear if it just says `Automatic`. Can also just be first in the list then.

At a glance seeing just `vX.XX (Stable)` (if you dont read the explanation above it) you might not understand this behaviour

---

Also, the `value` should be `''`? (both these comments apply to tokens also)

frontend/src/pages/(authenticated)/machines/installation-media/create/talos-version.vue
```diff
@@ -66,26 +77,58 @@ const joinTokens = computed(() =>
...
+})
+
+const resolvedTalosVersion = computed(() => resolveTalosVersion(formState.value.talosVersion))
 
-// Form defaults
-onBeforeMount(() => (formState.value.talosVersion ??= DefaultTalosVersion))
-watch(joinTokens, (v) => (formState.value.joinToken ??= v[0]?.value))
+// Form defaults — empty string means "stable" / "default"
+onBeforeMount(() => (formState.value.talosVersion ??= ''))
+watch(joinTokens, () => (formState.value.joinToken ??= ''))
```
Slessi: both can be inside `onBeforeMount` now, the `watch` isnt necessary anymore

frontend/src/pages/(authenticated)/machines/installation-media/create/talos-version.vue
```diff
@@ -66,26 +77,58 @@ const joinTokens = computed(() =>
...
+    formState.value.talosVersion = v === 'stable' ? '' : v
+  },
+})
+
+const selectedToken = computed({
+  get: () => formState.value.joinToken || 'default',
+  set: (v: string) => {
+    formState.value.joinToken = v === 'default' ? '' : v
+  },
+})
```
Slessi: i think this can be reverted to how it was before? it should just set blank string, and option value should be a blank string.

Unix4ever:
I'd call Talos version `Latest Stable`, and join token `Current Default`.
Then we won't need to explain what it does in the text below maybe.

Unix4ever (approved):

Slessi:
@Unix4ever 

I'm not sure that "Latest Stable" alone will be clear to the user that the version is not pinned to the current latest stable, and can change later. The "Automatic" + explanation was to make it explicit, but happy to remove explanation and keep "Latest Stable" only

Unix4ever:
> @Unix4ever
> 
> I'm not sure that "Latest Stable" alone will be clear to the user that the version is not pinned to the current latest stable, and can change later. The "Automatic" + explanation was to make it explicit, but happy to remove explanation and keep "Latest Stable" only

I don't have strong opinion here :)
Just a suggestion.

PR #2738, opened 2026-04-24
feat: download talosctl directly from factory

### LINKED ISSUE #2722: Omni UI should download `talosctl` from the Image Factory
Image Factory provides talosctl downloads from Talos 1.11+.

Slessi:
Download `talosctl` binaries from factory instead of Github.

Small change summary:
- Bump `talos/machinery` to 1.13, `image-factory` to 1.2
- Created a `Quirks` resource with `Get`/`List` to expose Talos quirk stuff to frontend (id = talos version)
- Updated E2E Omni config to talk to staging image-factory rather than prod
- Added a `downloadFile` helper to frontend to stop repeating this anchor creation/click/remove logic to trigger file downloads
- Migrated `DownloadTalosctl` to new modal format
- Replaced `/talosctl/downloads` with `/talosctl/downloads/{version}` which will list all the `talosctl` download links `image-factory` offers for that version (using quirks to filter list down to only the supported versions which is `1.11.0-alpha.3`+)
- Add E2E to try download a `talosctl` binary

image

Closes #2722 

Requires:
- https://github.com/siderolabs/talos/pull/13201
- https://github.com/siderolabs/image-factory/pull/431


frontend/e2e/talemu/installation_media.spec.ts
```diff
@@ -135,18 +135,20 @@ test('Download installation media', async ({ page }, testInfo) => {
 
     await expect(
       page.getByText(
-        `https://factory.talos.dev/image/${schematicId}/1.12.0/metal-arm64-secureboot.iso`,
+        `https://factory.staging.talos.dev/image/${schematicId}/1.12.0/metal-arm64-secureboot.iso`,
```
Unix4ever: I hope this won't bite us. Using staging on staging is one thing, but running every integration test against it might be a bit annoying if it's unstable.

Unix4ever (approved):

frontend/e2e/talemu/installation_media.spec.ts
Slessi: Easy to change it back if it is a problem later, I figured CI coverage can also catch staging issues. But also, my E2E test will fail on prod atm 😬 

It should be stable, we don't release factory often, and stage is only updated manually with new releases.

Slessi: Gave up with trying to fix this, taking too long. Just going to leave it as it is. But imo testing against prod and not staging is not great. Staging is where we should be testing things.

PR #2747, opened 2026-04-27
feat: support raw bytes in the inline fields for manifests/patches

### LINKED ISSUE #2683: [bug] Cluster export doesn't support multidoc patches
Current implementation of `omnictl cluster template export` parses only the first document in the list of patches defined in the patch resource.
It puts them into the template as `inline` patches and template doesn't support multiple inline documents in a single patch model, so we need to figure something out there.

We might need to add support for multiple inline documents in a single patch in the cluster templates. Or might need to split them to several patches if we don't want to do that.

Unix4ever:
Now inline supports all three variants:
- a single inline map (backward compatibility for config patches).
- a list of inline maps
- raw bytes, that can also contain multiple documents.

`omnictl cluster template export` command was updated to export config patches/manifests as raw bytes to ensure that multiple values are properly supported.

Fixes: https://github.com/siderolabs/omni/issues/2683

### DIFF
--- a/client/pkg/template/internal/models/inline.go
+++ b/client/pkg/template/internal/models/inline.go
@@ -1,0 +1,125 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+package models
+
+import (
+	"bytes"
+	"fmt"
+
+	"go.yaml.in/yaml/v4"
+)
+
+// InlineContent represents inline YAML content embedded into a template.
+//
+// It accepts three input forms:
+//   - a single YAML mapping,
+//   - a sequence of YAML mappings (multi-document),
+//   - a raw YAML literal block scalar (bytes).
+//
+//nolint:recvcheck
+type InlineContent struct {
+	raw  []byte
+	docs []map[string]any
+}
+
+// NewInlineContent builds InlineContent from one or more YAML mappings.
+func NewInlineContent(docs ...map[string]any) *InlineContent {
+	return &InlineContent{docs: docs}
+}
+
+// NewInlineContentBytes builds InlineContent from raw YAML bytes.
+func NewInlineContentBytes(raw []byte) *InlineContent {
+	return &InlineContent{raw: raw}
+}
+
+// IsZero reports whether the InlineContent is empty.
+func (i InlineContent) IsZero() bool {
+	return i.raw == nil && len(i.docs) == 0
+}
+
+// Bytes returns the YAML-encoded inline content.
+//
+// Multi-document content is encoded with the standard YAML multi-document framing.
+func (i InlineContent) Bytes() ([]byte, error) {
+	if i.raw != nil {
+		return i.raw, nil
+	}
+
+	if len(i.docs) == 0 {
+		return nil, nil
+	}
+
+	var buf bytes.Buffer
+
+	encoder := yaml.NewEncoder(&buf)
+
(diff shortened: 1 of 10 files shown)


client/pkg/template/internal/models/inline.go
```diff
@@ -0,0 +1,120 @@
...
+		var m map[string]any
+
+		if err := node.Decode(&m); err != nil {
+			return err
+		}
+
+		i.docs = []map[string]any{m}
+
+		return nil
+	case yaml.SequenceNode:
```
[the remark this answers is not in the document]
Unix4ever: Manifests/Config patches are validated on server. So we can ignore this issue.

client/pkg/template/internal/models/inline.go
```diff
@@ -0,0 +1,121 @@
...
+			return nil
+		}
+
+		return i.UnmarshalYAML(node.Content[0])
+	}
+
+	//nolint:exhaustive
+	switch node.Kind {
+	case yaml.ScalarNode:
+		i.raw = []byte(node.Value)
```
smira, line 81: I think we should check the type here, scalar is not only string

client/pkg/template/internal/models/inline.go
```diff
@@ -0,0 +1,121 @@
...
+	if err := encoder.Close(); err != nil {
+		return nil, err
+	}
+
+	return buf.Bytes(), nil
+}
+
+// UnmarshalYAML implements yaml.Unmarshaler.
+func (i *InlineContent) UnmarshalYAML(node *yaml.Node) error {
+	if node.Kind == yaml.DocumentNode {
```
smira: not sure why this is out the switch? and not sure how document node might be here?

client/pkg/template/internal/models/inline_test.go
```diff
@@ -0,0 +1,260 @@
...
+  kind: ConfigMap
+  metadata:
+    name: a
+- apiVersion: v1
+  kind: ConfigMap
+  metadata:
+    name: b
+`
+
+	var got *models.InlineContent
```
smira: as a nit, I would probably test it in a context closer to real usage, that is wrap into some struct with a field being inline content:

```
struct {
   Inline models.InlineContent `yaml:"inline"`
```

I think visually it would be easier to validate, and I would put it to `testdata` so it's easier to see it

client/pkg/template/internal/models/inline.go
```diff
@@ -0,0 +1,121 @@
...
+		i.docs = s
+
+		return nil
+	default:
+		return fmt.Errorf("unsupported inline node kind %d", node.Kind)
+	}
+}
+
+// MarshalYAML implements yaml.Marshaler.
+func (i *InlineContent) MarshalYAML() (any, error) {
```
smira: If we implement correct `IsZero()` (if I remember name correctly), we can stop doing `*InlineContent` and use it by value `InlineContent` and support `omitempty`

client/pkg/template/operations/export.go
```diff
@@ -157,25 +156,19 @@ func writeYAML(writer io.Writer, modelList models.List) (err error) {
...
 	}
 
 	defer buffer.Free()
 
-	patchData := buffer.Data()
-
-	if err = yaml.Unmarshal(patchData, &data); err != nil {
-		return models.Patch{}, fmt.Errorf("failed to unmarshal patch %q: %w", configPatch.Metadata().ID(), err)
-	}
+	raw := append([]byte(nil), buffer.Data()...)
```
smira: `slices.Clone` ?

smira (approved):

PR #2776, opened 2026-05-04
fix(frontend): only show machine patches for currently visible machine

### LINKED ISSUE #2763: Machine patch view shows all patches in cluster

### Current Behavior

Given a cluster managed using cluster templates which consists of:
- cluster patches
- control plane patches
- worker patches
- individual machine patches

When using the Omni UI -> Cluster -> **Machine** -> Patches,

all patches are displayed

### Expected Behavior

When viewing the patches for a given machine, I'd expect only to see patches related to that machine.

### Steps To Reproduce

1. Create a cluster using cluster templates
2. Use a mixture of:
  - cluster patches
  - control plane patches
  - worker patches
  - individual machine patches
3. View the patches for an individual machine

_No response_

### Anything else?

_No response_

Slessi:
Only show patches for the currently viewed machine on the Patches tab of a clustered or unclustered machine.

Closes #2763

### DIFF
--- a/frontend/src/views/Config/Patches.vue
+++ b/frontend/src/views/Config/Patches.vue
@@ -13,8 +13,7 @@ import type { RouteLocationRaw } from 'vue-router'
 import WordHighlighter from 'vue-word-highlighter'
 
 import { Runtime } from '@/api/common/omni.pb'
-import type { Resource } from '@/api/grpc'
-import type { ConfigPatchSpec } from '@/api/omni/specs/omni.pb'
+import type { ClusterMachineStatusSpec, ConfigPatchSpec } from '@/api/omni/specs/omni.pb'
 import {
   ClusterMachineStatusType,
   ConfigPatchDescription,
@@ -60,10 +59,10 @@ const { canManageConfigPatches: canManageClusterMachineConfigPatches } = useClus
 const selectors = computed(() => {
   const res: string[] = []
 
-  if (cluster) {
-    res.push(`${LabelCluster}=${cluster}`)
-  } else if (machine) {
+  if (machine) {
     res.push(`${LabelMachine}=${machine}`, `${LabelClusterMachine}=${machine}`)
+  } else if (cluster) {
+    res.push(`${LabelCluster}=${cluster}`)
   } else {
     return
   }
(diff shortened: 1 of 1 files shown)


frontend/src/views/Config/Patches.vue
```diff
@@ -60,10 +60,10 @@ const { canManageConfigPatches: canManageClusterMachineConfigPatches } = useClus
 const selectors = computed(() => {
   const res: string[] = []
 
-  if (cluster) {
-    res.push(`${LabelCluster}=${cluster}`)
-  } else if (machine) {
+  if (machine) {
     res.push(`${LabelMachine}=${machine}`, `${LabelClusterMachine}=${machine}`)
+  } else if (cluster) {
+    res.push(`${LabelCluster}=${cluster}`)
```
Slessi, line 65: this is an artifact from the file based router change, logic was a bit unclear how it was before, it didn't used to have `cluster` assigned here for the `ClusterMachine` view. Got added during refactor (thought it was a mistake that it was omitted).

Changed priority here so that selectors say if we specify a machine, only query by machine.

smira: I think these labels are still wrong now.

The logic should match the controller, and it's not trivial:

* machine-level patches should match on machine 
* machineset patches on machine set ID
* cluster-level patches which don't have a specific machine/machineset labels

smira: I might be wrong on the exact queries, but I think it's certainly not `if/else`

Slessi: How I changed it here is to reflect how it was before the unintentional change.
We have 3 places were we show patches and the previous logic I just reverted to is:

- `Cluster` page
  - `omni.sidero.dev/cluster=cluster`
- `Machine` page
  - `omni.sidero.dev/machine=machine,omni.sidero.dev/cluster-machine=machine`
- `ClusterMachine` page (this is the wrong one I fixed to be like below)
  - `omni.sidero.dev/machine=machine,omni.sidero.dev/cluster-machine=machine`

This way, machine set patches will only show on the `Cluster` patches page, and machines page will only show `Machine`/`ClusterMachine` patches, which is how it was before.

Does that make sense for the context of this PR to not show patches for different cluster machines, when looking at a specific machine? If you disagree holistically that this original approach is also wrong thats fine but for a different topic, we had started a discussion [here](https://talossystems.slack.com/archives/C06ETD99FJS/p1774622019822579) and with Sterling about it already and we do plan to overhaul things a bit to make everything more clear.

smira: I mean we should fix the queries if we're fixing it to be correct always.

Probably someone else might give you a better suggestion, but my 2 cents is to make it match the actual way backend applies patches.

Unix4ever (approved):

PR #2781, opened 2026-05-05
feat: expose provision step errors on machine request status

### LINKED ISSUE #2629: [bug] Infra provider provision errors not visible in UI — cleared on every retry
### Current Behavior

When an infra provider's provision step returns an error, the `MachineRequestStatus.Error` field is set and `Stage` becomes `FAILED` — but only for ~1 second. The provision controller immediately requeues the reconciliation (line 344 of `provision.go`), and on the next attempt, line 314 clears the error:

```go
machineRequestStatus.TypedSpec().Value.Error = ""
machineRequestStatus.TypedSpec().Value.Stage = specs.MachineRequestStatusSpec_PROVISIONING
```

The result: **the Omni UI permanently shows "Provisioning" with no error message**, even when the provider is failing on every retry. Users have no way to know what's wrong without checking provider logs or rapidly polling `omnictl get machinerequeststatus`.

### Expected Behavior

The last provision error should persist in the `MachineRequestStatus` during retries so users can see it in the Omni UI. For example:

- Stage: `PROVISIONING` (still retrying)
- Status: `Retrying step "uploadISO" (attempt 3/∞)`
- Error: `pool "previewk8" not found on TrueNAS — this looks like a dataset path, not a pool name`

This lets operators diagnose issues from the UI without needing access to provider logs.

### Steps To Reproduce

1. Deploy any infra provider (e.g., TrueNAS provider)
2. Create a MachineClass with an invalid configuration (e.g., a ZFS pool name that doesn't exist)
3. Create a cluster using that MachineClass
4. Observe the Omni UI — machines show "Provisioning" indefinitely
5. Check provider logs — errors are logged every 60 seconds:
   ```
   machine provision failed  error="pool \"previewk8\" not found on TrueNAS"  step="uploadISO"
   ```
6. Run `omnictl get machinerequeststatus -o yaml` — the `error` field is empty most of the time, only populated for ~1 second between failure and the next retry's status reset

### Additional Context

This affects all infra providers, not just the TrueNAS provider. The behavior is in the SDK's `ProvisionController` at:
- `pkg/infra/controllers/provision.go` line 314 (clears error on retry)
- `pkg/infra/controllers/provision.go` line 341 (sets error on failure)
- `pkg/infra/controllers/provision.go` line 344 (requeues after 1 minute)

**Suggested fix**: Preserve the last error in a separate field (e.g., `LastError`) that isn't cleared on retry, or keep the `Error` field populated while `Stage` is `PROVISIONING` to indicate "retrying after error."

This was discovered while building the [omni-infra-provider-truenas](https://github.com/bearbinary/omni-infra-provider-truenas) — common user mistakes like specifying a dataset name instead of a pool name produce clear error messages from the provider, but users never see them in Omni.

oguzkilcan:
Surface step errors via a new Error field on `ClusterMachineRequestStatus` so users can see why a request is stuck or has failed without scraping logs. Persist the error on both the failure and requeue paths in the provision controller (the previous in-memory mutations were never written), and constrain step mutations to Id and LabelMachineInfraID by passing a copy of `MachineRequestStatus` to step.Run instead of the controller's live object.

Resolves: #2629

### DIFF
--- a/client/pkg/infra/controllers/provision.go
+++ b/client/pkg/infra/controllers/provision.go
@@ -6,6 +6,7 @@ package controllers
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"slices"
 	"strings"
@@ -18,7 +19,6 @@ import (
 	"github.com/cosi-project/runtime/pkg/safe"
 	"github.com/cosi-project/runtime/pkg/state"
 	"github.com/siderolabs/gen/optional"
-	"github.com/siderolabs/gen/xerrors"
 	"go.uber.org/zap"
 
 	"github.com/siderolabs/omni/client/api/omni/specs"
@@ -316,10 +316,13 @@ func (ctrl *ProvisionController[T]) reconcileRunning(ctx context.Context, r cont
 		machineRequestStatus.TypedSpec().Value.Error = ""
 		machineRequestStatus.TypedSpec().Value.Stage = specs.MachineRequestStatusSpec_PROVISIONING
 
+		// Pass a copy to the step so mutations beyond Id and LabelMachineInfraID don't leak into our state.
+		mrsCopy := machineRequestStatus.DeepCopy().(*infra.MachineRequestStatus) //nolint:forcetypeassert,errcheck
+
 		if err = safe.WriterModify(ctx, r, res.(T), func(st T) error { //nolint:forcetypeassert,errcheck
 			err = step.Run(ctx, logger, provision.NewContext(
 				machineRequest,
-				machineRequestStatus,
+				mrsCopy,
 				st,
 				connectionParams,
 				ctrl.imageFactory,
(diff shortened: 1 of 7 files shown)


client/pkg/infra/controllers/provision.go
```diff
@@ -351,6 +362,12 @@ func (ctrl *ProvisionController[T]) reconcileRunning(ctx context.Context, r cont
 			) error {
 				res.TypedSpec().Value = machineRequestStatus.TypedSpec().Value
 
+				applyStepMutations(res, mrsCopy)
```
Unix4ever, line 365: do we really need both? They kinda intersect

Unix4ever (approved):

client/pkg/infra/controllers/provision.go
oguzkilcan: What do you mean by both? Applying user mutations on normal error and retry error or doing the copy from our instance and then applying user changes?

PR #2782, opened 2026-05-05
fix(frontend): fix revoking/deleting join tokens

Slessi:
Fix an issue were the delete/revoke button for join tokens was permanently disabled in the modals. Refactor modals to ConfirmModal. Add stories & tests.


frontend/src/views/Modals/components/JoinTokenWarnings.vue
```diff
@@ -35,8 +31,8 @@ const {
   runtime: Runtime.Omni,
 }))
 
-watch(running, () => {
-  if (running.value) {
+watchEffect(() => {
+  if (joinTokenStatus.value) {
```
Slessi, line 35: here was the problem, `running` is already `true` (and remained `true`) at this stage, so the `watch` never fired. I just removed `running` all together, it isn't a useful flag. Just need to know that token has loaded.

Unix4ever (approved):

PR #2794, opened 2026-05-06
feat(frontend): allow switching logs inside logs tab

### LINKED ISSUE #2712: Make service logs UX more clear
### Problem Description

Currently if you click on one of a a clustered machine's services from the cluster machine overview page, you are taken to the `Console Logs` tab, which looks identical to as if you entered `Console Logs` manually, but is actually a different view of logs. This is not clear, the only way to see a difference is in the URL.

Image

Image

### Solution

Maybe `Console Logs` should just be called `Logs` and inside the view there is a dropdown next to the search which has the list of services you can choose from, and it will be pre-selected with the service. One of the options can be "All" which would just shows all logs.

### Alternative Solutions

_No response_

### Notes

_No response_

Slessi:
Allow switching between machine and service logs from inside the logs tab via a dropdown of services. Also includes some refactoring to allow the same behaviour on both machine page and node page.

Closes #2712 

image


Unix4ever:
I think dropdown height is 2 pixels bigger than the input. We can either do `item-center` align there, or fix the difference. Maybe 2nd option is prefferable.

Unix4ever:
Lets also manually add `controller-runtime` to the list of services.
It's not reported as the running service, but you can get the logs for it still.

frezbo:
> Lets also manually add `controller-runtime` to the list of services. It's not reported as the running service, but you can get the logs for it still.

there;s also `auditd` and `syslogd`

Slessi:
@frezbo those ones are part of the service list already

@Unix4ever sure but _should_ it be part of the service list? is fixing the backend API the correct fix or shall i just manually hardcode in frontend?

hardcoded for now, had to do some regex for parsing and to strip the ANSI colours

frontend/src/pages/(authenticated)/clusters/[cluster]/machine/[machine]/logs/[[service]].vue
Comment (utkuozdemir) [inline frontend/src/pages/(authenticated)/clusters/[cluster]/machine/[machine]/logs/[[service]].vue:1]: are double square brackets intentional?

Comment (Slessi) [reply frontend/src/pages/(authenticated)/clusters/[cluster]/machine/[machine]/logs/[[service]].vue:1]: yea that is the syntax for optional params when using file routing

https://router.vuejs.org/file-based-routing/file-based-routing.html#Dynamic-Routes

> You can create [optional params](https://router.vuejs.org/guide/essentials/route-matching-syntax#optional-parameters) by wrapping the param name with an extra pair of brackets, e.g. `src/pages/users/[[id]].vue` will create a route with the following path: `/users/:id?`.

utkuozdemir (approved):
Don't count on my review on frontend-specific stuff, but logic-wise I don't see any issues.

frontend/src/pages/(authenticated)/clusters/[cluster]/machine/[machine]/logs/[[service]].vue
Comment (utkuozdemir) [reply frontend/src/pages/(authenticated)/clusters/[cluster]/machine/[machine]/logs/[[service]].vue:1]: Got it. Not used to such filenames :)

Unix4ever:
> @frezbo those ones are part of the service list already
> 
> @Unix4ever sure but _should_ it be part of the service list? is fixing the backend API the correct fix or shall i just manually hardcode in frontend?
> 
> hardcoded for now, had to do some regex for parsing and to strip the ANSI colours

We can't add it to the backend response as it's coming from Talos

PR #2796, opened 2026-05-06
feat: proxy image factory requests done by the providers through Omni

### LINKED ISSUE #2787: implement image factory create schematic API in Omni
It should be proxied to the correct image factory and use the correct auth parameters.

Currently the providers go to the IF directly. It has some issues:
- we have to pass extra credentials to each provider to let them talk to the enterprise image factory/
- having several image factories forces each infra provider support that separately.

Unix4ever:
Introduce the new API for creating schematics using raw YAML data.

Proxied API becomes useful when Omni has auth to the image factory, or if it gets support for several image factory endpoints.

Fixes: https://github.com/siderolabs/omni/issues/2787

### DIFF
--- a/client/api/omni/management/management.proto
+++ b/client/api/omni/management/management.proto
@@ -175,6 +175,10 @@ message CreateSchematicRequest {
   SchematicBootloader bootloader = 10;
 }
 
+message CreateSchematicFromRawRequest {
+  bytes raw_schematic = 1;
+}
+
 message CreateSchematicResponse {
   string schematic_id = 1;
   string pxe_url = 2;
@@ -356,6 +360,7 @@ service ManagementService {
   rpc KubernetesUpgradePreChecks(KubernetesUpgradePreChecksRequest) returns (KubernetesUpgradePreChecksResponse);
   rpc KubernetesSyncManifests(KubernetesSyncManifestRequest) returns (stream KubernetesSyncManifestResponse);
   rpc CreateSchematic(CreateSchematicRequest) returns (CreateSchematicResponse);
+  rpc CreateSchematicFromRaw(CreateSchematicFromRawRequest) returns (CreateSchematicResponse);
   rpc GetSupportBundle(GetSupportBundleRequest) returns (stream GetSupportBundleResponse);
   rpc ReadAuditLog(ReadAuditLogRequest) returns (stream ReadAuditLogResponse);
   rpc MaintenanceUpgrade(MaintenanceUpgradeRequest) returns (MaintenanceUpgradeResponse);
--- a/client/pkg/client/management/management.go
+++ b/client/pkg/client/management/management.go
@@ -154,6 +154,15 @@ func (client *Client) CreateSchematic(ctx context.Context, req *management.Creat
 	return schematic, nil
 }
 
+// CreateSchematicFromRaw creates a schematic from the raw schematic string using the image factory.
+func (client *Client) CreateSchematicFromRaw(ctx context.Context, schematic []byte) (*management.CreateSchematicResponse, error) {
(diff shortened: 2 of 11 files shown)


internal/backend/grpc/schematics.go
```diff
@@ -124,6 +124,30 @@ func (s *managementServer) CreateSchematic(ctx context.Context, request *managem
 	}, nil
 }
 
+// CreateSchematicFromRaw implements managementServer.
+func (s *managementServer) CreateSchematicFromRaw(ctx context.Context, request *management.CreateSchematicFromRawRequest) (*management.CreateSchematicResponse, error) {
+	if _, err := auth.CheckGRPC(ctx, auth.WithValidSignature(true)); err != nil {
```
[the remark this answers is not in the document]
Unix4ever: As this is a simple proxy request we shouldn't really protect it that much

client/api/omni/management/management.proto
utkuozdemir, line 186: As a side note, this will need to be set to whatever the factory returns, when it starts returning normalized schematics.

client/pkg/client/management/management.go
```diff
@@ -154,6 +156,20 @@ func (client *Client) CreateSchematic(ctx context.Context, req *management.Creat
 	return schematic, nil
 }
 
+// CreateSchematicFromRaw creates a schematic from the raw schematic string using the image factory.
+func (client *Client) CreateSchematicFromRaw(ctx context.Context, schematic *schematic.Schematic) (*management.CreateSchematicResponse, error) {
```
utkuozdemir: Should we actually use the structured type here as arg (`schematic *schematic.Schematic`), in the client lib? It is not really "raw". Maybe we make it accept []byte or string instead?

client/pkg/infra/imagefactory/client.go
```diff
@@ -53,7 +54,29 @@ func NewClient(options ClientOptions) (*Client, error) {
...
+
+// NewProxiedClient creates new proxied image factory client.
+func NewProxiedClient(client *omniclient.Client) (*ProxiedClient, error) {
+	return &ProxiedClient{
+		client: client,
+	}, nil
+}
+
+// EnsureSchematic creates the schematic in the image factory through Omni API.
+func (pc *ProxiedClient) EnsureSchematic(ctx context.Context, schematic schematic.Schematic) (string, error) {
```
utkuozdemir, line 75: Same question - should we accept raw bytes instead?

utkuozdemir (approved):
LGTM other than the question: should we pass the raw schematic all the way, never unmarshaling it to `schematic.Schematic` on the way? Maybe it serves us better in the future?

client/pkg/client/management/management.go
Unix4ever: Ok, lets change it. I don't really have a preference here.

client/pkg/infra/imagefactory/client.go
Unix4ever: This might require changes in many plases.

Unix4ever: Many places create schematic as a struct. It won't be convenient to use it with raw bytes here.

Unix4ever:
> LGTM other than the question: should we pass the raw schematic all the way, never unmarshaling it to schematic.Schematic on the way? Maybe it serves us better in the future?

all current users of this API are constructing it using `schematic.Schematic`. We actually need this struct, not the raw bytes.
Old `CreateSchematic` func is designed more for the frontend usage that doesn't have ready to use `schematic.Schematic`.

client/api/omni/management/management.proto
Unix4ever: that's for the next PRs

PR #2799, opened 2026-05-07
feat: periodically re-check FULL-mode kubernetes manifest groups

kevintijssen:
When a kubernetes manifest group is configured in FULL mode, we now requeue the cluster manifest status reconciler every 5 minutes even when everything looks in sync. This allows out-of-band deletions or edits on the workload cluster to be detected and reverted without a manual nudge, since the next reconcile pass will Get every manifest, mark the group OutOfSync, and let the existing 30s requeue drive it back to the desired state.


smira:
I think this needs a bigger discussion - our initial design assumed Omni should never do this, it has performance implications.

Slessi:
Will be fixed differently!

PR #2815, opened 2026-05-11
chore(frontend): bump dependencies

Slessi:
Bump frontend dependencies


smira (approved):

frontend/e2e/talemu/installation_media.spec.ts
```diff
@@ -105,33 +105,44 @@ test('Download installation media', async ({ page }, testInfo) => {
...
 
-    schematicId = await page.evaluate(() => navigator.clipboard.readText())
-    expect(schematicId, 'Expect schematic ID to be valid').toMatch(/[a-zA-Z0-9]{64}/)
-
+    await expect
+      .poll(async () => {
+        schematicId = await page.evaluate(() => navigator.clipboard.readText())
+        return schematicId
+      }, 'Expect schematic ID to be valid')
+      .toMatch(/[a-zA-Z0-9]{64}/)
```
Slessi, line 113: NB `@vueuse` `14.3.0` has an [update](https://github.com/vueuse/vueuse/pull/5369) to `useClipboard` which causes it to be async which isn't relevant to real users but broke tests here, so switched to poll for now

There is an aside issue of playwright not having [clipboard isolation](https://github.com/microsoft/playwright/issues/13097) but for now I think it is fine to ignore

PR #2827, opened 2026-05-12
chore: rekres, bump deps and default versions

oguzkilcan:
Rekres. Bump deps and default versions


hack/test/common.sh
```diff
@@ -22,14 +22,14 @@ echo "127.0.0.1 omni.localhost" | tee -a /etc/hosts
...
+export TALOS_VERSION=1.13.1
+export KUBERNETES_VERSION=1.36.0
 # To use in:
 # - Omni upgrade tests, to prevent Talos changes interfering with Omni changes
 # - Talos minor upgrade tests
-export STABLE_TALOS_VERSION=1.11.6
+export STABLE_TALOS_VERSION=1.12.6
 export ANOTHER_OMNI_VERSION="${ANOTHER_OMNI_VERSION:-$LATEST_STABLE_OMNI}"
-export ANOTHER_KUBERNETES_VERSION=1.34.6
+export ANOTHER_KUBERNETES_VERSION=1.35.4
```
utkuozdemir, line 32: Can use 1.35.5 if you want (seems just released)

Unix4ever (approved):

utkuozdemir (approved):

hack/test/common.sh
oguzkilcan: Wow, it was released soon after I checked it. I'll update.

oguzkilcan: I might not update if all the tests pass, this is just used for testing anyway.

Issue #2797, opened 2026-05-07
Rework/fix schematic ID calculation/normalization


utkuozdemir:
In Omni, in various places, we build a `schematic.Schematic`, and call `.ID()` on it to get its ID. And we assume that this is the final ID, and it would match what the image factory would return, if we submitted it to factory instead.

(One of these places is `SchematicConfigurationController`).

This assumption is wrong, as there is no guarantee of that. Image factory can add fields (one example of this is the `owner1 field of enterprise factory with auth), normalize the whitespace differently, and so on.

Because of this, we should always take the ID returned by the factory as the source of truth.
Factory will soon start returning the normalized schematic in the response body as well as ID, which should get the same treatment - take that as the final form of the schematic, not what we have built.

We need to fix the places where we use the locally-computed schematic ID.

We have a schematic cache in Omni (we cache them in a resource) - this cache probably already has potentially wrong keys.
We should change our usage of the cache to:
- always populate the cache keys/values from what the factory returns.
- if we locally compute a schematic ID, and if we have a cache it, that's fine. But we should never put that ID into the cache, as it might not be final.
- We might consider doing a cache cleanup (with a migration?) or maybe dropping the cache altogether (might make Omni much more chatty with factory and would be more affected by network failures etc).

1. Find all the places where we do things with the wrong assumption, and fix.
2. Rework the cache to also fix this assumption.

PR #2845, opened 2026-05-14
feat: support basic auth against the image factory

### LINKED ISSUE #2721: support for Image Factory authentication
Omni configuration should support configuration for Image Factory authentication: username/password.

Omni uses that internally when talking to the Image Factory API.
Download media uses these credentials transparent for the user, final download links might have username/password embedded into the URL.

Omni should patch Talos machine config with Image Factory credentials: [RegistryAuthConfig](https://docs.siderolabs.com/talos/v1.12/reference/configuration/cri/registryauthconfig). (Including maintenance mode).

Unix4ever:
Add `registries.imageFactoryUsername` and `registries.imageFactoryPassword` config options. When both are set, the server-side image factory client authenticates with HTTP basic auth, and the credentials are embedded into the image factory base URL exposed through the features config so that downstream consumers (e.g. PXE boot) can reach the same factory.

Also drops the redundant `omni.Schematic` state-cache lookup from the server-side client now that the factory call is the source of truth.

Fixes: #2721


deploy/helm/omni/values.yaml
```diff
@@ -511,6 +511,10 @@ config:
     #imageFactoryBaseURL: https://factory.talos.dev
     # ImageFactoryPXEBaseURL is the base URL of the Image Factory PXE endpoint used to build custom PXE boot images.
     #imageFactoryPXEBaseURL: ""
+    # ImageFactoryUsername is the username used to authenticate against the Image Factory service.
```
oguzkilcan: Shouldn't we say Image Factory Enterprise here, since normal Image Factory doesn't support auth?

client/pkg/omni/resources/virtual/image_factory_auth.go
```diff
@@ -0,0 +1,63 @@
...
+		DefaultNamespace: resources.VirtualNamespace,
+		PrintColumns: []meta.PrintColumn{
+			{
+				Name:     "Username",
+				JSONPath: "{.username}",
+			},
+			{
+				Name:     "Password",
+				JSONPath: "{.password}",
+			},
```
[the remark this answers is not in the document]
utkuozdemir: I agree with this - shouldn't be a print column imo.

client/pkg/omni/resources/omni/schematic.go
utkuozdemir, line 1: I was thinking about changing the cache to take what the factory returns as the truth instead, and cache that. I wonder if dropping this will make Omni less resilient to image factory issues.

internal/pkg/config/config.go
```diff
@@ -207,6 +207,13 @@ func (p *Params) Validate(schema *jsonschema.Schema) error {
 		return fmt.Errorf("failed to validate config against JSON schema: %w", err)
 	}
 
+	username := p.Registries.GetImageFactoryUsername()
+	password := p.Registries.GetImageFactoryPassword()
+
+	if (username == "") != (password == "") {
```
utkuozdemir: Wondering if we could express this in the JSON schema instead, and change the `formatLeafError` function to humanize the error the schema validator returns.

cmd/omni/cmd/cmd.go
```diff
@@ -326,6 +326,8 @@ func defineRegistriesFlags(b *FlagBinder, flagConfig *config.Params) {
 	b.StringVar("registries.kubernetes", &flagConfig.Registries.Kubernetes)
 	b.StringVar("registries.imageFactoryBaseURL", &flagConfig.Registries.ImageFactoryBaseURL)
 	b.StringVar("registries.imageFactoryPXEBaseURL", &flagConfig.Registries.ImageFactoryPXEBaseURL)
+	b.StringVar("registries.imageFactoryUsername", &flagConfig.Registries.ImageFactoryUsername)
+	b.StringVar("registries.imageFactoryPassword", &flagConfig.Registries.ImageFactoryPassword)
```
utkuozdemir, line 330: Can we provide a way to pass this via an  env var? Since it is a sensitive piece of information.

I actually prototyped a way to pass all config values via env a while ago, but it's not ready for a PR yet.
Maybe `OMNI_IMAGE_FACTORY_USERNAME and `OMNI_IMAGE_FACTORY_PASSWORD` or so? Wdyt?
At least for the password.

internal/pkg/features/features.go
```diff
@@ -24,6 +24,8 @@ import (
 // Params holds the values needed to update the features config resource.
 type Params struct {
 	ImageFactoryPXEBaseURL          *url.URL
+	ImageFactoryUsername            string
+	ImageFactoryPassword            string
```
utkuozdemir: This resource only needs a valid signature, no role. Do we actually want to add them to this resource? Sensitive info, no?

internal/backend/grpc/schematics.go
```diff
@@ -113,14 +113,19 @@ func (s *managementServer) CreateSchematic(ctx context.Context, request *managem
 		return nil, fmt.Errorf("failed to ensure schematic: %w", err)
 	}
 
-	schematicYml, err := yaml.Marshal(schematicRequest)
+	schematicResponse, err := s.imageFactoryClient.SchematicGet(ctx, schematicInfo.FullID)
```
utkuozdemir: Here you can bump the factory client instead, and it returns the schematic body in the create call. It also internally does the fallback smartly to doing a get, if the response doesn't contain the body.

cmd/omni/cmd/cmd.go
Unix4ever: sure, can additionally have an env var for that

frontend/src/methods/useImageFactoryAuth.ts
```diff
@@ -0,0 +1,54 @@
...
+  })!
+}
+
+// withImageFactoryAuth embeds basic auth credentials into a URL's userinfo so
+// that top-level browser navigation (anchor clicks) sends them. Useful only for
+// links the browser navigates to directly.
+export function withImageFactoryAuth(
+  url: string | undefined,
+  spec: ImageFactoryAuthSpec | undefined,
+): string | undefined {
```
Slessi: Using `thing?: string` rather than `thing: string | undefined` is clearer syntax for optional params, otherwise it feels like "required but could be undefined" which is weird.

That said is `url` ever actually optional? Seems required to me.

For return type just omit and let it be inferred by TS. Less manual types is more readable, and TS sometimes infers better ones. We use strict mode now so any inferred `any` will be an error, so that won't be missed.


```suggestion
export function withImageFactoryAuth(
  url: string,
  spec?: ImageFactoryAuthSpec,
) {
```

frontend/src/methods/useImageFactoryAuth.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  try {
+    const u = new URL(url)
+    u.username = spec.username
+    u.password = spec.password
+
+    return u.toString()
+  } catch {
+    return url
+  }
```
Slessi: wont this potentially suppress errors? i dont see any valid reasons for this to ever throw, i think better to drop the try/catch?

frontend/src/views/InstallationMedia/Confirmation.vue
```diff
@@ -124,13 +132,37 @@ const schematicId = computed(() => schematic.value?.id ?? '')
...
-const factoryUrl = computed(() => features.value?.spec.image_factory_base_url)
+const factoryHost = computed(() => {
+  const baseURL = features.value?.spec.image_factory_base_url
+  if (!baseURL) return ''
+
+  try {
+    return new URL(baseURL).host
+  } catch {
+    return baseURL
+  }
```
Slessi: here also, why should this throw? if it does throw, how would we know without deliberate inspection?

frontend/src/views/InstallationMedia/Confirmation.vue
```diff
@@ -60,7 +62,13 @@ const talosctlAvailable = computed(() => quirks.value?.spec.supports_factory_tal
 
 const { data: features } = useFeatures()
 
-const { data: talosctlPaths } = useTalosctlDownloads(() => resolvedTalosVersion.value)
+const imageFactoryAuth = useImageFactoryAuth()
+
+const { data: talosctlPathsRaw } = useTalosctlDownloads(() => resolvedTalosVersion.value)
+
+const talosctlPaths = computed(() =>
+  talosctlPathsRaw.value.map((path) => withImageFactoryAuth(path, imageFactoryAuth.value)!),
```
Slessi, line 69: we need auth for downloading talosctl too? 👀

frontend/src/views/InstallationMedia/Confirmation.vue
```diff
@@ -124,13 +132,37 @@ const schematicId = computed(() => schematic.value?.id ?? '')
...
+const clusterCreateCommand = computed(() => {
+  const parts = [
+    'talosctl cluster create qemu',
+    `--schematic-id=${schematicId.value}`,
+    `--talos-version=v${resolvedTalosVersion.value}`,
+  ]
+
+  const { username, password } = imageFactoryAuth.value ?? {}
+  if (username && password) {
+    parts.push(`--image-factory-auth=${shellQuote([`${username}:${password}`])}`)
```
Slessi: does it really need quoting here? if so, mabe a package for this not worth it, seems pretty simple, shall we just copy/paste the fn? 1 less npm dependency to get shai-hulu'd 😅 

https://github.com/ljharb/shell-quote/blob/main/quote.js

```ts
function quote(input: string) {
  return input
    .map(function (s) {
      if (s === "") {
        return "''";
      }
      if (s && typeof s === "object") {
        return s.op.replace(/(.)/g, "\\$1");
      }
      if (/["\s\\]/.test(s) && !/'/.test(s)) {
        return "'" + s.replace(/(['])/g, "\\$1") + "'";
      }
      if (/["'\s]/.test(s)) {
        return '"' + s.replace(/(["\\$`!])/g, "\\$1") + '"';
      }
      return String(s).replace(
        /([A-Za-z]:)?([#!"$&'()*,:;<=>?@[\\\]^`{|}])/g,
        "$1\\$2",
      );
    })
    .join(" ");
}

```

internal/backend/runtime/omni/migration/migrations.go
```diff
@@ -261,26 +261,40 @@ func createIdentityLastActiveForExistingIdentities(ctx context.Context, st state
...
-
-	list, err := st.List(ctx, kind)
-	if err != nil {
-		return err
-	}
+func moveSchematicCacheToEphemeral(_ context.Context, _ state.State, _ *zap.Logger, _ migrationContext) error {
+	// No-op: superseded by dropSchematicResource, which removes the Schematic
+	// resource type from the state entirely.
+	return nil
+}
```
Slessi, line 268: I understand the decision here, but is there a reason we need it this way? For one-time migration operations performance isn't the top priority

As a general practice modifying old migrations and having to handle multiple cases in the new migration (i.e. did or didnt run old migration) feels like just adding chances for mistakes

frontend/src/views/InstallationMedia/Confirmation.vue
```diff
@@ -124,13 +132,37 @@ const schematicId = computed(() => schematic.value?.id ?? '')
 
 const { links } = usePresetDownloadLinks(schematicId, resolvedPreset)
 
-const factoryUrl = computed(() => features.value?.spec.image_factory_base_url)
+const factoryHost = computed(() => {
+  const baseURL = features.value?.spec.image_factory_base_url
+  if (!baseURL) return ''
+
+  try {
+    return new URL(baseURL).host
```
Slessi: Was it wrong before to use full URL instead of just host?

internal/backend/runtime/omni/controllers/omni/internal/imagefactoryauth/imagefactoryauth.go
```diff
@@ -0,0 +1,58 @@
...
+// IsSupported reports whether the given Talos version supports the
+// RegistryAuthConfig multi-doc config document.
+func IsSupported(talosVersion string) bool {
+	v, err := semver.Parse(strings.TrimLeft(talosVersion, "v"))
+	if err != nil {
+		return false
+	}
+
+	return v.GTE(minVersionRegistryAuthConfig)
+}
```
Slessi: shouldnt stuff like this be part of machinery quirks?

internal/pkg/config/config.go
Slessi: For schema you can add to `schema.json` [dependentRequired](https://json-schema.org/understanding-json-schema/reference/conditionals#dependentRequired) like i did for `EulaAccept` to make validate in schema that its both or none

For `cmd.go` you can have `MarkFlagsRequiredTogether` which does the same for CLI flags

Then I think you shouldn't need this line

internal/pkg/features/features.go
Slessi: I think at least reader role should be required, there is no reason why someone without a role would ever need to read this information right?

internal/pkg/config/config.go
utkuozdemir: `MarkFlagsRequiredTogether` might not even be necessary.

Slessi: arent they separate flows? if you are using CLI flags vs the schema json? or is it all validated against schema json in the end? i added both for EULA, i feel like at the time i tried it and realised i needed both, but maybe i am remembering wrong

oguzkilcan: `MarkFlagsRequiredTogether` doesn't produce a nice user readable error. We should implement that check ourselves and return a nice error to users for better UX.

Slessi: Seems ok to me

```
2026/05/18 11:20:27 failed to run Omni: if any flags in the group [eula-accept-name eula-accept-email] are set they must all be set; missing [eula-accept-email]
```

frontend/src/views/InstallationMedia/Confirmation.vue
Unix4ever: yeah

internal/pkg/features/features.go
Unix4ever: yeah, will change that to reader

client/pkg/omni/resources/virtual/image_factory_auth.go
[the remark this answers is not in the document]
Unix4ever: removed

frontend/src/views/InstallationMedia/Confirmation.vue
Unix4ever: yes

frontend/src/views/InstallationMedia/Confirmation.vue
Unix4ever: if we add `!` or `$` to the password the shell command won't be valid, which will be annoying, so we should escape. I'm fine to drop the extra dependency, will do that

client/pkg/omni/resources/omni/schematic.go
Unix4ever: We can't do that. How would we calculate the cache ID if it differs locally? We can probably cache it using locally calculated ID without the extra return args, but that feels wrong.

internal/pkg/features/features.go
Unix4ever: Feels like UI will break though if I change it to Reader

utkuozdemir: Maybe let's introduce a new resource? I think it makes sense.

frontend/src/methods/useImageFactoryAuth.ts
Unix4ever: yeah, lets not suppress it actually

client/pkg/omni/resources/omni/schematic.go
utkuozdemir: I was thinking sth like:
- do a local calculation, get an id
- check if the cache has an entry. if yes, don't hit the factory
- if not, hit the factory, store the **returned id and normalized schematic** yaml into the cache.
- never put what we locally compute or marshal into the cache. always put what factory returns.
- in a migration, drop the existing cache (delete all existing cache), as we cannot guarantee their correctness. the new entries in the cache would work like I described above.

(just noticed I also described it in https://github.com/siderolabs/omni/issues/2797)

IMO the cache would work just fine if we did this.

internal/backend/runtime/omni/migration/migrations.go
utkuozdemir: It might be because a type (resource) is deleted, so it doesn't compile anymore. we had that in the past. downside of migrations being typed.

internal/pkg/features/features.go
Slessi: isnt it only part `ImageFactoryAuthSpec` in virtual? I saw it was part of this features `Params` thing but it wasnt included lower then when assigning the typed spec stuff so it shouldnt be returned to frontend anyway, no?

frontend/src/views/InstallationMedia/Confirmation.vue
Unix4ever: the function feels wrong though :)

Unix4ever: more like this I guess:
```
const quote = (input: string) => {
  return [...input]
    .map(function (s) {
      if (/["\s\\]/.test(s) && !/'/.test(s)) {
        return "'" + s.replace(/(['])/g, "\\$1") + "'";
      }

      if (/["'\s]/.test(s)) {
        return '"' + s.replace(/(["\\$`!])/g, "\\$1") + '"';
      }

      return String(s).replace(
        /([A-Za-z]:)?([#!"$&'()*,:;<=>?@[\\\]^`{|}])/g,
        "$1\\$2",
      );
    })
    .join('');
}
```

client/pkg/omni/resources/omni/schematic.go
Unix4ever: If we put `returned id` to the cache as the key we won't hit that cache when we compute the key locally as it's going to be different, that's the problem.

internal/pkg/features/features.go
Unix4ever: yeah, actually the place you pointed out needs to be updated. There shouldn't be these fields there.
And yes, just making the `ImageFactoryAuthSpec` only accessible by reader will work alright everywhere.

internal/backend/runtime/omni/migration/migrations.go
Unix4ever: yeah, the resource will be gone.

internal/backend/runtime/omni/controllers/omni/internal/imagefactoryauth/imagefactoryauth.go
Unix4ever: won't hurt to move it there.

client/pkg/omni/resources/omni/schematic.go
Unix4ever: I understand that this way it will keep working for the IF without the Auth, but that's until we add something that gets populated on IF side for the community version of it. 
I don't really want it to have 2 different flows here.

internal/backend/runtime/omni/migration/migrations.go
Slessi: aha... ok, i get it, unfortunate

frontend/src/views/InstallationMedia/Confirmation.vue
Slessi: yea i copied it wrong, it was in js and i added `: string` but actually its `: string[]` - the array args input, not an individual string, my bad, for one string their version is below, first two ifs not needed indeed

```ts
function quote(s: string) {
      if (s === "") {
        return "''";
      }
      if (s && typeof s === "object") {
        return s.op.replace(/(.)/g, "\\$1");
      }
      if (/["\s\\]/.test(s) && !/'/.test(s)) {
        return "'" + s.replace(/(['])/g, "\\$1") + "'";
      }
      if (/["'\s]/.test(s)) {
        return '"' + s.replace(/(["\\$`!])/g, "\\$1") + '"';
      }
      return String(s).replace(
        /([A-Za-z]:)?([#!"$&'()*,:;<=>?@[\\\]^`{|}])/g,
        "$1\\$2",
      );
    }
```

client/pkg/omni/resources/omni/schematic.go
utkuozdemir: Ok, fair, didn't have a strong opinion there actually. If we hit issues due to IF being not available or something, we can reintroduce something later.

Just I guess this PR also will close #2797 then, right :)

utkuozdemir: > If we put `returned id` to the cache as the key we won't hit that cache when we compute the key locally as it's going to be different, that's the problem.

BTW, it won't be different if factory doesn't add anything or change indentation and so on. When we use the factory client library and set all fields we know, it should, in many cases, be equal to what factory would return IMO.

One thing: if we don't put the `owner` field, factory will do, and the resulting ID will always be different. But I think we should probably always set `owner` field explicitly (it's the registry username).

client/pkg/infra/imagefactory/client.go
```diff
@@ -28,15 +28,15 @@ type ClientOptions struct {
...
+		err         error
+	)
 
 	callCtx, cancel := context.WithTimeout(ctx, 10*time.Second)
 	defer cancel()
 
-	if _, err = se.client.SchematicCreate(callCtx, schematic); err != nil {
+	if schematicID, _, err = se.client.SchematicCreate(callCtx, schematic); err != nil {
 		return "", fmt.Errorf("failed to create schematic: %w", err)
 	}
```
Slessi: just drop the inline `if err` imo

```suggestion
callCtx, cancel := context.WithTimeout(ctx, 10*time.Second)
defer cancel()

	schematicID, _, err := se.client.SchematicCreate(callCtx, schematic)
	if err != nil {
return "", fmt.Errorf("failed to create schematic: %w", err)
}
```

Slessi (approved):
LGTM! just left one nit

client/pkg/omni/resources/omni/schematic.go
Unix4ever: We would also have to set the RPI4 overlay config then. And any other stuff we might not yet support in the IF.

PR #2848, opened 2026-05-15
fix: wrong role promotion for some etcd APIs

### LINKED ISSUE #2846: Allow `talosctl etcd forfeit-leadership`
### Problem Description

During some Kubernetes resource churn, I discovered the Kubernetes etcd would benefit from a defrag. I was able to use `talosctl etcd defrag` on the two follower nodes without issue. I could not run `talosctl etcd forfeit-leadership` to move cluster leadership and defrag the remaining node, and was forced to defrag the leader directly, which is a blocking operation and not ideal.

### Solution

Allow the `talosctl etcd forfeit-leadership` API call for the `ops` role.

### Alternative Solutions

Add an Omni-managed defrag API which will take care of defragging etcd of the control plane nodes in a safe and reliable manner.

### Notes

_No response_

smira:
These APIs require `os:admin` on Talos side, so they should be moved to appropriate set of methods.

### DIFF
--- a/internal/backend/grpc/router/talos_backend.go
+++ b/internal/backend/grpc/router/talos_backend.go
@@ -35,10 +35,6 @@ var operatorMethodSet = xslices.ToSet([]string{
 	machine.MachineService_EtcdAlarmDisarm_FullMethodName,
 	machine.MachineService_EtcdAlarmList_FullMethodName,
 	machine.MachineService_EtcdDefragment_FullMethodName,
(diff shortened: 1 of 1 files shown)


smira:
Reference: https://github.com/siderolabs/talos/blob/fbcc5bfbf004cec836c5718f5aae54d36315ce0e/internal/app/machined/pkg/system/services/machined.go#L68-L81

shanduur (approved):

frezbo (approved):

PR #2858, opened 2026-05-19
feat:  Expose EC2 Tags as Machine Labels

### LINKED ISSUE #1059: [feature] Expose EC2 Labels as Machine Labels
### Problem Description

Streamline the usage of MachineClasses using my existing toolchain.

### Solution

I would like to be able to use my IaC to set labels on my ec2 instances and use those labels to power MachineClasses in Omni. 

### Alternative Solutions

I would prefer to use my IaC to set labels than to roll custom AMIs for every node type/tenant combo.

### Notes

_No response_

majabojarska:
Exposes Talos `PlatformMetadata` tags as user-editable (removable) machine labels in Omni.
- The labels based on `PlatformMetadata` tags are populated only once, when the machine joins. Once that succeeds, we set the `PlatformTagLabelsInitialized` annotation on the corresponding `MachineStatus` and `MachineLabels` resources, which prevent this reconciliation from repeating ever again.
    - A reboot will not recreate these labels.
    - Changing the `PlatformMetadata` tags (e.g. in EC2) will not be surfaced as labels in Omni if the machine already joined. 
    - The goal of annotating both is to have a failsafe if any of the resource writes fails. For example, if writing the annotation to `MachineStatus` fails, we still have the annotation on `MachineLabels` to catch us, preventing further reconciliation. Mind that it is the `MachineStatus` controller that modifies `MachineStatus` (and `MachineLabels`).
- If there's a label key conflict between custom user labels and labels based on PlatformMetadata tags, the custom user labels take priority. My assumption is that the user has to take intentional action to set the custom user label, and hence override a pre-existing label. This is similar to how `imageLabels` are handled ([src](https://github.com/majabojarska/omni/blob/ddfa70a9cf2af9028aaaeeb82df7fea0e3399ed6/internal/backend/runtime/omni/controllers/omni/internal/task/machine/poll.go#L427-L448)).  

This change is effective not only for EC2 instance tags, but for any platform that exposes `tags` through `PlatformMetadata`.

---

I've developed this under the local Omni compose stack, with a rigged Talos build ([diff](https://github.com/siderolabs/talos/commit/73200bd3736eab82d880e4071d4bef1b53163245)), to return mocked `PlatformMetadata` `tags` on _metal_.

```sh
omnictl get machinestatus -o yaml | yq '.spec.platformmetadata.tags'

[WARN] omnictl version "1.7.3" differs from the backend version "1.8.0-beta.1-16-gb469ce73".
usertag1: value1
usertag2: value2
usertag3emptyval: ""
---
# SNIP
```

The tag propagation logic is already in place for the aws provider ([src](https://github.com/majabojarska/talos/blob/3cc6c121ec37c4a920aadad5e29ec9247f7cd8f8/internal/app/machined/pkg/runtime/v1alpha1/platform/aws/aws.go#L260)), for example:
```sh
talosctl -n <publicly-routable-ipv4> get --insecure platformmetadata -oyaml | yq '.spec.tags'
Name: majabojarska-talos-aws
TagNoValue: ""
TagWithVal1: Val1
TagWithVal2: Val2
```

### DIFF
--- a/client/pkg/omni/resources/omni/machine_status.go
+++ b/client/pkg/omni/resources/omni/machine_status.go
@@ -113,10 +113,10 @@ func setLabelOptional(labels *resource.Labels, key string, valueFunc func() opti
 	}
 }
 
-// MachineStatusReconcileLabels builds a set of labels based on hardware/meta information.
+// ReconcileMachineStatusLabels builds a set of labels based on hardware/meta information.
 //
 //nolint:gocognit
-func MachineStatusReconcileLabels(machineStatus *MachineStatus) {
+func ReconcileMachineStatusLabels(machineStatus *MachineStatus) {
 	labels := machineStatus.Metadata().Labels()
 
 	setLabel(labels, MachineStatusLabelCores, func() string {
--- a/internal/backend/runtime/omni/controllers/omni/machine_status.go
+++ b/internal/backend/runtime/omni/controllers/omni/machine_status.go
@@ -118,6 +118,10 @@ func (ctrl *MachineStatusController) Settings() controller.QSettings {
 				Kind: controller.OutputExclusive,
 				Type: omni.MachineStatusType,
 			},
+			{
+				Kind: controller.OutputShared,
+				Type: omni.MachineLabelsType,
+			},
 		},
 		Concurrency: optional.Some[uint](4),
 		RunHook: func(ctx context.Context, _ *zap.Logger, r controller.QRuntime) error {
@@ -268,7 +272,9 @@ func (ctrl *MachineStatusController) reconcileRunning(ctx context.Context, r con
 
 		helpers.CopyUserLabels(m, ctrl.mergeLabels(m, inputs.machineLabels))
 
-		ctrl.reconcileLabels(m, inputs.infraMachineStatus)
+		if errLabels := ctrl.reconcileLabels(m, inputs.infraMachineStatus); errLabels != nil {
+			return errLabels
+		}
 
 		if err = ctrl.setClusterRelation(inputs, m); err != nil {
 			return err
@@ -330,8 +336,8 @@ func (ctrl *MachineStatusController) reconcileTearingDown(ctx context.Context, r
 
 //nolint:gocyclo,cyclop,gocognit
 func (ctrl *MachineStatusController) handleNotification(ctx context.Context, r controller.QRuntime, event machinetask.Info) error {
-	if err := safe.WriterModify(ctx, r, omni.NewMachineStatus(event.MachineID), func(m *omni.MachineStatus) error {
-		spec := m.TypedSpec().Value
+	if err := safe.WriterModify(ctx, r, omni.NewMachineStatus(event.MachineID), func(machineStatus *omni.MachineStatus) error {
+		spec := machineStatus.TypedSpec().Value
 
 		if event.LastError != nil {
 			spec.LastError = event.LastError.Error()
@@ -406,7 +412,7 @@ func (ctrl *MachineStatusController) handleNotification(ctx context.Context, r c
 		if event.ImageLabels != nil {
 			spec.ImageLabels = event.ImageLabels
 
-			m.Metadata().Labels().Do(func(temp kvutils.TempKV) {
+			machineStatus.Metadata().Labels().Do(func(temp kvutils.TempKV) {
 				for key, value := range spec.ImageLabels {
 					temp.Set(key, value)
 				}
@@ -414,9 +420,9 @@ func (ctrl *MachineStatusController) handleNotification(ctx context.Context, r c
 		}
 
 		if event.NoAccess {
-			m.Metadata().Labels().Set(omni.MachineStatusLabelInvalidState, "")
+			machineStatus.Metadata().Labels().Set(omni.MachineStatusLabelInvalidState, "")
 		} else {
-			m.Metadata().Labels().Delete(omni.MachineStatusLabelInvalidState)
+			machineStatus.Metadata().Labels().Delete(omni.MachineStatusLabelInvalidState)
 		}
 
 		if event.SecurityState != nil {
@@ -424,14 +430,24 @@ func (ctrl *MachineStatusController) handleNotification(ctx context.Context, r c
 		}
 
 		if event.Schematic != nil {
-			if err := ctrl.handleEventSchematic(ctx, m, event); err != nil {
+			if err := ctrl.handleEventSchematic(ctx, machineStatus, event); err != nil {
 				return fmt.Errorf("failed to handle schematic event: %w", err)
 			}
 		}
 
 		spec.Maintenance = event.MaintenanceMode
 
-		ctrl.reconcileLabels(m, event.InfraMachineStatus)
+		if err := ctrl.reconcileLabels(machineStatus, event.InfraMachineStatus); err != nil {
+			return fmt.Errorf("failed to reconcile labels: %w", err)
+		}
+
+		// Intentionally do this separately from the usual labels reconciliation, as these
+		// are based only on PlatformMetadata tags.
+		if machineStatus.TypedSpec().Value.PlatformMetadata != nil {
+			if err := ctrl.reconcilePlatformTagLabels(ctx, r, machineStatus); err != nil {
+				return err
+			}
+		}
 
 		return nil
 	}); err != nil && !state.IsPhaseConflictError(err) {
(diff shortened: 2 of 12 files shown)


rothgar:
I just want to make sure I understand this PR. ec2 labels will show up in Omni as machine labels. Is that right? I'm looking forward to this.

majabojarska:
> I just want to make sure I understand this PR. ec2 labels will show up in Omni as machine labels. Is that right? I'm looking forward to this.

Yes, EC2 ~labels~ **[tags](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/Using_Tags.html)** will show up as machine labels. They will be populated when the machine joins Omni, won't be automatically re-added later on. They will behave like custom user labels (modifiable).

Here's an example, the custom tags were populated automatically when the machines joined, from their PlatformMetadata `tags`.
image

client/pkg/omni/resources/omni/machine_status.go
```diff
@@ -113,10 +113,10 @@ func setLabelOptional(labels *resource.Labels, key string, valueFunc func() opti
 	}
 }
 
-// MachineStatusReconcileLabels builds a set of labels based on hardware/meta information.
+// ReconcileMachineStatusLabels builds a set of labels based on hardware/meta information.
 //
 //nolint:gocognit
-func MachineStatusReconcileLabels(machineStatus *MachineStatus) {
+func ReconcileMachineStatusLabels(machineStatus *MachineStatus) {
 	labels := machineStatus.Metadata().Labels()
```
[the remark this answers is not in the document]
majabojarska: We're the external consumers and the references are updated.

internal/backend/runtime/omni/controllers/omni/machine_status.go
```diff
@@ -532,8 +575,14 @@ func (ctrl *MachineStatusController) handleEventSchematic(ctx context.Context, m
 // This is because if there is a matching infra.MachineStatus, it means that the machine is managed by a static infrastructure provider (e.g., the bare-metal provider),
 // and it might be powered off by the provider. In that case, the block disks information on the MachineStatus might be not up to date,
 // and the information on infra.MachineStatus will be more reliable.
-func (ctrl *MachineStatusController) reconcileLabels(machineStatus *omni.MachineStatus, infraMachineStatus *infra.MachineStatus) {
-	omni.MachineStatusReconcileLabels(machineStatus)
+func (ctrl *MachineStatusController) reconcileLabels(ctx context.Context, r controller.QRuntime, machineStatus *omni.MachineStatus, infraMachineStatus *infra.MachineStatus) error {
```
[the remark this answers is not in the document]
majabojarska: Updated the remaining mentions.

internal/backend/runtime/omni/controllers/omni/machine_status_test.go
```diff
@@ -250,81 +251,164 @@ func (suite *MachineStatusSuite) TestMachineUserLabels() {
...
+	suite.Require().NoError(err)
 
+	rtestutils.AssertResources(ctx, suite.T(), suite.state, []string{testID}, func(status *omni.MachineStatus, assert *assert.Assertions) {
+		val, ok := status.Metadata().Labels().Get("imageLabel1")
+		assert.Truef(ok, "imageLabel1 doesn't exist")
+		assert.EqualValues("gasp", val)
+
+		_, ok = status.Metadata().Labels().Get("platformMetadataTag1")
+		assert.Falsef(ok, "platformMetadataTag1 should not be set after it's removed from MachineLabels")
+	})
```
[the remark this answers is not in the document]
majabojarska: This is correct, fixed. The test would've verified the functionality just fine until label removal, where it would check for non-existence of a different key, which would pass even if the create key was still present.

internal/backend/runtime/omni/controllers/omni/machine_status.go
```diff
@@ -458,6 +466,41 @@ func (ctrl *MachineStatusController) handleNotification(ctx context.Context, r c
...
+// We only populate these labels on Omni join; from there on the user can freely modify them.
+func (ctrl *MachineStatusController) reconcilePlatformTagLabels(ctx context.Context, r controller.QRuntime, machineStatus *omni.MachineStatus) error {
+	if _, isAlreadyInitialized := machineStatus.Metadata().Annotations().Get(omni.PlatformTagLabelsInitialized); isAlreadyInitialized {
+		return nil
+	}
+
+	tags := machineStatus.TypedSpec().Value.PlatformMetadata.GetTags()
+	if len(tags) == 0 {
+		machineStatus.Metadata().Annotations().Set(omni.PlatformTagLabelsInitialized, "")
+
```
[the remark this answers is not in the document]
majabojarska: That's intended. We act on whatever `tags` are included with the first `PlatformMetadata` update. Later edits to EC2 instace tags (or other platform equivalent) should have no effect.

internal/backend/runtime/omni/controllers/omni/machine_status.go
```diff
@@ -458,6 +466,41 @@ func (ctrl *MachineStatusController) handleNotification(ctx context.Context, r c
...
+		return nil
+	}
+
+	if err := safe.WriterModify(ctx, r, omni.NewMachineLabels(machineStatus.Metadata().ID()), func(machineLabels *omni.MachineLabels) error {
+		for k, v := range tags {
+			// Do not overwrite pre-existing user labels.
+			if _, ok := machineLabels.Metadata().Labels().Get(k); !ok {
+				machineLabels.Metadata().Labels().Set(k, v)
+			}
+		}
```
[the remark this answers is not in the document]
majabojarska: System-prefixed labels should not originate from `PlatformMetadata` tags. The current implementation is right to effectively ignore them, if present.

TimJones:
I'm not sure I agree with the labels being user-modifiable and not reflect the actual `PlatformMatadata`, IMO they should be a reflection of the tags in `PlatformMetada` rather than a shallow one-time copy.

I'm still happy for them to be filtered up into Omni though. Yey!

smira:
> I'm not sure I agree with the labels being user-modifiable and not reflect the actual `PlatformMatadata`, IMO they should be a reflection of the tags in `PlatformMetada` rather than a shallow one-time copy.

We discussed this, IMDS doesn't support watch interface, so Talos can't reflect tags as they are being changed.

internal/backend/runtime/omni/controllers/omni/machine_status.go
```diff
@@ -118,6 +118,10 @@ func (ctrl *MachineStatusController) Settings() controller.QSettings {
 				Kind: controller.OutputExclusive,
 				Type: omni.MachineStatusType,
 			},
+			{
+				Kind: controller.OutputShared,
+				Type: omni.MachineLabelsType,
+			},
```
Unix4ever, line 124: It's a bit scary to have this resource as the output here, as it's also used as an input in this controller, but I can't propose a better solution from the top of my head.
I guess it should be fine to have it this way as we gate it by the annotation.

internal/backend/runtime/omni/controllers/omni/machine_status.go
```diff
@@ -458,6 +466,41 @@ func (ctrl *MachineStatusController) handleNotification(ctx context.Context, r c
...
+	}
+
+	tags := machineStatus.TypedSpec().Value.PlatformMetadata.GetTags()
+	if len(tags) == 0 {
+		machineStatus.Metadata().Annotations().Set(omni.PlatformTagLabelsInitialized, "")
+
+		return nil
+	}
+
+	if err := safe.WriterModify(ctx, r, omni.NewMachineLabels(machineStatus.Metadata().ID()), func(machineLabels *omni.MachineLabels) error {
```
Unix4ever: I think safer option would be to also set `omni.PlatformTagLabelsInitialized` annotation on the `MachineLabels` resource.
Then even if we fail to update `MachineStatus` the next tags update won't cause `MachineLabels` update.

It makes it more atomic this way.
I think we can even move this `omni.PlatformTagLabelsInitialized` label here. 
The only issue it can cause is that as the user can delete `MachineLabels` and they might get re-populated from tags.

Unix4ever (approved):

internal/backend/runtime/omni/controllers/omni/machine_status.go
majabojarska: > The only issue it can cause is that as the user can delete MachineLabels and they might get re-populated from tags.

I've implemented your suggestion to annotate `MachineLabels` instead of `MachineStatus` (separate branch). Since the annotation's lifetime is bound to the `MachineLabels` resource, the machine's labels get repopulated from instance tags as soon as the last label is deleted.  I'm not convinced that's the desired behavior :thinking:.

The upside of annotating `MachineStatus` is that the annotation is removed only during machine deletion, at which point the annotation is not needed anymore, and so no accidental tag-based label repopulation can happen.

majabojarska: Here's the diff for reference: https://github.com/majabojarska/omni/commit/293f0ff60aeb6e5b70d9408f4e7e69047e01bd31

internal/backend/runtime/omni/controllers/omni/machine_status.go
majabojarska: I tried to avoid it, but `PlatformMetadata` arrives over a `machinetask.Info` channel, to which the `MachineStatusController` listens, and so that's the controller that needs to consume/use it first, in order to affect anything else (or at least that's my current understanding).

---

On another note, while analyzing the current implementation, I've noticed we could reduce the overhead slightly by factoring out `ctrl.reconcilePlatformTagLabels`  from `ctrl.reconcileLabels` ([src](https://github.com/majabojarska/omni/blob/88e7c1c59e770fc350f98ba9bdf56e2b5c541fbc/internal/backend/runtime/omni/controllers/omni/machine_status.go?plain=1#L582-L584)). My reasoning is that `reconcilePlatformTagLabels` can reconcile in any meaningful way only when `event.PlatformMetadata` is received, which can only happen in the `handleNotification` flow ([src](https://github.com/majabojarska/omni/blob/88e7c1c59e770fc350f98ba9bdf56e2b5c541fbc/internal/backend/runtime/omni/controllers/omni/machine_status.go?plain=1#L404-L406)), but not in the `reconcileRunning` flow (no PlatformMetadata change possible). Reconciling tag-based labels could happen [here](https://github.com/majabojarska/omni/blob/88e7c1c59e770fc350f98ba9bdf56e2b5c541fbc/internal/backend/runtime/omni/controllers/omni/machine_status.go?plain=1#L443), either before or after `ctrl.ReconcileLabels` (either order of operations should work).

This should save us some cycles on checking whether the annotation is set.

internal/backend/runtime/omni/controllers/omni/machine_status.go
Unix4ever: > I've implemented your suggestion to annotate `MachineLabels` instead of `MachineStatus` (separate branch). Since the annotation's lifetime is bound to the `MachineLabels` resource, the machine's labels get repopulated from instance tags as soon as the last label is deleted. I'm not convinced that's the desired behavior 🤔.

Yeah. I've mentioned that as well. Adding the annotations to both resources is more reliable.

oguzkilcan (approved):

internal/backend/runtime/omni/controllers/omni/machine_status.go
majabojarska: Yup, just wanted to explore that idea.

majabojarska: Reworked the implementation to:
- Set the annotation on both `MachineLabels` and `MachineStatus`. If either is set, we skip platform-tag-based label reconciliation.
- If just one of the resources has the annotation set, and the other doesn't, we ensure eventual consistency by setting the missing annotation on a subsequent reconciliation run.

internal/backend/runtime/omni/controllers/omni/machine_status.go
majabojarska: Refactored to reflect the above. `reconcilePlatformTagLabels` runs only when we receive `event.PlatformMetadata` (still gated by the annotation).

PR #2865, opened 2026-05-20
chore(hack/compose): Add Delve debugger for Omni

majabojarska:
Adds the option to enable the [Delve](https://github.com/go-delve/delve) debugger with the `hack/compose/docker-compose.yaml` stack.

- Leverages https://github.com/siderolabs/kres/pull/655
- Enablement is controlled through the existing `WITH_DEBUG=true` env var.
    - Binds by default to `0.0.0.0:12345`, but this is customizable.
    - Starts Omni right away, without waiting for the debugger to attach. The default behavior was to halt and wait for debug clients, which I think is too heavy-handed.
- `WITH_DEBUG=true` is now the recommended way to toggle debug mode. We had `WITH_DEBUG=1` in some places (minority). While Go code can handle the two variants gracefully through `strconv.ParseBool`, Makefiles and shell scripts become overcomplicated. All docs and sources were updated to recommend `true` as the enabling value.
- Includes a basic VS Code launch config.

---

### How do test this?

From Omni repo root dir:
```sh
make docker-compose-up WITH_DEBUG=true
```

From VS Code:
1. Set a breakpoint on [ResourceServer.Get:119](https://github.com/majabojarska/omni/blob/170246b6e19ec180682e88b76b0c34311e221006/internal/backend/grpc/resource.go#L119). 
2. Run the `Attach to Omni (Delve)` launch target. 
3. You should see the debug view now, and Omni runtime logs emitted in the `local-omni` container.
4. Open the locally-running Omni UI. This will attempt to fetch resources and execute the logic on which we've set the breakpoint in step 2.
5. Assert the breakpoint was triggered and the debugger stopped.

Now restart the compose stack without debug mode (`WITH_DEBUG` is `false` by default).
```sh
make docker-compose-down && make docker-compose-up
```

Delve should not be listening anymore:
```sh
sudo netstat -plntu | grep dlv
# Should be empty
```


hack/compose/launcher/go.mod
```diff
@@ -0,0 +1,3 @@
+module launcher
```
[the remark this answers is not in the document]
majabojarska: Removed `go.mod` altogether.

hack/compose/docker-compose.yml
```diff
@@ -4,25 +4,46 @@
...
 
 services:
+  dlv-init:
+    image: golang:1.26-alpine
+    restart: on-failure
+    volumes:
+      - dlv-tools:/output
+      - ./launcher:/launcher:ro
+    command: >-
+      sh -c "test -f /output/launcher || (cd /launcher && CGO_ENABLED=0 GOOS=linux go build -o /output/launcher .); [ '${WITH_DEBUG:-false}' != '1' ] || test -f /output/dlv || (CGO_ENABLED=0 go install github.com/go-delve/delve/cmd/dlv@latest && cp /go/bin/dlv /output/)"
```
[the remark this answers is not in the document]
majabojarska: Pinned to latest version-specific tag.

hack/compose/docker-compose.yml
```diff
@@ -4,25 +4,46 @@
...
+      - dlv-tools:/debug:ro
     container_name: local-omni
     restart: on-failure
     cap_add:
       - NET_ADMIN
+      - SYS_PTRACE
+    security_opt:
+      # Delve requires unconfined seccomp to allow ptrace and other syscalls for debugging.
+      - seccomp:unconfined
+    entrypoint:
```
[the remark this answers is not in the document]
majabojarska: `seccomp:unconfined` was in fact not necessary, removed.

DEVELOPMENT.md
```diff
@@ -133,6 +133,34 @@ Node.js development server can be used to get immediate feedback on frontend cha
...
 
+## Delve Debugger
+
+Passing `WITH_DEBUG=1` starts Omni under the [Delve](https://github.com/go-delve/delve) headless debug server.
+Without it, Omni runs directly with no debugger attached.
+
+`WITH_DEBUG=1` also:
+
+- keeps the symbol table (the default build strips it with `-s`)
+- disables compiler optimizations and inlining (`-gcflags=all=-N -l`) for clean step-through debugging
```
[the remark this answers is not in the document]
majabojarska: Any truthy value works, as we use `strconv.ParseBool`.

Dockerfile
utkuozdemir, line 1: Just in case you missed: Dockerfile is also a generated file by kres, and should not be edited manually.
If you need changes in it, you need to 
1. make https://github.com/siderolabs/kres capable of this customization instead
2. get your kres change merged to its `main`
3. come back here and do `make rekres`

Same for `.gitignore`.

Dockerfile
```diff
@@ -567,7 +648,16 @@ ARG GO_LDFLAGS
...
 ARG SHA
 ARG TAG
-RUN --mount=type=cache,target=/root/.cache/go-build,id=omni/root/.cache/go-build --mount=type=cache,target=/go/pkg,id=omni/go/pkg GOARCH=arm64 GOOS=linux go build ${GO_BUILDFLAGS} -ldflags "${GO_LDFLAGS} -X ${VERSION_PKG}.Name=omnictl -X ${VERSION_PKG}.SHA=${SHA} -X ${VERSION_PKG}.Tag=${TAG}" -o /omnictl-linux-arm64
+ARG WITH_DEBUG=false
+RUN --mount=type=cache,target=/root/.cache/go-build,id=omni/root/.cache/go-build \
+    --mount=type=cache,target=/go/pkg,id=omni/go/pkg \
+    GCFLAGS=""; \
+    case "$WITH_DEBUG" in true|1) GCFLAGS="all=-N -l" ;; esac; \
+    GOARCH=arm64 GOOS=linux go build \
+        -gcflags="$GCFLAGS" \
```
smira: +1 to Utku, this should be done in the Makefile to populate GO_BUILDFLAGS based on WITH_DEBUG and done in kres

utkuozdemir:
One question about usage of multiple commits in the PR: are you using them for organizing your own work, and intend to squash before merging?

Multiple commits in a PR is fine, but we prefer all commits on main to be atomic and complete, so if a single unit of work spans to multiple commits, we'd rather have them as a single squashed commit in main.

I'd say the work in this PR would be neater as a single commit (but no pressure).

If you intend to squash later, you need to do it locally using `git`, not use the **Squash and Merge** button of GitHub, since we sign commits ourselves.

smira:
And one small nit: this is definitely not `feat:`, as it brings nothing to the user, should be `chore:`.

majabojarska:
> One question about usage of multiple commits in the PR: are you using them for organizing your own work, and intend to squash before merging?
> 
> Multiple commits in a PR is fine, but we prefer all commits on main to be atomic and complete, so if a single unit of work spans to multiple commits, we'd rather have them as a single squashed commit in main.
> 
> I'd say the work in this PR would be neater as a single commit (but no pressure).
> 
> If you intend to squash later, you need to do it locally using `git`, not use the **Squash and Merge** button of GitHub, since we sign commits ourselves.

Yes, that's what I've done we my previous PRs. The point is to avoid having to `--amend` for every single change while addressing feedback/issues.

Unix4ever:
> One question about usage of multiple commits in the PR: are you using them for organizing your own work, and intend to squash before merging?
> 
> Multiple commits in a PR is fine, but we prefer all commits on main to be atomic and complete, so if a single unit of work spans to multiple commits, we'd rather have them as a single squashed commit in main.
> 
> I'd say the work in this PR would be neater as a single commit (but no pressure).
> 
> If you intend to squash later, you need to do it locally using `git`, not use the **Squash and Merge** button of GitHub, since we sign commits ourselves.

I would also prefer it squashed. It's great to review that with incremental changes, but as the whole it looks like it can land into a single commit.

majabojarska:
`WITH_DEBUG` handling in Makefile/Dockerfile will be implemented into Kres here: https://github.com/siderolabs/kres/pull/655

hack/generate-certs/main.go
```diff
@@ -270,7 +270,7 @@ const composeTemplate = `services:
       --bind-addr {{ .BindAddr }}
       --frontend-dst http://127.0.0.1:8121
       --frontend-bind 0.0.0.0:8120
-      --debug
+      --log-level debug
```
majabojarska, line 273: The `--debug` option was logged as deprecated, so I've updated it along the way.

Dockerfile
```diff
@@ -1,9 +1,4 @@
 # syntax = docker/dockerfile-upstream:1.24.0-labs
-
-# THIS FILE WAS AUTOMATICALLY GENERATED BY KRES, PLEASE DO NOT EDIT.
```
smira, line 3: I'm not sure why it got removed? kres should put it back?

smira (approved):

Dockerfile
majabojarska: I now ran `make rekres`, and it made no difference. Then I deleted a non-comment line, ran `make rekres` and it regenerated the file, including this section 🤔

smira: Don't remove it :)

Issue #2737, opened 2026-04-24
rate limit all resource operations done by Omni


smira:
similar too #2734, but applies to all resources (including those modified by Omni itself), prevents a bug in Omni from affecting the state/etcd in a bad way

Issue #2736, opened 2026-04-24
rate limit infra provider resource operations


smira:
Allow to put a rate limit on infra provider resource operations, might be a global one, or split by modify/read operations.

PR #2869, opened 2026-05-21
feat: add per-class etcd write-bytes rate limiting

### LINKED ISSUE #2734: rate limit user-managed resource mutations
Goal: apply rate-limit on user-managed resource mutations.

Reason: etcd keeps old revisions of the KV pairs, so even "no-op" create/destroy operation which doesn't change number of "active" resources impacts etcd database size until the next compaction.

Possible limits:

* limit API rate for mutating resource operations, allow some burst, but have a steady limit (req/s)
* limit rate of mutations via bytes/sec (needs to be done on lower level) - if the operation is going to write to etcd, calculate the size of the resource and apply the rate limit (bytes/sec)

oguzkilcan:
Throttle etcd writes by payload size, with independent budgets for end users, infra providers, and internal callers. Off by default, can be enabled via storage.rateLimits.etcd.*.

Exposes four new Prometheus series, all labeled by class:
  - omni_etcd_write_throttle_wait_seconds (histogram): time spent waiting for budget
  - omni_etcd_write_throttle_admitted_total: writes that got through after throttling
  - omni_etcd_write_throttle_failed_total: writes that timed out waiting
  - omni_etcd_write_throttle_failed_bytes_total: bytes of writes that timed out

Resolves: #2734, #2737 and #2736

### DIFF
--- a/internal/backend/ratelimit/classify.go
+++ b/internal/backend/ratelimit/classify.go
@@ -1,0 +1,42 @@
+// Copyright (c) 2026 Sidero Labs, Inc.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+
+// Package ratelimit throttles COSI mutations (Create/Update/Destroy) against the
+// etcd backend by marshaled-payload size, per caller class.
+//
+// Throttler is wired as the state-etcd WithLimiter hook: each mutation blocks
+// until enough bytes-budget is available for the marshaled payload, bounded by
+// min(ctx.Deadline-now, maxWait). On timeout it returns codes.DeadlineExceeded.
+// Service accounts and unknown callers share the `user` bucket.
+package ratelimit
+
+import (
+	"github.com/siderolabs/omni/internal/pkg/auth/actor"
+)
+
+const (
+	bucketInternal      = 0
+	bucketInfraProvider = 1
+	bucketUser          = 2
+	bucketCount         = 3
+)
+
+// bucketLabels are the Prometheus `class` label values, mirroring actor.Type names.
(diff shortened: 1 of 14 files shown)


internal/backend/ratelimit/classify.go
```diff
@@ -0,0 +1,42 @@
...
+
+import (
+	"github.com/siderolabs/omni/internal/pkg/auth/actor"
+)
+
+const (
+	bucketInternal      = 0
+	bucketInfraProvider = 1
+	bucketUser          = 2
+	bucketCount         = 3
```
shanduur, line 23: Is it worth having `iota` here?

shanduur (approved):

internal/backend/ratelimit/classify.go
oguzkilcan: bucketCount is an actual number here, it looks like iota would fit here but that's just a lucky coincidence.

internal/backend/ratelimit/throttler.go
```diff
@@ -0,0 +1,175 @@
...
+		return t.Gate(ctx, marshaledBytes, resourceType)
+	}
+}
+
+func buildLimiter(perSecond, burst uint64) *rate.Limiter {
+	if perSecond == 0 || burst == 0 {
+		return nil
+	}
+
+	return rate.NewLimiter(rate.Limit(perSecond), int(burst)) //nolint:gosec
```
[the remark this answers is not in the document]
oguzkilcan: Fixed

Unix4ever (approved):

Issue #2887, opened 2026-05-26
Show vulnerability information for existing clusters


Slessi:
Have a page in Omni which displays any known vulnerabilities with Talos/K8s/other currently in use by clusters/machines etc.

PR #2888, opened 2026-05-26
feat(frontend): show vulnerability items on installation media wizard

### LINKED ISSUE #2886: Show vulnerability scan on installation media wizard
Mirror the image factory logic for this

Slessi:
On the installation media wizard confirmation page, show SBOM & VEX links as well as the vulnerability scan & modal, as in factory. The feature will only be available when using enterprise image factory.

Closes #2886

[:book: Storybook](https://68d664245076e2bb623d98f3-qmvbkorhut.chromatic.com/?path=/story/pages-authenticated-machines-installation-media-create-confirmation--default)

---

**Changes Summary**:

- Added `is_enterprise_image_factory` to `FeaturesSpec`
  - Is filled on omni load with a `HEAD` call to factory on startup to determine if enterprise
- Added `/scans/` to list of non-frontend URLs
- Added `/scans/{schematicID}/{talosVersion}/{arch}/report.json` to get the report from factory
- On confirmation page of Installation Media:
  - Added checksum SHA links for image downloads
  - Added SBOM + VEX links to the bottom
  - Added vulnerability scan to the top, with a modal
  - Inside this modal show list of CVEs and scan report download links
- Renamed some icon names to more generic names


internal/backend/imagefactory/client.go
```diff
@@ -101,3 +110,34 @@ func (cli *Client) EnsureSchematic(ctx context.Context, inputSchematic schematic
 		Data:    data,
 	}, nil
 }
+
+// IsEnterprise reports whether the connected image factory is an Enterprise instance.
+func (cli *Client) IsEnterprise(ctx context.Context) (bool, error) {
```
Slessi: I call this on load, to fill `FeaturesSpec` - but if the call fails (maybe factory is down) then your Omni doesn't boot or if I change it to ignore the error Omni boots with the wrong flag set

This sucks.

Alternatives:
- Cached virtual resource instead
- Do the `HEAD` request, but only when features are requested, and cache it there
- Intercept any factory calls to set the resource on the fly? Prevents extra requests, but value will be wrong till at least 1 factory request happens

internal/backend/services/proxy.go
```diff
@@ -60,7 +60,7 @@ func (prx *httpProxy) Run(ctx context.Context, next http.Handler, logger *zap.Lo
 	srv := NewFromConfig(
 		&prx.config,
 		http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-			if hasPrefix(r.URL.Path, "/api/", "/omnictl/", "/talosctl/", "/image/") {
+			if hasPrefix(r.URL.Path, "/api/", "/omnictl/", "/talosctl/", "/image/", "/scans/") {
```
Slessi: Feels like this list can slowly get out of control as requirements grow

I usually prefer a namespace path to house all potential future APIs, but `/api` is already used for `gRPC` - any suggestions for alternatives?

Unix4ever:
Are we going to use Talos versions or we decided to keep the component type + its version for now?

Slessi:
@Unix4ever there wasn't a full decision from what i understood so my vision is to keep this 1:1 with factory for now, as its not the most important part people care about. Then for #2887 use the APIs I created in this for the cluster view, but show Talos version info. Then we get feedback, and can apply any decisions to installation media afterwards (if any).

Unix4ever (approved):

internal/backend/imagefactory/client.go
Slessi: moved to versions controller

internal/backend/services/proxy.go
Slessi: moved under /api

PR #2894, opened 2026-05-27
fix(frontend): remove lingering test code

Slessi:
Accidentally included test code in #2892, removing here.


majabojarska (approved):
LGTM 

Image

PR #2900, opened 2026-05-28
feat(frontend): add improvements to disks view

Slessi:
Filter out empty CD-ROMs from machine disks, hide device mapper entries for encrypted volumes, fix file system display for luks partitions, add a colour for luks partitions, and show a CD-ROM label in the header.

### DIFF
--- a/frontend/src/views/Machines/MachineDisks.vue
+++ b/frontend/src/views/Machines/MachineDisks.vue
@@ -157,7 +157,14 @@ const errCode = computed(
 
 const organizedDisks = computed(() =>
   disks.value
-    .filter((d) => !d.metadata.id?.startsWith('loop'))
+    .filter(
+      (d) =>
+        !d.metadata.id?.startsWith('loop') &&
+        // Device-mapper devices (dm-N) are synthetic kernel block devices created for
+        // things like LUKS encryption. They are represented through their parent
+        // partition's VolumeStatus and should not appear as top-level disk entries.
+        !d.metadata.id?.startsWith('dm-'),
+    )
     .map((disk) => ({
       disk,
       partitions: volumes.value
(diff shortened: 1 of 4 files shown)


frezbo (approved):

frontend/src/views/Machines/MachineDisks.vue
```diff
@@ -169,7 +176,9 @@ const organizedDisks = computed(() =>
         .sort(
           (a, b) => (a.volume.spec.partition_index || 0) - (b.volume.spec.partition_index || 0),
         ),
-    })),
+    }))
+    // Hide empty CD-ROMs
+    .filter((d) => !d.disk.spec.cdrom || d.partitions.length > 0),
```
smira: I don't think I get this condition fully... should be `d.name == ""` (if it's empty CDROM), but in this negated style I'm not sure what this means

Slessi: For context, `disk` here is `Disks.block.talos.dev` and `partitions` is a mapped array of `VolumeStatuses.block.talos.dev` & `DiscoveredVolumes.block.talos.dev` from earlier in the chain which belong to that disk

The filter will keep only disks that are either not CDROMS, or CDROMS with at least 1 partition.

Is `d.name == ""` is equivalent to the partition check? If so I think the partition check is a bit more clear, name being `""` doesn't self-explain as well

smira: my point is different. empty CDROM is `d.disk.spec.cdrom && d.name == ""` 

`d.name == ""` means that Talos failed to detect any filesystem on it.

Partition check doesn't make any sense, CDROMs don't have partitions.

Slessi: changed it

smira (approved):

PR #2903, opened 2026-05-29
test: bump Talos to 1.13.3

Unix4ever:
This should improve e2e-scaling test stability even more.

### DIFF
--- a/hack/test/common.sh
+++ b/hack/test/common.sh
@@ -22,8 +22,8 @@ echo "127.0.0.1 omni.localhost" | tee -a /etc/hosts
 # Settings.
 LATEST_STABLE_OMNI=$(git tag -l --sort=-version:refname HEAD "v*" | grep -E '^v?[0-9]+\.[0-9]+\.[0-9]+$' | head -n 1)
 
-export TALOS_VERSION=1.13.2
-export KUBERNETES_VERSION=1.36.0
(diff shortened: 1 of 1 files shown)


smira (approved):
I guess it should be promoted to scaling to verify?

Slessi (approved):

hack/test/common.sh
```diff
@@ -22,7 +22,7 @@ echo "127.0.0.1 omni.localhost" | tee -a /etc/hosts
 # Settings.
 LATEST_STABLE_OMNI=$(git tag -l --sort=-version:refname HEAD "v*" | grep -E '^v?[0-9]+\.[0-9]+\.[0-9]+$' | head -n 1)
 
-export TALOS_VERSION=1.13.2
+export TALOS_VERSION=1.13.3
 export KUBERNETES_VERSION=1.36.0
```
Slessi: `1.36.1` maybe?

utkuozdemir: yep

utkuozdemir: actually, shouldn't we bump the default versions in the constants as well?

Unix4ever:
> I guess it should be promoted to scaling to verify?

I guess it should be promoted several times as the test doesn't fail every time :)

shanduur (approved):

PR #2904, opened 2026-05-29
fix: do not downgrade nodes header to single node

utkuozdemir:
When Talos API client passes the "nodes" header to Omni, always use the cluster client, not a machine client, and let Talos do its proxying, as clients in some cases rely on the returned response structure, e.g., the hostname mapping.

This fixes the regression of `talosctl dashboard` not working with worker nodes, which was introduced in c7f60c0c6f26f869e1e46caeb01c29795142dbc1.


smira (approved):

smira:
Can we get this backported please?

utkuozdemir:
> Can we get this backported please?

Yes I'll do that.

PR #2905, opened 2026-05-29
Per-machine log ingestion has no rate limit / back-pressure; one noisy machine can saturate the log store

404prefrontalcortexnotfound:
## Problem
Currently, per-machine log ingestion has no rate limit or back-pressure; one noisy machine can saturate the log store.

The existing `siderolink/ratelimit` limiter is a global, packet-layer bandwidth bucket. While effective at protecting Omni's total NIC throughput, it has two gaps regarding log ingestion:
1. It does not isolate traffic per-tenant (a single machine exhausting the bucket starves control-plane traffic globally).
2. It sits below the log-write path. 

The physical path for machine logs is `ConnHandler.HandleConn` -> `LogHandler.HandleMessage` -> `cache.WriteMessage`. The loop reads off the socket and dispatches to the per-machine SQLite store as fast as TCP delivers it, without any bounds on store I/O from a single machine.

## Implementation Proposed
This PR targets the narrow, surgical spot: `LogHandler.HandleMessage`. Since it already resolves `srcAddress` to a `machineID`, it's the natural place for a per-machine token bucket.

The implementation mirrors the conventions established by the existing packet limiter:
- A `rate.Limiter` token bucket (constructed per-machine).
- Prometheus drop metrics (`omni_machine_logs_dropped_total{reason="rate_limit"}`).
- Debounced warn-logging (`zap` with `lastLogTime`).
- Nil-limiter-means-unlimited config convention, passed in via functional options (`WithLogIngestionLimiter`).
- Sane defaults via `schema.json` configuration (`MachineLogRateLimitBytes` and `MachineLogRateBurstBytes`).

## Questions for Reviewers
I'd love to collaborate on a couple of design choices to make sure this aligns with Sidero's long-term vision:
1. **Drop vs Back-pressure:** This PR drops log lines (with a metric and warn log) rather than slowing the reader (which would apply back-pressure). This protects Omni but loses observability data from the flooding node. Is a lossy drop with metrics preferred, or should we look at a bounded buffer with back-pressure in `HandleConn` instead?
2. **Per-Machine vs Per-Account:** This PR explicitly limits *per-machine* log volume. Per-account quotas as a follow-up might be necessary for multi-machine/multi-account scenarios. Let me know if you agree with phasing per-machine back-pressure as step one.

Any guidance on these would be appreciated!

### DIFF
--- a/cmd/omni/cmd/cmd.go
+++ b/cmd/omni/cmd/cmd.go
@@ -276,6 +276,8 @@ func defineLogsFlags(rootCmd *cobra.Command, b *FlagBinder, flagConfig *config.P
 	b.IntVar("logs.machine.storage.maxLinesPerMachine", &flagConfig.Logs.Machine.Storage.MaxLinesPerMachine)
 	b.Uint64Var("logs.machine.storage.maxSize", &flagConfig.Logs.Machine.Storage.MaxSize)
 	b.Float64Var("logs.machine.storage.cleanupProbability", &flagConfig.Logs.Machine.Storage.CleanupProbability)
+	b.Uint64Var("logs.machine.ingestionRateLimitBytesPerSecond", &flagConfig.Logs.Machine.IngestionRateLimitBytesPerSecond)
+	b.Uint64Var("logs.machine.ingestionRateBurstBytes", &flagConfig.Logs.Machine.IngestionRateBurstBytes)
 
 	if err := rootCmd.Flags().MarkHidden(b.mustFlagName("logs.machine.storage.cleanupProbability")); err != nil {
 		return err
--- a/internal/pkg/siderolink/loghandler.go
+++ b/internal/pkg/siderolink/loghandler.go
@@ -9,6 +9,7 @@ import (
 	"context"
 	"fmt"
 	"net/netip"
+	"time"
 
 	"github.com/cosi-project/runtime/pkg/state"
 	"github.com/cosi-project/state-sqlite/pkg/sqlitexx"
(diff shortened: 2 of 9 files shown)


cmd/omni/cmd/cmd.go
```diff
@@ -276,6 +276,8 @@ func defineLogsFlags(rootCmd *cobra.Command, b *FlagBinder, flagConfig *config.P
 	b.IntVar("logs.machine.storage.maxLinesPerMachine", &flagConfig.Logs.Machine.Storage.MaxLinesPerMachine)
 	b.Uint64Var("logs.machine.storage.maxSize", &flagConfig.Logs.Machine.Storage.MaxSize)
 	b.Float64Var("logs.machine.storage.cleanupProbability", &flagConfig.Logs.Machine.Storage.CleanupProbability)
+	b.Uint64Var("logs.machine.ingestionRateLimitBytes", &flagConfig.Logs.Machine.IngestionRateLimitBytes)
+	b.Uint64Var("logs.machine.ingestionRateBurstBytes", &flagConfig.Logs.Machine.IngestionRateBurstBytes)
```
utkuozdemir: better to name it like "bytes per second", in other words, include the time window.

internal/pkg/siderolink/loghandler.go
```diff
@@ -159,33 +174,33 @@ func (h *LogHandler) HandleMessage(ctx context.Context, srcAddress netip.Addr, r
...
 	}
-}
 
-func (h *LogHandler) writeMessage(ctx context.Context, ip string, data []byte) error {
-	id, err := h.machineMap.GetMachineID(ip)
-	if err != nil {
-		return fmt.Errorf("failed to get machine ID for ip address %q: %w", ip, err)
+	logger = logger.With(zap.String("machine_id", string(id)))
+
+	if h.limiter != nil {
```
utkuozdemir, line 198: maybe bring `writeMessage` function back, but pass the limiter into it?

internal/pkg/siderolink/loglimiter.go
```diff
@@ -0,0 +1,132 @@
...
+
+const logLimiterDebounceInterval = time.Minute
+
+// LogIngestionLimiter limits machine log ingestion rates per machine.
+//
+// Per-machine entries are created lazily and only removed via Forget when the omni.Machine resource is destroyed or Omni instance is restarted.
+type LogIngestionLimiter struct {
+	droppedLogs *prometheus.CounterVec
+	logger      *zap.Logger
+	limiters    sync.Map // map[MachineID]*machineLimiter
```
utkuozdemir: `gen` has typed version of this (and also lazy-loading I think)

utkuozdemir (approved):

Unix4ever (approved):

PR #2907, opened 2026-05-29
fix(frontend): fix typo in disk usage bar legend text color class

immanuwell:
`text-naturalsn-n12` (extra `n`) sneaked in with the disks view improvements. tailwind ignores unknown classes so the disk legend labels inherit `text-naturals-n11` from the parent div instead of `text-naturals-n12`.

ends up with partition names and "Unallocated" being the wrong shade of gray.

to reproduce:
1. open any machine's Disks page
2. check the legend below the disk bar, its a bit off

fixed both spots in the legend section of `DiskUsageBar.vue`


Slessi:
good catch 👍

Slessi (approved):

Slessi (changes requested):
@immanuwell could you just sign the CLA? 🙏

Slessi:
Will be fixed as part of a separate PR

PR #2910, opened 2026-05-29
feat(frontend): show join tokens in saved presets list

### LINKED ISSUE #2716: [feature] show join token status in installation media screen
Currently installation media presets are configured using the join token picked during the creation of the preset. We do not have a way to show  users whether the image preset is invalid because join token is expired or revoked.

Add a new column to Installation Media UI to show join token status (Active (expiring at xx), Revoked, Expired at xxx, Deleted). It should also have the default label if the used join token is the default join token.

Slessi:
On the saved presets list for installation media, show the selected join token and any relevant badges about it.

Closes #2716 

image

### DIFF
--- a/frontend/src/pages/(authenticated)/machines/installation-media/index.stories.ts
+++ b/frontend/src/pages/(authenticated)/machines/installation-media/index.stories.ts
@@ -9,7 +9,11 @@ import { http, HttpResponse } from 'msw'
 
 import type { DeleteRequest, DeleteResponse } from '@/api/omni/resources/resources.pb'
 import type { InstallationMediaConfigSpec } from '@/api/omni/specs/omni.pb'
-import { DefaultNamespace, InstallationMediaConfigType } from '@/api/resources'
+import {
+  type JoinTokenStatusSpec,
+  JoinTokenStatusSpecState,
+} from '@/api/omni/specs/siderolink.pb.ts'
+import { DefaultNamespace, InstallationMediaConfigType, JoinTokenStatusType } from '@/api/resources'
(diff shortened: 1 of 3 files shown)


frezbo:
what's an invalid token since it's omni which creates it?

Slessi:
@frezbo it means the token specified in the preset doesnt exist, its currently just a copy of the raw token itself inside the preset

maybe you deleted it, or maybe it was added through CLI as a yaml apply, and it never existed in the first place

name could be better... wasnt sure if "deleted" was correct since it might never have existed... could be "Not found" ?

Unix4ever:
I don't like that we have 3 different ways to display that:
Automatic (%token name%)
%token name% Expired
Invalid token

Lets maybe pick a single style for that.
Maybe something like:
%token name% Automatic
%token name% Expired
%token name% Revoked
%token name% Deleted

Automatic can be green, Expired, Revoked and Deleted can be red with ! in a triangle icon.

rothgar:
What does "Automatic" mean for join tokens?

Slessi:
@rothgar isn't fixed to a specific one and will always use default

Slessi:
@Unix4ever I can't show a token name for deleted because I can't know what the token name was, and there are also tokens that are directly specified (so no label to add), so it will be this

- %token name% 
- %token name% `Automatic`
- %token name% `Expired`
- %token name% `Revoked`
- `Deleted`

> Regarding `Deleted`: Might never have existed, as far as i see we have no join token validation, so you can add bad tokens that never existed, but im ok to ignore that nuance and use deleted

I will use green automatic label for talos version also, as I copied `Automatic (*)` from there

image

Unix4ever (approved):

PR #2919, opened 2026-06-02
test: mock clock in saml test

Slessi:
Also mock the clock in SAML tests as using real clock is now failing due to cert expiring in the test data

### DIFF
--- a/internal/backend/saml/session_test.go
+++ b/internal/backend/saml/session_test.go
@@ -17,6 +17,7 @@ import (
 	"github.com/cosi-project/runtime/pkg/state/impl/namespaced"
 	csaml "github.com/crewjam/saml"
 	"github.com/crewjam/saml/samlsp"
+	dsig "github.com/russellhaering/goxmldsig"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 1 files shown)


oguzkilcan (approved):

internal/backend/saml/session_test.go
```diff
@@ -62,6 +63,7 @@ func TestUserInfo(t *testing.T) {
 	} {
 		t.Run(tt.file, func(t *testing.T) {
 			fakeTime = tt.time
+			csaml.Clock = dsig.NewFakeClockAt(csaml.TimeNow())
```
frezbo, line 66: let's start using https://go.dev/blog/testing-time

this is way better than bringing in a new dependency

Slessi: its not a new dependency, this is the clock `csaml` is using internally so already a dependency, can't just swap it out with some other clock

for other tests that dont have fancy clock implementations and juse use `time.Time` sure

frezbo: ahh I see, assumed it was `time.Time`

Issue #1471, opened 2025-08-26
[UI] Tests for Watch/useWatch


Slessi:
As per https://github.com/siderolabs/omni/pull/1463#issuecomment-3214706708 lets have some more unit tests for watch related tasks. Either based off how the existing `Watch` class test works, or with something like MSW

PR #2925, opened 2026-06-02
refactor(frontend): drop watchjoin

Slessi:
Drop `WatchJoin` and related code and replace with manual merging as needed. The complexity it brings is easily solved with array maps, and is rarely needed. As a nice quick side benefit, `TList` items is typeable now and use cases have been typed.

Related #1471


frontend/src/pages/(authenticated)/clusters/[cluster]/nodes.vue
```diff
@@ -25,53 +25,75 @@ import TList from '@/components/List/TList.vue'
...
+  context,
+})
+
+function getNodeItem(
+  item: Resource<ClusterMachineStatusSpec>,
+): Resource<ClusterMachineStatusSpec & V1NodeSpec & TalosMemberSpec, V1NodeStatus> {
+  const itemID = item.metadata.labels?.[ClusterMachineStatusLabelNodeName] ?? item.metadata.id
+
+  const v1node = v1nodes.value.find((n) => n.metadata.name === itemID)
+  const talosMember = talosMembers.value.find((m) => m.metadata.id === itemID)
```
Comment (Unix4ever) [inline frontend/src/pages/(authenticated)/clusters/[cluster]/nodes.vue]: The complexity of that function is rather big now as we call find twice here.
If we go this path I think we should build a map where `n.metadata.name`/`m.metadata.id` is the key and the value is the resource. Then look up the extra data in that map.

frontend/src/pages/(authenticated)/clusters/[cluster]/scale.vue
```diff
@@ -129,30 +129,49 @@ onMounted(async () => {
...
+})
+
+const loading = computed(() => machineStatusesLoading.value || machineConfigGenOptionsLoading.value)
+const err = computed(() => machineStatusesErr.value || machineConfigGenOptionsErr.value)
+const data = computed(() =>
+  machineStatuses.value.map<Resource<MachineStatusSpec & MachineConfigGenOptionsSpec>>((m) => ({
+    ...m,
+    spec: {
+      ...m.spec,
+      ...machineConfigGenOptions.value.find((c) => c.metadata.id === m.metadata.id)?.spec,
```
Comment (Unix4ever) [inline frontend/src/pages/(authenticated)/clusters/[cluster]/scale.vue]: Same. Lets not use find there.
Maybe we can have a helper function `computedMap(items, idFunc)`.
Then:

```js
const machineConfigGenOptsMap = computedMap(machineConfigGenOpts, (c) => c.metadata.id)
...
...machineConfigGenOptsMap.value[m.metadata.id]?.spec,
```

frontend/src/pages/(authenticated)/clusters/[cluster]/nodes.vue
Comment (Slessi) [reply frontend/src/pages/(authenticated)/clusters/[cluster]/nodes.vue]: It's absolutely worse, and a map is absolutely better, but will we be dealing with datasets large enough to care? If every list had 1k items it would probably take like 5-10ms to iterate this at the most.

If we are dealing with large data like that then sure, but if not probably not important. Mostly wondering cause it would be good for me to know to expect large datasets like that in places like this.

(Keeping in mind there is also pagination here, so at most, it is going to be 100 items -> 200 finds, if page is set to 100, which is pretty cheap)

Comment (Slessi) [reply frontend/src/pages/(authenticated)/clusters/[cluster]/nodes.vue]: in any case, added now, easy change 🚀

Unix4ever (approved):

PR #2927, opened 2026-06-03
feat: preserve schematic contents

### LINKED ISSUE #2720: preserve schematic contents
Schematic in the Image Factory has some fields which Omni should never modify, in fact any field which is not managed by Omni.

Right now Omni manages (for a running machine) following fields:

* extensions list
* kernel args

Other fields in the schematic should be preserved as they were when the machine first joined, e.g.:

* owner of the schematic (Image Factory Enterprise)
* embedded machine config (also Enterprise)
* config for SBCs
* ... 

This might be a major rework of how Omni handles schematics, which should be easier/more consistent today rather with schematic GET API available at the Image Factory side.

### LINKED ISSUE #2797: Rework/fix schematic ID calculation/normalization

utkuozdemir:
When we modify the schematic of a machine, instead of rebuilding the whole schematic, we now only modify the fields we manage: extensions and kernel args. Other things stay untouched.

Modify `MachineStatus` resource and its controller to:
- Always store "raw" schematic:
  1. if we can read it from the machine via the meta extension, we use that.
  2. If we know the schematic ID but not the content, we do a GET request to the image factory to read the body.
  3. As a last fallback, we construct it artificially, like it was before.
- Drop the ".schematic.id" field: this ID field was called "plain id", which contained only the extensions, but not kernel args and other fields on the schematic. We should not use this field anymore, since:
  1. Extensions-only logic is not schematic-preserving: the schematics need to be preserved fully.
  2. With the kernel arg management feature landed, the "desired vs actual kernel args" logic already handles non-UKI <1.12 machines (whose kernel args cannot be updated via schematic update): `SchematicConfigurationController` already calculates the desired kernel args to be equal to the actual ones in this case, so the schematic will not be unnecessarily updated. The logic of "compare plain ID vs compare full ID" when determining `shouldUpdate` in `ClusterMachineConfigStatusController` is no longer needed.

Closes siderolabs/omni#2720.
Closes #2797.


utkuozdemir:
self-note: check if we also preserve contents on maintenance upgrade, i.e., through UI/CLI, cluster creation etc.

internal/backend/runtime/omni/controllers/testutils/factory_mock.go
```diff
@@ -7,44 +7,45 @@ package testutils
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/siderolabs/image-factory/pkg/schematic"
```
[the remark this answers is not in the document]
utkuozdemir: removed this completely, was unused.

internal/backend/runtime/omni/controllers/testutils/factory_mock.go
```diff
@@ -7,44 +7,45 @@ package testutils
...
+	}
 }
 
-// ImageFactoryMockClient is a mock image factory client.
-type ImageFactoryMockClient struct{}
+// ImageFactoryMockClient is a mock image factory client. It records every successful EnsureSchematic
+// upload in Schematics, keyed by the resulting ID, so tests can drive SchematicGet against it.
+type ImageFactoryMockClient struct {
+	Schematics map[string]*schematic.Schematic
+}
```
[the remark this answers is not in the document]
utkuozdemir: removed this

internal/backend/runtime/omni/controllers/testutils/factory_mock.go
```diff
@@ -7,44 +7,45 @@ package testutils
...
 
-	plainSchematic := schematic.Schematic{
-		Customization: schematic.Customization{
-			SystemExtensions: schematic.SystemExtensions{
-				OfficialExtensions: sch.Customization.SystemExtensions.OfficialExtensions,
-			},
-		},
-	}
+	stored := sch
+	i.Schematics[id] = &stored
```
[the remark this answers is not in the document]
utkuozdemir: removed this

internal/backend/runtime/omni/controllers/testutils/factory_mock.go
```diff
@@ -7,44 +7,45 @@ package testutils
...
-		return imagefactory.EnsuredSchematic{}, err
+	return id, &sch, nil
+}
+
+// SchematicGet returns a stored schematic by ID, or a 404-equivalent error if absent.
+func (i *ImageFactoryMockClient) SchematicGet(_ context.Context, id string) (*schematic.Schematic, error) {
+	sch, ok := i.Schematics[id]
+	if !ok {
+		return nil, fmt.Errorf("schematic %q not found", id)
 	}
```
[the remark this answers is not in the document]
utkuozdemir: removed

internal/backend/imagefactory/client.go
```diff
@@ -57,47 +54,16 @@ func (cli *Client) Host() string {
...
+// along with the normalized schematic as the factory persisted it.
 //
-// We do not call the image factory for the schematics that are already known to (cached by) Omni.
-func (cli *Client) EnsureSchematic(ctx context.Context, inputSchematic schematic.Schematic) (EnsuredSchematic, error) {
-	fullSchematicID, data, err := cli.SchematicCreate(ctx, inputSchematic)
+// The factory deduplicates by content: if the same schematic was uploaded before, it returns
+// the existing ID without creating a new one.
+func (cli *Client) EnsureSchematic(ctx context.Context, inputSchematic schematic.Schematic) (string, *schematic.Schematic, error) {
+	id, data, err := cli.SchematicCreate(ctx, inputSchematic)
 	if err != nil {
```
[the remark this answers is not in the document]
utkuozdemir: This was already the case - we dropped the cache on main. Not an issue of this PR.

internal/backend/runtime/omni/controllers/omni/machine_status.go
```diff
@@ -334,9 +337,51 @@ func (ctrl *MachineStatusController) reconcileTearingDown(ctx context.Context, r
...
+		info.Raw = existing.TypedSpec().Value.Schematic.Raw
+
+		return nil
+	}
+
+	sch, err := ctrl.ImageFactoryClient.SchematicGet(ctx, info.FullID)
+	if err != nil {
+		if factoryclient.IsHTTPErrorCode(err, http.StatusNotFound) {
+			return nil
+		}
```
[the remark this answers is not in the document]
utkuozdemir: Good point, fixed.

internal/pkg/config/validations/validations_test.go
```diff
@@ -96,7 +96,7 @@ func TestEnsureNoNonImageFactoryMachines(t *testing.T) {
 		{
 			name: "valid schematic",
 			machines: map[string]*specs.MachineStatusSpec_Schematic{
-				"m1": {Id: "abc", FullId: "abc123"},
+				"m1": {FullId: "abc123"},
 			},
```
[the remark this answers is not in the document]
utkuozdemir: `SchematicReady` is not based solely on `FullID` - `Invalid` schematics are also treated as "ready".

internal/backend/runtime/omni/controllers/omni/machine_status.go
```diff
@@ -334,9 +336,68 @@ func (ctrl *MachineStatusController) reconcileTearingDown(ctx context.Context, r
...
+		logger = logger.With(zap.String("id", existing.Metadata().ID()), zap.String("schematic_id", info.FullID))
+
+		logger.Info("machine does have a schematic ID but no raw schematic, get it from image factory")
+
+		sch, err := ctrl.ImageFactoryClient.SchematicGet(ctx, info.FullID)
+		if err != nil {
+			if factoryclient.IsHTTPErrorCode(err, http.StatusNotFound) {
+				logger.Warn("raw schematic not found in image factory", zap.String("host", ctrl.ImageFactoryClient.Host()))
+
+				return nil
```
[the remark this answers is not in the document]
utkuozdemir: No, this is intentional. If we cannot get raw in any way, we "give up" - downstream will not reconcile.

client/api/omni/specs/machine_status.go
```diff
@@ -21,5 +21,5 @@ func (spec *MachineStatusSpec) SchematicReady() bool {
 		return true
 	}
 
-	return spec.Schematic.Id != "" && spec.Schematic.FullId != ""
+	return spec.Schematic.FullId != ""
```
utkuozdemir, line 24: Note to reviewer:
Previously, due to partially historical reasons, we had two schematic IDs:
1. Extensions-only ID: what we also called "plain id", the `.schematic.id` field on the `MachineStatus` spec.
2. Full ID: `.schematic.fullid` field on the `MachineStatus` spec. Containing kernel args, overlays and all other things in a schematic.

The plain ID made sense back then, due to couple of reasons:
- There was no kernel args mgmt in Omni, and for the machines which did not support updating kernel args, we had to compare "extensions-only schematics".
- A custom schematic **only** meant extensions. In other words, extensions were the only thing that made a schematic. Hence the field name is "id", not "plain id".

This field does not make sense anymore, because:
- We now cannot operate on extensions-only, as our intention is to always preserve a schematic, not trim it down (e.g., to schematics-only)
-  Kernel args support feature, which landed a while ago, already handles the kernel args part of a schematic correctly for both "kernel args mgmt supporting (`UKI || >=1.12`)" and "not supporting (non-UKI, grub)" types of machines. For the non-UKI ones, the desired kernel args are always set to be equal to the actual args, which means they won't be touched when computing a new schematic.

Therefore, we could simply drop the plain ID, and the whole handling of "compare full vs compare plain" logic in the `ClusterMachineConfigStatusController. Greatly simplified some parts.

client/api/omni/specs/omni.proto
```diff
@@ -159,10 +159,7 @@ message MachineStatusSpec {
       repeated string extensions = 1;
     }
 
-    // Id is the image factory schematic id used for the image generation.
-    //
-    // This must be be plain id computed solely from the system extensions, not including the kernel command line arguments, META content etc.
-    string id = 1;
+    reserved 1;
```
utkuozdemir, line 162: Note to reviewer: see https://github.com/siderolabs/omni/pull/2927/changes#r3380826154

client/api/omni/specs/omni.proto
```diff
@@ -177,19 +174,14 @@ message MachineStatusSpec {
 
     reserved 5;
 
-    // Overlay contains the information about the overlay used during the machine image generation.
-    // It is used by SBC images.
-    Overlay overlay = 6;
+    reserved 6;
```
utkuozdemir, line 177: Note to reviewer: overlay is now preserved without needing to extract it to a dedicated field - it is on the `.raw` field as-is.

client/api/omni/specs/omni.proto
```diff
@@ -177,19 +174,14 @@ message MachineStatusSpec {
...
 
     // KernelArgs is the kernel command line arguments used during the machine image generation.
     repeated string kernel_args = 7;
 
-    // MetaValues is the list of meta values used during the machine image generation.
-    repeated MetaValue meta_values = 8;
+    reserved 8;
```
utkuozdemir, line 182: Note to reviewer: meta values are now preserved without needing to extract it to a dedicated field - it is on the .raw field as-is.

cmd/omni/pkg/app/app.go
```diff
@@ -83,7 +83,6 @@ func Run(ctx context.Context, state *omni.State, cfg *config.Params, logger *zap
 	}
 
 	imageFactoryClient, err := imagefactory.NewClient(
-		state.Default(),
```
utkuozdemir, line 86: Note to reviewer: state is no more needed/used by the image factory client, as we dropped the schematic cache. So, simply removed it.

internal/backend/grpc/schematics.go
```diff
@@ -108,18 +108,18 @@ func (s *managementServer) CreateSchematic(ctx context.Context, request *managem
...
 	}
 
-	schematicInfo, err := s.imageFactoryClient.EnsureSchematic(ctx, schematicRequest)
+	schematicID, schematicData, err := s.imageFactoryClient.EnsureSchematic(ctx, schematicRequest)
 	if err != nil {
 		return nil, fmt.Errorf("failed to ensure schematic: %w", err)
 	}
 
-	schematicYML, err := yaml.Marshal(schematicInfo.Data)
+	schematicYML, err := schematicData.Marshal()
```
utkuozdemir, line 116: Note to reviewer: I replaced all places where we mistakenly didn't use the `schematic.Marshal()` / `.Unmarshal()` functions with those.

internal/backend/imagefactory/client.go
```diff
@@ -91,47 +93,16 @@ func (cli *Client) CachedIsEnterprise() bool {
...
-}
-
-// EnsureSchematic ensures that the given schematic exists in the image factory.
-//
-// It ensures two schematics: one with the full content of the schematic.Schematic, and another with only the system extensions.
+// EnsureSchematic uploads the given schematic to the image factory and returns its ID
+// along with the normalized schematic as the factory persisted it.
 //
-// We do not call the image factory for the schematics that are already known to (cached by) Omni.
-func (cli *Client) EnsureSchematic(ctx context.Context, inputSchematic schematic.Schematic) (EnsuredSchematic, error) {
```
utkuozdemir, line 106: Note to reviewer: we don't ensure plain + full schematics anymore, do only "full". See  https://github.com/siderolabs/omni/pull/2927/changes#r3380826154

internal/backend/imagefactory/patch.go
```diff
@@ -0,0 +1,44 @@
...
+//
+// All other fields (owner, secureboot, bootloader, meta, overlay incl.
+// options, and any field added by future image-factory versions) are
+// preserved by going through the same schematic.Schematic struct the factory
+// uses for its canonical representation.
+//
+// The unmarshal is strict: any field the bundled image-factory library does
+// not know about causes an error. When that happens, the image-factory module
+// version in this repo must be bumped to catch up with the factory.
+func PatchSchematic(rawYAML string, extensions, kernelArgs []string) (schematic.Schematic, error) {
```
utkuozdemir, line 30: Note to reviewer: this is the core idea of this PR: patching a schematic while preserving its contents. We only manage two fields: extensions and kernel args.

internal/backend/runtime/omni/controllers/omni/internal/talos/schematic_test.go
```diff
@@ -0,0 +1,245 @@
...
+func schematicYAML(t *testing.T, s schematic.Schematic) string {
+	t.Helper()
+
+	data, err := s.Marshal()
+	require.NoError(t, err)
+
+	return string(data)
+}
+
+func TestGetSchematicInfo(t *testing.T) {
```
utkuozdemir, line 72: Note to reviewer: decided to add some unit tests for `GetSchematicInfo`.

internal/backend/runtime/omni/controllers/omni/machineconfig/reconciliation_context.go
utkuozdemir, line 1: Note to reviewer: dropped the field and all logic around `compareFullSchematicID`, see: https://github.com/siderolabs/omni/pull/2927/changes#r3380826154

internal/backend/runtime/omni/controllers/omni/machineconfig/status.go
utkuozdemir, line 1: Note to reviewer: dropped the plain id field and all logic around full vs plain id comparison, see: [#2927 (changes)](https://github.com/siderolabs/omni/pull/2927/changes#r3380826154)

internal/backend/runtime/omni/controllers/omni/schematic/configuration.go
utkuozdemir, line 1: Note to reviewer:
Multiple things happened here: 
1. I moved this controller to its own package, and changed the constructor to do the `ctrl.QController = ...`  trick, so we can have methods on the controller.
2. Rewrote its tests using `testutils.WithRuntime`
3. **Most important:** actual logical changes which now patch the schematic of a machine instead of constructing it from scratch.

internal/backend/runtime/omni/controllers/omni/schematic/configuration.go
```diff
@@ -199,70 +180,88 @@ func (helper *schematicConfigurationHelper) reconcile(
...
-		},
-		Overlay: overlay,
-	}
-
-	id, err := config.ID()
+	// Patch only the fields Omni manages (extensions list, kernel args). Everything else
+	// the user/factory put into the schematic (owner, secureboot, bootloader, meta,
+	// overlay incl. options, future fields) is preserved by going through the raw YAML
+	// the machine reports.
+	patched, err := imagefactory.PatchSchematic(rawSchematic, customization.extensionsList, kernelArgs)
```
utkuozdemir, line 216: Note to reviewer: this is the core change.

internal/backend/runtime/omni/controllers/omni/schematic/configuration.go
```diff
@@ -199,70 +180,88 @@ func (helper *schematicConfigurationHelper) reconcile(
...
-		return nil, err
+		return fmt.Errorf("failed to patch schematic: %w", err)
 	}
 
-	if _, err = helper.imageFactoryClient.EnsureSchematic(ctx, config); err != nil {
-		return nil, err
+	// TODO(preserve-schematic): when the patched schematic is content-equal to the
+	// source raw (i.e. no Omni-driven customization changed anything), short-circuit
+	// the factory call and publish ms.Schematic.FullId directly.
+	id, _, err := ctrl.imageFactoryClient.EnsureSchematic(ctx, patched)
```
utkuozdemir, line 224: Note to reviewer: one of the core changes: now we store the returned ID from the factory into a variable, and set it on the `SchematicId` field of this resource.

internal/backend/runtime/omni/controllers/omni/schematic_configuration.go
```diff
@@ -276,16 +275,6 @@ func newMachineCustomization(ctx context.Context, r controller.Reader, ms *omni.
...
 
-	// Leave meta values as-is to prevent any unwanted upgrades.
-	// They are no-op for non-UKI machines, but we still preserve them to not cause a schematic ID change, which would upgrade (reboot) the machine.
-	mc.meta = xslices.Map(ms.TypedSpec().Value.Schematic.MetaValues,
-		func(v *specs.MetaValue) schematic.MetaValue {
-			return schematic.MetaValue{
-				Key:   uint8(v.GetKey()),
-				Value: v.GetValue(),
-			}
-		})
```
utkuozdemir, line 287: Note to reviewer: no need to reconstruct meta values anymore - we pass it through as-is (never trim it from the schematic, never construct a schematic).

internal/backend/runtime/omni/controllers/omni/schematic/configuration_test.go
utkuozdemir, line 1: Note to reviewer: same controller tests as before, but:
1. migrated to use `testutils.WithRuntime`
2. Accounts for the changes in the controller (schematic patching / field preservation)

internal/backend/runtime/omni/controllers/omni/schematic/configuration_test.go
```diff
@@ -0,0 +1,615 @@
...
+// TestSchematicConfigurationPreservesRawFields verifies that the controller preserves all schematic fields
+// that Omni does not typed-manage (overlay options, embedded machine config, bootloader, secureboot,
+// disk image options) when it patches the machine's raw schematic and uploads the result to the factory.
+//
+// It also verifies that the controller uses the schematic id returned by the factory as the source of truth.
+// The fake factory is configured with an Owner, which it stamps onto every upload (mirroring the Enterprise
+// factory). The controller-published SchematicId must therefore match the id of the owner-tagged schematic,
+// not the id of the un-tagged patched schematic the controller computed locally. We verify this by looking
+// up the published id in the factory's storage and asserting the stored body carries the injected owner.
+func TestSchematicConfigurationPreservesRawFields(t *testing.T) {
```
utkuozdemir, line 440: Note to reviewer: this is the new test asserting field preservation.

internal/backend/runtime/omni/controllers/omni/cluster_machine_encryption.go
```diff
@@ -80,6 +81,10 @@ func (ctrl *ClusterMachineEncryptionController) Run(ctx context.Context, r contr
 
 		params, err := safe.ReaderGetByID[*siderolink.APIConfig](ctx, r, siderolink.ConfigID)
 		if err != nil {
+			if state.IsNotFoundError(err) { // not initialized yet, skip
+				continue
+			}
```
utkuozdemir, line 86: Note to reviewer: this is an unrelated fix to the PR, I just noticed this controller producing an error log on startup, and the fix was trivial, so just included it.

internal/backend/runtime/omni/controllers/omni/schematic_configuration_test.go
utkuozdemir, line 1: Actually moved, not deleted, but changes are a lot, so it's not detected as a move.

internal/backend/runtime/omni/controllers/omni/machineupgrade/status.go
```diff
@@ -140,11 +136,6 @@ func (ctrl *StatusController) transform(ctx context.Context, r controller.Reader
 		return nil
 	}
 
-	schematicEqual, schematicEqualWithoutKernelArgs, err := ctrl.checkSchematicEquality(currentSchematic, schematicConfiguration)
-	if err != nil {
-		return fmt.Errorf("failed to get schematic diff: %w", err)
-	}
```
utkuozdemir, line 146: Note to reviewer: I realized that here we had logically dead code - the `schematicEqualWithoutKernelArgs`, which compares desired vs actual schematic to determine "is kernel arg update requested". This would not work, because for the machines which do not support kernel args updates would have their desired schematic already have the same kernel args as actual. This fact made this and the removed code below useless. And that rendered `SchematicConfiguration.kernel_args` useless, therefore dropped.

client/api/omni/specs/omni.proto
```diff
@@ -1298,9 +1290,7 @@ message TalosExtensionsSpec {
 message SchematicConfigurationSpec {
   string schematic_id = 1;
   string talos_version = 2;
-
-  // KernelArgs are the kernel args which were used to compute the SchematicId.
-  repeated string kernel_args = 3;
+  reserved 3;
```
utkuozdemir, line 1292: Note to reviewer: dropped because it was logically useless/dead code - see https://github.com/siderolabs/omni/pull/2927/changes#r3382502896

internal/backend/runtime/omni/controllers/omni/machine_status.go
```diff
@@ -334,9 +336,68 @@ func (ctrl *MachineStatusController) reconcileTearingDown(ctx context.Context, r
 	return r.RemoveFinalizer(ctx, machine.Metadata(), ctrl.Name())
 }
 
+// populateSchematicRaw fills schematicInfo.Raw for the "meta extension present but no ExtraInfo" case (older Talos)
```
utkuozdemir, line 339: > "meta extension present but no ExtraInfo" case (older Talos)

This is a very rare case (maybe 1% of all machines), but IMO still worth doing this backfill, since now the **raw schematic** matters more for us (essentially required).

internal/backend/runtime/omni/controllers/omni/internal/talos/schematic_test.go
```diff
@@ -0,0 +1,245 @@
...
+
+		assert.Equal(t, []string{"siderolabs/gvisor"}, info.Extensions)
+	})
+
+	t.Run("extension name without official prefix gets prefixed", func(t *testing.T) {
+		t.Parallel()
+
+		st := buildState(t, []extension{
+			{name: constants.SchematicIDExtensionName, version: factorySchematicID, extraInfo: factorySchematicYAML},
+			{name: "gvisor"}, // missing "siderolabs/" prefix
```
Unix4ever, line 222: Was it like that before? Didn't know that we add the prefix to the extensions if it's missing.

internal/backend/runtime/omni/controllers/omni/machineupgrade/status.go
Unix4ever: We don't even allow to update the kernel args for machines below 0.12, so this condition won't be ever satisfied. Do I understand that correctly?

internal/backend/runtime/omni/controllers/omni/internal/talos/schematic_test.go
utkuozdemir: Yep, it seems I introduced it long ago: https://github.com/siderolabs/omni/commit/176f9d9f57530832a9ebbb64d008bc98300b2cc7#diff-e24b8662fb0c9ae1636e4e064ce7f2d3e55c937efc48c7c7aff2569e39282204R58-R60

Don't remember the exact reasoning though.

Unix4ever:
This code throws away so much of spaghetti code we had around the schematic generation 🥳 Great job 💯

internal/backend/runtime/omni/controllers/omni/machineupgrade/status.go
utkuozdemir: My original idea here was to compare actual vs desired schematic two-fold: one ignoring kernel args, one including kernel args. We were then determining `schematicEqualWithoutKernelArgs`. If that was true, we knew that user tried to update the kernel args (in retrospect, this is not a good way in any case).

When this flag was true, we then checked if that machine supported updating kernel args (`UKI` or `>=1.12 + grubUseUKICmdline`). If it didn't, we could determine "the user wants kernel args to be updated, but it's not possible, because there is an unmet condition". We were surfacing this info on `MachineUpgradeStatus` (a status resource only relevant in maintenance mode).

I realized that this would never work, because for the kernel-args non-supporting machines, the desired schematic ID's kernel args would always be equal to the actual ones. This is because in the `SchematicConfigurationController`, we do the same exact check on the machine, "is kernel args update supported?", and if it returns false, we say "desired = actual" and compute the desired schematic that way.

So, `schematicEqualWithoutKernelArgs` was meaningless in this controller - even if you attempted to do a maintenance upgrade by updating kernel-args only, you'd see "machine is up to date" in the `MachineUpgradeStatus` resource.

To be clear: the same warning that is dead code here is also produced on the `KernelArgsStatus` resource, following a similar logic, but there it actually works, and is surfaced to the UI:

image

utkuozdemir:
> This code throws away so much of spaghetti code we had around the schematic generation 🥳 Great job 💯

Yep I like getting rid of a lot of stuff, especially the `compareFullSchematicID` mess :)

internal/backend/runtime/omni/controllers/omni/machineupgrade/status.go
Unix4ever: ah I see

Unix4ever (approved):

internal/backend/runtime/omni/controllers/omni/schematic/configuration.go
```diff
@@ -199,70 +180,88 @@ func (helper *schematicConfigurationHelper) reconcile(
...
+	if err != nil {
+		return err
 	}
 
 	schematicConfiguration.TypedSpec().Value.SchematicId = id
-	schematicConfiguration.TypedSpec().Value.KernelArgs = kernelArgs
 	machineExtensionsStatus.TypedSpec().Value.Extensions = computeMachineExtensionsStatus(ms, &customization)
```
oguzkilcan, line 230: Not sure if this is needed anymore. Wouldn't it be better to use the ensured schematic to update extensions?

oguzkilcan (approved):

oguzkilcan:
🆒

PR #2928, opened 2026-06-03
feat(frontend): generate talos types for frontend

### LINKED ISSUE #2200: Generate types for Talos proto files
Currently all Talos proto files are untyped on the frontend.

As examples, the below types have no corresponding specs generated:
- `CPUStats.perf.talos.dev`
- `Services.v1alpha1.talos.dev`

Slessi:
Generate Talos types for frontend with a go script that pulls correct field names from yaml tags.

Closes #2200


internal/internal/generate.go
```diff
@@ -8,6 +8,9 @@ package internal
 
 //go:generate go run -tags=sidero.tools github.com/siderolabs/omni/internal/internal/tools/tsgen -out ../../frontend/src/api/resources.ts ../../,../../client/
 
+// Generate TypeScript types for Talos COSI resource specs from yaml struct tags.
```
Unix4ever, line 11: So that's how we do it :D

Unix4ever (approved):

frontend/src/api/talos/runtime.pb.ts
```diff
@@ -0,0 +1,188 @@
...
+}
+
+export type WatchdogTimerConfigSpec = {
+  device?: string
+  timeout?: number
+}
+
+export type WatchdogTimerStatusSpec = {
+  device?: string
+  timeout?: number
```
smira: this is `time.Duration`, it will be something "30s" in YAML, so I guess `string` unless you convert it somewhere else?

internal/internal/generate.go
```diff
@@ -8,6 +8,9 @@ package internal
...
+//go:generate go -C tools/talostsgen run -tags=sidero.tools . ../../../../frontend/src/api/talos
```
smira: how does it get Talos source code?

Slessi: requires it in https://github.com/Slessi/omni/blob/eff40ef4a03dcfab140e163b04279a2c0d2e1570/internal/internal/tools/talostsgen/deps.go but im trying to drop that file

smira: ok... I never knew it works this way

frontend/src/api/talos/runtime.pb.ts
Slessi: fixed!

PR #2932, opened 2026-06-03
refactor: drop unused field in create schematic gRPC request

utkuozdemir:
The field `secure_boot` was unused when creating a schematic. Drop that field.

### DIFF
--- a/frontend/src/views/InstallationMedia/usePresetSchematic.ts
+++ b/frontend/src/views/InstallationMedia/usePresetSchematic.ts
@@ -58,7 +58,6 @@ export function usePresetSchematic(presetRef: MaybeRefOrGetter<InstallationMedia
             : undefined,
         overlay,
         talos_version: preset.talos_version,
-        secure_boot: preset.secure_boot,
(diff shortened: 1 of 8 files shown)


frontend/src/views/InstallationMedia/usePresetSchematic.ts
```diff
@@ -58,7 +58,6 @@ export function usePresetSchematic(presetRef: MaybeRefOrGetter<InstallationMedia
             : undefined,
         overlay,
         talos_version: preset.talos_version,
-        secure_boot: preset.secure_boot,
```
Slessi, line 61: there is a checkbox for this in the UI, and factory has this option also, why are we removing it?

image

smira: ScureBoot is not part of the schematic, it's encoded in the URL of the boot asset, like arch.

smira (approved):

Unix4ever (approved):

Issue #2805, opened 2026-05-07
Pending config changes never apply for one machine until Omni is restarted


utkuozdemir:
## Summary

In a cluster managed by Omni, one machine stopped receiving config updates. The desired `ClusterMachineConfig` content was up to date and the machine was online and connected, but `ClusterMachineConfigStatus` for that machine never advanced and `MachineConfigDiff` entries stopped being produced for it. The rest of the cluster kept applying config normally.

Restarting Omni cleared the issue. The machine started getting its config applied again on its own, with no other intervention.

## What we observed

- `ClusterMachineStatus.configuptodate = false` and `configapplystatus = PENDING`, with `lastconfigerror = ""`.
- `ClusterMachineConfigStatus.clustermachineconfigversion` for the affected machine was far behind the rest of the cluster (peers were tens of versions ahead).
- `MachineConfigDiff` stopped producing entries for the affected machine while peers kept getting them.
- `ClusterMachine` was the only resource in the cluster whose metadata `updated` timestamp had not moved in days.
- Locking and unlocking the `MachineSetNode` did not unblock anything.
- The machine itself was fine. `MachineStatus.connected = true`, no generation error, desired `ClusterMachineConfig` content was current.
- In Omni controller logs, `ClusterMachineConfigStatusController` produced no entries at all for the affected machine over a long observation window. For every other `ClusterMachineConfig` in the same cluster it produced the usual `reconcile succeeded`, `operations locked for machine`, or `status update skipped` lines.
- Other controllers keyed off the same machine ID (`ClusterMachineConfigController`, `ClusterMachineConfigPatchesController`, `ClusterMachineStatusController`, `SchematicConfigurationController`, and so on) kept reconciling for that machine normally.
- Even direct triggers that should always queue a reconcile via `MapperSameID`, like toggling the `MachineSetNode` lock annotation, did not produce any `ClusterMachineConfigStatusController` log line for that ID.

## What we suspect (not confirmed)

We do not know the actual root cause. Our suspicion, based on the log signal above, is that `ClusterMachineConfigStatusController` got stuck for that one input ID while the same controller kept reconciling for other IDs. Possible mechanisms we could not rule out:

- An external call (Talos client, image factory, etc.) without a bounded timeout, holding a worker slot.
- A finalizer or output-teardown path that qtransform is waiting on.
- A goroutine or lock leak from some failure mode that never gets cleaned up.
[...]

PR #2936, opened 2026-06-04
fix: release config update slot while a machine waits to upgrade

utkuozdemir:
A machine that needs both a config change and a Talos upgrade could pin its config update slot indefinitely and starve the rest of the machine set. A machine processes its upgrade before applying its config, since the config cannot be applied until the machine is on the right Talos version. If the upgrade could not start because no upgrade slot was free, the machine kept holding the config update slot it had taken earlier while parked waiting for an upgrade turn. With that slot held, other machines that only needed a config change could never get a turn, and the set stayed stuck until Omni was restarted.

A machine now releases its config update slot whenever it still needs an upgrade, whether the upgrade is in progress or waiting for a free slot, and re-takes the slot later once it is back in sync and actually applies its config. This mirrors the upgrade slot already being released before the config apply phase, so the two rollouts never block each other.

Related to siderolabs/omni#2805.

### DIFF
--- a/internal/backend/runtime/omni/controllers/omni/machineconfig/status.go
+++ b/internal/backend/runtime/omni/controllers/omni/machineconfig/status.go
@@ -202,32 +202,8 @@ func (ctrl *ClusterMachineConfigStatusController) reconcileRunning(
 		machineConfigStatus.TypedSpec().Value.SchematicId = ""
 	}
 
-	if rc.shouldUpgrade {
-		inSync, syncErr := ctrl.upgrade(ctx, logger, r, rc)
-		if syncErr != nil {
-			return syncErr
-		}
-
-		if !inSync {
-			logger.Info(
-				"the machine talos version is out of sync, the config is not applied",
-				zap.String("machine", machineConfig.Metadata().ID()),
-			)
-
-			return xerrors.NewTaggedf[qtransform.SkipReconcileTag]("machine talos version is out of sync: %s", machineConfig.Metadata().ID())
-		}
-	}
-
-	// At this point the machine is in sync: it either needed no upgrade, or the upgrade just
-	// completed. Release the upgrade lock here, before entering the config-apply phase, so the
-	// upgrade and config-update locks are never held simultaneously by the same machine.
-	//
-	// Holding the upgrade lock into the config-apply phase allows a lock-ordering inversion to
-	// deadlock the whole machine set: one machine holds the config-update lock and waits for the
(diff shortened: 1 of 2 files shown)


frezbo (approved):
Do we have the upgrade process documented in omni docs, how locks are held, this should give users information on how to schedule patches and upgrades

internal/backend/runtime/omni/controllers/omni/machineconfig/status.go
```diff
@@ -204,6 +204,18 @@ func (ctrl *ClusterMachineConfigStatusController) reconcileRunning(
...
 	if rc.shouldUpgrade {
 		inSync, syncErr := ctrl.upgrade(ctx, logger, r, rc)
+
+		// While the machine still needs an upgrade (the upgrade is in progress, or is blocked
+		// waiting for a free upgrade slot), it is not applying its config, so it must not keep
+		// holding the config-update lock. Otherwise a machine parked waiting for an upgrade slot
+		// starves the other machines in the set that only need a config change. The config-update
+		// lock is re-acquired later, once the machine is back in sync and applies its config.
+		if syncErr != nil || !inSync {
+			if err = ctrl.releaseConfigUpdateLock(ctx, r, rc.clusterMachine); err != nil {
```
oguzkilcan: What if we are config update caused a reboot and we are waiting the machine to become ready before removing the lock? This change would remove the lock prematurely.

Unix4ever: yeah, that feels wrong. Also the condition combination is a bit messy here now. 3 different checks that do kinda the same. Maybe we could extract that into a function and use premature returns to make it easier to read?

utkuozdemir:
> Do we have the upgrade process documented in omni docs, how locks are held, this should give users information on how to schedule patches and upgrades

good point, wdyt about: https://github.com/siderolabs/docs/pull/583

internal/backend/runtime/omni/controllers/omni/machineconfig/status.go
utkuozdemir: Good points. I addressed both your comments. Pls have another look.

frezbo:
> > Do we have the upgrade process documented in omni docs, how locks are held, this should give users information on how to schedule patches and upgrades
> 
> good point, wdyt about: [siderolabs/docs#583](https://github.com/siderolabs/docs/pull/583)

This is good 👍

Unix4ever (approved):

PR #2940, opened 2026-06-05
feat(frontend): add kubespan status view

### LINKED ISSUE #1139: [feature] Incorporate kubespan status in Omni
### Problem Description

We had kubespan go down on some nodes, but the only way we could determine it was via `talosctl get kubespanpeerstatuses.kubespan.talos.dev`.

### Solution

Omni should display kubespan status for nodes, and ideally make it obvious when the kubespan status is `down`

### Alternative Solutions

_No response_

### Notes

_No response_

Slessi:
Add a graphical view of KubeSpan statuses for a given cluster machine

[🎨 Figma](https://www.figma.com/design/4qvWHbfWdFKcDc29wPRn3l/Sidero-UI?node-id=6106-66159&t=T8PFC5tyVVO8rcoN-4)
[📖 Storybook](https://kubespan-status--68d664245076e2bb623d98f3.chromatic.com/?path=/story/views-kubespanstatus--default)

Closes #1139 

image


rothgar:
A bit of confusion on my part, isn't KubeSpan a full mesh network so each node has a dedicated connection to every other node? This visual makes it seem like a one to many.

I really like it :+1:

Slessi:
@rothgar it is a mesh, but showing the full mesh can get a bit crazy, so we are going to start with showing it only as 1->n on a per-machine perspective for now, you will be able to switch between different machine perspectives

You can checkout the [figma](https://www.figma.com/design/4qvWHbfWdFKcDc29wPRn3l/Sidero-UI?node-id=6106-66159&t=T8PFC5tyVVO8rcoN-4) for more deets

Unix4ever:
I think it might be nice to add worker/control plane role labels to the nodes on the graph.
Maybe as a second row to the nodes on the left side of the graph.

For the current node we can also add it as:
```
Role
Control Plane
```
Same way as we could have it for the IP, but have the role there.

Unix4ever (approved):

PR #2946, opened 2026-06-05
feat: cap Talos upgrade targets at the latest supported release

oguzkilcan:
Introduce a LatestSupportedTalosVersion constant so each Omni release advertises which Talos minors it can support end-to-end, and expose the cap on the TalosVersion resource. Cluster create/update, the MaintenanceUpgrade RPC, the cluster TalosUpgradeStatus computation, and every FE picker now use the same flag.


internal/backend/grpc/management.go
```diff
@@ -593,13 +593,30 @@ func (s *managementServer) MaintenanceUpgrade(ctx context.Context, req *manageme
...
 	if err != nil {
 		return nil, err
 	}
 
 	ctx = actor.MarkContextAsInternalActor(authCtx)
 
+	targetTalosVersion, err := safe.StateGetByID[*omnires.TalosVersion](ctx, s.omniState, strings.TrimPrefix(req.Version, "v"))
+	if err != nil {
+		if state.IsNotFoundError(err) {
+			return nil, status.Errorf(codes.FailedPrecondition, "talos version %q is not tracked by this Omni release", req.Version)
```
utkuozdemir, line 610: "not tracked" reads a bit weird, maybe you mean "not supported"?

utkuozdemir: or "not known".

internal/backend/runtime/omni/controllers/omni/internal/talos/versions.go
```diff
@@ -23,31 +25,43 @@ import (
...
+		return v.Metadata().ID(), v
+	})
+
+	sourceID := strings.TrimPrefix(currentVersion, "v")
+
+	candidates, err := getCandidates(talosVersionByID, sourceID, currentVersion)
 	if err != nil {
-		return nil, err
+		// Failed to calculate upgrade candidates. Assume that there are no upgrade versions available.
+		return make([]string, 0), nil //nolint:nilerr
```
utkuozdemir, line 50: Is swallowing the error a good idea here? We don't even check the type of the error.

internal/backend/runtime/omni/controllers/omni/talosupgrade/status_test.go
```diff
@@ -230,13 +230,16 @@ func TestStatusController(t *testing.T) {
 					anotherTalosVersion: {"1.32.0", "1.33.0", "1.34.2"},
 				}
 
+				talosVersionIDs := xmaps.Keys(talosVersions)
```
utkuozdemir, line 234: You probably want to sort this slice (with semver comparator). Otherwise the resource can get updated with different order each time.

utkuozdemir: Sorry, just noticed it's a test, so it doesn't matter probably.

internal/backend/runtime/omni/controllers/omni/versions.go
```diff
@@ -276,16 +276,29 @@ func (ctrl *VersionsController) reconcileTalosVersions(ctx context.Context, r co
...
+	if err != nil {
+		return fmt.Errorf("failed to parse latest supported Talos version %q: %w", constants.LatestSupportedTalosVersion, err)
+	}
+
 	err = forAllCompatibleVersions(talosVersions, k8sVersions, func(talosVer string, compatibleK8sVersions []string) error {
 		talosVersion := omni.NewTalosVersion(talosVer)
 		if writeErr := safe.WriterModify(ctx, r, talosVersion, func(res *omni.TalosVersion) error {
 			res.TypedSpec().Value.Version = talosVer
 			res.TypedSpec().Value.CompatibleKubernetesVersions = compatibleK8sVersions
+			res.TypedSpec().Value.UpgradableTalosVersions = upgradeTargets[talosVer]
```
utkuozdemir, line 294: is this value sorted (by semver)? if not, it should be.

internal/backend/runtime/omni/controllers/omni/internal/talos/versions.go
oguzkilcan: This is intentional. If we fail to parse current version, we can't calculate upgrade paths, so we don't return any upgrade path but do not break resource creation.

client/pkg/constants/constants.go
```diff
@@ -19,6 +19,11 @@ const DefaultTalosVersion = "1.13.3"
 // tsgen:MinTalosVersion
 const MinTalosVersion = "1.9.0"
 
+// LatestSupportedTalosVersion is the highest Talos release Omni officially supports as an upgrade target.
+// Only the major.minor portion is significant, the patch is included so the string parses as a complete semver.
+// tsgen:LatestSupportedTalosVersion
+const LatestSupportedTalosVersion = "1.13.0"
```
utkuozdemir: I'd actually suggest doing this `major.minor` and doing the parsing in our internal code. Since this is our public client package, the `.0` can be confusing for the users of the library. Better to communicate what we mean exactly - patch doesn't play a role, so I would omit that from this constant.

internal/backend/grpc/management.go
oguzkilcan: We might have discovered the talos version from image factory but if it's part of the deny list in Omni. We do not create a TalosVersion resource for that. So it's known but we don't keep track of it anymore. It's also not supported but with this PR we are adding a new field (`unsupported`) to TalosVersion resource. I didn't want to use "not supported" here because with actually not supported Talos versions, a more recent Omni might bring support. That's not the case here. This talos version was broken for some reason so Omni doesn't track it anymore.
Do you have a better way to express it, except "not known" and "not supported"?

internal/backend/runtime/omni/controllers/omni/versions.go
oguzkilcan: I will check

internal/backend/runtime/omni/controllers/omni/versions.go
```diff
@@ -309,6 +322,86 @@ func (ctrl *VersionsController) reconcileTalosVersions(ctx context.Context, r co
...
+		})
+
+		result[source.version] = targets
+	}
+
+	return result, nil
+}
+
+// targetAllowed decides whether a Talos version may appear in an upgrade-target list given the global cap on supported versions and the pre-release feature flag.
+func targetAllowed(target, latestSupported semver.Version, enablePreReleases bool) bool {
```
utkuozdemir: It'd be good to add a comment on this function, giving examples of what it'd return for which inputs.

Also, it feels like it can be simplified with early returns.

And the `enablePreReleases` is a bit confusing - do we want to allow minor versions beyond the supported version when that flag is enabled? If so, maybe we should allow all upgrades (to newer versions) when it is true.

internal/backend/runtime/omni/controllers/omni/talosupgrade/status_test.go
oguzkilcan: Currently we don't use it in the test but if we do in the future, this part might lead to flakiness. I'll fix it.

internal/backend/runtime/omni/controllers/omni/internal/talos/versions.go
utkuozdemir: But parsing error is unexpected? It should be surfaced to us - what failed on parsing no? We need at least an error I guess.

internal/backend/runtime/omni/controllers/omni/versions.go
oguzkilcan: Regarding `enablePreReleases`, if it's enabled only pre-releases for next minor will be shown. For example, you are on Omni 1.8 and Talos 1.14 is going through it's alpha/rc stages. You still want to test it. In the past `enablePreReleases` was sufficient to see those versions. Now with the cap this is not possible because Omni was officially tested with Talos 1.13. So I instead of adding a new flag to let Omni support these Talos versions, I integrated it into `enablePreReleases` flag. If user wants pre-release versions then they can get them even though they are not fully supported.

utkuozdemir: Why not show all future versions, be it prerelease or final, when `enablePreReleases` flag is enabled. Kinda short-circuit to returning true. Because the way I see it, `enablePreReleases` flag is kinda "use at your own risk" developer flag.

internal/backend/runtime/omni/controllers/omni/versions.go
oguzkilcan: It is sorted.

internal/backend/runtime/omni/controllers/omni/versions.go
oguzkilcan: I disagree. The flag let's you see prerelease versions. Allowing access to stable but unsupported versions is not within it's scope.

Slessi: Both sides of that argument sound valid 👀

Unix4ever (approved):

PR #2947, opened 2026-06-05
feat: use discovery endpoint from the generated machine config

utkuozdemir:
Omni used to read the discovery service endpoint back from each node and rely on it when removing the node from the discovery service during teardown. Use the discovery endpoint that Omni resolves from the machine configuration it generates and applies instead.

The endpoint is recorded only after the configuration has been applied to the machine, so it reflects what Omni pushed. When discovery is disabled the endpoint is empty and the discovery service cleanup is skipped, as before.


utkuozdemir:
We decided to not do this for various reasons.

PR #2950, opened 2026-06-08
refactor: move state validations into their own package

utkuozdemir:
Move the state-level resource validations into their own package, with one file per resource family.

Authorization checks are now applied ahead of value validations, so unauthorized callers see a permission error first.


internal/backend/runtime/omni/validations/auth.go
frezbo, line 1: no tests for these?

internal/backend/runtime/omni/validations/cluster.go
frezbo, line 1: i think we need to have tests for all validations

oguzkilcan: I think only validations were split into their own package. Tests are still in `state_validation_test.go`, which was moved.

oguzkilcan (approved):

internal/backend/runtime/omni/validations/auth.go
utkuozdemir: I only moved code around, which is the goal of this PR. no tests were added or removed.

internal/backend/runtime/omni/validations/cluster.go
utkuozdemir: Yep, I only moved code around, which is the goal of this PR. no tests were added or removed.

PR #2951, opened 2026-06-09
feat: validate resource metadata at the state layer

utkuozdemir:
Resource IDs, labels, and annotations are now validated at the state layer for every resource type.

IDs are checked on create only, since they are immutable. Labels and annotations are checked on create and update with a diff-aware strategy: only entries that changed from the existing resource are re-validated, so previously stored offending values do not block unrelated updates.

### DIFF
--- a/internal/backend/runtime/omni/validations/metadata.go
+++ b/internal/backend/runtime/omni/validations/metadata.go
@@ -1,0 +1,154 @@
+// Copyright (c) 2026 Sidero Labs, Inc.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+
+package validations
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"strings"
+	"unicode"
+
+	"github.com/cosi-project/runtime/pkg/resource"
+	"github.com/cosi-project/runtime/pkg/state"
+
+	"github.com/siderolabs/omni/internal/backend/runtime/omni/validated"
+)
+
+const (
+	// MaxResourceIDLength caps the byte length of a resource ID.
+	MaxResourceIDLength = 1024
+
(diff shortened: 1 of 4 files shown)


internal/backend/runtime/omni/validations/metadata.go
```diff
@@ -0,0 +1,153 @@
...
+	if strings.ContainsFunc(value, isControlChar) {
+		return fmt.Errorf("%s value for key %q must not contain control characters", kind, key)
+	}
+
+	return nil
+}
+
+func isControlChar(r rune) bool {
+	return r < 0x20 || r == 0x7F
+}
```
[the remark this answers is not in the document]
utkuozdemir: fixed

internal/backend/runtime/omni/validations/metadata.go
```diff
@@ -0,0 +1,154 @@
...
+	"strings"
+	"unicode"
+
+	"github.com/cosi-project/runtime/pkg/resource"
+	"github.com/cosi-project/runtime/pkg/state"
+
+	"github.com/siderolabs/omni/internal/backend/runtime/omni/validated"
+)
+
+const (
```
oguzkilcan, line 21: Should we make these limits public and expose them to FE? I think it makes sense to have a single source of truth for these kind of validations.

oguzkilcan (approved):

internal/backend/runtime/omni/validations/metadata.go
utkuozdemir: We can do it in a separate PR as we need it. i think frontend can even use them to show the validation on the fields directly, but it should belong to another PR. As the FE needs those vars, we can simply add `tsgen` directive to these vars to get them as constants. I'll create an issue for it.

PR #2952, opened 2026-06-09
feat: install/upgrade Talos in maintenance mode

### LINKED ISSUE #2531: use install API in maintenance mode
Use install API in maintenance mode to install Talos to disk

oguzkilcan:
Adds a streaming MaintenanceLifecycle management RPC backed by Talos's LifecycleService.{Install,Upgrade} for machines booted in maintenance mode, plus omnictl install/upgrade subcommands and FE modals that stream installer progress live.

Resolves: #2531


.github/workflows/e2e-misc-upgrades-cron.yaml
```diff
@@ -59,7 +59,7 @@ jobs:
           sops -d .secrets.yaml | yq -e '.secrets | to_entries[] | .key + "=" + .value' >> "$GITHUB_ENV"
       - name: run-integration-qemu
         env:
-          INTEGRATION_TEST_ARGS: --test.run TestIntegration/Suites/(CleanState|KernelArgsUpdate|MaintenanceUpgrade)$
+          INTEGRATION_TEST_ARGS: --test.run TestIntegration/Suites/(CleanState|KernelArgsUpdate|MaintenanceUpgrade|MaintenanceLifecycle)$
```
utkuozdemir, line 62: Wondering if we should merge `MaintenanceUpgrade` and `MaintenanceLifecycle` and just call it `MaintenanceLifecycle`.

client/pkg/client/management/management.go
```diff
@@ -312,6 +312,45 @@ func (client *Client) ReadAuditLog(ctx context.Context, req *management.ReadAudi
 	}
 }
 
+// MaintenanceLifecycle installs Talos to disk or upgrades the existing on-disk Talos on a machine running in maintenance mode.
+func (client *Client) MaintenanceLifecycle(
+	ctx context.Context,
+	machineID string,
+	operation management.MaintenanceLifecycleRequest_Operation,
+	version, disk string,
+) iter.Seq2[*management.MaintenanceLifecycleResponse, error] {
```
utkuozdemir, line 321: Good use case for sequence 🆒

internal/backend/grpc/management.go
utkuozdemir, line 1: I could just suggest moving the new code in this file into a separate file (in the same package). This file is too bloated imo.

internal/backend/grpc/management.go
```diff
@@ -113,18 +115,19 @@ func newManagementServer(cfg *config.Params, omniState state.State, jwtSigningKe
...
+	jwtSigningKeyProvider        JWTSigningKeyProvider
+	auditor                      AuditLogger
+	talosconfigProvider          TalosconfigProvider
+	talosRuntime                 TalosRuntime
+	logHandler                   *siderolinkinternal.LogHandler
+	logger                       *zap.Logger
+	dnsService                   *dns.Service
+	imageFactoryClient           *imagefactory.Client
+	omniconfigDest               string
+	maintenanceLifecycleInFlight sync.Map
```
utkuozdemir, line 130: Use the concurrent map in `gen` instead?

.github/workflows/e2e-misc-upgrades-cron.yaml
oguzkilcan: MaintenanceUpgrade is using the legacy upgrade api. MaintenanceLifecycle is using the new lifecycle api. I wanted to keep them separate to easily discontinue legacy api later.

internal/backend/grpc/management.go
oguzkilcan: I'll move them to their own package in my next PR, when I'm integrating lifecycle apis to Cluster operations.

Slessi (approved):
frontend ✅ 

machines list needs a redesign IMO. nothing broken there atm but can be made nicer

modals are ok, maybe upgrade/update can be merged, or just leave it as is and drop update later when legacy api is dropped

utkuozdemir (approved):
🚀

frontend/src/views/Machines/Machines.vue
```diff
@@ -66,6 +74,16 @@ const { data: infraProviderStatuses } = useResourceWatch<InfraProviderStatusSpec
   runtime: Runtime.Omni,
 })
 
+const { data: snapshots } = useResourceWatch<MachineStatusSnapshotSpec>({
```
Unix4ever: I think the better solution would be to add current machine status phase to the `MachineStatusLink` resource.
We compute it anyways already by merging link and machine status.

Not a big deal though.

internal/integration/talos_test.go
```diff
@@ -827,6 +828,170 @@ func AssertMachineShouldBeUpgradedInMaintenanceMode(
...
+		}
+
+		ctx, cancel := context.WithTimeout(testCtx, 15*time.Minute)
+		defer cancel()
+
+		omniClient := options.omniClient
+		omniState := omniClient.Omni().State()
+
+		// pick a machine in maintenance mode and does not yet have Talos installed on disk.
+		uninstalledMaintenance := func(machineStatus *omni.MachineStatus, _ []*omni.MachineStatus) bool {
```
Unix4ever:
```suggestion
		notInstalledMaintenance := func(machineStatus *omni.MachineStatus, _ []*omni.MachineStatus) bool {
```

internal/integration/talos_test.go
```diff
@@ -827,6 +828,170 @@ func AssertMachineShouldBeUpgradedInMaintenanceMode(
...
+		require.NoError(t, err)
+
+		parsedVersion.Pre = nil
+		parsedVersion.Build = nil
+
+		if parsedVersion.LT(maintenanceLifecycleMinTalosVersion) {
+			t.Skipf("maintenance lifecycle requires Talos %s or newer, upgrade target is %q", maintenanceLifecycleMinTalosVersion, targetVersion)
+		}
+
+		require.NotEmpty(t, *machineID, "machineID must be set by a preceding install assertion")
```
Unix4ever, line 933: I think to make this check complete we also need nil check here
```suggestion
		require.NotNil(t, machineID, "machineID must be set by a preceding install assertion")
		require.NotEmpty(t, *machineID, "machineID must be set by a preceding install assertion")
```

internal/backend/grpc/management.go
```diff
@@ -680,6 +679,471 @@ func (s *managementServer) MaintenanceUpgrade(ctx context.Context, req *manageme
...
+	default:
+		return status.Error(codes.InvalidArgument, "operation must be OPERATION_INSTALL or OPERATION_UPGRADE")
+	}
+
+	authCtx, _, err := s.checkAuthorization(ctx, req.MachineId, role.Operator)
+	if err != nil {
+		return err
+	}
+
+	ctx = actor.MarkContextAsInternalActor(authCtx)
```
Unix4ever: I wonder if we should check auth before doing any validation

oguzkilcan: We do request validation on other management APIs before auth. But it makes sense to do auth first.

internal/backend/grpc/management.go
```diff
@@ -680,6 +679,471 @@ func (s *managementServer) MaintenanceUpgrade(ctx context.Context, req *manageme
...
+		}
+	}
+
+	if _, loaded := s.maintenanceLifecycleInFlight.LoadOrStore(req.MachineId, struct{}{}); loaded {
+		return status.Errorf(codes.FailedPrecondition, "a maintenance lifecycle operation is already in progress for machine %q", req.MachineId)
+	}
+
+	defer s.maintenanceLifecycleInFlight.Delete(req.MachineId)
+
+	platform := machineStatus.TypedSpec().Value.GetPlatformMetadata().GetPlatform()
```
Unix4ever, line 790: This code can panic: if MachineStatus doesn't get populated with platform metadata.
We need to check that platform metadata is not nil

oguzkilcan: How can it panic using the getter?

internal/backend/grpc/management.go
```diff
@@ -680,6 +679,471 @@ func (s *managementServer) MaintenanceUpgrade(ctx context.Context, req *manageme
...
+		}
+
+		if sendErr := srv.Send(resp); sendErr != nil {
+			sendDisconnectLogOnce.Do(func() {
+				s.logger.Info(
+					"maintenance lifecycle client disconnected; operation continues server-side",
+					zap.String("machine_id", req.MachineId),
+					zap.Error(sendErr),
+				)
+			})
```
Unix4ever: I doubt we need goroutine-safe primitives here. It can be a simple boolean flag

internal/backend/grpc/management.go
Unix4ever: this is a protobuf generated getter, it doesn't protect you from getting nils in the middle
https://github.com/siderolabs/omni/blob/bfa805431868250f9ff7b72e4d58089fdb7f3dad/client/api/omni/specs/omni.pb.go#L1893-L1898

Unix4ever: actually ignore me. 2 getters in row can handle nils gracefully

utkuozdemir: I think getters are nil-safe, no?

internal/backend/grpc/management.go
utkuozdemir: Good point, we should.

Unix4ever (approved):

PR #2956, opened 2026-06-10
feat: allow opt-in skip of Kubernetes node audit

shanduur:
KubernetesNodeAuditController deletes any node not backed by a
ClusterMachine after one minute, which tears down virtual nodes
(e.g. VirtualKubelet) on every NotReady transition.

Add a two-part opt-out:
- omni.sidero.dev/node-audit-skip annotation on the Kubernetes node
  surfaces as ClusterSpec.Features.EnableNodeAuditSkip -> NodeStatus.SkipAudit
- ClusterSpec.Features.EnableNodeAuditSkip gates whether the annotation
  is honored

Gating on the cluster Feature prevents a workload with nodes/patch RBAC
from self-annotating to escape the audit; the cluster owner must enable
the feature in Omni first.

Signed-off-by: Mateusz Urbanek <mateusz.urbanek@siderolabs.com>

### DIFF
--- a/internal/backend/runtime/omni/controllers/omni/kubernetes_node_audit.go
+++ b/internal/backend/runtime/omni/controllers/omni/kubernetes_node_audit.go
@@ -66,9 +66,34 @@ func NewKubernetesNodeAuditController(getKubernetesClientFunc GetKubernetesClien
 				}
 
 				cluster := nodes.Metadata().ID()
-				actualNodes := xslices.Map(status.TypedSpec().Value.GetNodes(), func(nodeStatus *specs.KubernetesStatusSpec_NodeStatus) string {
-					return nodeStatus.Nodename
-				})
+
+				// EnableNodeAuditSkip is a cluster-level feature flag that must be set by the cluster owner
+				// in Omni before the omni.sidero.dev/node-audit-skip annotation on a Kubernetes node is honored.
+				// This prevents a workload with nodes/patch RBAC from hijacking the audit by self-annotating.
+				honorSkipAudit := false
+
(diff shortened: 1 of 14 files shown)


internal/backend/runtime/omni/controllers/omni/kubernetes_node_audit.go
```diff
@@ -66,9 +66,29 @@ func NewKubernetesNodeAuditController(getKubernetesClientFunc GetKubernetesClien
...
+				if err != nil && !state.IsNotFoundError(err) {
+					return fmt.Errorf("failed to get cluster: %w", err)
+				}
+
+				if clusterRes != nil {
+					honorSkipAudit = clusterRes.TypedSpec().Value.GetFeatures().GetEnableNodeAuditSkip()
+				}
+
+				actualNodes := xslices.Map(
+					xslices.Filter(status.TypedSpec().Value.GetNodes(), func(nodeStatus *specs.KubernetesStatusSpec_NodeStatus) bool {
```
Unix4ever: We could probably short circuit here if audit is disable for the whole cluster and skip map/filter calls

Unix4ever:
LGTM in general, but I wonder if we really need to gate it with another feature flag on the Cluster resource. What's the motivation for that?

smira:
Wild idea - what if we just label/annotate Nodes which should be exempt from the audit? No features needed in this case? Omni just ignores them completely (on the Kubernetes side)

shanduur:
Security boundary - without the feature flag, anyone who can patch Node objects (intentionally or via RBAC drift / compromise) could set the annotation and opt out of the audit, effectively bypassing the cleanup.

Unix4ever (approved):

PR #2983, opened 2026-06-16
feat: validate Talos extensions against the catalog

utkuozdemir:
Extension names on installation media configs, machine request sets, and extensions configurations are now validated against the Talos extensions catalog for the relevant Talos version. Unknown names, duplicates, and oversized lists are rejected. When no Talos version is set, the default version's catalog is used so the names still get checked. Names without a namespace are looked up under siderolabs/ so older clients that send the documented short form keep working.

The omnictl installation media create command also resolves short or partial extension names to canonical form before sending, replacing the client-side catalog check it used to do.

### DIFF
--- a/internal/backend/runtime/omni/validations/extensions.go
+++ b/internal/backend/runtime/omni/validations/extensions.go
@@ -1,0 +1,78 @@
+// Copyright (c) 2026 Sidero Labs, Inc.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+
+package validations
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"github.com/cosi-project/runtime/pkg/safe"
+	"github.com/cosi-project/runtime/pkg/state"
+
+	"github.com/siderolabs/omni/client/pkg/constants"
+	"github.com/siderolabs/omni/client/pkg/omni/resources/omni"
+	"github.com/siderolabs/omni/internal/backend/extensions"
+)
+
+// MaxExtensionsCount caps the number of entries in a repeated extensions list. The value is
+// arbitrary, picked well above what real callers would send. Bump if needed.
+const MaxExtensionsCount = 256
+
+// validateExtensions checks each requested extension name against the TalosExtensions catalog. An
+// empty list is accepted. An empty Talos version is the "automatic" sentinel used by resources
+// like InstallationMediaConfig, in which case the default Talos version's catalog is used.
+// Names without a slash are looked up under the official extensions namespace so older clients
+// that send the documented short form keep working. The list is rejected if it exceeds MaxExtensionsCount or
+// contains duplicate entries.
+func validateExtensions(ctx context.Context, st state.State, talosVersion string, names []string) error {
+	if len(names) == 0 {
+		return nil
+	}
+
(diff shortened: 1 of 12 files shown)


internal/backend/runtime/omni/validations/installation_media_config.go
```diff
@@ -78,7 +78,7 @@ func installationMediaConfigValidationOptions(st state.State) []validated.StateO
 			}
 		}
 
-		return nil
+		return validateExtensions(ctx, st, spec.GetTalosVersion(), spec.GetInstallExtensions())
```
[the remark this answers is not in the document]
utkuozdemir: addressed.

internal/backend/runtime/omni/validations/extensions.go
```diff
@@ -0,0 +1,77 @@
...
+	"github.com/cosi-project/runtime/pkg/safe"
+	"github.com/cosi-project/runtime/pkg/state"
+
+	"github.com/siderolabs/omni/client/pkg/constants"
+	"github.com/siderolabs/omni/client/pkg/omni/resources/omni"
+	"github.com/siderolabs/omni/internal/backend/extensions"
+)
+
+// MaxExtensionsCount caps the number of entries in a repeated extensions list.
+const MaxExtensionsCount = 64
```
frezbo: why this artificial limit?

utkuozdemir: The idea is to limit user inputs in general to reasonable numbers - other such examples: https://github.com/siderolabs/omni/blob/main/internal/backend/runtime/omni/validations/metadata.go#L21-L42

frezbo: but we've more than that many extensions 

```
❯ crane export ghcr.io/siderolabs/extensions:v1.13.4 | tar x -O image-digests | wc -l
83
```

majabojarska: It would be nice to have the rationale documented alongside this kind of magic numbers. Later on it might be difficult to remember how and why a value was chosen (well, apart from going back in the git log).

utkuozdemir: Tbh it's kinda arbitrary - how realistic is it to have more than 64 extensions (I don't want to make up a reason :))
What I was thinking was a reasonable number which would in real world we would most likely never see. And the reason for the limits is to act as a user-input size limit (we'll also have a limit at a lower, storage layer, but it's another thing). We discussed whether to have such limits or not in Slack, and there were no objections.

So if we document them, the explanation for the most would be simply "a reasonable, safe limit to cap the input size".

If you folk have a better number for this or for any other limits in mind, I am open to change them.

majabojarska: Knowing the choice was arbitrary is valuable information too! Not a hill I'll die on though, just something I've noticed.

utkuozdemir: Good point actually, I'm convinced - having the reason "**why**" as a comment would make it easier in the future if we ever need to bump them. Added it to all limits.

Also bumped the extensions limit to 256 to happily accommodate all of our extensions today.

oguzkilcan (approved):

PR #2987, opened 2026-06-16
fix: properly handle empty provider data in the common module

Unix4ever:
Looks like the behavior has changed from v3 -> v4 and now YAML parser crashes if the provider data is the empty string.

Update the common module to not try to parse the data if it's empty.

### DIFF
--- a/hack/test/provisionconfig.yaml
+++ b/hack/test/provisionconfig.yaml
@@ -2,6 +2,8 @@
 count: 15
 provider:
   id: talemu
+  data: |
+    secure_boot: false
 ---
 count: 15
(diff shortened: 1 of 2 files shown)


utkuozdemir (approved):

hack/test/provisionconfig.yaml
```diff
@@ -2,6 +2,8 @@
 count: 15
 provider:
   id: talemu
+  data: |
+    secure_boot: false
```
Unix4ever, line 6: Chicken and egg problem so making the tests happy for the previous state of the common module.

PR #2989, opened 2026-06-17
feat: install/upgrade maintenance-mode Talos during cluster create/scale-up

### LINKED ISSUE #2838: Allow using older versions of talos on cluster operations (creation/scaling)
Using the "install API" introduced in Talos 1.13, now we can allow users to install an ephemeral Talos, e.g., running from an ISO, to the disk.

When creating or scaling up a cluster, should use maintenance install code to allow users to be able to pick machines with different minor Talos versions. Currently, user can pick machines with a different minor version only if they are installed, since only those can be "maintenance-upgraded".

oguzkilcan:
Cluster creation and scale-up can now bring a maintenance-mode machine to the cluster's Talos version when they differ by a minor version.

Resolves: #2838

### DIFF
--- a/client/api/omni/specs/omni.proto
+++ b/client/api/omni/specs/omni.proto
@@ -618,6 +618,12 @@ message ClusterMachineConfigStatusSpec {
   //
   // Deprecated: use accessor methods GetUncompressedData/SetUncompressedData to manage this field.
   bytes compressed_redacted_machine_config = 10;
+
+  // PreRebootBootId is the machine boot ID captured when an operation that requires the
+  // machine to reboot (e.g. an install/upgrade via LifecycleService) was last issued. It
+  // debounces such operations: the controller waits until the machine reboots and its boot
+  // ID changes before acting again.
+  string pre_reboot_boot_id = 11;
 }
 
 message MachinePendingUpdatesSpec {
@@ -932,6 +938,8 @@ message MachineStatusSnapshotSpec {
 
   machine.MachineStatusEvent machine_status = 1;
   PowerStage power_stage = 2;
+  // BootId is the machine's kernel boot identifier (/proc/sys/kernel/random/boot_id).
+  string boot_id = 3;
 }
 
 enum ConditionType {
--- a/client/pkg/omni/resources/omni/version_compat.go
+++ b/client/pkg/omni/resources/omni/version_compat.go
@@ -1,0 +1,94 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+package omni
+
+import (
+	"strings"
+
+	"github.com/blang/semver/v4"
+	"github.com/cosi-project/runtime/pkg/resource"
+)
+
+// LifecycleServiceMinTalosVersion is the minimum Talos version that supports the LifecycleService API
+// tsgen:LifecycleServiceMinTalosVersion
+const LifecycleServiceMinTalosVersion = "1.13"
+
+// MachineCompatibleWithCluster reports whether a machine can join a cluster running clusterVersion.
+func MachineCompatibleWithCluster(
+	ms *MachineStatus,
+	clusterVersion semver.Version,
+) (ok bool, willAutoInstall bool, reason string) {
+	if ms.TypedSpec().Value.GetSchematic().GetInAgentMode() {
+		return true, false, ""
+	}
+
+	return MachineLabelsCompatibleWithCluster(ms.Metadata().Labels(), clusterVersion)
+}
+
+// MachineLabelsCompatibleWithCluster is a labels-only variant of MachineCompatibleWithCluster for callers with only a machine's labels (e.g. the MachineSetNode allocator).
+func MachineLabelsCompatibleWithCluster(labels *resource.Labels, clusterVersion semver.Version) (ok bool, willAutoInstall bool, reason string) {
+	versionStr, hasVersion := labels.Get(MachineStatusLabelTalosVersion)
+	if !hasVersion {
+		return false, false, "the machine has no Talos version label"
+	}
+
+	machineVersion, err := semver.Parse(strings.TrimPrefix(versionStr, "v"))
+	if err != nil {
+		return false, false, "the machine's Talos version could not be determined"
+	}
+
+	_, installed := labels.Get(MachineStatusLabelInstalled)
+
+	return versionCompatibleWithCluster(machineVersion, installed, clusterVersion)
+}
+
+// ParseTalosVersionLifecycleSupport parses a Talos version and reports whether it supports the LifecycleService install/upgrade APIs.
+func ParseTalosVersionLifecycleSupport(version string) (semver.Version, bool) {
+	parsed, err := semver.ParseTolerant(version)
+	if err != nil {
+		return semver.Version{}, false
+	}
+
+	return parsed, talosVersionSupportsLifecycleService(parsed)
+}
+
+// versionCompatibleWithCluster applies the version/installed compatibility rules to extracted inputs.
+func versionCompatibleWithCluster(machineVersion semver.Version, installed bool, clusterVersion semver.Version) (ok bool, willAutoInstall bool, reason string) {
+	// Refuse downgrade (e.g. installed at 1.14 trying to join a 1.13 cluster).
+	if machineVersion.Major > clusterVersion.Major || (machineVersion.Major == clusterVersion.Major && machineVersion.Minor > clusterVersion.Minor) {
+		return false, false, "the machine has a newer Talos version installed: downgrade is not allowed. Upgrade the machine or change the cluster Talos version"
+	}
+
+	if installed {
+		return true, false, ""
+	}
+
+	// Same major.minor: Omni applies config and Talos installs through the normal config path.
+	if machineVersion.Major == clusterVersion.Major && machineVersion.Minor == clusterVersion.Minor {
+		return true, false, ""
+	}
+
+	// Cross-minor and not-installed: Omni can support the install only if both sides are >= 1.13.
+	if talosVersionSupportsLifecycleService(machineVersion) && talosVersionSupportsLifecycleService(clusterVersion) {
+		return true, true, ""
+	}
+
+	return false, false, "the machine running from ISO or PXE must have the same major and minor version as the cluster " +
+		"it is going to be added to. Please use another ISO or change the cluster Talos version"
+}
+
+// talosVersionSupportsLifecycleService reports whether the Talos version exposes the LifecycleService install/upgrade APIs.
+func talosVersionSupportsLifecycleService(v semver.Version) bool {
+	minTalosVersion, _ := semver.ParseTolerant(LifecycleServiceMinTalosVersion) //nolint:errcheck
+	if v.Major > minTalosVersion.Major {
+		return true
+	}
+
+	if v.Major < minTalosVersion.Major {
+		return false
+	}
+
+	return v.Minor >= minTalosVersion.Minor
+}
--- a/frontend/src/methods/compat.ts
+++ b/frontend/src/methods/compat.ts
@@ -1,0 +1,96 @@
+// Copyright (c) 2026 Sidero Labs, Inc.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+
+import { parse, SemVer } from 'semver'
+
+import type { Resource } from '@/api/grpc'
+import type { MachineStatusSpec } from '@/api/omni/specs/omni.pb'
+import { LifecycleServiceMinTalosVersion, MachineStatusLabelInstalled } from '@/api/resources'
+
+const [minTalosMajor, minTalosMinor] = LifecycleServiceMinTalosVersion.split('.').map(Number)
+
+export type CompatResult = {
+  ok: boolean
+  willAutoInstall: boolean
+  reason: string | null
+}
+
+const okResult: CompatResult = { ok: true, willAutoInstall: false, reason: null }
+const okWithInstall = (version: string): CompatResult => ({
+  ok: true,
+  willAutoInstall: true,
+  reason: `Omni will install Talos v${version} to this machine when it joins the cluster.`,
+})
+
+function supportsMaintenanceInstall(v: SemVer): boolean {
+  if (v.major > minTalosMajor) return true
+  if (v.major < minTalosMajor) return false
+  return v.minor >= minTalosMinor
+}
+
+export function machineCompatibleWithCluster(
+  machine: Resource<MachineStatusSpec>,
+  clusterVersionStr: string,
+): CompatResult {
+  const clusterVersion = parse(clusterVersionStr)
+  const machineVersion = parse(machine.spec.talos_version)
+
+  if (!machineVersion) {
+    return {
+      ok: false,
+      willAutoInstall: false,
+      reason: "The machine's Talos version could not be determined.",
+    }
+  }
+
+  if (!clusterVersion) {
+    return {
+      ok: false,
+      willAutoInstall: false,
+      reason: 'The cluster Talos version could not be determined.',
+    }
+  }
+
+  const inAgentMode = !!machine.spec.schematic?.in_agent_mode
+  if (inAgentMode) {
+    return okResult
+  }
+
+  // Refuse downgrade.
+  if (
+    machineVersion.major > clusterVersion.major ||
+    (machineVersion.major === clusterVersion.major && machineVersion.minor > clusterVersion.minor)
+  ) {
+    return {
+      ok: false,
+      willAutoInstall: false,
+      reason:
+        'The machine has a newer Talos version installed: downgrade is not allowed. Upgrade the machine or change the cluster Talos version.',
+    }
+  }
+
+  const installed = machine.metadata.labels?.[MachineStatusLabelInstalled] !== undefined
+  if (installed) {
+    return okResult
+  }
+
+  if (
+    machineVersion.major === clusterVersion.major &&
+    machineVersion.minor === clusterVersion.minor
+  ) {
+    return okResult
+  }
+
+  if (supportsMaintenanceInstall(machineVersion) && supportsMaintenanceInstall(clusterVersion)) {
+    return okWithInstall(clusterVersionStr.replace(/^v/, ''))
+  }
+
+  return {
+    ok: false,
+    willAutoInstall: false,
+    reason:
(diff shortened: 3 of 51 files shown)


internal/backend/runtime/talos/clients.go
```diff
@@ -126,6 +126,18 @@ func (c *Client) Connected(ctx context.Context, r controller.Reader) (bool, erro
...
+func NewMaintenanceClient(ctx context.Context, address string) (*client.Client, error) {
+	opts := GetSocketOptions(address)
+	opts = append(
+		opts,
+		client.WithTLSConfig(&tls.Config{InsecureSkipVerify: true}),
+		client.WithEndpoints(address),
+	)
+
+	return client.New(ctx, opts...)
+}
```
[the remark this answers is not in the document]
oguzkilcan: Will fix.

internal/backend/runtime/omni/controllers/omni/machineconfig/status.go
```diff
@@ -477,6 +494,85 @@ func (ctrl *ClusterMachineConfigStatusController) upgrade(
...
+
+	switch {
+	case errors.Is(err, lifecycle.ErrAlreadyInFlight):
+		// The management RPC is operating on this machine; retry once it releases the slot.
+		logger.Info("maintenance lifecycle already in flight", zap.String("machine", machineID))
+
+		return false, controller.NewRequeueInterval(maintenanceLifecyclePollInterval)
+	case err != nil:
+		rc.machineConfigStatus.TypedSpec().Value.LastConfigError = lifecycle.WrapErr(err, "").Error()
+
```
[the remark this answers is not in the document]
oguzkilcan: I forgot to add the prefix to WrapErr. Will fix.

internal/backend/talos/lifecycle/types.go
```diff
@@ -0,0 +1,76 @@
...
+	if err == nil {
+		return nil
+	}
+
+	if s, ok := status.FromError(err); ok {
+		return status.Errorf(s.Code(), "%s: %s", prefix, s.Message())
+	}
+
+	return status.Errorf(codes.Internal, "%s: %v", prefix, err)
+}
```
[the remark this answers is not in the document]
oguzkilcan: Will fix

internal/backend/runtime/omni/controllers/omni/machineconfig/status.go
```diff
@@ -477,6 +494,85 @@ func (ctrl *ClusterMachineConfigStatusController) upgrade(
...
+	if ctrl.lifecycleManager == nil {
+		return false, errors.New("lifecycle manager not configured")
+	}
+
+	machineID := rc.ID()
+
+	if rc.bootID == "" {
+		logger.Info("waiting for machine boot ID before running maintenance lifecycle operation", zap.String("machine", machineID))
+
+		return false, controller.NewRequeueInterval(maintenanceLifecyclePollInterval)
```
Unix4ever: I don't think we need to requeue as the controller will wake up on bootID changes.

internal/backend/runtime/omni/controllers/omni/machineconfig/status.go
```diff
@@ -477,6 +494,85 @@ func (ctrl *ClusterMachineConfigStatusController) upgrade(
...
+	}
+
+	lastRebootID := rc.machineConfigStatus.TypedSpec().Value.LastRebootId
+	if rc.bootID == lastRebootID {
+		logger.Info("waiting for machine to reboot after maintenance lifecycle operation", zap.String("machine", machineID))
+
+		return false, controller.NewRequeueInterval(maintenanceLifecyclePollInterval)
```
Unix4ever, line 536: Same story here. BootID should be updated after a resource change.

internal/backend/runtime/omni/controllers/omni/machineconfig/status.go
oguzkilcan: Will replace it with skip reconcile.

internal/backend/runtime/omni/controllers/omni/machineconfig/status.go
```diff
@@ -477,6 +494,85 @@ func (ctrl *ClusterMachineConfigStatusController) upgrade(
...
+	}))
+
+	switch {
+	case errors.Is(err, lifecycle.ErrAlreadyInFlight):
+		// The management RPC is operating on this machine; retry once it releases the slot.
+		logger.Info("maintenance lifecycle already in flight", zap.String("machine", machineID))
+
+		return false, controller.NewRequeueInterval(maintenanceLifecyclePollInterval)
+	case err != nil:
+		rc.machineConfigStatus.TypedSpec().Value.LastConfigError = lifecycle.WrapErr(err, "maintenance lifecycle failed").Error()
```
Unix4ever: The error sounds cryptic. Maybe we could make it say update or install failed

frontend/src/views/Overview/components/OverviewContent.vue
```diff
@@ -201,6 +217,14 @@ const machineLockedForSecretRotation = computed(() => {
         All operations on this cluster are currently disabled. Config patches can be created,
         updated or deleted but these changes will not be applied while the cluster is locked.
       </TAlert>
+      <TAlert
+        v-if="machinesWithErrors.length > 0"
+        :title="`${machinesWithErrors.length} ${machinesWithErrors.length === 1 ? 'machine has' : 'machines have'} configuration errors`"
+        type="error"
+        class="mb-4"
+      >
+        Open the machine detail page for each affected node to view the error and recover.
```
Slessi: We can include be a link to the machine detail page we are directing them to. That said im not sure if "machine detail page" is clear, maybe "cluster machine detail page" ? because on the non-cluster version it does not show.

Alternatively we can just list the errors out here per machine, we have the data there already anyway.

internal/backend/runtime/omni/controllers/omni/machineconfig/status.go
```diff
@@ -477,6 +494,84 @@ func (ctrl *ClusterMachineConfigStatusController) upgrade(
...
+	}))
+
+	switch {
+	case errors.Is(err, lifecycle.ErrAlreadyInFlight):
+		// The management RPC is operating on this machine; retry once it releases the slot.
+		logger.Info("maintenance lifecycle already in flight", zap.String("machine", machineID))
+
+		return false, controller.NewRequeueInterval(maintenanceLifecyclePollInterval)
+	case err != nil:
+		rc.machineConfigStatus.TypedSpec().Value.LastConfigError = lifecycle.WrapErr(err, fmt.Sprintf("%s failed", operationName)).Error()
```
Unix4ever, line 582: I _think_ `LastConfigError` won't be updated on the resource if we return the error from this method.
As it skips updating the resource in case of failure.

oguzkilcan: Good catch.

frontend/src/views/Clusters/Management/ClusterMachineItem.vue
```diff
@@ -145,6 +147,8 @@ const options = computed(() => {
     if (versionMismatch) {
       disabled = true
       tooltip = versionMismatch
+    } else if (autoInstallNotice && !tooltip) {
```
Slessi: Is this `autoInstallNotice` not relevant if there is a `versionMisMatch`? the `!tooltip` feels odd also

We check multiple conditions and always overwrite the last checked condition, but then this one we dont overwrite. Feels to me like we could just show a tooltip that includes content from all matching conditions

frontend/src/views/ClusterMachines/ClusterMachine.vue
```diff
@@ -121,6 +121,10 @@ const updateLock = async () => {
     </div>
 
     <div class="flex items-center justify-end">
+      <Tooltip v-if="machine.spec.last_config_error" :description="machine.spec.last_config_error">
+        <TIcon icon="error" class="mx-1.5 h-4 w-4 text-red-400" />
+      </Tooltip>
+
```
Slessi, line 127: Adding this extra icon here compresses things next to it if there are >2 items, so i think we need to increase the reserved space here or find another home for this icon

Image

frontend/src/methods/compat.ts
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  if (
+    machineVersion.major === clusterVersion.major &&
+    machineVersion.minor === clusterVersion.minor
+  ) {
+    return okResult
+  }
+
+  if (supportsMaintenanceInstall(machineVersion) && supportsMaintenanceInstall(clusterVersion)) {
```
Slessi, line 86: the `supportsMaintenanceInstall(clusterVersion)` is because of that thing u showed in the call where u tried to use maintenance install on a 1.13 to install 1.9?

oguzkilcan: No, it's checking whether both cluster and machine being added supports LifecycleService APIs that's used by maintenance install,

Unix4ever (approved):

Slessi (approved):
frontend LGTM

@oguzkilcan i made a change to add the extra spacing change i mentioned, and updated stories, FYI

client/api/omni/specs/omni.proto
```diff
@@ -932,6 +934,8 @@ message MachineStatusSnapshotSpec {
 
   machine.MachineStatusEvent machine_status = 1;
   PowerStage power_stage = 2;
+  // boot_id is the machine's kernel boot identifier (/proc/sys/kernel/random/boot_id).
```
utkuozdemir:
```suggestion
  // BootId is the machine's kernel boot identifier (/proc/sys/kernel/random/boot_id).
```

client/api/omni/specs/omni.proto
```diff
@@ -618,6 +618,8 @@ message ClusterMachineConfigStatusSpec {
   //
   // Deprecated: use accessor methods GetUncompressedData/SetUncompressedData to manage this field.
   bytes compressed_redacted_machine_config = 10;
+
+  string last_reboot_id = 11;
```
utkuozdemir: This name is not clear, also, I use the same name for randomly-generated uuids in BM provider related stuff - a uuid generated by Omni completely.

I'd suggest:
- adding a godoc to clarify its purpose
- renaming it to `boot_id_before_last_reboot` or sth like that.

client/api/omni/specs/omni.proto
oguzkilcan: Name of the variable is boot_id, do you want me to change that?

client/pkg/omni/resources/omni/version_compat.go
```diff
@@ -0,0 +1,103 @@
...
+
+// LifecycleServiceMinTalosVersion is the minimum Talos version that supports the LifecycleService API
+// tsgen:LifecycleServiceMinTalosVersion
+const LifecycleServiceMinTalosVersion = "1.13"
+
+// MachineCompatibleWithCluster reports whether a machine can join a cluster running clusterVersion.
+func MachineCompatibleWithCluster(
+	ms *MachineStatus,
+	clusterVersion semver.Version,
+) (ok bool, willAutoInstall bool, reason string) {
```
utkuozdemir, line 22: Does this func exist only to check agent mode? Because otherwise you could remove this and just use `MachineLabelsCompatibleWithCluster` below (pass labels of MachineStatus). It'd be simpler.

If you certainly want to keep it, after the agent mode check, you can call `MachineLabelsCompatibleWithCluster` from here to not have code duplication.

internal/backend/talos/lifecycle/manager.go
```diff
@@ -0,0 +1,193 @@
...
+func WithAudit(fn AuditFunc) Option {
+	return func(c *runConfig) {
+		if fn != nil {
+			c.audit = fn
+		}
+	}
+}
+
+// CheckTalosVersion verifies the machine runs a Talos version exposing the LifecycleService APIs.
+func (m *Manager) CheckTalosVersion(version string) error {
```
utkuozdemir: I'd suggest renaming this function to `SupportsLifecycleManagement` or something like that - the name itself is not clear about what it actually does.

internal/backend/talos/lifecycle/lifecycle.go
utkuozdemir, line 1: We usually put such structs into a file named after the package, i.e., `lifecycle.go`.

client/api/omni/specs/omni.proto
utkuozdemir: no, the go code it generates will have camel-case, it'll be `BootId`, and the godoc comment should start exactly with the field name.

internal/backend/runtime/omni/controllers/omni/internal/task/snapshot/snapshot.go
```diff
@@ -136,6 +141,34 @@ func (spec CollectTaskSpec) RunTask(ctx context.Context, _ *zap.Logger, notifyCh
 	}
 }
 
+// readBootID reads the machine's kernel boot identifier from the node. It is best-effort: on any error it returns "".
+func readBootID(ctx context.Context, c *client.Client, logger *zap.Logger) string {
```
utkuozdemir: Can you add **why** it is best-effort to the comment. Also, **should** it be best-effort, actually? How would things break/behave when empty string is returned?

internal/backend/runtime/omni/controllers/omni/machineconfig/status.go
```diff
@@ -57,19 +57,20 @@ const (
...
 	ongoingResets    *ongoingResets
+	lifecycleManager *lifecycle.Manager
 	imageFactoryHost string
 	talosRegistry    string
 	acquireLockMu    sync.Mutex
 }
 
 // NewClusterMachineConfigStatusController initializes ClusterMachineConfigStatusController.
-func NewClusterMachineConfigStatusController(imageFactoryHost, talosRegistry string) *ClusterMachineConfigStatusController {
+func NewClusterMachineConfigStatusController(imageFactoryHost, talosRegistry string, lifecycleManager *lifecycle.Manager) *ClusterMachineConfigStatusController {
```
utkuozdemir: I could suggest making this accept an interface, passing a mock implementation from `status_test.go`, remove nil checks from this file (I mean the `if ctrl.lifecycleManager == nil`), and unit testing that code path as well. This is a critical controller, we must cover it as much as we can (maybe that test can even find bugs :))

utkuozdemir: I tell this especially because the new code path acquires upgrade locks - we must should make sure that any acquired lock gets released at all relevant exit points - otherwise we can get hit by bugs last the last deadlock one.

internal/backend/runtime/omni/controllers/omni/secrets/rotation_status_test.go
```diff
@@ -124,7 +124,7 @@ func Test_TalosCARotation(t *testing.T) {
 				))
 				require.NoError(t, testContext.Runtime.RegisterQController(secretsctrl.NewSecretsController(nil)))
 				require.NoError(t, testContext.Runtime.RegisterQController(omnictrl.NewClusterMachineConfigController("test.factory", nil, "ghcr.io/siderolabs/installer", omnicfg.Registries{})))
-				require.NoError(t, testContext.Runtime.RegisterQController(machineconfig.NewClusterMachineConfigStatusController("test.factory", "ghcr.io/siderolabs/installer")))
+				require.NoError(t, testContext.Runtime.RegisterQController(machineconfig.NewClusterMachineConfigStatusController("test.factory", "ghcr.io/siderolabs/installer", nil)))
```
utkuozdemir: See my comment for this file - I'm not a big fan of passing a crucial component as nil to this controller from the tests. Let's have a common mock implementation and pass it in every test?

internal/backend/runtime/talos/clients.go
```diff
@@ -126,6 +126,18 @@ func (c *Client) Connected(ctx context.Context, r controller.Reader) (bool, erro
 	return clusterStatus.TypedSpec().Value.GetAvailable(), nil
 }
 
+// NewMaintenanceClient opens an insecure Talos client to a machine's maintenance API.
+func NewMaintenanceClient(ctx context.Context, address string) (*client.Client, error) {
```
utkuozdemir, line 130: I'd suggest not using this (can remove), and instead, passing TalosClientFactory to its callers, and calling `GetMaintenance` on it - it makes sure that it gives you maintenance clients, and they are cached :)

internal/backend/runtime/omni/controllers/omni/machineconfig/reconciliation_context.go
```diff
@@ -36,11 +53,18 @@ type ReconciliationContext struct {
...
 	redactedMachineConfig []byte
 
-	shouldUpgrade        bool
 	configUpdatesAllowed bool
 	locked               bool
+	lifecycleOp          LifecycleOp
+}
+
+// shouldUpgrade returns true when an upgrade/install action needs to run before config can be applied.
+func (rc *ReconciliationContext) shouldUpgrade() bool {
```
utkuozdemir: we use `shouldUpgrade` to mean other things - maybe rename this to be `hasPendingLifecycleOperation` or something like that?

internal/backend/runtime/omni/controllers/omni/machineconfig/reconciliation_context.go
```diff
@@ -224,7 +250,51 @@ func BuildReconciliationContext(ctx context.Context, r controller.Reader,
...
+
+// DecideLifecycleOp picks the upgrade/install path from the live machine state. Exported for tests.
+func DecideLifecycleOp(machineStatus *omni.MachineStatus, installImage *specs.MachineConfigGenOptionsSpec_InstallImage, schematicMismatch, talosVersionMismatch bool) LifecycleOp {
+	hasSystemDisk := omni.GetMachineStatusSystemDisk(machineStatus) != ""
+
+	machineVersion, machineSupportsLifecycle := omni.ParseTalosVersionLifecycleSupport(machineStatus.TypedSpec().Value.TalosVersion)
+	targetVersion, targetSupportsLifecycle := omni.ParseTalosVersionLifecycleSupport(installImage.TalosVersion)
+
+	// Maintenance path: both the machine and the target must support the LifecycleService API.
+	if machineStatus.TypedSpec().Value.Maintenance && machineSupportsLifecycle && targetSupportsLifecycle {
```
utkuozdemir, line 271: nit: can extract this expression into a variable for better readability

internal/backend/runtime/omni/controllers/omni/machineconfig/reconciliation_context.go
oguzkilcan: Legacy upgrade is not utilizing lifecycle API but I'll update is as such.

utkuozdemir (approved):

Issue #2972, opened 2026-06-14
omni does not clean up leftover machineextensionsstatuses resources


fsgh42:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

I created a cluster in February. This cluster has since been destroyed.
Yesterday, I created a fresh cluster with the same name and the same nodes.

The machines were originally provisioned with a schematic, that contained extensions such as `tailscale`. This caused the machines to install these extensions every time I had them join a cluster, even if those extensions were not defined in the cluster-template. This is an issue because extensions that require a config to reach `ready` state, such as the tailscale extensions, can keep a worker or controlplane from reaching ready state.

I tried to hack my around having to physically reprovision these machines by issuing this first:
```
$ talosctl-linux-amd64-v1.12.5 -n <uuid> upgrade --image https://factory.talos.dev/metal-installer/<schematicID>:1.13.4 --wait=false
```
Using talosctl v1.12.5 on purpose - I can't remember why but at the time for some reason the talosctl v1.13.x binary I had installed wasn't working.
Any way, at the time, this command succeeded and reprovisioned these nodes with a schematic that did not contain those extensions. I confirmed in the UI that they did in fact not have these installed.

When I then joined these nodes to the (similarily named cluster) yesterday, the extensions were reinstalled, which seems like a bug.
I did a little digging:
```
$ omnictl get machineextensionsstatuses <uuid> -oyaml
metadata:
    namespace: default
    type: MachineExtensionsStatuses.omni.sidero.dev
    id: <uuid>
    version: 29
    owner: SchematicConfigurationController
    phase: running
    created: 2026-02-26T19:35:18Z
    updated: 2026-06-13T15:47:17Z
    labels:
        omni.sidero.dev/cluster: <name>
        omni.sidero.dev/machine-set: <name>-worker
        omni.sidero.dev/role-worker:
spec:
    extensions:
        - name: siderolabs/amd-ucode
          immutable: false
          phase: 0
        - name: siderolabs/intel-ucode
          immutable: false
          phase: 0
        - name: siderolabs/iscsi-tools
          immutable: false
          phase: 0
        - name: siderolabs/tailscale
          immutable: false
          phase: 0
        - name: siderolabs/util-linux-tools
          immutable: false
          phase: 0
    talosversion: v1.13.4
```
[...]

PR #2990, opened 2026-06-17
fix: allow empty list of extensions in cluster templates

utkuozdemir:
Cluster templates could not distinguish an unset extensions list from an explicitly empty one, so extensions could not be cleared through a template. The list is now optional, like kernel arguments: omitting it keeps the current extensions, while setting it to empty clears them. The exporter preserves this distinction too.

Also, a machine removed from a cluster kept its cluster and machine set labels on the extensions status, because label updates never dropped stale entries. They are now synchronized on each update.

Part of siderolabs/omni#2972.

### DIFF
--- a/client/pkg/template/internal/models/models.go
+++ b/client/pkg/template/internal/models/models.go
@@ -38,18 +38,19 @@ type TranslateContext struct {
 	FileContext
 	LockedMachines            map[MachineID]struct{}
 	MachineDescriptors        map[MachineID]Descriptors
-	MachineSetLevelKernelArgs map[MachineID]KernelArgs
-	ClusterLevelKernelArgs    KernelArgs
+	MachineSetLevelKernelArgs map[MachineID]OptionalList
 	ClusterName               string
+	ClusterLevelKernelArgs    OptionalList
 }
 
-// SystemExtensions is embedded in Cluster, MachineSet and Machine objects.
-type SystemExtensions struct {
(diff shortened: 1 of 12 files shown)


oguzkilcan (approved):

client/pkg/template/internal/models/models.go
```diff
@@ -44,12 +44,16 @@ type TranslateContext struct {
 }
 
 // SystemExtensions is embedded in Cluster, MachineSet and Machine objects.
+//
+// The value is a pointer so an unset list can be told apart from an explicitly empty one. An unset
+// list leaves the machine's current extensions untouched, while an empty list clears them instead of
+// falling back to the set of extensions discovered when the machine first connected.
 type SystemExtensions struct {
-	SystemExtensions []string `yaml:"systemExtensions,omitempty"`
+	SystemExtensions *[]string `yaml:"systemExtensions,omitempty"`
```
Unix4ever: I don't quite like pointer to slice here. Maybe we could use a custom type there instead?

Not insisting though.

Unix4ever (approved):

client/pkg/template/internal/models/models.go
utkuozdemir: Sure, `KernelArgs` was also like this, also changed it. Introduced the new `OptionalList` type. Turns out there's a `IsZeroer` interface which is recognized respected by YAML library, which makes such clean implementations possible.

PR #3003, opened 2026-06-19
release(v1.9.0-beta.0): prepare release

oguzkilcan:
This is the official v1.9.0-beta.0 release.


shanduur (approved):

hack/release.toml
```diff
@@ -6,220 +6,128 @@ github_repo = "siderolabs/omni"
...
 """
 
-[notes.talos-end-of-support]
-title = "Talos Version End-of-Support Notifications"
+[notes.maintenance-install-upgrade]
+title = "Install and Upgrade Talos in Maintenance Mode"
 description = """\
-Omni now tracks machines running Talos versions approaching or past end of support relative to `MinTalosVersion`, \
-emits two new notifications (approaching end of support, end of support reached), and exposes Prometheus metrics for both.
+A new streaming management API installs or upgrades Talos on machines booted in maintenance mode. It comes with `omnictl` install and upgrade \
```
smira, line 23: probably Talos 1.13+

hack/release.toml
```diff
@@ -6,220 +6,128 @@ github_repo = "siderolabs/omni"
...
 """
 
-[notes.audit-log-filters]
-title = "Additional Audit Log Filters"
+[notes.cluster-healthchecks]
+title = "Per-Cluster Kubernetes Healthchecks"
 description = """\
-Audit logs gain a generic search box and sortable columns in the UI, plus CLI filters for `event_type`, `resource_type`, `resource_id`, `cluster_id`, and `actor`.
+Omni gains per-cluster Kubernetes healthchecks. You define them as `KubernetesHealthCheck` resources, either declared in a cluster template or applied \
+directly with the cluster label set. Omni runs each one before it upgrades a node during a Talos upgrade and retries a failed job until it passes. \
```
smira: I think this sounds confusing (I understand what it is), but I would rather leave the "cluster templates" option for the release notes

I think we need to expand to say that a healthcheck is a Kubernetes Job configured by the user, and if the job succeeds, the healthcheck is "green".

Also we need to say when they run, and which operations they gate

hack/release.toml
```diff
@@ -6,220 +6,128 @@ github_repo = "siderolabs/omni"
...
+A new graphical view shows KubeSpan peer status for a cluster machine.
 """
 
-[notes.frontend-resave-support-bundle]
-title = "Re-Saving the Omni Support Bundle"
+[notes.supply-chain]
+title = "Signed Images and SBOM Release Artifacts"
 description = """\
-The frontend now allows re-saving a previously generated Omni support bundle without regenerating it.
+Omni releases now ship an SBOM built from the Go modules as a release artifact, and Omni signs the published container images during release.
```
smira:
```suggestion
Omni releases now ship an SBOM built from the Go modules as a release artifact, and Sidero Labs signs the published container images during release.
```

hack/release.toml
utkuozdemir: Btw, Artem is going to write docs for this, we can add a link to that page when that is merged (doesn't need to be in this beta release).

PR #3036, opened 2026-06-24
fix: prevent removal of node unique tokens that still have a link

utkuozdemir:
The node unique token cleanup removes tokens that have no associated link for longer than a grace period. It looked up the link through the cached reader, which is populated by a separate watch and can trail real state. On a loaded runner the link could already exist while being invisible through the cache, so a token that had just regained its link was wrongly treated as orphaned and torn down.

Read the link bypassing the cache so an existing link is always observed even when the watch is behind. This fixes the flaky cleanup test that timed out when the recreated token got removed.

### DIFF
--- a/internal/backend/runtime/omni/controllers/omni/node_unique_token_cleanup.go
+++ b/internal/backend/runtime/omni/controllers/omni/node_unique_token_cleanup.go
@@ -21,6 +21,7 @@ import (
 	"github.com/siderolabs/omni/client/pkg/cosi/helpers"
 	"github.com/siderolabs/omni/client/pkg/omni/resources"
 	"github.com/siderolabs/omni/client/pkg/omni/resources/siderolink"
+	"github.com/siderolabs/omni/internal/backend/runtime/omni/controllers/uncached"
 )
 
 // NodeUniqueTokenCleanupControllerName is the name of the NodeUniqueTokenCleanupController.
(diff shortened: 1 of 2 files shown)


oguzkilcan (approved):

Unix4ever (approved):

shanduur (approved):

smira:
I see the intention to fix flaky unit-tests, but this is making performance worse to address unit-test (which could be addressed with retries in the test).

These two resources don't have a proper link, so that there's still a reverse race that a machine can be let in while the token is being cleaned up.

I think uncached readers to fix the test is the wrong way, unless this issue actually fixes something in production.

I don't think with the current design of two races a race can be prevented in general.

utkuozdemir:
> I see the intention to fix flaky unit-tests, but this is making performance worse to address unit-test (which could be addressed with retries in the test).
> 
> These two resources don't have a proper link, so that there's still a reverse race that a machine can be let in while the token is being cleaned up.
> 
> I think uncached readers to fix the test is the wrong way, unless this issue actually fixes something in production.
> 
> I don't think with the current design of two races a race can be prevented in general.

After discussing this at the standup and talking with @Unix4ever about it, I merged it as-is.
I also left the `orphanTTL` in place, because it serves an another case than this race: `NodeUniqueToken` is created before the `Link` in SideroLink provisioning code, so there's a real time window between their creation. We don't want the controller to immediately wake up and see that the Link is not there yet and clean up the token.

I just wanted to address this race without rearchitecting things much. For performance, what I could do was: to the regular (cached) read, and if it returned not found, do the same read but uncached this time to make sure that the observation was correct. We discussed this with @Unix4ever, and since it's a single GetByID, and the inputs wouldn't get updated too often, decided to skip that (if you disagree, I can do it that way).

We can decide to rework this (turn this into a `NodeUniqueTokenController`, not a cleanup-only controller), but that's a big change I wanted to skip.

PR #3040, opened 2026-06-24
feat(frontend): add more visual distinction for offline kubespan peers

Slessi:
Add more visual distinction between online and offline peers in the KubeSpan view. Offline nodes will have hollow squares and dashes lines, in addition to being red. This makes it more clear for red/green colorblind users.

image


smira (approved):
❤️

Unix4ever (approved):

PR #3043, opened 2026-06-24
fix: move timeout for factory requests to controllers

Slessi:
Scan requests to factory take longer than 30 seconds and are timing out from the backend's baked in 30 second timeout. Moved the timeouts up to their controller calls to cater for its original purpose, but without affecting other factory calls.

### DIFF
--- a/internal/backend/imagefactory/client.go
+++ b/internal/backend/imagefactory/client.go
@@ -20,7 +20,8 @@ import (
 
 // requestTimeout caps every call to the image factory. Without it, the factory's HTTP client has
 // no internal timeout, so a stuck connection would pin a controller reconcile slot indefinitely.
-const requestTimeout = 30 * time.Second
+// 30 minutes as the scan report request can be particularly slow.
+const requestTimeout = 30 * time.Minute
 
 // serverSnifferTransport wraps an http.RoundTripper and records whether the image factory
 // identifies itself as an Enterprise instance via the Server response header.
--- a/internal/backend/runtime/omni/controllers/omni/machine_status.go
+++ b/internal/backend/runtime/omni/controllers/omni/machine_status.go
@@ -11,6 +11,7 @@ import (
 	"maps"
 	"net/http"
 	"strings"
+	"time"
 
 	"github.com/cosi-project/runtime/pkg/controller"
 	"github.com/cosi-project/runtime/pkg/controller/generic"
@@ -361,7 +362,11 @@ func (ctrl *MachineStatusController) populateSchematicRaw(ctx context.Context, i
 
 		logger.Info("machine does have a schematic ID but no raw schematic, get it from image factory")
 
-		sch, err := ctrl.ImageFactoryClient.SchematicGet(ctx, info.FullID)
+		factoryCtx, cancel := context.WithTimeout(ctx, time.Second*30)
+		sch, err := ctrl.ImageFactoryClient.SchematicGet(factoryCtx, info.FullID)
+
+		cancel()
+
 		if err != nil {
 			if factoryclient.IsHTTPErrorCode(err, http.StatusNotFound) {
 				logger.Warn("raw schematic not found in image factory", zap.String("host", ctrl.ImageFactoryClient.Host()))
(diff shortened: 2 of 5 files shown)


frezbo (approved):

smira (commented):
imho it's overcomplication, and client.Timeout should not be used at all

the caller should set its own deadline on the context as it talks to the Image Factory (or anything else).

if the frontend talks to IF via Omni, the frontend sets its own timeout, and Omni will cancel the request IF via context cancelation via the client.

I don't see why backend should set the pre-defined timeouts.

Slessi:
@smira seems to be originally because of this comment:

```
// requestTimeout caps every call to the image factory. Without it, the factory's HTTP client has
// no internal timeout, so a stuck connection would pin a controller reconcile slot indefinitely.
```

i can try to change it to your version, i guess backend calls can also themselves set their timeouts

smira:
> @smira seems to be originally because of this comment:
> 
> ```
> // requestTimeout caps every call to the image factory. Without it, the factory's HTTP client has
> // no internal timeout, so a stuck connection would pin a controller reconcile slot indefinitely.
> ```
> 
> i can try to change it to your version, i guess backend calls can also themselves set their timeouts

yep, this makes sense, but the controller should set a timeout on the context when it talks to the client, much cleaner imho

internal/backend/imagefactory/client.go
```diff
@@ -12,16 +12,11 @@ import (
...
 	"sync/atomic"
-	"time"
 
 	"github.com/siderolabs/image-factory/pkg/client"
 	"github.com/siderolabs/image-factory/pkg/schematic"
 )
 
-// requestTimeout caps every call to the image factory. Without it, the factory's HTTP client has
-// no internal timeout, so a stuck connection would pin a controller reconcile slot indefinitely.
-const requestTimeout = 30 * time.Second
```
Slessi, line 23: @utkuozdemir I removed the timeout you added [here](https://github.com/siderolabs/omni/pull/2927/changes#diff-2364e8346692d48dec645ba102d865295dd25f842629aa36f45fcd79f8acd185)

It was making scan requests from factory (which take long) time-out. I moved the timeouts up manually to the controller call sites.

Originally I just added a separate timeout in here for scan sites but andrey made a [good point](https://github.com/siderolabs/omni/pull/3043#pullrequestreview-4564048944) and i dont think hardcoding a timeout here makes sense.

WDYT?

utkuozdemir: I added it as a last resort protection in case the call sites forget to add it - we had issues with forgetting to set timeouts to external API calls. Since here is a central point that we control, I think it's a good safety measure.

Of course we can bump this to account for the longest possible operation + headroom. But I wouldn't remove it. But no big deal, just my 2c.

Slessi: If its just a fail-safe parameter I guess I could justify adding it back with something silly like 30 minutes?

But it should really just be a fail-safe. I think Andrey's point that call-sites should be specifying their own timeout (if relevant) makes sense. I do understand the "we might forget" point though. Wonder if there is a nicer way to deal with that.

smira: We could make the client fail if the context doesn't have a deadline, but I think it's too much if the context is bound to the incoming HTTP request.

So a failsafe is fine imho, but it should be high enough, and call sites should do their own timeouts/context chaining. 

Client.Timeout is a deprecated way to do this (it is there before contexts were added to Go).

Slessi: Ehh I can't a nice solution I am happy with, I will leave like this for now. All call sites have timeouts added.

internal/backend/imagefactory/client.go
```diff
@@ -53,7 +48,7 @@ func NewClient(imageFactoryBaseURL, username, password string) (*Client, error)
 	sniffer := &serverSnifferTransport{wrapped: http.DefaultTransport}
 
 	clientOptions := []client.Option{
-		client.WithClient(http.Client{Transport: sniffer, Timeout: requestTimeout}),
```
Unix4ever, line 56: Lets still have some sane bigger timeout here, because if for some reason the timeout will be missing from the place where the client is called it will hang forever

Unix4ever: ...and then I read your and Andrey conversation :)

utkuozdemir: I vote for keeping it, a sane big timeout.

Btw, the client-wide timeout (this field) is **not** deprecated - in fact, for such use case as ours, it is the recommended/only method. Feel free to look it up.

Slessi: I added it back at 30 minutes. We should still have timeouts at all relevant call sites because 30 minutes is a long time. In practice we should never rely on this timeout, but it's there.

PR #3051, opened 2026-06-26
WIP: refactor: use Talos client cache for most Talos client usages

utkuozdemir:
Instead of creating cold clients in many places, use the client cache, as it now can also provide maintenance-mode clients.


utkuozdemir:
proved to be non-trivial, willgo for another take later.

PR #3054, opened 2026-06-26
fix: recover a machine from a reverted reboot-requiring config patch

utkuozdemir:
When a reboot-requiring config patch is applied to a machine and then removed before the machine finishes rebooting into it, the machine could stay stuck and never return to a running state. Omni treated it as already in sync, because the desired config matched the config last confirmed as applied, so it never re-applied the reverted config to bring the machine back. The machine recovered by itself only after a long delay, which is how this surfaced as a flaky integration test.

Omni now re-applies the config when the config it last pushed to a machine differs from the desired one, not just when the last confirmed config differs from it. A machine that rebooted into a config that was then reverted now gets the reverted config re-applied and returns to a running state.

### DIFF
--- a/internal/backend/runtime/omni/controllers/omni/machineconfig/status.go
+++ b/internal/backend/runtime/omni/controllers/omni/machineconfig/status.go
@@ -189,7 +189,8 @@ func (ctrl *ClusterMachineConfigStatusController) reconcileRunning(
 		return err
 	}
 
-	if err = ctrl.computePendingUpdates(ctx, r, rc); err != nil {
+	configChanged, err := ctrl.computePendingUpdates(ctx, r, rc)
+	if err != nil {
 		return fmt.Errorf("failed to compute pending updates: %w", err)
 	}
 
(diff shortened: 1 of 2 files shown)


utkuozdemir:
The explanation:


# Reverted reboot-requiring config patch: the bug and the fix

The bug comes down to one detail: Omni keeps **two** records of a machine's config, not one.

- **Confirmed config** is the config Omni is sure the machine finished applying.
- **Last-pushed config** is the last config Omni actually sent to the machine.

Most of the time these are equal. They only drift apart in one window, and that window is where the bug happened.

The other thing to know: a config change that writes a file is reboot-requiring. When Omni sends one, the machine reboots into it, and Omni only marks the config as confirmed once the machine comes back up. So while the machine is rebooting, the config has been pushed but is not yet confirmed.

## The bug, step by step

1. The machine is running normally. Confirmed = clean config, desired = clean config. They match, so Omni does nothing.
2. The test adds the broken patch (it writes a file to `/tmp`). Desired is now the broken config.
3. Omni compares desired against confirmed: broken does not equal clean, so it pushes the broken config. Talos reports that this needs a reboot, and the machine reboots into it.
4. Because it was a reboot, Omni saves "last-pushed = broken" but does not yet mark it confirmed. Confirmed is still "clean" (stale). Omni would only update confirmed later, on a follow-up check after the machine is back.
5. The test removes the broken patch before that follow-up runs. Desired goes back to the clean config.
6. Omni compares desired against confirmed again: clean equals clean. It decides nothing changed, so it sends nothing and never reboots the machine.
7. But the machine is actually on the broken config. On a real node it is wedged mid-boot. Nothing brings it back, so it stays stuck until Talos self-heals much later, well after the test gives up. That is the flake.

The outcome was a race: did Omni mark the broken config confirmed before the test removed it? If yes, the removal looked like a real change and the machine recovered. If no, which is common under load in CI, it wedged.

## What we do differently now

1. The "should I re-apply?" decision now checks two things instead of one: re-apply if desired differs from confirmed, **or** if desired differs from last-pushed.
2. Replay the same revert. After the patch is removed, desired = clean and confirmed = clean (they match, as before), but last-pushed = broken, which does not equal clean.
3. That second check now trips, so Omni re-sends the clean config to the machine.
4. The machine reboots back into the clean config and returns to running.

So instead of trusting only "what did the machine last confirm," Omni also asks "does what I last sent the machine still match what I want?" If a reboot-requiring change was reverted mid-flight, the answer is no, and Omni corrects it. Timing no longer decides the outcome.

smira (approved):
good that we don't have apply + reboot in Talos anymore

utkuozdemir:
> good that we don't have apply + reboot in Talos anymore

Yep, I thought about that + the try mode were gonna implement, but this fix is still relevant today, and gonna stay relevant for older Talos versions.

PR #3055, opened 2026-06-29
refactor: use ImageServiceClient for pulling images

Slessi:
Use ImageServiceClient instead of the deprecated methods to pull images

### DIFF
--- a/internal/backend/runtime/omni/controllers/omni/image/client.go
+++ b/internal/backend/runtime/omni/controllers/omni/image/client.go
@@ -13,6 +13,8 @@ import (
 
 	"github.com/siderolabs/talos/pkg/machinery/api/common"
 	"github.com/siderolabs/talos/pkg/machinery/api/machine"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 
 	"github.com/siderolabs/omni/internal/backend/grpc/router"
 	"github.com/siderolabs/omni/internal/backend/runtime/talos"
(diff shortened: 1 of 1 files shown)


frezbo (approved):

internal/backend/runtime/omni/controllers/omni/image/client.go
```diff
@@ -42,15 +43,20 @@ func (c *TalosImageClient) ListImagesOnNode(ctx context.Context, cluster, node s
 		return nil, fmt.Errorf("failed to get talos client for node %q: %w", node, err)
 	}
 
-	imageListStream, err := talosCli.ImageList(ctx, common.ContainerdNamespace_NS_CRI) //nolint:staticcheck
+	imageListStream, err := talosCli.ImageClient.List(ctx, &machine.ImageServiceListRequest{
```
smira: the new method is not supported through every Talos version, so _probably_, this need to have a fall back with Unimplemented error

Slessi: Fallback to an error though? Not a fallback to the deprecated functions?

smira: If you call new method (this one), and you get `Unimplemented` error back, you fall back to the old API.

Please note that with streaming APIs this error will be returned on the first read.

https://github.com/siderolabs/talos/blob/47d4bd87e687bdf4288873962c5db14cedd5bb50/cmd/talosctl/cmd/talos/image.go#L151-L155

Slessi: aha, i misunderstood, clear

internal/backend/runtime/omni/controllers/omni/image/client.go
```diff
@@ -42,15 +44,44 @@ func (c *TalosImageClient) ListImagesOnNode(ctx context.Context, cluster, node s
...
 	}
 
-	imageListStream, err := talosCli.ImageList(ctx, common.ContainerdNamespace_NS_CRI) //nolint:staticcheck
+	stream, err := talosCli.ImageClient.List(ctx, &machine.ImageServiceListRequest{
+		Containerd: &common.ContainerdInstance{
+			Driver:    common.ContainerDriver_CRI,
+			Namespace: common.ContainerdNamespace_NS_CRI,
+		},
+	})
+	if err == nil {
```
smira: this flow looks a bit strange to me, the error will only come on the first read, i.e. `Recv`, so if you get an `err != nil`, it means something else happened, and you should abort

I think this would simplify things a bit if we abort here early, then call `readImagesFromStream`, and if that returns Unimplemented, we finally fall back to legacy. I would probably move legacy into a separate function to make it easier to track it. The new APIs are Talos 1.13+

Slessi: its correct how it is, maybe not the most clear, ill try make it more readable

its: `err == nil` not `err != nil`, then the err is updated with the stream error inside the block

after this, we have (potentially) the stream err which might be `codes.Unimplemented` or if the stream didnt start at all, whatever error `List` returned

but anyways, will simplify

smira: It is correct, but hard to follow as it has so much nesting. The Unimplemented error comes on first Recv.

internal/backend/runtime/omni/controllers/omni/image/client.go
```diff
@@ -79,9 +107,36 @@ func (c *TalosImageClient) PullImageToNode(ctx context.Context, cluster, node, i
...
+	stream, err := talosCli.ImageClient.Pull(ctx, &machine.ImageServicePullRequest{
+		ImageRef: image,
+		Containerd: &common.ContainerdInstance{
+			Driver:    common.ContainerDriver_CRI,
+			Namespace: common.ContainerdNamespace_NS_CRI,
+		},
+	})
+	if err != nil {
 		return fmt.Errorf("failed to pull image %s: %w", image, err)
 	}
```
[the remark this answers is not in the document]
Slessi: Its returned from first RECV so not valid

internal/backend/runtime/omni/controllers/omni/image/client.go
```diff
@@ -42,15 +44,41 @@ func (c *TalosImageClient) ListImagesOnNode(ctx context.Context, cluster, node s
...
-	imageListStream, err := talosCli.ImageList(ctx, common.ContainerdNamespace_NS_CRI) //nolint:staticcheck
+	stream, err := talosCli.ImageClient.List(ctx, &machine.ImageServiceListRequest{
+		Containerd: &common.ContainerdInstance{
+			Driver:    common.ContainerDriver_CRI,
+			Namespace: common.ContainerdNamespace_NS_CRI,
+		},
+	})
 	if err != nil {
 		return nil, fmt.Errorf("failed to list images: %w", err)
 	}
```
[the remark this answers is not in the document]
Slessi: Its returned from first RECV so not valid

PR #3061, opened 2026-06-29
fix(frontend): fix double command in processes and format CPU chart better

### LINKED ISSUE #3059: Display CPU graphs stacked on the node overview page
Image

Also add % symbol to the legend also.

### LINKED ISSUE #3060: Command line of the processes has an extra executable name in the beginning of the line
See the screenshot:

Image

`clang++ /usr/bin/clang++` should be `/usr/bin/clang++`.

Slessi:
- Stack the CPU usage chart areas and add a `%` symbol
- Remove duplicate command output from processes list command clumn
- Also add [stories](https://frontend-monitor-fixes--68d664245076e2bb623d98f3.chromatic.com/?path=/story/pages-authenticated-clusters-cluster-machine-machine-monitor--default)

Closes #3059
Closes #3060

### DIFF
--- a/frontend/src/pages/(authenticated)/clusters/[cluster]/machine/[machine]/monitor.vue
+++ b/frontend/src/pages/(authenticated)/clusters/[cluster]/machine/[machine]/monitor.vue
@@ -95,9 +95,8 @@ const getProcs = async () => {
         state: proc.state!,
         virtualMemory: parseInt(proc.virtual_memory || '0'),
         residentMemory: parseInt(proc.resident_memory || '0'),
-        command: proc.command!,
+        command: proc.args!,
         cpuTime: proc.cpu_time || 0,
-        args: proc.args!,
       }
     }),
   )
@@ -294,7 +293,7 @@ const sortBy = (id: keyof Proc) => {
           v-for="process in sortedProcesses"
           :key="process.pid"
           class="grid grid-cols-12 py-2 text-xs text-naturals-n12"
-          :title="process.command + ' ' + process.args"
+          :title="process.command"
         >
           <div>
             {{ process.pid }}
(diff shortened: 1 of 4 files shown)


frontend/src/pages/(authenticated)/clusters/[cluster]/machine/[machine]/monitor.vue
```diff
@@ -320,7 +333,7 @@ const sortBy = (id: keyof Proc) => {
           <div>
             {{ process.cpuTime }}
           </div>
-          <div class="col-span-4 truncate">{{ process.command }} {{ process.args }}</div>
```
Comment (smira) [inline frontend/src/pages/(authenticated)/clusters/[cluster]/machine/[machine]/monitor.vue:323]: this feels wrong to me... 

See https://github.com/siderolabs/talos/blob/47d4bd87e687bdf4288873962c5db14cedd5bb50/cmd/talosctl/cmd/talos/processes.go#L106-L116

Comment (smira) [reply frontend/src/pages/(authenticated)/clusters/[cluster]/machine/[machine]/monitor.vue:323]: or am I reading it wrong, not sure 🤷

Comment (smira) [reply frontend/src/pages/(authenticated)/clusters/[cluster]/machine/[machine]/monitor.vue:323]: so the proper way is patch in args[0] with the full executable name, it'd match what other tools output

Comment (Slessi) [reply frontend/src/pages/(authenticated)/clusters/[cluster]/machine/[machine]/monitor.vue:323]: What do we want here, just the binary name, or the full command with all arguments?

In my change here, I made it just `args`, so you will see this kinda thing:
- `/bin/kubelet --config=/etc/kubernetes/kubelet.yaml`

I realise it reads there as `process.command` which might lead you to believe it is going to be only `kubelet` but just to clarify thats cause I'm mapping to "columns" and in an earlier place I mapped args into the `process.command` "column" if that makes sense

Comment (smira) [reply frontend/src/pages/(authenticated)/clusters/[cluster]/machine/[machine]/monitor.vue:323]: I posted above the "golden" code that this should follow

Comment (Slessi) [reply frontend/src/pages/(authenticated)/clusters/[cluster]/machine/[machine]/monitor.vue:323]: Done!

Unix4ever (approved):

PR #3065, opened 2026-06-30
feat: validate undefined group references in AccessPolicy rules

ohauer:
Add validation in `Validate()` that catches two classes of AccessPolicy misconfiguration:

1. **Typo detection**: Rejects entries like `groups/my-group` (plural) with a helpful suggestion:
   `invalid user reference "groups/my-group" in rule: did you mean "group/my-group"?`
2. **Undefined references**: Rejects `group/nonexistent` when the referenced group is not defined.

Both checks apply to `rules[].users[]` and `rules[].clusters[]`.

The reserved prefix for group references is `group/` (singular). Using `groups/` (plural)
causes a silent failure: the entry is treated as a literal user identity that never matches.
The documentation example that led to this was fixed in https://github.com/siderolabs/docs/pull/623.

Tested on self-hosted Omni v1.9.0 with SAML/Keycloak:
- `omnictl apply` with `groups/ns-admins` → rejected with helpful error ✅
- `omnictl apply` with `group/ns-admins` → accepted ✅

### DIFF
--- a/internal/pkg/auth/accesspolicy/accesspolicy.go
+++ b/internal/pkg/auth/accesspolicy/accesspolicy.go
@@ -133,6 +133,38 @@ func Validate(accessPolicy *auth.AccessPolicy) error {
 				validationErrs = multierror.Append(validationErrs, err)
 			}
 		}
+
+		for _, ruleUser := range rule.GetUsers() {
+			if strings.HasPrefix(ruleUser, "group") && !strings.HasPrefix(ruleUser, GroupPrefix) {
+				validationErrs = multierror.Append(validationErrs, fmt.Errorf(
+					"invalid user reference %q in rule: did you mean %q?",
+					ruleUser, GroupPrefix+strings.TrimPrefix(ruleUser, "groups/")))
+			}
+
+			if strings.HasPrefix(ruleUser, GroupPrefix) {
+				groupName := ruleUser[len(GroupPrefix):]
+				if _, ok := accessPolicySpec.GetUserGroups()[groupName]; !ok {
+					validationErrs = multierror.Append(validationErrs, fmt.Errorf(
+						"rule references undefined user group %q", groupName))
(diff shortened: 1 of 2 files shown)


internal/pkg/auth/accesspolicy/accesspolicy.go
```diff
@@ -133,6 +133,38 @@ func Validate(accessPolicy *auth.AccessPolicy) error {
...
+					ruleUser, GroupPrefix+strings.TrimPrefix(ruleUser, "groups/")))
+			}
+
+			if strings.HasPrefix(ruleUser, GroupPrefix) {
+				groupName := ruleUser[len(GroupPrefix):]
+				if _, ok := accessPolicySpec.GetUserGroups()[groupName]; !ok {
+					validationErrs = multierror.Append(validationErrs, fmt.Errorf(
+						"rule references undefined user group %q", groupName))
+				}
+			}
```
Slessi: I don't think the first check for `HasPrefix(ruleUser, "group")` is a great move. It can fail for perfectly valid things like `group-admin`. Arguably could change it to `groups/` but I also don't see the value in checking for only one specific kind of typo. Wouldn't catch `gorup/`, `gruop/`, or any other kind of typo.

Lets keep only the 2nd portion which checks if the user group is actually defined.

Slessi: @ohauer kind bump

ohauer:
Thanks for the review @Slessi — you're right that the `HasPrefix(ruleUser, "group")` check would false-positive on valid identities like `group-admin`.

I've updated the PR to remove the typo detection entirely and keep only the undefined group reference check, as you suggested. Force-pushed with the simplified version.

Slessi (approved):

Slessi:
Thanks for the contribution!

PR #3070, opened 2026-07-01
fix: update COSI runtime to fix hanging TeardownAndDestroy calls

Unix4ever:
See https://github.com/cosi-project/runtime/pull/668


smira (approved):
let's update client as well!

Unix4ever:
> let's update client as well!

oh, good point

oguzkilcan (approved):

PR #3081, opened 2026-07-02
refactor(frontend): merge Watch class into useResourceWatch composable

### LINKED ISSUE #1471: [UI] Tests for Watch/useWatch

Slessi:
Merge the Watch class into the useResourceWatch composable. Comes with a lot of code simplifications, and is the first step in having watch singletons so that a given resource is never watched more than once across multiple components.

Basic design:

- `useResourceWatch` entry-point, determines if its a single item or list watch
  - `useWatchSingle` if single, goes here. Doesn't do much other than set/remove the watched item.
  - `useWatchMulti` if a list, goes here. All list logic encapsulated here (sorting, bootstrapping, total)
- `useWatchStream` both single/multi will call this final composable, which handles opening and closing the stream, setting request options, loading/error state, forwarding the stream messages, parsing JSON.

Closes #1471


frontend/src/components/OngoingTasks/OngoingTasks.vue
```diff
@@ -95,7 +94,7 @@ const getCurrentComponent = (item: Resource<OngoingTaskSpec>) => {
       >
         <div
           v-for="item in data"
-          :key="itemID(item)"
+          :key="item.metadata.id"
```
Unix4ever, line 97: in general using itemID is safer as it includes the namespace. Just ids from different namespaces might clash.
We don't use different namespaces for COSI, but it might happen in Kubernetes watches

Slessi: The only place I saw that could benefit from `itemID` was [`pods.vue`](https://github.com/Slessi/omni/blob/9ae983308143e377155af5f9641139c645f0056d/frontend/src/pages/(authenticated)/clusters/%5Bcluster%5D/pods.vue), but it is doing it manually already with ``:key="`${item.metadata.namespace}/${item.metadata.name || idx}`"``

I kept it inside `useResourceWatch` because no way to know what resource we are watching, but for the external locations where I removed it, it doesn't look applicable

I can revert it back for these other places, but I think it is misleading to use it where it doesn't apply (IMO)

Unix4ever: I guess it's fine as far as we are careful :)

Unix4ever (approved):

PR #3093, opened 2026-07-06
feat: manage Talos install and upgrade via LifecycleService

### LINKED ISSUE #2530: rework the Upgrade sequence to use new APIs
Use new image pull, upgrade and (old) reboot APIs to reimplement resilient upgrade.

Should also work in maintenance mode if needed.

oguzkilcan:
Talos introduced LifecycleService API with version 1.13. Omni now uses this API to manage install/upgrade operations. It runs sequence itself: pull the installer, install or upgrade to disk, forfeit etcd leadership on control planes, cordon and drain the node, then reboot. The classic path stays for older Talos and same-minor installs, and each machine's live version and schematic decide which path it takes.

Resolves: #2530


internal/backend/runtime/omni/controllers/omni/machineconfig/status.go
```diff
@@ -391,6 +414,36 @@ func (ctrl *ClusterMachineConfigStatusController) reconcileUpgrade(
...
+			if nodeName != "" {
+				if err = ctrl.lifecycleManager.FinalizeReboot(ctx, lifecycle.WithUncordon(clusterName, nodeName)); err != nil {
+					// Keep PreRebootBootId so the next reconcile retries the uncordon.
+					return controller.NewRequeueError(fmt.Errorf("failed to finalize reboot for machine %q: %w", rc.ID(), err), lifecycleOperationPollInterval)
+				}
+			}
+		}
+	}
+
+	rc.machineConfigStatus.TypedSpec().Value.PreRebootBootId = ""
```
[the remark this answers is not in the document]
oguzkilcan: We can't cordon a node when we can't determine cluster and node name. So if we could cordon a node then we will be able to uncordon it after the reboot.

internal/backend/runtime/omni/controllers/omni/machineconfig/reconciliation_context.go
```diff
@@ -267,34 +270,48 @@ func DecideLifecycleOp(machineStatus *omni.MachineStatus, installImage *specs.Ma
...
+		}
+
+		if hasSystemDisk {
+			if !machineVersion.EQ(targetVersion) || schematicDiffers(machineStatus, installImage) {
+				return LifecycleOpClusterUpgrade
+			}
+
 			return LifecycleOpNone
 		}
 	}
```
Unix4ever, line 296: This condition is hard to read, but I can't suggest a better one so far.

oguzkilcan: We have 4 paths to choose from depending on similar but different inputs. It's tricky to get it right and easily readable. I'm always open to suggestions.

Unix4ever:
Can't leave feedback for `machineconfig/status.go` controller as the file is too big :melting_face: 

I see the comment there:

```golang
+       case err != nil:
+               // Returning a raw error would make qtransform drop the output write and lose LastConfigError.
+               rc.machineConfigStatus.TypedSpec().Value.LastConfigError = lifecycle.WrapErr(err, "upgrade failed").Error()
+
+               logger.Error("upgrade failed", zap.String("machine", machineID), zap.Error(err))
+
+               return false, controller.NewRequeueInterval(lifecycleOperationPollInterval)
```

At the same time it breaks back-off. Maybe not a big deal though.

Unix4ever (approved):

PR #3096, opened 2026-07-06
feat: send initial instance user to signup page

### LINKED ISSUE #2061: [feature] Send user to auth0 signup page on first visit, not login page
### Problem Description

Currently, when an Omni account is first provisioned, the email the customer receives to tell them their Omni account is ready takes them to 
https://steve4.na-west-1.omni.siderolabs.io/
which then redirects them to https://sidero-omni.us.auth0.com/u/login?state=....

However, at this point the user has not signed up, so from that page they have to notice and click the small text "First time logging in? [Create your Omni username/password]" which then takes them to the signup page.

### Solution

If the user has never logged in, redirect them to the signup page immediately.
(If we dont know at this point if the user has never logged in, we DO know if any user has ever logged in to the account. If not, send them to the signup page.)

### Alternative Solutions

_No response_

### Notes

_No response_

Slessi:
Currently we always send users to the login page. With this change, the very first user of any instance will always be sent instead to the sign up page. We can't do it on a per-user basis as until we reach auth0 and the user attempts to login, we do not know who is attempting the login.

Closes #2061

### DIFF
--- a/frontend/src/methods/useMachineServices.ts
+++ b/frontend/src/methods/useMachineServices.ts
@@ -10,6 +10,7 @@ import type { RuntimeContext } from '@/api/options'
 import { withAbortController, withContext, withRuntime } from '@/api/options'
 import { MachineService, type ServiceEvent } from '@/api/talos/machine/machine.pb'
 import { TCommonStatuses } from '@/constants'
+import { showError } from '@/notification'
 
 interface Service {
   name?: string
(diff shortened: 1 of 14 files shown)


Unix4ever:
It might be nice to see how the flow looks like: some video.

Slessi:
@Unix4ever its pretty much 100% the same, only change is you will see the sign-up page instead of login page, if there are no users yet in the instance. 

You can see that new page already by visiting any of our SaaS Omni's and during login click "Sign up" (highlighted the lines here):
image

This takes you to the sign up page that initial users will see with my change. The request from Steve was basically to not require them to have to click "Sign up" manually.
image

frontend/src/methods/useMachineServices.ts
```diff
@@ -29,25 +30,31 @@ export function useMachineServices(context: MaybeRefOrGetter<RuntimeContext>) {
...
+              ? TCommonStatuses.HEALTHY
+              : TCommonStatuses.UNHEALTHY,
+          events: service.events?.events,
+        })),
+      )
+    } catch (e) {
+      if (abortController.signal.aborted) return
+
+      showError('Error', e instanceof Error ? e.message : String(e))
+    }
```
Slessi, line 57: unrelated to this PR, i included a separate commit to fix this tiny issue, was missing a catch

Unix4ever (approved):

PR #3097, opened 2026-07-06
feat(frontend): add a quickstart page for kubespan

### LINKED ISSUE #3087: Automatic Kubespan enable from tab
When kubespan is disable on a cluster (default) the tab is not selectable and you can't learn more about what it does.

Image

It would be nice if the tab could still be selected, tell you some information about what it does (and maybe a preview of what it looks like) and then have a simple way to apply a patch to the cluster.

This would help people discover what it is for, enable it quickly, and prevent users from applying a patch incorrectly for their version of Talos.

If you want to enable it today you typically have to go to the docs, search for the feature (example is only in reference pages) copy the section and go back to omni and create a new patch (properly scoped to cluster)

Slessi:
When a cluster does not have KubeSpan enabled, instead of simply disabling the tab, we will instead show a quick start page with a basic explanation, docs link, the YAML for the patch, and an "Enable KubeSpan" button to get started, which simply applies the patch. Warnings about network overhead (especially on >=50 node clusters) will be shown.

Closes #3087 

image

image

image

### DIFF
--- a/frontend/src/views/KubeSpanStatus/KubeSpanStatus.vue
+++ b/frontend/src/views/KubeSpanStatus/KubeSpanStatus.vue
@@ -16,10 +16,13 @@ import {
   ClusterMachineIdentityType,
   DefaultNamespace,
   LabelCluster,
+  TalosConfigNamespace,
+  TalosKubespanConfigID,
+  TalosKubeSpanConfigType,
   TalosKubeSpanNamespace,
   TalosKubeSpanPeerStatusType,
 } from '@/api/resources'
-import type { PeerStatusSpec } from '@/api/talos/kubespan.pb'
+import type { ConfigSpec, PeerStatusSpec } from '@/api/talos/kubespan.pb'
 import IconButton from '@/components/Button/IconButton.vue'
 import TIcon from '@/components/Icon/TIcon.vue'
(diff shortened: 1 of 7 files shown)


rothgar (approved):

frontend/src/views/KubeSpanStatus/KubeSpanStatus.vue
```diff
@@ -97,7 +114,7 @@ function onPeerClick(peer: Resource<PeerStatusSpec>) {
 </script>
 
 <template>
-  <PageContainer class="@container flex h-full flex-col gap-4">
+  <PageContainer v-if="kubeSpanConfig?.spec.enabled" class="@container flex h-full flex-col gap-4">
```
Unix4ever, line 117: I wonder if that's related or it's just storybooks acting up for me, but the KubeSpan peers storybook doesn't open now.

Unix4ever (approved):

frontend/src/views/KubeSpanStatus/KubeSpanStatus.vue
Slessi: my bad, was a new watched resource and forgot to mock the `enabled: true` - fixed now

PR #3098, opened 2026-07-07
feat(frontend): persist search and filters as query strings

### LINKED ISSUE #2492: [feature] add search labels, string search query to the URI when changed
- Preserve current search in query string
- Possibly allow negating conditions like in CLI
- Improve UX of the control (e.g. allow clicking on label to select from dropdown)

Slessi:
When searching/filtering/sorting in Omni, we will now persist the fields in the URL bar as query strings. For the labels part (e.g. from cluster/machine lists) it required a bit of refactoring first. Refactors are separated into their own commits.

Closes #2492

### DIFF
--- a/frontend/src/methods/labels.ts
+++ b/frontend/src/methods/labels.ts
@@ -3,6 +3,8 @@
 // Use of this software is governed by the Business Source License
 // included in the LICENSE file.
 
+import { useRouteQuery } from '@vueuse/router'
+
 import {
   InfraProviderLabelPrefix,
   LabelCluster,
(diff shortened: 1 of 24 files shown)


frontend/src/methods/labels.ts
```diff
@@ -82,12 +84,35 @@ const labelClasses: Record<string, string> = {
...
 export const getLabelClass = (labelKey: string) => labelClasses[labelKey]
 
+export function useLabelRouteQuery() {
+  return useRouteQuery<string, Label[]>('labels', '', {
+    transform: {
+      get(val) {
+        if (!val) return []
+
+        try {
+          const parsed = JSON.parse(window.atob(val))
```
Unix4ever, line 94: I think it would be nicer to have human readable query here, but I know about the complication it brings with the need to escape stuff.

Unix4ever (approved):

frontend/src/methods/labels.ts
Slessi: I tried, the problem is that the source information (label list) is completely detached from here.

Its passed as a resource MD definition then queried repeatedly inside the child component, i believe partly because it isnt watchable.

Its possible to fix this... but i fell into a hole of refactors already trying to untangle label logic here and didnt wanna go too crazy

Current solution ugly one but works for now

PR #3099, opened 2026-07-07
feat: expose node audit skip cluster feature

shanduur:
Surface ClusterSpec.Features.EnableNodeAuditSkip in the UI so the
cluster owner can opt in - the backend gates the node-audit-skip
annotation on this feature to prevent workloads with nodes/patch
RBAC from self-exempting.

- toggle on cluster create
- toggle on cluster overview, gated on canManageClusterFeatures

Signed-off-by: Mateusz Urbanek <mateusz.urbanek@siderolabs.com>


Slessi (approved):
Code LGTM ✅

PR #3102, opened 2026-07-08
feat(frontend): update home page charts and show ongoing tasks

### LINKED ISSUE #1635: [feature] Simplify the Omni Home page cluster widget.
### Problem Description

Currently the Omni home page clusters widget looks like:

Image

But this is displaying both briefly transient info (Destroying, scaling up, scaling down) with higher level health. And the presentation makes it hard to interpret.

It would be clearer if it was just a pie chart, and either omitted Scaling and Destroying numbers, or, if those clusters do not fall within Healthy or Unhealthy, just combine all 3 as "Scaling"

Image

### Solution

_No response_

### Alternative Solutions

_No response_

### Notes

_No response_

### LINKED ISSUE #1636: [feature] Simplify the Omni home page machines widget.
### Problem Description

The machines widget presents two independent classes of information (machine connectivity, and whether the machines are allocated) in a single widget in a confusing manner.

### Solution

Either break out the two classes into separate widgets, or combine them like:

Image
All machines are either connected/not connected, and allocated, not allocated or pending.
So the totals of those two columns will always match.

### Alternative Solutions

_No response_

### Notes

_No response_

Slessi:
Replace the radial bar charts on the homepage with segmented bars which more appropriately display the information. Also splits the machine information into two separate points, connection state and allocation state.

Also added the ongoing tasks from the dropdown to the home page for better visibility, as in the dropdown it is convenient but can be missed. Added storybook stories for all home page items.

Closes #1635
Closes #1636

image

### DIFF
--- a/frontend/src/components/OngoingTasks/OngoingTasks.vue
+++ b/frontend/src/components/OngoingTasks/OngoingTasks.vue
@@ -8,64 +8,14 @@ included in the LICENSE file.
 import { PopoverContent, PopoverPortal, PopoverRoot, PopoverTrigger } from 'reka-ui'
 import { ref } from 'vue'
 
-import { Runtime } from '@/api/common/omni.pb'
-import { type Resource } from '@/api/grpc'
-import {
-  KubernetesUpgradeStatusSpecPhase,
-  type OngoingTaskSpec,
-  SecretRotationSpecComponent,
-} from '@/api/omni/specs/omni.pb'
-import { EphemeralNamespace, OngoingTaskType } from '@/api/resources'
 import TIcon from '@/components/Icon/TIcon.vue'
 import IconHeaderDropdownLoading from '@/components/icons/IconHeaderDropdownLoading.vue'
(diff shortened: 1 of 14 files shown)


Unix4ever:
I wonder do we really need pending on connected/disconnected chart?

Slessi:
@Unix4ever doesnt need to be, its technically duplicate data as it appears on both bars

just not clear which bar should own it, pending could arguably be both "waiting to allocate" or "waiting to connect" 

if u have a preference to keeping it on the allocation one i can do that

Unix4ever:
Lets keep it only on allocation one.

Slessi:
@Unix4ever done! 

one minor consequence, total will say for example 44 machines (includes 2 pending) but the connections bar will only sum up to 42 (doesnt include pending)

updated screenshot in desc to show this

image

Unix4ever (approved):

shanduur (approved):

PR #3119, opened 2026-07-14
fix: don't treat unset S3 endpoint as an empty override

oguzkilcan:
When the S3 endpoint field was left unset, Omni still passed an empty string to the AWS SDK as a literal endpoint override, and the SDK rejected it as an invalid URI. Configuring S3 backup storage against real AWS S3, as opposed to a custom endpoint like MinIO, always failed as a result. resolveBaseEndpoint now returns nil when no endpoint is configured, so the SDK falls back to its normal endpoint resolution.

### DIFF
--- a/internal/backend/runtime/omni/controllers/omni/etcdbackup/store/s3store.go
+++ b/internal/backend/runtime/omni/controllers/omni/etcdbackup/store/s3store.go
@@ -199,7 +199,7 @@ func S3ClientFromResource(ctx context.Context, s3Conf *omni.EtcdBackupS3Conf) (*
 
 	client := s3.NewFromConfig(loadedCfg, func(o *s3.Options) {
 		o.UsePathStyle = true
-		o.BaseEndpoint = aws.String(baseEndpoint)
+		o.BaseEndpoint = resolveBaseEndpoint(baseEndpoint)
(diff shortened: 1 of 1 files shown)


majabojarska (approved):

internal/backend/runtime/omni/controllers/omni/etcdbackup/store/s3store.go
```diff
@@ -199,7 +199,7 @@ func S3ClientFromResource(ctx context.Context, s3Conf *omni.EtcdBackupS3Conf) (*
 
 	client := s3.NewFromConfig(loadedCfg, func(o *s3.Options) {
 		o.UsePathStyle = true
-		o.BaseEndpoint = aws.String(baseEndpoint)
+		o.BaseEndpoint = resolveBaseEndpoint(baseEndpoint)
```
utkuozdemir, line 202: maybe as simple as:
```
if baseEndpoint != "" {
  o.BaseEndpoint = new(baseEndpoint)
}
```

since using `new()` is now _the_ way to do this (I think running `go fix ./...` might even convert it automatically, but maybe misses this one).

Unix4ever (approved):

utkuozdemir (approved):

PR #3124, opened 2026-07-14
feat(frontend): add a /logout route for auth0 logouts

### LINKED ISSUE #3107: [feature] Implement logout handler for Auth0
Omni currently has logout handling for SAML and OIDC. It should have logout handler for Auth0 to close the gap so logout handling is provided for all auth providers.

Slessi:
Add a /logout route to handle logging out from auth0

Closes #3107

### DIFF
--- a/frontend/typed-router.d.ts
+++ b/frontend/typed-router.d.ts
@@ -619,6 +619,13 @@ declare module 'vue-router/auto-routes' {
       Record<never, never>,
       | never
     >,
+    'Logout': RouteRecordInfo<
+      'Logout',
+      '/logout',
+      Record<never, never>,
+      Record<never, never>,
+      | never
+    >,
     'OIDC Login': RouteRecordInfo<
       'OIDC Login',
       '/oidc-login/:authRequestId',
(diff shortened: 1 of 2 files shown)


frontend/typed-router.d.ts
```diff
@@ -619,6 +619,13 @@ declare module 'vue-router/auto-routes' {
       Record<never, never>,
       | never
     >,
+    'Logout': RouteRecordInfo<
```
Unix4ever, line 622: I think this logout route should be added only if auth0 is enabled.
Can we do that dynamically here?

Slessi: I don't really think its possible to do that, with file-based routing routes are defined at build time, not runtime, but omni flags are at runtime

Is there a specific concern you wanna avoid? Maybe I can add runtime code to avoid the concern

Unix4ever: I want to avoid the frontend catching `/logout` URL before backend is able to react.
Maybe it's not a problem at all. I'm going to checkout your branch and test OIDC there.

Unix4ever (approved):

PR #3125, opened 2026-07-14
fix: correct boot ID and Talos version tracking for maintenance installs

oguzkilcan:
The maintenance-mode Talos install/upgrade lifecycle recorded machine state before it was actually confirmed. A fresh same-minor install could get its pre-install Talos version written as the confirmed one, and TalosUpgradeStatusController would later read that and report a bogus downgrade.

A machine downgraded below 1.13 kept a stale boot ID forever because the snapshot merge couldn't tell the collect task's own empty value apart from a source that never carries one. This also skips the boot_id Read RPC on Talos versions that never implement it, instead of retrying a call with certain error every 30 seconds.

### DIFF
--- a/internal/backend/runtime/omni/controllers/omni/internal/task/snapshot/snapshot.go
+++ b/internal/backend/runtime/omni/controllers/omni/internal/task/snapshot/snapshot.go
@@ -14,6 +14,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/blang/semver/v4"
 	"github.com/cosi-project/runtime/pkg/resource"
 	"github.com/cosi-project/runtime/pkg/state"
 	"github.com/siderolabs/gen/channel"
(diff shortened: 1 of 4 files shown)


frezbo (approved):

internal/backend/runtime/omni/controllers/omni/internal/task/snapshot/snapshot.go
```diff
@@ -125,6 +134,10 @@ func (spec CollectTaskSpec) RunTask(ctx context.Context, logger *zap.Logger, not
 		case <-ctx.Done():
 			return nil
 		case <-bootIDTicker.C:
```
Unix4ever: Maybe instead of skipping bootID ticker we can skip starting it.
Like

```go
vat bootIDTickerChan chan struct{}

if bootIDSupported {
	bootID = spec.readBootID(ctx, client, logger)

	bootIDTicker := time.NewTicker(bootIDProbeInterval)
	defer bootIDTicker.Stop()

	bootIDTickerChan = bootIDTicker.C
}

...

case <-bootIDTickerChan: // just never enters here if the chan is nil
```

oguzkilcan: This might be even clearer.

Unix4ever (approved):

Issue #2985, opened 2026-06-16
Omni audit log exporter


smira:
Similar to #2984, stream Omni audit logs with safe restarts from Omni and write them to stdout/stderr.

Coler-e:
Would happily make an Elastic integration for it once that's out!

PR #3130, opened 2026-07-15
feat: support following the audit log over the management API

utkuozdemir:
The audit log read API gains a follow mode: after serving the backlog, the stream stays open and delivers new events as they are written, in insertion order. The request accepts a millisecond-precision start position for resuming from where a previous stream left off, and each response carries the event timestamp, so a client can track its position without parsing the payloads.

A follow stream starts with an event-less acknowledgment response, letting a client tell a server that honors the follow mode apart from an older one that would serve a bounded read and stop. Streams end cleanly after a bounded duration, and reconnecting re-runs authentication and authorization, so an identity that has been revoked or downgraded stops being served promptly. Follow access is recorded in the audit log like any other audit log access.

Follow reads are served in bounded batches that do not hold a database connection while the stream is blocked on a slow client.

omnictl audit-log gains a --follow flag to tail the log live.

Part of siderolabs/omni#2985.

### DIFF
--- a/client/pkg/client/management/management.go
+++ b/client/pkg/client/management/management.go
@@ -87,12 +87,15 @@ func WithOIDCCacheIsolation(value bool) KubeconfigOption {
 // Client for Management API .
 type Client struct {
 	conn management.ManagementServiceClient
+
+	followReconnectFloor time.Duration
 }
 
 // NewClient builds a client out of gRPC connection.
 func NewClient(conn *grpc.ClientConn) *Client {
 	return &Client{
-		conn: management.NewManagementServiceClient(conn),
+		conn:                 management.NewManagementServiceClient(conn),
+		followReconnectFloor: auditLogFollowReconnectFloor,
 	}
 }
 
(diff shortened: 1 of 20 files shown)


client/pkg/client/management/management.go
```diff
@@ -312,6 +318,111 @@ func (client *Client) ReadAuditLog(ctx context.Context, req *management.ReadAudi
...
+// followAuditLogOnce consumes a single follow stream, yielding its events and advancing the
+// start position on the request for the next stream. It reports whether following should
+// continue, which it should not when the stream failed, the server turned out not to support
+// following, or the consumer stopped the iteration.
+func (client *Client) followAuditLogOnce(
+	ctx context.Context, req *management.ReadAuditLogRequest, yield func(*management.ReadAuditLogResponse, error) bool,
+) bool {
+	gotAck := false
+
+	for resp, err := range client.ReadAuditLog(ctx, req) {
```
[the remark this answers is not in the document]
utkuozdemir: done

internal/backend/grpc/management.go
```diff
@@ -600,6 +640,132 @@ func (s *managementServer) ReadAuditLog(req *management.ReadAuditLogRequest, srv
 	return closeFn()
 }
 
+// followAuditLog streams the audit log: the backlog from the requested start position first,
+// then new events as they are written, until the stream lease expires. A client keeps
+// following by reconnecting and resuming from the id of the last received event.
+func (s *managementServer) followAuditLog(req *management.ReadAuditLogRequest, srv grpc.ServerStreamingServer[management.ReadAuditLogResponse]) error {
```
Unix4ever, line 640: Could we instead of polling introduce events in the auditor?
Not insisting though.

Unix4ever (approved):

internal/backend/grpc/management.go
utkuozdemir: good point, I changed it to be same as machine logs following.

PR #3131, opened 2026-07-15
fix: dont clear sa expiration when no keys are left

### LINKED ISSUE #3128: Service account shows incorrect date for Expiration date for expired keys

### Current Behavior

When expired service account shows expiration as 2025 years ago

### Expected Behavior

should show date key actually expired

### Steps To Reproduce

NA

Chrome

### Anything else?

Image

Slessi:
Currently when all public keys are removed from a service account, the expiration gets zero'd out. When all public keys have been removed from a service account, instead leave the expiration value on the spec as is.

Closes #3128

### DIFF
--- a/internal/backend/runtime/omni/controllers/omni/service_account_status.go
+++ b/internal/backend/runtime/omni/controllers/omni/service_account_status.go
@@ -34,7 +34,7 @@ const serviceAccountStatusControllerName = "ServiceAccountStatusController"
 
 // NewServiceAccountStatusController instantiates the ServiceAccountStatus controller.
 //
-//nolint:gocognit
+//nolint:gocognit,gocyclo,cyclop
 func NewServiceAccountStatusController() *ServiceAccountStatusController {
 	return qtransform.NewQController(
 		qtransform.Settings[*auth.Identity, *auth.ServiceAccountStatus]{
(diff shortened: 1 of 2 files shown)


internal/backend/runtime/omni/controllers/omni/service_account_status.go
```diff
@@ -116,7 +120,10 @@ func NewServiceAccountStatusController() *ServiceAccountStatusController {
 				}
 
 				status.TypedSpec().Value.Role = user.TypedSpec().Value.Role
-				status.TypedSpec().Value.Expiration = timestamppb.New(expiration)
+
+				if hasRunningKey {
+					status.TypedSpec().Value.Expiration = timestamppb.New(expiration)
```
smira, line 125: this would keep the last seen expiration date forever (?) I think `nil` Expiration says it better that there are no valid keys? We should not show `nil` as zero timestamp of course, but say something about it having no keys or what not

Slessi: > this would keep the last seen expiration date forever 

thats what i tried to achieve yes - when looking at service account resources, dont you want to know when an expired account actually expired?

Unix4ever (approved):
I think it's a good change. As the expired keys get deleted it's the only way to keep the service account status sane

internal/backend/runtime/omni/controllers/omni/service_account_status.go
smira: well kind of, but it breaks the reconciliation logic a bit - the output resource will preserve last state which can't be reconciled from the inputs. so either one expired key should be left always (the last one), or this is a bit "it works as long as it works"

smira: not that it's so much important here, it's a minor thing in the end.

PR #3133, opened 2026-07-15
test(frontend): use dex oidc instead of auth0 as default for e2e tests

Slessi:
- Use dex OIDC as the default auth provider for all E2E suites
- Add a new small dedicated `e2e-auth0` suite to run login/logout flow with auth0 so that this path remains tested
- Wait for minio more reliably in E2E
- Made frontend dev server use same backend routing as non-dev
- Change factory URL to the new staging one in vite dev server
- Drop flaky ISO download test from installation media tests (just the actual downloading from factory)


utkuozdemir (approved):
🆒

Issue #3134, opened 2026-07-15
Report omni account's Total core count to Omni billing dashboard


steverfrancis:
### Problem Description

Given we considering counting cores for billing purposes, we need to know the current core count of a customer in our Grafana, along with host count.

PR #3145, opened 2026-07-16
fix: expose the machines cores and virtual state watches metrics

utkuozdemir:
The gauge counting the total number of CPU cores was updated on every reconcile but never included in the collected metrics, so it did not show up on the Prometheus endpoint. The gauge tracking active virtual state watches had a similar problem, its collector was never registered. Expose both and add tests that check the exposed metrics.

Part of siderolabs/omni#3134.


smira (approved):

Unix4ever (approved):
Oh good catch

utkuozdemir:
This PR was accidentally not opened from my fork, but I'll keep it, not going to close and re-open.

PR #3153, opened 2026-07-17
test: run browser e2e tests off the host network

utkuozdemir:
The e2e jobs ran the browser container on the host network, while machine provisioning constantly adds and removes network interfaces on that same host. Chromium treats every interface change as a network change and aborts all in-flight requests, including the streams that feed the UI, so pages intermittently showed stale or missing data and the tests flaked, most often on the cluster create and scale flows.

Run the browser container in its own network namespace instead, reaching Omni and Dex through a host gateway mapping. The OIDC issuer now uses the same hostname as the Omni endpoint so the login flow also works from outside the host network namespace.

### DIFF
--- a/hack/test/e2e-talemu.sh
+++ b/hack/test/e2e-talemu.sh
@@ -62,6 +62,11 @@ SIDEROLINK_DEV_JOIN_TOKEN="${JOIN_TOKEN}" \
 
 # Run the e2e test.
 # the e2e tests are in a submodule and need to be executed in a container with playwright dependencies
+#
+# Keep this container off the host network. Machine provisioning adds and removes host
+# network interfaces, and Chromium aborts every in-flight request (including the resource
+# watch streams feeding the UI) with ERR_NETWORK_CHANGED when it sees an interface change
+# in its own network namespace, so on the host network the UI tests randomly flake.
 cd frontend/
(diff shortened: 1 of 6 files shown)


Unix4ever (approved):

Slessi (approved):

hack/test/e2e-talemu.sh
```diff
@@ -62,6 +62,11 @@ SIDEROLINK_DEV_JOIN_TOKEN="${JOIN_TOKEN}" \
 
 # Run the e2e test.
 # the e2e tests are in a submodule and need to be executed in a container with playwright dependencies
+#
+# Keep this container off the host network. Machine provisioning adds and removes host
+# network interfaces, and Chromium aborts every in-flight request (including the resource
+# watch streams feeding the UI) with ERR_NETWORK_CHANGED when it sees an interface change
+# in its own network namespace, so on the host network the UI tests randomly flake.
```
Slessi, line 69: this comment is only replicated here so kinda missable but i dont mind

frezbo (approved):

oguzkilcan (approved):

hack/test/e2e-talemu.sh
utkuozdemir: replicated

PR #3154, opened 2026-07-17
chore: bump talos machinery to v1.14.0-alpha.2

oguzkilcan:
Bump talos machinery to v1.14.0-alpha.2. Add appropriate comments for each deprecated field/method.

Rework integration tests to migrate from the deprecated WithNodes API to per-node WithNode calls.


majabojarska (approved):

internal/backend/runtime/talos/talos.go
```diff
@@ -118,17 +117,15 @@ func (r *Runtime) Get(ctx context.Context, setters ...runtime.QueryOption) (any,
...
+		c   *Client
+		err error
 	)
 
 	switch len(opts.Machines) {
 	case 0:
 		c, err = r.GetClientForCluster(ctx, opts.Context)
 	case 1:
 		c, err = r.GetClientForMachine(ctx, opts.Machines[0])
-		responseMetadata = &taloscommon.Metadata{Hostname: opts.Machines[0]}
```
smira, line 131: this seems to be different than what Talos does for what it's worth; 

This is COSI API (from Talos point of view), which was never supported with `WithNodes`, so this is Omni inserting something in the response which has never been there with "raw" Talos API. 

So whether to remove or keep it is up to Omni

oguzkilcan: Let's go with removal. I couldn't find any place where it's utilized.

smira (approved):

PR #3155, opened 2026-07-17
feat(frontend): dim machine stage status when machine is not healthy

Slessi:
When machine is not healthy, dim the MachineStage status like ClusterMachinePhase to indicate this. Last known status will continue to be shown.

image

### DIFF
--- a/frontend/src/components/Status/MachineStage.vue
+++ b/frontend/src/components/Status/MachineStage.vue
@@ -5,8 +5,12 @@ Use of this software is governed by the Business Source License
 included in the LICENSE file.
 -->
 <script setup lang="ts">
+import { computed } from 'vue'
+
 import type { Resource } from '@/api/grpc'
 import type { MachineStatusLinkSpec } from '@/api/omni/specs/ephemeral.pb'
+import { MachineStatusSnapshotSpecPowerStage } from '@/api/omni/specs/omni.pb'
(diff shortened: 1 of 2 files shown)


Unix4ever (approved):

frontend/src/components/Status/MachineStage.vue
```diff
@@ -25,12 +39,12 @@ const { status } = useDerivedMachineStage(() => machine.spec.snapshot)
       cn(
         'inline-flex items-center gap-1 text-xs',
         status.class,
-        { 'opacity-50': machine.spec.tearing_down },
+        { 'brightness-50': !isHealthy },
```
majabojarska: I'm wondering whether a change of color (hue) would be a better way to communicate this. The change in brightness is very subtle in my opinion. Maybe yellow-ish as a warning?

Slessi: I'm just not sure if its warning level information, the machine may actually be running, Omni just doesn't know the current state and is displaying the last known state.

Mostly just doing this for feature parity with the cluster page (it is already this way, machines did not have anything)

@Unix4ever WDYT? Is it something that we should "warn" to user about?

majabojarska (approved):

frontend/src/components/Status/MachineStage.vue
majabojarska: Fair, approved

PR #3158, opened 2026-07-17
feat: label Talos Enterprise machines

### LINKED ISSUE #2724: label Talos Enterprise machines
Talos 1.13+ in the `talosctl get version` reports the `Name` which is `Talos Enterprise` for Enterprise builds.

The FIPS status is reported in `securitystate`. 

We should label machines with "Enterprise" label which are Talos Enterprise. 

We can bubble up FIPS status as well.

utkuozdemir:
Machines running Talos Enterprise now get an enterprise label, and machines running Talos in FIPS mode get a fips label that also shows whether the strict mode is on. Like the other machine labels managed by Omni, they cannot be removed by the user, and they are cleared when a machine is downgraded to a regular build.

The version name and the FIPS state are read over the Talos API and stored in the machine status. Both are watched for changes, as they follow the installed image and can change when the machine is upgraded to a different build. Older Talos versions do not report this information, so machines running them simply get no labels.

Closes siderolabs/omni#2724.


utkuozdemir:
How it looks:
enterprise-fips-labels

frezbo:
> How it looks: enterprise-fips-labels

only nit is it's not obvious at first glance, should we highlight it more?

if it's supposed to be a label, then it's fine

utkuozdemir:
> > How it looks: enterprise-fips-labels
> 
> only nit is it's not obvious at first glance, should we highlight it more?
> 
> if it's supposed to be a label, then it's fine

Yes that's a question to @Slessi and @sterlingkoch I guess - my intention was to do the groundwork - design is to be iterated on.

client/api/omni/specs/omni.proto
```diff
@@ -22,12 +22,21 @@ message MachineSpec {
   bool use_grpc_tunnel = 5;
 }
 
-// SecurityState describes the UKI and secure boot information of the machine.
+// SecurityState describes the security state of the machine: UKI, secure boot and FIPS information.
 message SecurityState {
+  // FIPSState describes the FIPS state of the machine.
+  enum FIPSState {
+    FIPS_STATE_DISABLED = 0;
```
oguzkilcan, line 29: Do these need to be this explicit? We already have `FIPSState` in the enum name.

oguzkilcan: I saw this later:
```
fipsState = specs.SecurityState_FIPS_STATE_DISABLED
```
Ignore this comment. It makes things readable.

oguzkilcan (approved):

Slessi:
> only nit is it's not obvious at first glance, should we highlight it more?

Is that a problem? Making one piece of content more obvious makes the surrounding content less obvious. Don't know if this label is important enough for that, but maybe it is.

I would expect in the future after machines are migrated, and for new users, all machines would be enterprise/fips if that is relevant to your organisation, right? In which case, it isn't really notifying you of much.

PR #3162, opened 2026-07-20
test: fix Omni upgrade workload proxy DNS

utkuozdemir:
The previous-version integration container needs both the Omni instance hostname for Dex and the base hostname used by workload proxy requests. Add the missing base hostname mapping so the upgrade preparation suite reaches the embedded Omni server.


utkuozdemir:
Already passed e2e-omni-upgrade - removing label so it won't run after rebase.

frezbo (approved):

oguzkilcan (approved):

PR #3165, opened 2026-07-20
feat: allow disabling config patches

### LINKED ISSUE #3144: add a button to disable/enable config patches
Backend side should also get some changes. We can introduce a new label `omni.sidero.dev/disabled`, then when the controller gets the patches to apply it can filter them by `!omni.sidero.dev/disabled` query.
Related file for the change on the backend `internal/backend/runtime/omni/controllers/omni/internal/layeredresource/layeredresource.go`

The UI will then just add/remove the label on the config patch resource. The button can exist on the list items and when the patch is open.

Slessi:
Allow disabling config patches for a machine, which keeps it around as a resource but prevents it from ever getting applied.

Closes #3144

https://github.com/user-attachments/assets/675ce1f5-f249-4cc2-88cc-2a41a958588f

### DIFF
--- a/internal/backend/runtime/omni/controllers/omni/internal/layeredresource/layeredresource.go
+++ b/internal/backend/runtime/omni/controllers/omni/internal/layeredresource/layeredresource.go
@@ -15,6 +15,7 @@ import (
 	"github.com/cosi-project/runtime/pkg/controller/generic"
 	"github.com/cosi-project/runtime/pkg/resource"
 	"github.com/cosi-project/runtime/pkg/safe"
+	"github.com/cosi-project/runtime/pkg/state"
 	"github.com/siderolabs/gen/xslices"
 
 	"github.com/siderolabs/omni/client/pkg/omni/resources/omni"
@@ -27,7 +28,7 @@ type Helper[T generic.ResourceWithRD] struct {
 
 // NewHelper creates a new helper for managing layered resources.
 func NewHelper[T generic.ResourceWithRD](ctx context.Context, r controller.Reader) (*Helper[T], error) {
-	allResources, err := safe.ReaderListAll[T](ctx, r)
+	allResources, err := safe.ReaderListAll[T](ctx, r, state.WithLabelQuery(resource.LabelExists(omni.LabelDisabled, resource.NotMatches)))
 	if err != nil {
 		return nil, err
 	}
--- a/client/pkg/omni/resources/omni/labels.go
+++ b/client/pkg/omni/resources/omni/labels.go
@@ -49,6 +49,10 @@ const (
 	// tsgen:LabelSystemPatch
 	LabelSystemPatch = SystemLabelPrefix + "system-patch"
 
+	// LabelDisabled marks a config patch as disabled, so it is skipped when the config patches are applied to the machines.
+	// tsgen:LabelDisabled
+	LabelDisabled = SystemLabelPrefix + "disabled"
+
 	// LabelSystemManifest marks the manifest as the system manifest, so it shouldn't be editable by the user.
 	// tsgen:LabelSystemManifest
 	LabelSystemManifest = SystemLabelPrefix + "system-manifest"
--- a/frontend/eslint-suppressions.json
+++ b/frontend/eslint-suppressions.json
@@ -1,9 +1,4 @@
 {
-  "src/components/ActionsBox/TActionsBoxItem.vue": {
-    "vue/define-props-destructuring": {
-      "count": 1
-    }
-  },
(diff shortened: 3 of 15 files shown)


Unix4ever:
I would suggest to use a toggle like this one:
image

Slessi:
@Unix4ever for where? on the row in the patch list? or for inside the patch edit? (or both?)

Unix4ever:
> @Unix4ever for where? on the row in the patch list? or for inside the patch edit? (or both?)

For both. I guess inside patch edit it can also have text `disabled/enabled`

Slessi:
@Unix4ever so like this? 

image

image

I think list will feel quite noisy with a lot of orange switches in a column

image

Unix4ever:
> @Unix4ever so like this? 
> 
> image
> 
> image
> 
> I think list will feel quite noisy with a lot of orange switches in a column
> 
> image
> 
> 

I think we shouldn't have the text on the list view, only on the edit page.

We can also make it a bit smaller in the list.

Unix4ever:
Also to make it less noisy we could use a different color in the list. Some shade of gray instead of primary.

But dropping text and making the toggle smaller should already make it much better.

As another option I guess we could keep using the icon, but I think play/pause doesn't fit well here.

@sterlingkoch what do you think?

Slessi:
@Unix4ever I don't like the idea of using gray because it makes it harder to understand the current state of the switch. IMO all controls which are on/enabled/active etc. should be using primary color to indicate this, for consistency and quick consumption.

I think it's just the wrong tool for the job. Its doing two jobs here:
1. Display current state
2. Toggle the state

But:
1. The state is almost always going to be enabled, and rarely disabled.
2. Users don't need the convenience of quickly toggling multiple patches.

So better is something that _only_ shows the unusual state, which is disabled, and reports nothing for enabled. And the toggle action is in a place that doesn't distract from content.

Lets do like in machine/cluster lists with a `...` dropdown button and move disable/enable and also the delete action into that dropdown, and only display "Disabled" or some icon with a tooltip for disabled patches. I think that is better than also the play/pause thing I tried initially. `PatchEdit` can keep the switch i think that is perfectly fine.

Unix4ever:
> @Unix4ever I don't like the idea of using gray because it makes it harder to understand the current state of the switch. IMO all controls which are on/enabled/active etc. should be using primary color to indicate this, for consistency and quick consumption.
> 
> I think it's just the wrong tool for the job. Its doing two jobs here:
> 
> 1. Display current state
> 2. Toggle the state
> 
> But:
> 
> 1. The state is almost always going to be enabled, and rarely disabled.
> 2. Users don't need the convenience of quickly toggling multiple patches.
> 
> So better is something that _only_ shows the unusual state, which is disabled, and reports nothing for enabled. And the toggle action is in a place that doesn't distract from content.
> 
> Lets do like in machine/cluster lists with a `...` dropdown button and move disable/enable and also the delete action into that dropdown, and only display "Disabled" or some icon with a tooltip for disabled patches. I think that is better than also the play/pause thing I tried initially. `PatchEdit` can keep the switch i think that is perfectly fine.

That also works :+1:

Slessi:
@Unix4ever done, updated video on PR desc to reflect changes

Unix4ever:
Should be aligned here:

image

Was on the same line before

internal/backend/runtime/omni/controllers/omni/internal/layeredresource/layeredresource.go
```diff
@@ -80,6 +80,15 @@ func (h *Helper[T]) Get(machine resource.Resource, machineSet *omni.MachineSet)
 	resources = slices.AppendSeq(resources, machineLevel.All())
 
 	return xslices.Filter(resources, func(res T) bool {
-		return res.Metadata().Phase() == resource.PhaseRunning
+		if res.Metadata().Phase() != resource.PhaseRunning {
+			return false
+		}
+
+		// skip disabled resources (e.g. config patches disabled via the UI)
+		if _, disabled := res.Metadata().Labels().Get(omni.LabelDisabled); disabled {
```
Unix4ever: Can be more efficient.
On the line 30 where we list all config patches we can do:

```golang
allResources, err := safe.ReaderListAll[T](ctx, r, state.WithLabelQuery(resource.LabelExists(omni.LabelDisabled, resource.NotMatches)))
```

Then disabled patches won't even appear in the list we filter here.

Slessi: done

Unix4ever (approved):

PR #3169, opened 2026-07-21
feat: factory migration support

### LINKED ISSUE #2760: support for two Image Factories
A user has an Omni instance with has running clusters today. A user wants to migrate to Talos Enterprise.

Talos Enterprise is only available for versions >= v1.13.0.

Omni has connections to two Image Factories:

* Enteprise one serving Talos >= 1.13.0 (new IF)
* Open-source one serving all Talos versions (old/legacy IF)

Omni pulls a list of Talos versions from both Image Factories, and "prefers" Enterprise for overlapping version.

As Omni shows list of versions, it labels them with "Enterprise" if they would come from an Image Factory Enterprise.

A user doesn't have a choice to pick an open-source Image Factory for a Talos version once it is available in Image Factory Enterprise (>= 1.13.0). If a user already runs Talos >= 1.13.0 open-source, the next machine upgrade (including kernel args change, system extension list change) will automatically upgrade Talos to Talos Enterprise.

As the cluster is upgraded, as it moves to Enterprise version, it gets upgraded from an Enterprise IF, and gets Talos Enterprise.

It is prohibited to downgrade from Talos Enterprise to Talos open-source.

---

If Omni loses connection to the Image Factory it had previously - it considers machines which came from that Image Factory as orphaned, and it doesn't allow to upgrade those machines. The only operation allowed on such machines is to destroy them.

---

Notes: Image Factory now embeds its own identity (`Image Factory` or `Image Factory Enteprise`) and its own URL (e.g. factory.talos.dev) into the Author field of the virtual schematic extension.

Notes: Image Factory reports its own flavor in the `Server:` header in the response.

Unix4ever:
Omni now supports running against two image factories: one primary and one secondary.
This allows migration from one factory to another.

Fixes: https://github.com/siderolabs/omni/issues/2760


client/pkg/omnictl/internal/download/download.go
```diff
@@ -620,12 +625,9 @@ func DownloadImageTo(ctx context.Context, client *client.Client, image ImageInfo
...
-	if err != nil {
-		return err
-	}
+	fmt.Fprintf(os.Stderr, "Using image factory at %q\n", image.FactoryURL)
 
-	username, password, err := ImageFactoryCredentials(ctx, client.Omni().State(), features)
+	username, password, err := ImageFactoryCredentials(ctx, client.Omni().State(), image.FactoryURL)
 	if err != nil {
 		return err
 	}
```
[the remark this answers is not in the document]
Unix4ever: It will just have empty credentials.

internal/pkg/config/schema.json
```diff
@@ -146,7 +146,8 @@
       "required": [
         "talos",
         "kubernetes",
-        "imageFactoryBaseURL"
+        "imageFactoryBaseURL",
+        "factories"
       ],
```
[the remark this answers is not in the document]
Unix4ever: @utkuozdemir can you help me a bit here? How to do that properly?

internal/backend/runtime/omni/controllers/omni/cluster_machine_config.go
```diff
@@ -374,7 +440,26 @@ func (helper clusterMachineConfigControllerHelper) generateConfig(clusterMachine
 		return nil, fmt.Errorf("talos version is not set on the resource %s", clusterConfigVersion.Metadata())
 	}
 
-	installImage, err := installimage.Build(helper.imageFactoryHost, configGenOptions.Metadata().ID(), configGenOptions.TypedSpec().Value.InstallImage, helper.talosRegistry)
+	installImageSpec := configGenOptions.TypedSpec().Value.InstallImage
+
+	// Migration code, if the factory URL is not populated yet, use the secondary factory URL if configured
+	// As the second factory should be the old one we're migrating from
```
Slessi, line 417: this feels easy to get wrong as a user... not possible to have `image_factory_url` and `enterprise_image_factory_url` instead of primary vs secondary?

frontend/src/views/InstallationMedia/Confirmation.vue
```diff
@@ -61,11 +64,27 @@ const { data: quirks } = useResourceGet<QuirksSpec>(() => ({
...
+    id: resolvedTalosVersion.value,
+  },
+}))
 
-const imageFactoryAuth = useImageFactoryAuth()
+const { base: factoryBaseURL, credentials: factoryCredentialsRef } = useResolvedFactory(
+  () => talosVersion.value?.spec.image_factory_url,
+)
+
+const isEnterpriseFactory = computed(() => talosVersion.value?.spec.is_enterprise)
```
Slessi, line 83: I wonder does this make the `ClusterSecurity` page check wrong, and we need to check it per-cluster? Or I guess it is fine to run security scan on the legacy clusters that were created without enterprise, if they have enterprise configured in general?

frontend/src/pages/(authenticated)/machines/installation-media/create/talos-version.vue
```diff
@@ -82,6 +82,22 @@ const joinTokens = computed(() => [
...
+      .image_factory_url,
+)
+
+watch(
+  resolvedFactoryURL,
+  (url) => {
+    formState.value.imageFactoryUrl = url
+  },
+  { immediate: true },
+)
```
Slessi, line 99: This `imageFactoryUrl` field isn't used by the form, or set by the user. I don't think it belongs here? Sounds like a backend only concern

internal/backend/runtime/omni/controllers/omni/cluster_machine_config.go
Unix4ever: Primary and secondary doesn't necessarily mean that primary factory is enterprise.

frontend/src/pages/(authenticated)/machines/installation-media/create/talos-version.vue
Unix4ever: We need to know the correct image factory URL when we create the preset. I guess we could detect that on backend, but this resource is user managed, so it's easier to populate it fully on the frontend.

frontend/src/views/InstallationMedia/Confirmation.vue
Unix4ever: I guess it should be fine

internal/backend/runtime/omni/controllers/omni/cluster_machine_config.go
Slessi: Isn't the purpose of this just to support migration from non-enterprise to enterprise? If it has scope beyond that, I wonder if it should just be a list of N factories, rather than exactly 1 primary and optionally 1 secondary

This PR looks to me _specifically_ to be 1 primary enterprise factory, 1 secondary non-enterprise factory

frontend/src/pages/(authenticated)/machines/installation-media/create/talos-version.vue
Slessi: If its resolved by the backend at the moment it needs it, it would be dynamic in the event that factory URLs change. With this implementation the preset can go out of sync with reality and requires you to re-save it without changing a single thing just to fix this field you can't even see

internal/pkg/config/schema.json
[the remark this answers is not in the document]
utkuozdemir: I checked, and this is a false positive. The config validation runs only after all config sources are merged. We put those `required` properties just to make the config code generator make it a struct, not a pointer to a struct. It doesn't force its fields to be populated. We mark every intermediary object as required, and it works. It'll keep working with this - no issues here.

frontend/src/pages/(authenticated)/machines/installation-media/create/talos-version.vue
Unix4ever: Different factories might have different schematics, so we need to know which factory was used when the preset was created initially.

So backend doesn't resolve it dynamically. If the factory URL is no longer configured in Omni preset download should be blocked. That's why we have this field here.

internal/backend/runtime/omni/controllers/omni/cluster_machine_config.go
Unix4ever: It's migration from one factory to another one, not necessarily enterprise. You can migrate from the official one to your self hosted factory for example.

So it's always two.

internal/pkg/config/schema.json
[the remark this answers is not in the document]
Unix4ever: thanks :)

internal/backend/runtime/omni/controllers/omni/cluster_machine_config.go
Unix4ever: Having more than 2 doesn't really make much sense. As the first one always wins if it has the version available.
If we provided a way to pick the factory to use then having N factories could make sense. But that's not the goal here.

frontend/src/pages/(authenticated)/machines/installation-media/create/talos-version.vue
Slessi: I thought we fixed it already so that Omni never assumes factory has a given schematic here, and will always ask for it, but ok, clear

internal/backend/grpc/management.go
```diff
@@ -763,6 +764,33 @@ func validateAuditLogFollowRequest(req *management.ReadAuditLogRequest) error {
...
+			return "", "", fmt.Errorf("failed to patch schematic: %w", patchErr)
+		}
+
+		ensureCtx, cancel := context.WithTimeout(ctx, time.Minute)
+		defer cancel()
+
+		if id, _, err = factoryClient.EnsureSchematic(ensureCtx, patched); err != nil {
+			return "", "", fmt.Errorf("failed to ensure schematic on factory %q: %w", factoryClient.URL(), err)
+		}
+	}
```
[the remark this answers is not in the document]
Unix4ever: the concern is valid, but the fix incorrect.

client/pkg/imagefactory/clients.go
```diff
@@ -0,0 +1,185 @@
...
+
+// FactoryClient is the contract Omni controllers rely on for interacting with a single image factory.
+type FactoryClient interface {
+	EnsureSchematic(ctx context.Context, inputSchematic schematic.Schematic) (string, *schematic.Schematic, error)
+	SchematicGet(ctx context.Context, id string) (*schematic.Schematic, error)
+	Host() string
+	URL() string
+	OverlaysVersions(context.Context, string) ([]client.OverlayInfo, error)
+	Versions(context.Context) ([]string, error)
+	ExtensionsVersions(context.Context, string) ([]client.ExtensionInfo, error)
```
oguzkilcan, line 27: Small nit: you can name these arguments, so all methods in the interface follow the same pattern

Unix4ever: Here I mostly need it for mocking the image factory client.

oguzkilcan: It's not big a deal. Just some method args are named and some are not. I just meant that.

client/pkg/omnictl/installationmedia/create.go
```diff
@@ -158,6 +160,16 @@ func createPreset(ctx context.Context, cmd *cobra.Command, client *client.Client
 		}
 	}
 
+	imageFactoryClients, err := imagefactory.NewClientsFromState(ctx, client.Omni().State())
+	if err != nil {
+		return err
+	}
+
+	imageFactoryClient, err := imageFactoryClients.ForTalosVersion(ctx, talosVersion)
```
oguzkilcan: `talosVersion` here can be empty you should use `resolveTalosVersion`

internal/backend/runtime/omni/controllers/omni/machineconfig/status.go
```diff
@@ -215,15 +217,22 @@ func (ctrl *StatusController) reconcileRunning(
...
-
-	stage := rc.machineStatusSnapshot.TypedSpec().Value.GetMachineStatus().GetStage()
-	if stage == machineapi.MachineStatusEvent_BOOTING || stage == machineapi.MachineStatusEvent_RUNNING {
-		if err = ctrl.deleteUpgradeMetaKey(ctx, logger, r, rc); err != nil {
+	// A high priority config change (e.g. the registry auth for a newly configured image factory)
+	// must reach the machine before any upgrade: the upgrade pulls the installer image from that
+	// factory, so the credentials have to be applied to the machine config first. Skip the upgrade
+	// for this reconcile and apply the config immediately; the upgrade proceeds on a later reconcile
+	// once the machine already has the new auth.
+	if !computeDiffResult.hasHighPriorityConfigChanges {
```
oguzkilcan: Applying the config doesn't install talos on disk for >= 1.13. If machine was booted from with non-enterprise factory iso, and on cluster creation screen a talos version from IFE was selected, I presume this check will win and we will attempt to apply full cluster config to a talos on memory. Talos on memory might be a lower minor, not supporting all the fields present in the config. We were installing/upgrading the machine to match the cluster's talos version before applying full cluster config for this reason.

internal/pkg/config/config.go
```diff
@@ -254,7 +235,11 @@ const (
 )
 
 // applyEnvOverrides overrides config values from environment variables, when set.
-// Env overrides win over all other sources (defaults, files, flags).
+//
+// These set the deprecated flat imageFactory* fields, which GetPrimaryFactory treats as the
```
oguzkilcan: Aren't we going to offer env vars for username/password for the new structure?

client/pkg/clusterimport/client.go
```diff
@@ -255,22 +252,3 @@ func BuildTalosClient(ctx context.Context, config, context, sideroV1KeysDir stri
 		Client: c,
 	}, nil
 }
-
-// BuildImageFactoryClient builds an Image Factory client using the Image Factory base url from Omni configuration.
-func BuildImageFactoryClient(ctx context.Context, omniState state.State) (*imagefactory.DirectClient, error) {
```
utkuozdemir, line 260: why removed

internal/backend/grpc/management.go
```diff
@@ -763,6 +764,33 @@ func validateAuditLogFollowRequest(req *management.ReadAuditLogRequest) error {
...
+	factoryClient, err := s.imageFactoryClients.ForTalosVersion(ctx, talosVersion)
+	if err != nil {
+		return "", "", fmt.Errorf("failed to get image factory client for Talos version %q: %w", talosVersion, err)
+	}
+
+	schematicSpec := machineStatus.TypedSpec().Value.GetSchematic()
+
+	id := schematicSpec.GetFullId()
+
+	if schematicSpec.GetFullId() != "" && schematicSpec.GetRaw() != "" && !schematicSpec.GetInvalid() && !schematicSpec.GetInAgentMode() {
```
utkuozdemir, line 777: maybe we can reuse here the `.SchematicReady()` function on the MachineStatus spec - this is a common logic. But still, need to check if it matches the check you want here exactly.

internal/backend/grpc/schematics.go
```diff
@@ -137,12 +154,22 @@ func (s *managementServer) CreateSchematicFromRaw(ctx context.Context, request *
...
+		return nil, fmt.Errorf("failed to get image factory client for Talos version: %w", err)
+	}
+
+	schematicID, _, err := factoryClient.EnsureSchematic(ctx, *schematicRequest)
 	if err != nil {
 		return nil, fmt.Errorf("failed to ensure raw schematic: %w", err)
 	}
 
-	s.logger.Info("ensure raw schematic succeeded", zap.String("schematic_id", schematicID), zap.String("factory_host", s.imageFactoryClient.Host()))
+	s.logger.Info("ensure raw schematic succeeded", zap.String("schematic_id", schematicID), zap.String("factory_host", s.imageFactoryClients.Primary().Host()))
```
utkuozdemir: Is this intentional?

internal/backend/runtime/omni/controllers/omni/machineconfig/status.go
```diff
@@ -1477,6 +1505,69 @@ func (ctrl *StatusController) computePendingUpdates(ctx context.Context, r contr
 	})
 }
 
+// computeHighPriorityConfigChanges reports whether the transition from oldConfig to newConfig adds
+// or modifies an image factory registry auth document (cri.RegistryAuthConfig). Such changes are
+// high priority: a machine that just gained credentials for a newly configured image factory must
+// receive them before it upgrades, since the upgrade pulls the installer image from that factory.
+func computeHighPriorityConfigChanges(oldConfig, newConfig []byte) (bool, error) {
```
utkuozdemir: ❤️

internal/backend/runtime/omni/controllers/omni/machineconfig/status_test.go
utkuozdemir, line 1: Can we add a controller unit test here to cover the high priority flow? To make sure there's no gaps, deadlocks etc.

internal/backend/runtime/omni/controllers/omni/machine_status.go
```diff
@@ -530,6 +536,13 @@ func (ctrl *MachineStatusController) handleNotification(ctx context.Context, r c
 	// preserves all fields the user/factory set on the original schematic (owner,
 	// secureboot, bootloader, meta, overlay incl. options, future fields).
 	if event.Schematic != nil && event.Schematic.FullID != "" && event.Schematic.Raw != "" {
+		talosVersion := pointer.SafeDeref(event.TalosVersion)
+
+		factoryClient, err := ctrl.imageFactoryClients.ForTalosVersion(ctx, talosVersion)
```
utkuozdemir, line 541: Is empty value for `talosVersion` expected here? What happens when we resolve the factory client with an empty string?

internal/pkg/config/config.go
utkuozdemir: +1 - at least the passwords should be able to passable via env.

internal/backend/runtime/omni/controllers/omni/cluster_machine_config.go
```diff
@@ -321,6 +338,55 @@ func (helper clusterMachineConfigControllerHelper) buildRegistryAuthPatch() (str
 	return string(authBytes), nil
 }
 
+// configMissingFactoryAuth reports whether more than one image factory is configured with
+// credentials and the existing machine config is missing the registry auth document for any of
+// them. It is used to force a config regeneration for machines whose config was generated before
+// the additional factory was configured, since the factory configuration is not tracked as a
+// resource input.
+func (helper clusterMachineConfigControllerHelper) configMissingFactoryAuth(machineConfig *omni.ClusterMachineConfig) (bool, error) {
```
Unix4ever: We can make ImageFactoryAuth resources not virtual and then use them as an input to this controller instead.

utkuozdemir:
@Unix4ever Here's the review from multiple agents:

## Must fix

### 1. Fresh machines apply config before install
`machineconfig/status.go:225` skips the upgrade path when the pending diff contains a high priority config change. On the first reconcile of any machine the old config is empty, so any configured factory credentials count as "added" and the check fires (see the "auth added" case in `high_priority_test.go`). The result: every new machine on an instance with factory credentials applies the full cluster config in maintenance mode before installing. If the boot media is an older minor than the cluster version, the apply fails validation, the high priority flag never clears, and the machine deadlocks. If the apply succeeds on Talos 1.13 or newer, the machine can leave maintenance running from RAM with nothing on disk, after which the maintenance install skips forever with "not in maintenance stage".

The planned MaintenanceConfigStatus ready flag is a sound fix, with three requirements:

- It has to land together with item 4. The maintenance config only carries primary factory credentials today, so machines that pull from the secondary would wait on credentials that never arrive.
- It needs an explicit "not applicable" state for clusters created on Talos 1.11 or older. Those never get auth docs at all (see item 10), so a plain wait would hang forever.
- The readiness gate has to replace the skip at status.go:225 for machines with no applied config, not sit next to it. Otherwise a fresh machine still applies full config first.

Also worth adding the controller-level test I already asked for on the PR, covering exactly the fresh-machine-in-maintenance flow.

### 2. Empty image factory host is preserved forever and stalls upgrades
`machine_config_gen_options.go:70-74` keeps the stored install image host when the desired version and schematic are unchanged. For every machine that predates this Omni version the stored host is empty, and both compared values come from ClusterMachineTalosVersion, so they match at steady state and the empty value is preserved indefinitely. `cluster_machine_config.go` has a fallback for this, but `installimage.Build` now hard-errors on an empty host (`installimage.go:69`) and is called with the gen-options spec directly from `machineconfig/status.go:512` and `lifecycle/client.go`, with no fallback. A machine that is behind its desired version at the moment Omni is upgraded gets stuck in a Build error loop until someone bumps the version or extensions. Fix: never preserve an empty host, apply the same secondary-else-primary fallback in gen options.

### 3. A transient secondary factory outage deletes its Talos versions
`fetchTalosVersions` logs and continues when the secondary fetch fails, and `tracker.cleanup` at `versions.go:374` then destroys every TalosVersion resource that was not visited. A single 503 from the secondary wipes all secondary-only versions, after which `ForTalosVersion` silently routes those versions to the primary. Before this PR, any factory error aborted the whole reconcile, so this destructive path is new. Fix: fail the reconcile when either configured factory cannot be enumerated, or at least never prune based on an incomplete observation. The two fetches also share one 30 second context, so a slow secondary eats the primary's entire budget. Give them separate deadlines or run them concurrently.

## Should fix

### 4. Maintenance config ships only primary credentials
`maintenance_config_status.go:169` still uses `BuildDoc` (primary only) while the cluster config path moved to `BuildDocs` (both factories). A maintenance machine whose install pulls from the secondary has no credentials for it. This matters for the self-hosted to self-hosted migration Artem says is in scope, and it undermines the item 1 fix, so the two belong in the same change.

### 5. Config fallback can pair old credentials with a new factory URL
`factories.go:24-31` resolves each field of the primary factory independently, falling back to the deprecated flat fields per field. Setting only `factories.primary.url` on a config that has legacy credentials sends the old factory's basic auth to the new host, and derives the PXE URL from the old flat field so PXE requests route to the wrong factory. The credential leak is self-limiting (the new factory rejects them), the PXE misrouting is the concrete harm. Fix: only consult the deprecated fields when `factories.primary.url` itself is unset.

### 6. Enterprise detection latches on the first response, including errors
`client/pkg/imagefactory/client.go:34-41` records the enterprise state from the first round trip that returns without a transport error, which includes a 503 or an LB error page without the Server header. A misdetected enterprise factory then has Owner stripped from every schematic for the process lifetime (`client.go:120-123`). Latch only on a 2xx. There is also a small race: `detected` and `isEnterprise` are two separate atomics, so a concurrent EnsureSchematic can see detected as true before the enterprise bit lands. Publish both under one synchronization point.

### 7. Schematic lookups miss the machine's actual factory during migration
`machine_status.go:366` resolves the factory from the machine's Talos version, and versions present on both factories are always attributed to the primary. A machine booted from the secondary's ISO whose schematic exists only there gets a 404 and never populates its raw schematic. The same root cause sits in `management.go` ensureSchematic: when Raw is empty it returns the stored full ID without verifying it exists on the target factory, so the built install image can 404 (Copilot raised this on the PR, Artem agreed the concern is valid). Fix for both: on 404, try the other configured factory, and fail fast with a clear error instead of building an image the machine cannot pull.

### 8. The legacy empty-host fallback ignores the resolved primary
`cluster_machine_config.go:448` falls back to the deprecated `GetImageFactoryBaseURL()` instead of `GetPrimaryFactory().GetUrl()`. An install configured only through `factories.primary.url` regenerates old empty-host resources with the default community host.

### 9. Old artifacts with no factory identity resolve to the wrong factory
Pre-upgrade installation media presets have an empty factory URL, and `installationmedia/download.go:143` interprets that as the current primary. Those presets were created against the old factory, which after a migration is the secondary. Old omnictl binaries have the same problem on the server side: `schematics.go` resolves their requests by Talos version, possibly creating the schematic on one factory while the old binary downloads from another. Empty identity should follow the same secondary-else-primary rule as the config fallback, or presets should be backfilled once.

### 10. configMissingFactoryAuth: never converges for old clusters, blind to rotation, and expensive
Three related problems in `cluster_machine_config.go`, best fixed by one mechanism.

- The auth patch is only added to generated configs when the cluster's creation-version contract supports multidoc network config, meaning Talos 1.12 or newer (`cluster_machine_config.go:553`). The missing-auth check has no such gate, so for any cluster created on Talos 1.9 to 1.11 with credentials configured it reports missing auth forever and permanently defeats the input-versions short circuit. No write loop thanks to configsEqual, but full config regeneration plus parse on every reconcile, for the cluster's lifetime.
- The check compares document names only, so credential rotation never triggers regeneration. Rotation did not propagate before this PR either, but the redacted-config comparison also makes the high priority detector blind to password changes, so the ordering machinery this PR adds cannot see the very change it exists to order.
- The check decompresses and parses the full machine config on every reconcile even when no factory has credentials at all.

Fix: compute a non-secret fingerprint of the desired auth documents, store it as an annotation on the output, gate it on the same version contract as the patch itself, and compare fingerprints instead of parsing configs. That covers regen triggering, rotation, ordering, and cost in one place.

### 11. Nothing stops two factories on the same host
The config schema accepts any strings for the factory URLs. If both resolve to the same host and both have credentials, `BuildDocs` produces two auth documents with the same name and `container.New` rejects the duplicate, which breaks config generation for every machine. Validate at startup that the two factories have distinct normalized hosts and absolute http(s) URLs.

### 12. The test helper cannot even construct a primary plus secondary pair
`testutils/factory.go:29` replaces the first client with an empty mock whenever two clients are passed, so no controller test actually exercises a two-factory setup. There are also no tests for `client/pkg/imagefactory/clients.go`, the version merge, or outage behavior. Ties into the controller test requested in item 1.

### 13. Client module API breaks
`BuildContext`, `NewInstallImage`, `NewImageFactoryAuth`, `CreateSchematicFromRaw` all changed signatures and `BuildImageFactoryClient` was deleted, all in the published client module. Acceptable under the lockstep versioning policy, but it should be a conscious call. The now-dead `clusterimport.ImageFactoryClient` interface (modified in this PR yet unused by anything) should be deleted either way.

### 14. omnictl download edge cases
`DownloadImageTo` must never see an empty factory URL: when it is empty and no secondary is configured, the PXE branch matches empty against empty and picks the empty secondary PXE URL, and the direct download builds a broken request. Fall back to the primary inside `DownloadImageTo`, which also resolves Copilot's original comment. Separately, `omnictl download --talos-version v1.9.0` fails the resource lookup because TalosVersion IDs have no leading v. The default value works, so this only hits user-typed versions. TrimPrefix before the lookup.

## Nits

- `schematics.go:172` logs the primary host instead of the resolved client's host (already commented on the PR).
- The `ImageFactoryClientProvider` doc comment references `ForURLOrPrimary`, which does not exist.
- The schematic subpackage imports its parent `controllers/omni` package for the provider interface. Declaring the small interface on the consumer side fixes both the layering and the producer-side interface smell.
- `Clients.ForURL/ForHost/ForTalosVersion` compare URLs exactly while the virtual state and validations trim trailing slashes. The one client-supplied URL (`CreateSchematicRequest.ImageFactoryUrl`) should be normalized in Clients.
- The content-equality skip in `schematic/configuration.go:246` never assigns SchematicId in the skip branch, though its comment claims it publishes the machine's full ID. A stale empty ID survives, for example after an invalid-to-valid transition. Add the else assignment.
- The TalosVersion mapper in the schematic controller re-reconciles every machine per TalosVersion event. It converges cheaply thanks to the content-equality skip, but it is O(versions times machines) on startup and every refresh.
- `ClusterMachineConfigStatusSpec.image_factory_url` is documented as a URL, gets fed a host, is only written on the legacy upgrade path, and has no readers. Populate it consistently or drop it.
- No env var support for the new factory credential fields (already commented on the PR, see item 5 for the related fallback subtlety).
- `download_test.go` dropped the test case for non-NotFound errors from the auth resource lookup.

## Design note, for discussion rather than as a defect

The stateful `ForTalosVersion` is the root of several findings: it does a hidden uncached state read per call from hot controller transforms, and it gives controllers an undeclared, non-reactive dependency on TalosVersion (gen options can stamp the wrong factory during the startup window and the migration guard then preserves it). The simpler shape is to keep `Clients` a pure URL-to-client map and carry the resolved factory identity through declared controller inputs, for example on ClusterMachineTalosVersion. That would remove the per-reconcile etcd reads and make factory changes reactive. It is a bigger refactor, so it is a menu item, not a blocker.

client/pkg/clusterimport/client.go
Unix4ever: It was using the factory client from the infra providers. I think it's wrong.
I switched it to use a generic image factory client (clientset actually, which automatically picks the right factory).

internal/backend/grpc/schematics.go
Unix4ever: Nope. Need to fix that.

internal/backend/runtime/omni/controllers/omni/machine_status.go
Unix4ever: If talos version is empty it falls back to the primary factory.

oguzkilcan (approved):

utkuozdemir (approved):

PR #3179, opened 2026-07-22
fix: allow SAMLLabelRule to downgrade user role to None

### LINKED ISSUE #3178: SAMLLabelRule with assignRole: None does not downgrade existing user roles
## Description

When using SAMLLabelRules with `assignRole: None` and `updateOnEachLogin: true`, existing users are never downgraded to `None` on login. All other role transitions (upgrades and downgrades between Reader/Operator/Admin) work correctly.

## Reproduction

**Prerequisites:** Self-hosted Omni with SAML (Keycloak), SAMLLabelRules with `updateOnEachLogin: true`:
- `assign-reader`: matchlabels `saml.omni.sidero.dev/role/omni-reader`, assignRole: Reader
- `assign-none`: matchlabels `saml.omni.sidero.dev/role/omni-none`, assignRole: None

**Steps:**
1. User has KC role `omni-reader` → logs in → Omni role = Reader ✅
2. User logs out → admin changes KC role to `omni-none` → user logs in → Omni role = **still Reader** ❌

**Expected:** Role should be downgraded to None.

## Root Cause

`MatchSAMLLabelRule()` in `internal/backend/saml/session.go` returns `nil` when the highest matched role is `None`.

The function starts with `samlRole := role.None` and only assigns the matched rule when `samlRole.Compare(maxRole) == -1`. When a rule matches with `assignRole: None`:
- `maxRole = Max(None, None) = None`
- `None.Compare(None) = 0` (not `-1`)
- `rule` is never assigned → returns `nil`

This causes `ensureUser()` to skip reading `updateOnEachLogin` from the matched rule, so the user's existing role is never updated.

## Role transition matrix

| From → To | None | Reader | Operator | Admin |
|-----------|------|--------|----------|-------|
| None | - | ✅ | ✅ | ✅ |
| Reader | ❌ | - | ✅ | ✅ |
| Operator | ❌ | ✅ | - | ✅ |
| Admin | ❌ | ✅ | ✅ | - |

## Fix

One-line change in `MatchSAMLLabelRule()`:

```go
// Before:
if samlRole.Compare(maxRole) == -1 {
    rule = labelRule
}

// After:
if rule == nil || samlRole.Compare(maxRole) == -1 {
    rule = labelRule
}
```

This ensures the first valid match always assigns `rule` (even when role is None), while preserving the existing "highest role wins" behavior for all other cases.

**No impact on non-None role scenarios.**

## Environment

- Omni v1.9.0 (self-hosted)
- Keycloak SAML with group/role mappers
- `updateOnEachLogin: true` confirmed set on the SAMLLabelRule

## Workaround

Manually set user role to None via Omni UI or `omnictl`.

ohauer:
Fixes #3178

## Summary

`MatchSAMLLabelRule()` returns `nil` when the highest matched role is `None`, because `samlRole.Compare(maxRole) == -1` is never true when both values are `None` (Compare returns 0). This causes `ensureUser()` to skip reading `updateOnEachLogin` from the matched rule, preventing role downgrades to `None`.

## Fix

One-line change: `if samlRole.Compare(maxRole) == -1` → `if rule == nil || samlRole.Compare(maxRole) == -1`

This ensures the first valid match always assigns `rule` (even when role is None), preserving the existing highest-role-wins behavior for all other cases.

## Testing

- Added `TestMatchSAMLLabelRuleNone`: verifies a rule with `assignRole: None` is returned (not nil)
- Added `TestMatchSAMLLabelRuleNoneWithHigherRulePresent`: verifies None-only → returns None rule; None+Reader → Reader wins
- Existing `TestRoleInSAMLLabelRules` passes unchanged
- Live-tested on self-hosted Omni v1.9.0 with Keycloak SAML

### DIFF
--- a/internal/backend/saml/session.go
+++ b/internal/backend/saml/session.go
@@ -399,7 +399,7 @@ func MatchSAMLLabelRule(samlLabelRules []*auth.SAMLLabelRule, samlLabels map[str
 				continue
 			}
 
-			if samlRole.Compare(maxRole) == -1 {
+			if rule == nil || samlRole.Compare(maxRole) == -1 {
 				rule = labelRule
 			}
 
--- a/internal/backend/saml/session_test.go
+++ b/internal/backend/saml/session_test.go
@@ -270,3 +270,59 @@ func TestRoleCompare(t *testing.T) {
 
 	require.Equal(t, -1, role.Operator.Compare(role.Admin))
 }
+
+func TestMatchSAMLLabelRuleNone(t *testing.T) {
+	logger := zaptest.NewLogger(t)
+
+	noneRoleToInactive := auth.NewSAMLLabelRule("assign-none-to-inactive")
+	noneRoleToInactive.TypedSpec().Value.MatchLabels = []string{"saml.omni.sidero.dev/role/omni-none"}
+	noneRoleToInactive.TypedSpec().Value.AssignRole = string(role.None)
+	noneRoleToInactive.TypedSpec().Value.UpdateOnEachLogin = true
+
+	// None role rule should be returned (not nil)
+	matchedRule := saml.MatchSAMLLabelRule(
+		[]*auth.SAMLLabelRule{noneRoleToInactive},
+		map[string]string{
+			"saml.omni.sidero.dev/role/omni-none": "",
+		}, logger,
+	)
+
+	require.NotNil(t, matchedRule)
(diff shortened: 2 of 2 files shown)


oguzkilcan:
Thanks for the contribution. LGTM 👍

oguzkilcan (approved):

PR #3186, opened 2026-07-24
chore: change virtual ImageFactoryAuth resource to be persistent

Unix4ever:
The idea behind it is that we can then make `ClusterMachineConfig` controller react on this resource version being updated. It will be also able to list the existing factory credentials.


internal/backend/runtime/omni/controllers/omni/maintenance_config_status.go
```diff
@@ -161,36 +166,77 @@ func newMaintenanceConfigStatusControllerHelper(maintenanceClientFactory Mainten
...
 
-	patchWithRegistryAuth := sync.OnceValues(func() (maintenanceBasePatch, error) {
-		authDoc, err := imagefactoryauth.BuildDoc(registries)
+	getPatchWithRegistryAuth := func(ctx context.Context) (maintenanceBasePatch, error) {
+		patchWithRegistryAuthMu.Lock()
+		defer patchWithRegistryAuthMu.Unlock()
+
+		if patchWithRegistryAuth != nil {
+			return *patchWithRegistryAuth, nil
+		}
```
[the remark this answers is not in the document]
Unix4ever: it's updated only on the process restart, so all good

cmd/omni/pkg/app/app.go
```diff
@@ -217,3 +222,37 @@ func Run(ctx context.Context, state *omni.State, cfg *config.Params, logger *zap
...
+		return fmt.Errorf("failed to list ImageFactoryAuth resources: %w", err)
+	}
+
+	for auth := range auths.All() {
+		if _, ok := visited[auth.Metadata().ID()]; !ok {
+			if err := state.Default().Destroy(ctx, auth.Metadata()); err != nil {
+				return fmt.Errorf("failed to delete ImageFactoryAuth resource %s: %w", auth.Metadata().ID(), err)
+			}
+		}
+	}
```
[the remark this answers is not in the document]
Unix4ever: it's meant to be always synced with what's set in the config. So deleting no longer defined auths is the goal

internal/backend/runtime/omni/controllers/omni/cluster_machine_config.go
```diff
@@ -210,6 +220,30 @@ func reconcileClusterMachineConfig(
 		machineJoinConfig,
 	}
 
+	var imageFactories safe.List[*omni.ImageFactoryAuth]
```
oguzkilcan, line 223: Nit: this should say enterpriseImageFactories or authenticatedImageFactories. This way a reader might think we mean all configured image factories.

frontend/src/methods/useImageFactoryAuth.ts
```diff
@@ -19,16 +21,15 @@ export function useImageFactoryAuth() {
...
       resource: {
-        namespace: VirtualNamespace,
+        namespace: DefaultNamespace,
         type: ImageFactoryAuthType,
-        id: ImageFactoryAuthID,
       },
     })
 
-    return computed(() => data.value?.spec)
+    return computed(() => data.value[0]?.spec)
```
Slessi, line 31: shouldnt we get auth by URL?

client/pkg/omnictl/internal/download/download.go
```diff
@@ -684,7 +684,7 @@ func ImageFactoryCredentials(ctx context.Context, st state.State, features *omni
 		return "", "", nil
 	}
 
-	auth, err := safe.ReaderGetByID[*virtual.ImageFactoryAuth](ctx, st, virtual.ImageFactoryAuthID)
+	auth, err := safe.ReaderGetByID[*omni.ImageFactoryAuth](ctx, st, omni.ImageFactoryAuthID)
```
Slessi: this `ImageFactoryAuthID` doesnt point to anything, right? Needs to be URL here, and we should drop the constant entirely

internal/backend/runtime/omni/controllers/omni/cluster_machine_config.go
```diff
@@ -454,27 +493,17 @@ func (helper clusterMachineConfigControllerHelper) generateConfig(clusterMachine
...
+	authPatch, authErr := helper.buildRegistryAuthPatch(imageFactories)
+	if authErr != nil {
+		return nil, authErr
+	}
 
-			if authPatch != "" {
-				patchList = append(patchList, authPatch)
-			}
-		}
+	if authPatch != "" {
```
oguzkilcan, line 505: `quirks.New(initialTalosVersion).SupportsMultidoc()` is still a valid check that should be used here. These auth patches are multidoc.

internal/backend/runtime/omni/controllers/omni/cluster_machine_config.go
Unix4ever: Well, actually I'm not sure about that. I don't think we should name it enterprise. What if we ever get non enterprise images with auth? And `ImageFactoryAuth` is not that different from `authenticatedImageFactories`.

internal/backend/runtime/omni/controllers/omni/cluster_machine_config.go
Unix4ever: It's used when we compose the list of the image factory auth resources.
We need to take auths as the inputs to the UpdateInputVersions. If it doesn't support multidoc, we don't list them, don't add them to the inputs and pass the empty list to the generateConfig method.

internal/backend/runtime/omni/controllers/omni/internal/imagefactoryauth/imagefactoryauth.go
```diff
@@ -39,3 +35,27 @@ func BuildDoc(registries omnicfg.Registries) (*cri.RegistryAuthConfigV1Alpha1, e
 
 	return doc, nil
 }
+
+// BuildDocs returns RegistryAuthConfig docs for every configured image factory that has
+// credentials set: the primary factory and, when configured, the secondary factory.
+// Factories without credentials are skipped. Returns nil if none are configured.
```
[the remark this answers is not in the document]
Unix4ever: I can keep that intact. This will become true later.

frontend/src/methods/useImageFactoryAuth.ts
Unix4ever: For the current state of things before multiple factories support it's going to always be a single item.
It will go away when I add the two factories support back here.

client/pkg/omnictl/internal/download/download.go
Unix4ever: yeah, that's true. Will do that, thanks

oguzkilcan (approved):

PR #3187, opened 2026-07-24
fix: verify maintenance install/upgrade against the live machine

oguzkilcan:
The maintenance install/upgrade path trusted the cached MachineStatus once it saw a reboot, so during the window where the status still lagged behind the node it could re-run the operation. A re-run install got rejected as AlreadyExists and fell back to a spurious upgrade and reboot, and a re-run upgrade ran a second time. The controller now reads the live version before acting and waits for the status to report the installed disk before deciding, the same way the cluster path already does.

### DIFF
--- a/internal/backend/runtime/omni/controllers/omni/machineconfig/status_test.go
+++ b/internal/backend/runtime/omni/controllers/omni/machineconfig/status_test.go
@@ -1098,6 +1098,80 @@ func TestMachineConfigStatusController(t *testing.T) {
 		)
 	})
 
+	// maintenanceUpgradeConvergesOnLiveVersion verifies that after a maintenance upgrade reboots, a cached
+	// MachineStatus still reporting the pre-upgrade version (the post-reboot lag) does not cause a second
+	// upgrade: the controller reads the live version, sees the machine already at the target, and converges
+	// on it. Without the live re-check the stale version mismatch would re-run the upgrade and reboot the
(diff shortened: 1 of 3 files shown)


Unix4ever (approved):

internal/backend/runtime/omni/controllers/omni/machineconfig/status_test.go
```diff
@@ -1098,6 +1098,135 @@ func TestMachineConfigStatusController(t *testing.T) {
...
+				rmock.Mock[*omni.MachineStatusSnapshot](ctx, t, st, options.WithID(id), options.Modify(func(res *omni.MachineStatusSnapshot) error {
+					res.TypedSpec().Value.MachineStatus = &machine.MachineStatusEvent{Stage: machine.MachineStatusEvent_MAINTENANCE}
+					res.TypedSpec().Value.BootId = "boot-after-install"
+
+					return nil
+				}))
+
+				// Negative assertion: with the boot ID advanced but the disk not yet reported, the controller
+				// must wait rather than re-install, so there is no new state to poll on.
+				time.Sleep(time.Second)
```
utkuozdemir: Can we do this without time.sleep? There are other ways of doing it, via an optional channel to the controller or via changing something on the output resource and asserting on that. If not possible I'd drop the negative assertion altogether.

PR #3192, opened 2026-07-28
feat: rework discovery service configuration for Talos 1.14

### LINKED ISSUE #3024: Discovery service config multiple endpoints
1. Omni in the discovery service audit should handle multiple endpoints (remove from all).
2. Omni's embedded discovery service should now be a tri-state: only public discovery, only embedded or embedded + public.

oguzkilcan:
Talos 1.14 replaces the single-endpoint discovery block with multi-document DiscoveryServiceConfig, so a machine can register with more than one discovery service at once. Omni now removes a departing machine's affiliate from every endpoint it used. Each cluster can also keep the public service, use the embedded one, run both, or turn the public service off, though the extra choices are only available for clusters created with Talos 1.14 or newer.

Resolves: #3024


internal/backend/runtime/omni/controllers/omni/discovery_affiliate_delete_task.go
```diff
@@ -79,34 +79,43 @@ func (ctrl *DiscoveryAffiliateDeleteTaskController) Reconcile(ctx context.Contex
...
 		)
 	}
 
 	isTearingDown := res.Metadata().Phase() == resource.PhaseTearingDown
 	affiliateAlreadyDeleted := expiredOnDiscoveryService || isTearingDown
 
 	if !affiliateAlreadyDeleted {
-		endpoint := res.TypedSpec().Value.DiscoveryServiceEndpoint
 		clusterID := res.TypedSpec().Value.ClusterId
 		affiliateID := res.Metadata().ID()
```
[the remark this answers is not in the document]
oguzkilcan: Discovery service affiliates have TTL of 30 mins. Affiliate will eventually get deleted by discovery service.

frontend/src/views/Clusters/EmbeddedDiscoveryServiceCheckbox.vue
```diff
@@ -9,27 +9,65 @@ import TCheckbox from '@/components/Checkbox/TCheckbox.vue'
...
+            Running it alongside the public service adds redundancy: discovery keeps working if one
+            of them becomes unavailable.
+          </p>
+        </div>
+      </template>
+      <TCheckbox
+        v-model="useEmbedded"
+        :disabled="disabled || !embeddedAvailable"
+        label="Use Embedded Discovery Service"
+      />
```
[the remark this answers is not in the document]
oguzkilcan: Fixed.

frontend/src/methods/compat.ts
```diff
@@ -30,6 +30,18 @@ function supportsMaintenanceInstall(v: SemVer): boolean {
...
+const multidocDiscoveryMajor = 1
+const multidocDiscoveryMinor = 14
+
+export function supportsMultidocDiscovery(version?: string): boolean {
+  const v = parse(version)
+  if (!v) return false
+  if (v.major !== multidocDiscoveryMajor) return v.major > multidocDiscoveryMajor
+
+  return v.minor >= multidocDiscoveryMinor
+}
```
Slessi: Can we expose `VersionContract` the same way i exposed `Quirks`? Avoids duplicating the contract. Will be useful for future requirements also

frontend/src/pages/(authenticated)/clusters/create.vue
```diff
@@ -337,10 +348,6 @@ const list = useTemplateRef('list')
           v-model="state.cluster.features.enableWorkloadProxy"
           :disabled="!features?.spec.enable_workload_proxying"
         />
-        <EmbeddedDiscoveryServiceCheckbox
-          v-model="state.cluster.features.useEmbeddedDiscoveryService"
-          :disabled="!isEmbeddedDiscoveryServiceAvailable"
-        />
         <NodeAuditSkipCheckbox v-model="state.cluster.features.enableNodeAuditSkip" />
         <ClusterEtcdBackupCheckbox
```
Slessi, line 352: checkbox order change deliberate? also in OverviewContent

client/pkg/template/internal/models/cluster.go
```diff
@@ -52,13 +52,16 @@ type Cluster struct { //nolint:govet
...
+type Features struct { //nolint:govet // field order matches the serialized YAML, do not reorder for alignment
 	// DiskEncryption enables KMS encryption.
 	DiskEncryption bool `yaml:"diskEncryption,omitempty"`
 	// EnableWorkloadProxy enables workload proxy.
 	EnableWorkloadProxy bool `yaml:"enableWorkloadProxy,omitempty"`
 	// UseEmbeddedDiscoveryService enables the use of embedded discovery service.
 	UseEmbeddedDiscoveryService bool `yaml:"useEmbeddedDiscoveryService,omitempty"`
+	// DisablePublicDiscoveryService turns the public discovery service off. It is on by default (unset or false),
+	// and can only be turned off for clusters created with Talos 1.14+.
+	DisablePublicDiscoveryService *bool `yaml:"disablePublicDiscoveryService,omitempty"`
```
smira, line 64: I think this is a bit confusing when combined with `UseEmbeddedDiscoveryService` ? 

We never had the ability to disable public discovery before... 

I wonder if `useEmbeddedDiscovery` should become more of a tri-state:

* `no` (leave Talos defaults)
* `yes` (backwards compatible, disable Talos default, use embedded)
* `both` (1.14+ only, leave default and inject embedded?)

oguzkilcan: Since this is a bool, we can't make it tri-state in a backwards compatible way. So we need to add a new field anyway to be able to configure all states. I opted to add a new field while keeping existing field working as is. On top of this, even though it is not offered via Omni UI, disabling discovery service is possible via a ConfigPatch. Any option other than leaving Talos default would end in Omni creating a patch, which then might conflict with user owned patch.

I can still rework the new field to accept multiple values to hold multiple states, not offer disabling all discovery services in UI and deprecate `useEmbeddedDiscoveryService` field for clusters created with Talos >= 1.14.

frontend/src/pages/(authenticated)/clusters/create.vue
oguzkilcan: I thought operations on public discovery service should come before embedded config but I can move it to last place if it feels wrong.

frontend/src/methods/compat.ts
oguzkilcan: We can have a virtual resource similar to Quirks but cluster level to expose these kind of support checks. Let me think on it.

client/pkg/template/internal/models/cluster.go
smira: Why can't we? it's YAML, so we need only to keep backwards compat for the text.

We can just add custom type which marshals/unmarshals the way we want?

oguzkilcan: Hmm, that's true. I'll look into it.

frontend/src/pages/(authenticated)/clusters/create.vue
Slessi: Nah just wondered if deliberate or not 👍

frontend/src/methods/compat.ts
Slessi: I just use talos version as the `ID` for the quirks, I would expect same for contract - dont think it needs to be on a cluster, might need it outside of a cluster context

oguzkilcan: Then we'd need to do it for all talos versions. Still possible but more heavy on omni. Nevertheless, I see we need a better approach here, but this PR already got too big. Would it be OK if I tackle it on a different PR?

client/pkg/template/internal/models/cluster.go
oguzkilcan: That option works for Cluster model. However we still have `UseEmbeddedDiscoveryService` under `Cluster.Features` and `ClusterStatus` resources as a bool. These are proto definitions, I can't change them.

I can deprecate this bool field from model and resources and add a new string/enum type for `discoveryServiceMode` allowing values DEFAULT, EMBEDDED, BOTH. What do you think?

frontend/src/methods/compat.ts
Slessi: I had to fix it for the incorrect kubespan patch, so I added it here as part of the same PR, we can merge, rebase, and use it here too? just need to add the supports discovery flag for your PR (or whatever other contract items)

- https://github.com/siderolabs/omni/pull/3196

client/pkg/template/internal/models/cluster.go
Unix4ever: Oh yeah, I was thinking about making it tri-state and forgot to suggest 🫠

client/pkg/template/operations/testdata/export/cluster-resources.yaml
```diff
@@ -35,6 +36,22 @@ spec:
...
+  type: ClusterStatuses.omni.sidero.dev
+  id: export-test
+  version: 1
+  owner: ClusterStatusController
+  phase: running
+  created: 2023-12-07T13:36:21Z
+  updated: 2023-12-08T15:43:10Z
+spec:
+  initialtalosversion: 1.14.0
+---
```
utkuozdemir, line 51: This is confusing - what's its purpose? Is it because the export operation now depend on reading ClusterStatus?

oguzkilcan: Yes, to make the decision to populate `disablePublicDiscoveryService` we need initial talos version.

internal/backend/runtime/omni/controllers/omni/discovery_affiliate_delete_task.go
```diff
@@ -79,34 +79,43 @@ func (ctrl *DiscoveryAffiliateDeleteTaskController) Reconcile(ctx context.Contex
...
 
-		if err = ctrl.discoveryClientCache.AffiliateDelete(ctx, endpoint, clusterID, affiliateID); err != nil {
-			return fmt.Errorf("error deleting affiliate %q/%q from %q: %w", clusterID, affiliateID, endpoint, err)
+		// remove the affiliate from every endpoint it was registered with; AffiliateDelete is
+		// idempotent, so a retry after a partial failure re-deletes already-removed affiliates harmlessly
+		for _, endpoint := range endpoints {
```
utkuozdemir: Two things here:
1. one error would block the further deletions. You should probably use multierror here (or errors.Join etc)
2. You can consider removing them in parallel using an errgroup or something (I'm torn - your call) - but it should still be error tolerant - a failure should not cancel others.

client/pkg/template/internal/models/cluster.go
oguzkilcan: Implemented by adding validation to prevent disabling all discovery service endpoints. We not only support three states; public, embedded and both.

internal/backend/runtime/omni/controllers/omni/discovery_affiliate_delete_task.go
oguzkilcan: Fixed.

utkuozdemir (approved):
LGTM

PR #3193, opened 2026-07-28
feat: add Auditor role for reading the audit log

utkuozdemir:
Reading the audit log required the Admin role, so the audit log exporter had to be granted user management and full administrative access to every managed Kubernetes cluster just to consume events.

The new Auditor role grants read access plus the audit log, and nothing else. It is checked by exact role rather than by rank, so an Operator, which outranks an Auditor, still cannot read the audit log. It can be assigned to users and service accounts, but not by access control policies or SAML label rules, which now reject roles that are not meant to be assigned that way. The settings navigation is no longer administrator only, so an Auditor can reach the audit log view.

Part of siderolabs/omni#2985.

### DIFF
--- a/frontend/src/components/SideBar/TSideBar.vue
+++ b/frontend/src/components/SideBar/TSideBar.vue
@@ -47,7 +47,8 @@ const route = useRoute()
 const { avatar, fullname, identity } = useIdentity()
 
 const { data: featuresConfig } = useFeatures()
-const { canManageBackupStore, canManageUsers, canReadClusters, canReadMachines } = usePermissions()
+const { canManageBackupStore, canManageUsers, canReadAuditLog, canReadClusters, canReadMachines } =
+  usePermissions()
 const isEnterpriseFactory = useIsEnterprise()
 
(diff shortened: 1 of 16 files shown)


frontend/src/components/SideBar/TSideBar.vue
```diff
@@ -268,39 +269,46 @@ const rootItems = computed(() => {
...
-      {
-        name: 'Backups',
-        route: getRoute('Backups', '/settings/backups'),
-        icon: 'rollback',
-      },
-    ]
+  const canManageSettings =
+    canManageUsers.value || (backupStatus.value.configurable && canManageBackupStore.value)
+
+  {
```
Slessi: im pretty sure this brace is a mistake?

Slessi (approved):
frontend LGTM

frontend/src/components/SideBar/TSideBar.vue
utkuozdemir: yep it was :)

PR #3196, opened 2026-07-29
feat: for kubespan quick start use correct patch for the cluster

Slessi:
For the KubeSpan quick start component, check if the cluster supports multi-doc kubespan configs, and if so, suggest that as the patch.

image


frontend/src/views/KubeSpanStatus/KubeSpanStatusQuickStart.stories.ts
```diff
@@ -26,12 +36,52 @@ const clusterHandler = createWatchStreamHandler<ClusterSpec>({
...
+
+      if (id !== talosVersion || type !== VersionContractType || namespace !== VirtualNamespace)
+        return
+
+      return HttpResponse.json({
+        body: JSON.stringify({
+          metadata: {
+            namespace: VirtualNamespace,
+            type: VersionContractType,
+            id: clusterId,
```
oguzkilcan: Shouldn't this be talosVersion?

oguzkilcan (approved):

PR #3200, opened 2026-07-30
chore: bump talos machinery

oguzkilcan:
Talos 1.14 moves several sections out of the v1alpha1 document and drops .machine.install entirely, so the config accessors and the os:admin role stripping move to the multi-doc equivalents. GrubUseUkiCmdline now comes from the version contract, since the config no longer carries it.

Migrate TestClusterMachineConfigSuite from testify.Suite to tesutils.WithRuntime


internal/backend/runtime/omni/controllers/omni/maintenance_config_status_test.go
```diff
@@ -284,8 +284,9 @@ func (suite *MaintenanceConfigStatusControllerSuite) TestMachineConfigPatchPrese
 
 	// a machine-level config patch which carries a partial config document to preserve (TrustedRootsConfig) and a v1alpha1 document which must be stripped in maintenance mode
 	const patchData = `machine:
-  install:
-    disk: /dev/sda
+  features:
+    hostDNS:
+      enabled: true
```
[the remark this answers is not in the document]
oguzkilcan: Fixed

client/pkg/machineconfig/machineconfig.go
```diff
@@ -58,6 +58,11 @@ func Generate(input GenerateInput) (GenerateOutput, error) {
 
 	genOptions = append(genOptions, generate.WithSecretsBundle(input.Secrets))
 
+	// For Talos 1.14+, skip generating unattended config
+	if versionContract.Greater(config.TalosVersion1_13) {
+		genOptions = append(genOptions, generate.WithSkipUnattendedInstallConfig(true))
+	}
```
Slessi, line 64: Wasn't there an actual method on `versionContract` for `UnattendedInstallConfig`? shouldn't need to specify version manually

oguzkilcan: Actually there is. I'll update that. I will wait for e2e tests to pass before committing this change though.

Slessi (approved):

Issue #3199, opened 2026-07-30
Manage the install disk with a first-class resource instead of a config patch


utkuozdemir:
### Background

The user's install disk selection (UI dropdown, cluster template `machine.install.disk`) is stored as a system ConfigPatch that sets `.machine.install.disk`. This worked because Talos read the install disk from the applied config and performed the initial install itself.

Two things changed:

- `.machine.install` is deprecated for removal in Talos. Installs and upgrades are moving to the imperative lifecycle API (Talos 1.13+).
- Omni now drives installs and upgrades of Talos 1.13+ machines through that API and passes the install disk explicitly, so the disk inside the config is no longer what decides where Talos gets installed.

This made the config patch mechanism a dead end: the user's selection lives in config, but the install no longer reads config.

### Interim fix (1.9/1.10)

The controller driving lifecycle installs reads the effective install disk from the merged machine config at the point of use, so the patch-based selection is respected again. This keeps the patch as the source of truth and adds no new API surface. See the TODOs referencing this issue.

### Proper fix candidate

This is a candidate direction, not a settled design. From the engineering discussion (2026-07-30): there are three user profiles to support, (a) users who let Omni pick the disk, (b) users who explicitly pick a disk, (c) users who automate disk selection with a matcher. Since Omni manages the installation, it should support all three.

- Introduce a first-class user-owned resource for the install disk selection (working name `MachineInstallDiskConfig`), replacing the config patch as the source of truth. Frontend and cluster templates write this resource instead of a patch.
- The resource should carry both an explicit disk field and a disk selector in the new CEL expression form (the same expression language Talos uses for volume disk selection). Machinery has full support for these expressions: offline syntax validation, evaluation against disk resources, and YAML/protobuf marshalling. Omni evaluates the selector against the machine's disks at the install decision moment and passes the resolved disk to the lifecycle install API. Resolution problems (no match, multiple matches) are surfaced to the user instead of silently picking something.
- Software RAID as the install target needs no special handling beyond the selector: the user provisions RAID through Talos machine config (machine-level patches applied in maintenance), then points the install disk resource at the resulting device by its stable name or with a selector. The selector flow gives wait-for-ready behavior for free, no install happens until the expression resolves to a device. Managing the disks themselves stays out of scope for this resource, it only answers which disk Talos is installed to.
- Omni keeps full imperative control over installs: the Talos unattended install document will not be used, config generation should skip or drop it, and config patches should be blocked from adding it.
[...]

PR #3203, opened 2026-07-31
feat(frontend): add talos 1.14 patch variants

Slessi:
- Add Talos 1.14 schema
- Add Talos 1.14 patch variants in frontend
  - :warning: `machine.install` patch is still there pending #3199
- Small CSS fix for mobile `h-screen` -> `h-dvh`


frontend/src/pages/(authenticated)/clusters/create.vue
```diff
@@ -88,10 +92,6 @@ const removeLabels = (_: string, ...keys: string[]) => {
   state.value.removeClusterLabels(keys)
 }
 
-const supportsEncryption = computed(
-  () => compare(state.value.cluster.talosVersion ?? '', 'v1.5.0') >= 0,
-)
```
Slessi, line 93: dropped this for cluster create as it shouldnt be possible to have 1.5.0 machines during creation, right?

frontend/src/methods/patches/kubeSpanMultiDoc.yaml
utkuozdemir, line 1: Shouldn't this also contain then a:
```
apiVersion: v1alpha1
kind: DiscoveryServiceConfig
name: default # Name of the discovery service configuration.
endpoint: https://discovery.talos.dev/ # Discovery service endpoint to use.
```
to be equivalent of the one above? (do `talosctl gen config` using talosctl 1.14 for an example)

Slessi: I tested locally (both 1.13 and 1.14) and it accepted the config without complaints - and seemed to work.

As far as I know discovery service is enabled by default, and it was part of the legacy config only for legacy reasons and the validation required it

frontend/src/methods/patches/machineInstallDisk.yaml
utkuozdemir, line 1: Note: this will soon not be a thing anymore. But I guess in this PR it's fine.

Slessi: Yea, I left a note in PR description, I am waiting on #3199 to see what to do about this

Dockerfile
```diff
@@ -278,6 +278,7 @@ ADD https://raw.githubusercontent.com/siderolabs/talos/v1.10.9/pkg/machinery/con
 ADD https://raw.githubusercontent.com/siderolabs/talos/v1.11.6/pkg/machinery/config/schemas/config.schema.json frontend/src/schemas/config_1_11.schema.json
 ADD https://raw.githubusercontent.com/siderolabs/talos/v1.12.8/pkg/machinery/config/schemas/config.schema.json frontend/src/schemas/config_1_12.schema.json
 ADD https://raw.githubusercontent.com/siderolabs/talos/v1.13.3/pkg/machinery/config/schemas/config.schema.json frontend/src/schemas/config_1_13.schema.json
+ADD https://raw.githubusercontent.com/siderolabs/talos/v1.14.0-beta.0/pkg/machinery/config/schemas/config.schema.json frontend/src/schemas/config_1_14.schema.json
```
utkuozdemir, line 281: Wondering if there'll be any difference between this and today's beta.1

Slessi: I can rebase the commit to `.1` if we wanna wait 👍

Makefile
```diff
@@ -362,6 +363,19 @@ lint-fmt: lint-eslint-fmt lint-golangci-lint-client-fmt lint-golangci-lint-fmt
 image-omni:  ## Builds image for omni.
 	@$(MAKE) registry-$@ IMAGE_NAME="omni"
 
+.PHONY: sign-image-omni
```
utkuozdemir, line 365: Since you rekres'ed and these came, you can drop the `sign-images` custom step and its references `.kres.yaml` - that custom step is not needed anymore.

frontend/src/methods/patches/kubeSpanMultiDoc.yaml
oguzkilcan: I'd say no. Discovery will always be enabled unless user went and disabled with a config patch explicitly. Above patch would conflict with multi discovery endpoint changes if only embedded discovery service is enabled.

utkuozdemir: Alright then, you're more involved with the discovery service lately.

Makefile
Slessi: Done

utkuozdemir (approved):

Issue #3104, opened 2026-07-09
Block submission of `UnattendedInstallConfig`


shanduur:
Submission of `UnattendedInstallConfig` may lead to race between `Install Lifecycle API` and the `UnattendedInstallController`. It makes no sense to use the `UnattendedInstallConfig` with Omni, as Omni is managing the installation and upgrades of Talos.

PR #3204, opened 2026-07-31
feat: validate multi-doc config patches

### LINKED ISSUE #3025: machine config deny list should be updated
Talos refactor Kubernetes configuration into multi-doc, so a new list of denied list to cover new documents.

oguzkilcan:
Talos 1.14 moves cluster identity, CAs and encryption secrets into documents of their own. Those documents are now rejected outright, and the os:admin restriction follows Talos API access into its own document.

Resolves: #3025, #3104

### DIFF
--- a/client/pkg/omni/resources/omni/config_patch.go
+++ b/client/pkg/omni/resources/omni/config_patch.go
@@ -18,6 +18,9 @@ import (
 	"github.com/hashicorp/go-multierror"
 	"github.com/siderolabs/gen/pair"
 	"github.com/siderolabs/talos/pkg/machinery/config/configloader"
+	taloscluster "github.com/siderolabs/talos/pkg/machinery/config/types/cluster"
+	talosk8s "github.com/siderolabs/talos/pkg/machinery/config/types/k8s"
+	talosruntime "github.com/siderolabs/talos/pkg/machinery/config/types/runtime"
 	talosrole "github.com/siderolabs/talos/pkg/machinery/role"
 	"go.yaml.in/yaml/v4"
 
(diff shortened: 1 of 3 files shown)


client/pkg/omni/resources/omni/config_patch_test.go
```diff
@@ -81,6 +81,76 @@ machine:
...
+installer:
+  image: example.com/installer:v1.14.0
+`),
+			expectedError: "1 error occurred:\n\t* the \"UnattendedInstallConfig\" document is not allowed in the config patch\n\n",
+		},
+		{
+			name: "cluster identity document",
+			config: strings.TrimSpace(`
+apiVersion: v1alpha1
+kind: DiscoveryIdentityConfig
```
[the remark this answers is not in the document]
oguzkilcan: Added a new test to cover all forbidden documents.

client/pkg/omni/resources/omni/config_patch.go
```diff
@@ -41,12 +44,31 @@ var forbiddenFields = []string{
 	"machine.acceptedCAs",
 }
 
-var forbiddenSliceElements = map[string]map[any]struct{}{
+var forbiddenDocuments = map[string]struct{}{
+	talosruntime.UnattendedInstallConfigKind: {}, // not supported in Omni
+	taloscluster.DiscoveryIdentityKind:       {}, // cluster.id, cluster.secret
+	talosk8s.KubeClusterConfig:               {}, // cluster.clusterName, cluster.controlPlane.endpoint
+	talosk8s.KubeEtcdEncryptionConfig:        {}, // cluster.secretboxEncryptionSecret, cluster.aescbcEncryptionSecret
```
smira, line 51: should be fine for now (to match previous), but I think in general we should allow people to modify this freely to rotate etcd encryption keys

oguzkilcan: How would such a config patch look like? Omni currently owns etcd encryption keys. Shouldn't Omni be responsible for rotating them?

client/pkg/omni/resources/omni/config_patch.go
```diff
@@ -41,12 +44,31 @@ var forbiddenFields = []string{
...
+	talosk8s.KubeAPIServerCAConfig:           {}, // the Kubernetes CA, private key included
+	talosk8s.KubeAggregatorCAConfig:          {}, // the aggregator CA, private key included
```
smira, line 53: https://docs.siderolabs.com/talos/v1.14/reference/configuration/kubernetes/kubeserviceaccountconfig if we previously protected service account (?)

smira (approved):

client/pkg/omni/resources/omni/config_patch.go
smira: First of all, this is all user's domain imho - it's how data is encrypted in etcd by the kube-apiserver. So an admin should be able to see/modify this.

Second, rotation here is kind of user-assisted rotation - you can only add new encryption keys, but the moment to remove old ones depends on api-server activity.

Also there are some other settings there which might be user-managed - e.g. what Kubernetes resources to encrypt in etcd - today the default is only Secrets

client/pkg/omni/resources/omni/config_patch.go
oguzkilcan: No, we weren't. Should we though?

PR #3205, opened 2026-07-31
fix(frontend): use cluster config version for version contract in scale

Slessi:
When scaling clusters, use ClusterConfigVersion to get the correct VersionContract for generating patches.

### DIFF
--- a/frontend/src/pages/(authenticated)/clusters/[cluster]/scale.vue
+++ b/frontend/src/pages/(authenticated)/clusters/[cluster]/scale.vue
@@ -12,12 +12,14 @@ import { useRoute, useRouter } from 'vue-router'
 import { Runtime } from '@/api/common/omni.pb'
 import type { Resource } from '@/api/grpc'
 import type {
+  ClusterConfigVersionSpec,
   ClusterSpec,
   MachineConfigGenOptionsSpec,
   MachineStatusSpec,
 } from '@/api/omni/specs/omni.pb'
 import type { VersionContractSpec } from '@/api/omni/specs/virtual.pb'
 import {
+  ClusterConfigVersionType,
   DefaultNamespace,
   LabelNoManualAllocation,
   MachineConfigGenOptionsType,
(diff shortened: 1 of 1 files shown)


frontend/src/components/CodeEditor/CodeEditor.vue
```diff
@@ -293,5 +313,11 @@ watch(editorOptions, (options) => instanceRef?.updateOptions(options))
 </script>
 
 <template>
-  <div ref="editor"></div>
+  <div class="flex flex-col gap-1 bg-naturals-n0">
+    <p class="shrink-0 self-end pt-1 pr-1 text-right text-xs/none opacity-50">
+      Config version {{ majorMinorVersion(schemaVersion.format()) }}
```
smira: We could link it to docs.siderolalbs.com config reference for that version?

frontend/src/views/Clusters/Management/MachineSetConfig.vue
```diff
@@ -207,7 +207,7 @@ const labelId = useId()
       :id="`Machine Set ${modelValue.name}`"
       v-model:open="configPatchEditModalOpen"
       :config="patches[PatchID.Default]?.data ?? ''"
-      :talos-version="talosVersion"
+      :talos-version="clusterConfigVersion"
```
smira: my only concern here - if we implement this _before_ we allow config version migration, is it good or bad UX?

Slessi: Good if my understanding is correct:

> Config is validated against config version of the cluster.
If I create a `1.12` cluster, and upgrade to `1.13`, I can still only use configs that are supported by `1.12`.
If I add a block that only works on `1.13` it will be an error

Is that the case?

frontend/src/components/CodeEditor/CodeEditor.vue
Slessi: Yea, this is actually done in some places already, but now that you mention it, I see it isn't everywhere. I will add that.

e.g. in cluster create, the modal patch flow has it:

image

frontend/src/views/Clusters/Management/MachineSetConfig.vue
smira: Nope, that is not true. Talos 1.13 is backwards compatible with 1.12 (or any earlier) config, but it also supports new config.

The issue was a bit different. I had Talos 1.12 cluster, and Talos 1.12 config, where I used in my patch some document `XYZConfig`.

I update to Talos 1.13. Talos 1.13 reads `XYZConfig` and supports it, but let's imagine Talos 1.13 deprecated `XYZConfig` and provided a new document `BarConfig` which is better. Talos 1.13 config schema drops `XYZConfig` (as it's deprecated), so if I pull config schema for 1.13, Omni's editor will highlight `XYZConfig` in red.

But with this change alone, it means that any new Talos 1.13 config would be highlighted in red. 

So the solution was supposed to be two steps: 

1. This PR change (lock schema in the editor to cluster's config version).
2. Plus an option/button to bump cluster's config version to the Talos version (after upgrading to 1.13, modernize machine config to 1.13).

These combined should provide better UX: the base machine config is lifted to the new config contract, and editor's validation is consistent with my patches. 

If I decide to never use new 1.13 machine configs and stay on 1.12 => I can do so. I want use new features => I modernize, fix issues in the process, and move on to the brave new world.

Slessi: Yea, that changes things 🤔 I'll have a think

oguzkilcan (approved):

frontend/src/views/Clusters/Management/MachineSetConfig.vue
Slessi: Going to skip this for now

PR #3210, opened 2026-07-31
fix: forbid cluster CA and service account key in config patches

oguzkilcan:
Add the Kubernetes CA and the service account signing key to forbiddenFields list. Also create a new documentForbiddenFields list to forbid fields coming from multi-doc documents.

### DIFF
--- a/client/pkg/omni/resources/omni/config_patch.go
+++ b/client/pkg/omni/resources/omni/config_patch.go
@@ -36,7 +36,9 @@ var forbiddenFields = []string{
 	"cluster.secret",
 	"cluster.aescbcEncryptionSecret",
 	"cluster.secretboxEncryptionSecret",
+	"cluster.ca",
 	"cluster.acceptedCAs",
+	"cluster.serviceAccount",
 	"machine.token",
 	"machine.ca",
 	"machine.type",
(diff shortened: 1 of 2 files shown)


client/pkg/omni/resources/omni/config_patch.go
```diff
@@ -266,6 +283,17 @@ func sanitizeConfigPatch(config map[string]any) map[string]any {
 	return config
 }
 
+// isEmptyDocument reports whether the document carries nothing beyond its meta header.
+func isEmptyDocument(document map[string]any) bool {
+	for key := range document {
+		if key != "apiVersion" && key != "kind" {
```
smira, line 289: I wonder if `name` should be here as well

smira (approved):

client/pkg/omni/resources/omni/config_patch.go
oguzkilcan: Good question. Do we have any document that only has `apiVersion`, `kind` and `name` fields?

oguzkilcan: I checked. There are such documents, e.g. `KernelModuleConfig`. If we add `name` to that condition, a config patch with such documents would be considered empty. Checking kind and apiVersion is sufficient.

PR #3213, opened 2026-08-03
fix(frontend): also check version contract in cluster scale

Slessi:
Also check VersionContract in cluster scaling view, not just cluster create.

### DIFF
--- a/frontend/src/pages/(authenticated)/clusters/[cluster]/scale.vue
+++ b/frontend/src/pages/(authenticated)/clusters/[cluster]/scale.vue
@@ -16,6 +16,7 @@ import type {
   MachineConfigGenOptionsSpec,
   MachineStatusSpec,
 } from '@/api/omni/specs/omni.pb'
+import type { VersionContractSpec } from '@/api/omni/specs/virtual.pb'
 import {
   DefaultNamespace,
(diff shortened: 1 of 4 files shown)


frontend/src/pages/(authenticated)/clusters/[cluster]/scale.vue
```diff
@@ -137,12 +139,33 @@ const {
...
+  data: versionContract,
+  loading: versionContractLoading,
+  error: versionContractErr,
+} = useResourceGet<VersionContractSpec>(() => ({
+  skip: !state.value.cluster.talosVersion,
+  runtime: Runtime.Omni,
+  resource: {
+    namespace: VirtualNamespace,
+    type: VersionContractType,
+    id: state.value.cluster.talosVersion!,
```
Comment (oguzkilcan) [inline frontend/src/pages/(authenticated)/clusters/[cluster]/scale.vue:152]: Initial Talos version should either come from ClusterConfigVersion or ClusterStatus.

Comment (Slessi) [reply frontend/src/pages/(authenticated)/clusters/[cluster]/scale.vue:152]: Yea, for this PR it is just to fix the bug in cluster scaling, the correct version will be fixed in #3205

oguzkilcan (approved):

PR #3218, opened 2026-08-03
refactor: drop resource cache staleness workarounds

utkuozdemir:
All controller reads now go through the controller runtime resource cache. The reads that bypassed the cache were workarounds for the cache serving mutually inconsistent data across resource types, which is now fixed at the state layer: all watches are served from a single shared watch stream, so cached reads are consistent with each other.

Unit tests now run with no resources marked as cached, which effectively disables the cache: the in-memory state they use does not provide the cross-resource event ordering the cache needs, and the cache is exercised by the integration tests against etcd. This can be reverted once the in-memory state provides the same ordering guarantees.

Requires the shared watch stream support in the etcd state backend.


utkuozdemir:
Not stale

PR #3228, opened 2026-08-06
fix: log out of Omni and the IdP when /logout is opened directly

oguzkilcan:
Only the frontend can revoke the public key that authenticates a browser against Omni, so for SAML and OIDC, where the backend owned /logout outright, opening the URL ended the session at the identity provider and left the user signed in to Omni. A bare /logout now serves the frontend, and the identity provider logout runs once the frontend comes back with flow=frontend.

### DIFF
--- a/frontend/src/pages/logout.vue
+++ b/frontend/src/pages/logout.vue
@@ -9,7 +9,6 @@ import { onBeforeMount } from 'vue'
 
 import PageContainer from '@/components/PageContainer/PageContainer.vue'
 import TSpinner from '@/components/Spinner/TSpinner.vue'
-import { AuthType, authType } from '@/methods'
 import { useLogout } from '@/methods/auth'
 
 definePage({
@@ -19,10 +18,6 @@ definePage({
 const logout = useLogout()
 
 onBeforeMount(() => {
-  // For SAML and OIDC the backend registers its own /logout handler
-  // Hard reload to hit backend forcibly if we are here accidentally
-  if (authType.value !== AuthType.Auth0) window.location.reload()
-
   logout()
 })
 </script>
--- a/internal/backend/server.go
+++ b/internal/backend/server.go
@@ -889,13 +889,7 @@ func makeMux(
 
 	muxHandle("/", frontendHandler, "static")
 
-	if err := registerAuthHandlers(
-		mux,
-		samlHandler,
-		oidcProvider,
-		logger,
-		cfg,
-	); err != nil {
+	if err := registerAuthHandlers(mux, samlHandler, oidcProvider, logger, cfg); err != nil {
 		return nil, err
 	}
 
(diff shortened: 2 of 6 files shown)


frontend/src/pages/logout.vue
```diff
@@ -19,10 +18,9 @@ definePage({
 const logout = useLogout()
 
 onBeforeMount(() => {
-  // For SAML and OIDC the backend registers its own /logout handler
-  // Hard reload to hit backend forcibly if we are here accidentally
-  if (authType.value !== AuthType.Auth0) window.location.reload()
-
+  // Every provider lands here first, whether the user typed the URL or used the
+  // logout control. Only the browser can revoke the public key, and that has to
+  // happen before the backend hands off to the identity provider.
```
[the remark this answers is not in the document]
oguzkilcan: Fixed.

internal/backend/server.go
```diff
@@ -981,14 +989,34 @@ func registerAuthHandlers(mux *http.ServeMux, samlHandler *samlsp.Middleware, oi
...
 
+// routeLogoutThroughFrontend serves the frontend for a bare /logout and runs the
+// identity provider logout only once the frontend asks for it.
+//
+// A browser authenticates against Omni with a registered public key, so only the
+// frontend can end an Omni session: it revokes the key, clears its storage,
+// and comes back here with flow=frontend. Sending a bare /logout straight
+// to the identity provider ends the session upstream but leaves the browser still
+// signed in to Omni.
+func routeLogoutThroughFrontend(idpLogout http.HandlerFunc, frontendHandler http.Handler) http.HandlerFunc {
```
Slessi: it feels still like this is unnecessarily complicated, to have `/logout` potentially served by frontend/backend depending on a query string

we can drop this method entirely if we change the backend path to e.g. `/idp-logout`, and change frontend redirect to `/idp-logout`

then its simple:
- `/logout` -> always frontend, always full logout
- `/idp-logout` -> backend only oidc/saml session clearing

Unix4ever: I think changing URI might break backward compatibility with the existing OIDC/SAML configs.

Slessi: The URI will still be `/logout` - it will just always pass through frontend. The extra `/idp-logout` (or other naming) URI is to avoid the code branching, and will be used by frontend in `/logout` (user only needs to know `/logout`.)

If that _also_ is a problem, we might have a problem with this PR no? Because it is going to send `/logout` to frontend, and only send `/logout?flow=frontend` to backend (which is confusing btw)

oguzkilcan: I think having a different path to log out from IdPs makes sense. Let me see how that actually work in real life and has any implications for existing configurations.

frontend/src/pages/logout.vue
```diff
@@ -19,10 +18,9 @@ definePage({
 const logout = useLogout()
 
 onBeforeMount(() => {
-  // For SAML and OIDC the backend registers its own /logout handler
-  // Hard reload to hit backend forcibly if we are here accidentally
-  if (authType.value !== AuthType.Auth0) window.location.reload()
-
+  // /logout is served by the frontend for every auth type, so opening it directly (typed
+  // URL, bookmark) ends the session the same way the user menu does. Only the browser can
+  // revoke the public key, and that has to happen before the identity provider logout.
```
Slessi: i think we can drop this comment

Slessi (approved):

utkuozdemir (approved):

PR #3230, opened 2026-08-06
test: add e2e-saml test class backed by a local Keycloak IdP

oguzkilcan:
Covers the SAML login flow, the identity and full name mapped out of the assertion attributes, and Single Logout.


Slessi (approved):
<3

PR #3231, opened 2026-08-06
feat: manage the install disk with a first-class resource

### LINKED ISSUE #3199: Manage the install disk with a first-class resource instead of a config patch

utkuozdemir:
The high-level diagram:

```mermaid
flowchart LR
    subgraph selection["Disk selection (user-owned)"]
        direction TB
        User([User<br/>UI / cluster template / omnictl])
        MIDC["MachineInstallDiskConfig 🆕<br/>disk | disk_selector"]
    end

    subgraph resolution["Disk resolution"]
        direction TB
        MS["MachineStatus<br/>(the machine's disks)"]
        Resolver{{"MachineInstallDiskStatusController 🆕"}}
        MIDS["MachineInstallDiskStatus 🆕<br/>resolved disk, candidate disks,<br/>message, selection_hash"]
    end

    subgraph generation["Config generation"]
        direction TB
        Generator{{"ClusterMachineConfigController"}}
        CMC["ClusterMachineConfig<br/>config data (.machine.install.disk)<br/>install_disk 🆕"]
    end

    subgraph install["Install"]
        direction TB
        Decision{{"machineconfig.StatusController"}}
        Talos["Talos machine"]
    end

    User -- "writes the selected disk<br/>or a disk selector" --> MIDC
    MIDS -. "dropdown: candidate disks,<br/>resolved disk, message" .-> User

    MS --> Resolver
    MIDC --> Resolver
    Resolver -- "writes the resolved disk +<br/>the hash of the disk selection" --> MIDS

    MIDS --> Generator
    MIDC -. "reads the disk selection fresh,<br/>compares its hash vs the one on the status,<br/>skips if stale" .-> Generator
    Generator -- "generates the config data,<br/>records the resolved disk<br/>as install_disk" --> CMC

    CMC -- "primary input" --> Decision
    Decision -- "Talos ≥ 1.13: install API call<br/>with the recorded disk" --> Talos
    Decision -- "Talos < 1.13: config apply<br/>(the disk is inside the config data)" --> Talos
```

---

The install disk selection moves from a generated config patch to a dedicated resource pair: a user-owned selection carrying either a static disk or a CEL disk selector, and the install disk resolution computed for every machine from its disks. The UI dropdowns, cluster templates and omnictl work against these resources, and the previously generated install disk patches are migrated automatically. The dropdowns show what the automatic selection currently resolves to, and tell an explicit disk, a disk selector and the automatic selection apart.

The selector uses the same disk expression language Talos does, so disks can be targeted by stable properties such as the serial number, and an explicit selector can install onto disks built on top of other disks, such as RAID arrays. The automatic selection never picks such disks or their member disks.

The config generation checks that the install disk resolution is not stale before generating a config, and records the resolved disk on the config itself. The install always uses the disk recorded on the config version it works on, so the disk passed to the installer and the config data cannot disagree. Config patches can no longer change the install disk: new patches carrying the field are rejected, and the existing ones are overridden at config generation time.

The backend API version is bumped, because older omnictl versions would sync cluster templates through the removed patch mechanism and silently drop selections on export. The version check works in both directions, so omnictl and Omni must be upgraded together.

Closes #3199.


frontend/src/pages/(authenticated)/clusters/[cluster]/scale.vue
```diff
@@ -201,28 +191,9 @@ const {
...
-    ...m,
-    spec: {
-      ...m.spec,
-      ...machineConfigGenOptionsMap.value[m.metadata.id!]?.spec,
-    },
-  })),
-)
+const loading = computed(() => machineStatusesLoading.value || versionContractLoading.value)
+const err = computed(() => machineStatusesErr.value || versionContractErr.value)
+const data = computed(() => machineStatuses.value)
```
Comment (Slessi) [inline frontend/src/pages/(authenticated)/clusters/[cluster]/scale.vue]: `data` redundant now, just rename `machineStatuses` to `data` (or vice versa)

oguzkilcan (approved):

Slessi (approved):

PR #3245, opened 2026-08-10
fix(frontend): be more defensive about signup errors

### LINKED ISSUE #3243: Upgrading Omni self-hosted from 1.9.3 to 1.10.0 causes authentication issues

### Current Behavior

I've been running Omni on-prem for probably a few years now, with no issues. I initially set it up using your Auth0 guidance and am using Google Auth via Auth0.  I upgraded my SideroLabs Omni instance to from 1.9.3 to 1.10.0 over the weekend. The container starts, but when I browse to my Omni instance, it continually flashes back and forth between:
 
`https://omni.laskonet.com/` 
and 
`https://omni.laskonet.com/?error=invalid_request&error_description=connection%20not%20found%20for%20the%20client&state=<redacted>%3D%3D`

The container logs had similar error messages. Looking at Auth0, there were numerous errors like this:

```
{
  "date": "2026-08-10T15:55:07.557Z",
  "type": "f",
  "description": "connection not found for the client",
  "connection_id": "",
  "client_id": "<redacted>",
  "client_name": "Omni",
  "ip": "142.181.117.158",
  "user_agent": "Firefox 153.0.0 / Windows 10.0.0",
  "details": {
    "body": {},
    "qs": {
      "client_id": "<redacted>",
      "scope": "openid profile email",
      "redirect_uri": "https://omni.laskonet.com/",
      "max_age": "120",
      "screen_hint": "signup",
      "response_type": "code",
      "response_mode": "query",
      "state": "********************************************************dw==",
      "nonce": "<redacted>+<redacted>==",
      "code_challenge": "<redacted>",
      "code_challenge_method": "S256",
      "auth0Client": "<redacted>"
    },
    "error": {
      "message": "connection not found for the client",
      "oauthError": "connection not found for the client",
      "type": "invalid_request",
      "uri": null
    },
    "session_id": "<redacted>",
    "riskAssessment": null,
    "transaction_id": "IrxgWfyQkgjRNrjDB5dWB-dY3mPg-ToE"
  },
  "hostname": "ucdialplans.us.auth0.com",
  "audience": "https://ucdialplans.us.auth0.com/userinfo",
  "scope": "openid profile email",
  "auth0_client": {
    "name": "auth0-vue",
    "version": "2.3.1"
  },
  "location_info": {
    "latitude": 43.5204,
    "longitude": -80.2245,
    "country_code": "CA",
    "country_name": "Canada",
    "city_name": "Guelph",
    "subdivision_code": "ON",
    "subdivision_name": "Ontario",
    "continent_code": "NA",
    "time_zone": "America/Toronto",
    "country_code3": "CAN"
  },
  "$event_schema": {
    "version": "1.0.0"
  },
  "environment_name": "prod-us-4",
  "log_id": "90020260810155507593911000000000000001223372121781584157",
  "tenant_name": "ucdialplans",
  "_id": "90020260810155507593911000000000000001223372121781584157",
  "isMobile": false,
  "id": "90020260810155507593911000000000000001223372121781584157"
}
```
Based on a conversation with Claude, it suggested the issue was caused by Omni sending `screen_hint=signup` in the auth request message body and to enable `Username-Password-Authentication` in the `Connections` tab of my Auth0 Omni application. After doing this, I was able to successfully login to Omni via Auth0, but now I have the option (which I don't want) to login with username/password. 

### Expected Behavior

I expect to not have to enable `Username-Password-Authentication`  in order for Auth0 to function properly with Omni on v1.10.0

### Steps To Reproduce

1. Upgrade Omni to 1.10.0 on an environment where 1.9.3 was working fine via Auth0 and Google Auth
2. Browse to your Omni instance 
3. Note that it continually flashes between 2 URLs. You never see the authentication screen nor the Omni dashboard
4. In Auth0 Application settings, navigate to Connections and enable `Username-Password-Authentication`
5. Note that Omni now shows the authentication screen and allows you to login as before

Chrome, Firefox

### Anything else?

_No response_

Slessi:
Be more defensive about handling auth0 signup errors, to ensure we don't softlock any unexpected auth0 implementations

Closes #3243

### DIFF
--- a/frontend/src/main.ts
+++ b/frontend/src/main.ts
@@ -79,9 +79,9 @@ async function setupApp() {
     authType.value = AuthType.Auth0
 
     const searchParams = new URLSearchParams(window.location.search)
-    const signupDisabled =
-      searchParams.get('error') === 'invalid_request' &&
-      searchParams.get('error_description') === 'signup is disabled'
+
+    // If there is any kind of error response from auth0, defensively assume signup isn't supported
+    const signupDisabled = typeof searchParams.get('error') === 'string'
 
     const requestSignup = !authConfig.spec.has_initial_user && !signupDisabled
 


frontend/src/main.ts
```diff
@@ -79,9 +79,9 @@ async function setupApp() {
     authType.value = AuthType.Auth0
 
     const searchParams = new URLSearchParams(window.location.search)
-    const signupDisabled =
-      searchParams.get('error') === 'invalid_request' &&
-      searchParams.get('error_description') === 'signup is disabled'
+
+    // If there is any kind of error response from auth0, defensively assume signup isn't supported
+    const signupDisabled = typeof searchParams.get('error') === 'string'
```
smira, line 84: I might be misreading, but isn't it too broad to assume that if someone lands me on omni with `?error=foo`, it means that signup is disabled?

Slessi: You are correct. For context the only time we see `?error=` (unless hand-written) is when auth0 reports a problem back after an auth redirect

Initially I tested locally by disabling signups and I got that specific error and I thought that covered it, but it seems as per #3243 you can get different errors if other requirements are disabled e.g. username/password authentication in general.

So I kinda felt since this is a very _minor_ user convenience at best as it just saves 1 click on the login flow, and only for the very first Omni user, that better to just defensively just skip trying `signup` in case it causes whatever problem, since it is hard to know every case that can cause this. Whereas conversely the fallout could be soft-locking a user from their Omni instance which is pretty bad.

smira: got you, I guess it's fine, but it also feels fragile a bit

smira (approved):

frontend/src/main.ts
Slessi: ye, its very much "best effort"

PR #3246, opened 2026-08-11
fix: bound the machine config diff by bytes, not just lines

oguzkilcan:
The diff operation could produce a diff larger than etcd's write limit, leading to failures when storing it on etcd.

Compute now also compares byte sizes and returns the short summary when either the combined inputs or the resulting diff exceed 512 KiB.

### DIFF
--- a/client/pkg/diff/diff.go
+++ b/client/pkg/diff/diff.go
@@ -18,9 +18,12 @@ import (
 // human can meaningfully review it and the algorithmic cost becomes prohibitive.
 const MaxLines = 75_000
 
+// MaxBytes is the maximum size of the diff returned.
+const MaxBytes = 512 * 1024
+
 // Compute returns a unified diff (without the --- / +++ header) between two
-// byte slices. For inputs whose combined line count exceeds MaxLines it returns
-// a short summary instead.
+// byte slices. Inputs past MaxLines, or a diff past MaxBytes, yield a short
+// summary instead.
 func Compute(previousData, newData []byte) (string, error) {
 	if bytes.Equal(previousData, newData) {
 		return "", nil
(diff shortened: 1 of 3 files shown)


client/pkg/diff/diff.go
```diff
@@ -50,10 +54,20 @@ func Compute(previousData, newData []byte) (string, error) {
 		return "", err
 	}
 
+	// Two inputs that each fit within the bounds above can still produce a diff bigger
+	// than both, so the result needs its own check.
+	if len(result) > MaxBytes {
+		return tooLargeSummary(prevLines, newLines), nil
+	}
```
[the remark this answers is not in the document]
oguzkilcan: We would be saving like 10 bytes here. But I'll fix it for consistency.

client/pkg/diff/diff.go
```diff
@@ -29,8 +33,8 @@ func Compute(previousData, newData []byte) (string, error) {
 	prevLines := bytes.Count(previousData, []byte("\n"))
 	newLines := bytes.Count(newData, []byte("\n"))
 
-	if prevLines+newLines > MaxLines {
-		return fmt.Sprintf("@@ -%d,%d +%d,%d @@ diff too large to display\n", 1, prevLines, 1, newLines), nil
+	if prevLines+newLines > MaxLines || len(previousData)+len(newData) > MaxBytes {
```
smira:
```suggestion
	if prevLines+newLines > MaxLines {
```

I would not put the check here - even two large inputs can produce a tiny diff. The diff algorithm itself only cares about lines, so the size doesn't matter at all.

The check on the result diff size makes sense imho on line 62-66

oguzkilcan: I guess I was a bit too aggressive with defensive code there.

smira (approved):

Issue #3166, opened 2026-07-20
Serve EIF security artifacts through Omni (API + UI) instead of client-side factory fetches


sterlingkoch:
### Problem Description

Enterprise customers can only reach their per-schematic security artifacts (SPDX 2.3 SBOM, VEX JSON, scan reports) through the Omni UI. `v1.9.0` put them in the install-media wizard; `v1.10.0` added a security page for what a cluster is *actually running*, with the upgrade paths that clear each vulnerability (#3048). Both fetch from the factory client-side, so the browser needs factory credentials — today they are embedded in the URL's userinfo.

Two problems follow:

1. **No programmatic path.** A fleet-wide pull ("give me the SBOM + VEX for every installed image in this cluster") means reconstructing schematic/version/arch combinations by hand and hitting the factory directly.
2. **Factory credentials reach the browser.** Every client-side artifact fetch needs them, so they end up in URLs rendered on screen.

Part of TEL September readiness (M2 in the [Build Spec](https://app.notion.com/p/siderolabs/Build-Spec-TEL-September-2026-Launch-39eb1211badf8102b317ee1635efcae9)). The definition of done includes: script a fleet-wide artifact pull — for a cluster, enumerate installed (schematic, version) pairs and retrieve the matching SBOM + VEX via the Omni API.

### Solution

Omni fetches security artifacts server-side using the factory credentials it already holds, and serves them to the frontend. No factory credentials in the browser.

- Machine- and cluster-level artifact access keyed to installed state (schematic, version, arch), in the UI and via the Omni API.
- API: enumerate a cluster's installed (schematic, version) pairs and resolve each to its SBOM / VEX / scan report, so a fleet-wide pull is a script rather than a spreadsheet.
- Omni serves the artifact itself rather than handing the browser a credentialed or tokened link to the factory.

### Alternative Solutions

**Redirect the browser to the factory with a signed download token (siderolabs/image-factory#509) — not viable.** Download tokens are accepted only on `GET`/`HEAD` `/image/`; the factory middleware rejects them on every other path, deliberately ("Tokens are only accepted on GET/HEAD /image/ to prevent use on other endpoints"). Artifact paths would 401.

It is also the wrong fit. Download tokens exist so a URL can leave the browser — copy it, carry it through a bastion, curl it into a restricted network. Security artifacts are read in the UI or pulled by a script against the Omni API, so they don't need a portable URL; they need Omni to fetch on the user's behalf.

### Notes

[...]

Issue #3138, opened 2026-07-15
Support time-based signed download URLs


sterlingkoch:
### Problem Description

Enterprise download URLs currently embed the Image Factory username and password in plaintext. Per the July 15 planning meeting, these are replaced with time-based signed URLs. This ticket covers the Omni side; the factory mechanism shipped in siderolabs/image-factory#509 (merged July 27), which resolved the original open question of where signing happens — the factory mints.

How it works: a token is requested from the factory's authenticated `POST /download-token` endpoint and appended to download URLs as `?token=…`. Tokens are identity-scoped JWTs, accepted only on `GET`/`HEAD` `/image/` paths, with validity returned as `expires_in`. Part of TEL September readiness.

sterlingkoch:
Design guidance...

**Where this lives** 
The schematic output screen of the install-media wizard (the screen listing ISO / disk images / installer image / extra assets). Signed URLs replace the embedded-credential URLs in place — same layout, same copy buttons.

**One validity message for the whole screen, not per link.** 
All links on this screen are minted together at render, so they share one expiry. One line above the asset list:

> "These links expire in 4:32..."

**Expired state: screen-level, not per-link.** When expiry passes while the dialog is open, flip the whole screen with a client-side timer icon rather than letting someone copy a dead link:

Message: "These download links have expired."
One action: "Generate fresh links" button — re-requests tokens and re-renders the same screen.
Copy buttons disabled in the expired state.

Something like this:

```
Schematic ready   412ef446…
──────────────────────────────────────────────
ⓘ These links expire in 4:32...

ISO          https://factory.siderolabs.com/image/…?token=…   [copy]
Disk (raw)   https://factory.siderolabs.com/image/…?token=…   [copy]
…
──────────────────────────────────────────────
⚠ These download links have expired.
  [ Generate fresh links ]        (copy buttons disabled)

```

PXE: the PXE row disappears for enterprise per #3137, so it needs no expired-state treatment.

Happy to do a Figma pass on the two states if that would help, but I'm guessing the above is enough to go on.

sterlingkoch:
- Default token TTL is 5 minutes (factory-configurable, `DownloadTokenTTL`). Whether that clears the bastion use case above is an open question — the UI should drive everything off `expires_in` and assume nothing about the duration.
[...]

PR #3248, opened 2026-08-11
fix(frontend): add missing credentials to security pages

Slessi:
- Consolidate credential URL logic into useResolvedFactory.
- Add missing factory auth credentials to enterprise security pages in the frontend.
- Remove credentials from talosctl downloads as they are not required there.
- Add messages to inform user about orphaned presets rather than just disabling buttons in some cases, and rendering undefined URLs in others.


sterlingkoch:
For the record since this will outlive the DM it was discussed in: this consolidates the existing embedded-credential URL pattern and extends it to the security pages as an interim fix for the 401s. The pattern itself is on the way out — #3138 replaces it with signed tokens for image downloads, and #3166 moves the security artifacts behind Omni entirely.

utkuozdemir (approved):

Issue #3109, opened 2026-07-10
[bug] infra provider common library doesn't provide a method to get image factory URLs


Unix4ever:
As we don't have a common method for getting an image URL in the providers they generate it on their own.
But as factory address is now picked by the common infra provider module, the rest of the code should follow it.
So if it selects another image factory (which is configured in Omni), infra provider should also generate the image URL using the correct address, not static one.

To fix it we can introduce a method on the provision context struct, that will use the correct base image factory address and the auth credentials for the image factory.

Coler-e:
This should take care of Proxmox Infra provider currently being incompatible with on prem factories because it has the factory url hardcoded right?

PR #3250, opened 2026-08-11
feat: serve infra provider boot assets through Omni

oguzkilcan:
Every infra provider reached the image factory itself, at the hardcoded public URL, so an on-prem or authenticated factory was out of reach, and each provider carried its own copy of the factory's URL layout and file naming.

A provision step now names the asset it wants by kind, platform, architecture and format, and Omni does the rest: it creates the schematic, picks the factory serving that Talos version, works out the filename including the secure boot variant, and answers with a URL plus whatever headers the fetch has to carry.

Providers hold no factory URL, no factory credentials, and no knowledge of how assets are named, so an authenticated or on-prem factory needs no provider-side configuration at all. Where something else performs the fetch and cannot be handed a header, such as PXE firmware or a hypervisor download API given a bare URL, the provider asks for a self-contained URL instead.

Related: #3109


client/pkg/imagefactory/endpoint.go
```diff
@@ -0,0 +1,166 @@
...
+		if pxeBaseURL, err = derivePXEBaseURL(baseURL); err != nil {
+			return Endpoint{}, err
+		}
+	}
+
+	// An Omni that predates ImageFactoryAuth, or one whose role cannot read it, denies the read outright.
+	// Tolerating that keeps provisioning working against older servers, but only where the factory does not
+	// need credentials in the first place.
+	username, password, err := Credentials(ctx, st, baseURL)
+	if err != nil && (isEnterprise || status.Code(err) != codes.PermissionDenied) {
```
Unix4ever: This "if" is hard to understand. I wonder if we could hide this complexity inside the Credentials method. Maybe we can omit isEnterprise flag there. Any factory might have auth technically

smira (approved):

smira:
I guess this doesn't resolve #3109 - as we need to update all providers first? 

Do we have a way to make them start only with newer Omni?

client/pkg/imagefactory/endpoint.go
oguzkilcan: We discussed this and it seems Image Factory is acting differently when accessed with auth headers. Attempting to access a normal Image Factory with auth headers might lead to issues. @shanduur please correct me if I am wrong.

client/pkg/infra/provision/context.go
```diff
@@ -91,18 +93,20 @@ type ConnectionParams struct {
 func NewContext[T resource.Resource](
 	machineRequest *infra.MachineRequest,
 	machineRequestStatus *infra.MachineRequestStatus,
-	state T,
+	providerState T,
 	connectionParams ConnectionParams,
 	imageFactory FactoryClient,
 	runtime controller.QRuntime,
+	omniState state.State,
```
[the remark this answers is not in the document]
oguzkilcan: We will release this with a new minor tag. So users of the library can be aware that there are breaking changes.

oguzkilcan:
> I guess this doesn't resolve #3109 - as we need to update all providers first?
> 
> Do we have a way to make them start only with newer Omni?

You are right, this doesn't resolve the issue. Each infra provider will need to bump their omni client library to have access to these changes. I'll update the description.

client/pkg/imagefactory/endpoint.go
Slessi: I think it makes sense to be dumb about this here and just use auth if Omni says it has auth.

If Omni or factory is configured wrong that wouldn't be a concern of this function.

client/api/omni/management/management.proto
```diff
@@ -424,6 +471,7 @@ service ManagementService {
   rpc KubernetesSyncManifests(KubernetesSyncManifestRequest) returns (stream KubernetesSyncManifestResponse);
   rpc CreateSchematic(CreateSchematicRequest) returns (CreateSchematicResponse);
   rpc CreateSchematicFromRaw(CreateSchematicFromRawRequest) returns (CreateSchematicResponse);
+  rpc GetBootAssetURL(BootAssetRequest) returns (BootAssetResponse);
```
utkuozdemir:
```suggestion
  rpc GetBootAsset(BootAssetRequest) returns (BootAssetResponse);
```

Since we send a `BootAssetRequest` and get a `BootAssetResponse`, not `BootAssetURL[Request|response]`?

oguzkilcan: I first implemented both method and request/response as BootAsset, I updated the method name later, forgot to update request and response. Will do so now.

client/pkg/imagefactory/bootasset.go
```diff
@@ -0,0 +1,304 @@
...
+		switch {
+		case r >= 'a' && r <= 'z', r >= 'A' && r <= 'Z', r >= '0' && r <= '9':
+			return false
+		case r == '.', r == '_', r == '-':
+			return false
+		default:
+			return true
+		}
+	}) == -1 && value != "." && value != ".."
+}
```
utkuozdemir, line 79: Wondering if we should use what stdlib provides here, instead of doing this. For example, build the url, pass it to `URL.Parse()` - if it passes, we are good. There can be also alternative/better ways.

oguzkilcan: Let me look into that.

oguzkilcan: URL Parser would still see a valid url though but it might not be a correct asset url. This function makes sure the value passed doesn't effect other path segments.

client/pkg/imagefactory/clients.go
```diff
@@ -89,17 +98,33 @@ func NewClientsFromState(ctx context.Context, st state.State) (*Clients, error)
...
-	auth, err := safe.ReaderGetByID[*omni.ImageFactoryAuth](ctx, st, normalizeFactoryURL(url))
-	if err != nil && !state.IsNotFoundError(err) {
+// credentialsAllowingDenied reads the credentials for a factory, treating a denied read as "no credentials".
+//
+// An Omni that predates ImageFactoryAuth, or a caller whose role cannot read it, denies the read outright.
+// Callers that only need to reach a factory serving assets anonymously should still get there, and one that
+// does need credentials fails on its own with a 401 naming the factory.
+func credentialsAllowingDenied(ctx context.Context, st state.State, factoryURL string) (username, password string, err error) {
+	username, password, err = Credentials(ctx, st, factoryURL)
+	if err != nil && status.Code(err) != codes.PermissionDenied {
```
utkuozdemir, line 108: I don't understand this - `Credentials` only does a resource read, how can it return PermissionDenied GRPC code?

client/pkg/infra/infra.go
```diff
@@ -207,57 +206,128 @@ func (provider *Provider[T]) Run(ctx context.Context, logger *zap.Logger, opts .
...
-
-	if useProxied {
-		supported, err := supportsProxiedImageFactory(ctx, st, logger)
-		if err != nil {
-			return nil, err
-		}
+// minOmniVersion is the oldest Omni an infra provider built against this library can talk to. Older
+// servers do not serve CreateSchematicFromRaw, and there is no longer a path around it: schematics are
+// created through Omni so that the image factory stays Omni's business alone.
+const minOmniVersion = "1.9"
```
utkuozdemir: Why not simply rely on `Unimplemented` of GRPC here? I'm not a big fan of the pattern of putting Omni versions into Omni itself.

internal/backend/runtime/omni/state_access.go
```diff
@@ -397,7 +397,8 @@ func filterAccess(ctx context.Context, access state.Access) error {
 	switch access.ResourceType {
 	case omni.MachineType, // infra provider needs to be able to read machines to find out force-deleted ones and deprovision them
 		siderolink.APIConfigType,        // infra provider needs to be able to read connection params to join nodes to Omni
-		siderolink.ConnectionParamsType: // infra provider needs to be able to read connection params to join nodes to Omni
+		siderolink.ConnectionParamsType, // infra provider needs to be able to read connection params to join nodes to Omni
+		omni.ImageFactoryAuthType:       // infra provider needs the credentials to pull boot assets from an authenticated factory
```
utkuozdemir: Don't get this - shouldn't the credentials be returned only on the `BootAssetResponse`? If yes, why should infra providers need access to this resource?

oguzkilcan: This is for backwards compatibility, new management endpoint will be valid for Omni 1.11. Older versions will fallback to read the Image Factory credentials and this happens on provider code so Infra provider SA needs read access.

client/pkg/imagefactory/clients.go
oguzkilcan: If the caller doesn't have correct Role to read `ImageFactoryAuth` resource, then state_access would return it, no?

client/pkg/infra/infra.go
oguzkilcan: It was just for the error message, we already use Unimplemented. I'll embed it directly in to the error.

client/pkg/imagefactory/bootasset.go
```diff
@@ -0,0 +1,304 @@
...
+		talosVersion = constants.DefaultTalosVersion
+	}
+
+	if !validPathSegment(schematicID) {
+		return BootAsset{}, fmt.Errorf("%w: invalid schematic ID %q", ErrInvalidInput, schematicID)
+	}
+
+	version := "v" + strings.TrimLeft(talosVersion, "v")
+
+	if !validPathSegment(version) {
```
utkuozdemir: This only checks that the version is a valid path segment, so a value that is not a version at all still resolves successfully. `not-a-version` becomes `.../vnot-a-version/nocloud-amd64.raw.xz` and the caller gets a success response with a URL that can only 404 at the factory, so the useful error appears at download or PXE boot time instead of on this call. `CreateSchematic` already rejects such values with `semver.ParseTolerant`.

Reproducer: save this beside `bootasset_test.go` and run:

```console
cd client && go test ./pkg/imagefactory/ -run TestReviewRejectsInvalidTalosVersion -count=1
```

```go
func TestReviewRejectsInvalidTalosVersion(t *testing.T) {
	t.Parallel()

	ctx := t.Context()
	st := newTestState(t)

	createFeaturesConfig(ctx, t, st, nil)

	_, err := imagefactory.ResolveBootAsset(ctx, st, "not-a-version", diskSpec(), schematicID, false)
	require.ErrorIs(t, err, imagefactory.ErrInvalidInput)
}
```

At this head it fails with `Expected error with "invalid boot asset request" in chain but got nil`.

client/api/omni/management/management.proto
```diff
@@ -189,6 +189,53 @@ message CreateSchematicResponse {
...
+  // TalosVersion selects the image factory the same way asset URLs do.
+  string talos_version = 1;
+  // SchematicId is the ID the image factory returned for a schematic, e.g. from CreateSchematicFromRaw.
+  string schematic_id = 2;
+  // StandaloneUrl makes Omni return a URL containing everything needed, e.g. the credentials or token
+  // in the URL itself, instead of returning them in the headers.
+  //
+  // When this is set, Headers on the response will be empty. It is implied for PXE assets, which are
+  // fetched by firmware that cannot send headers.
+  bool standalone_url = 3;
```
utkuozdemir, line 217: A standalone URL carries the credentials, so wherever it ends up written, an iPXE script, a hypervisor task log, it hands them to whoever can read that. Worth saying on the field, since providers will pass these URLs to systems that log them.

```suggestion
  // StandaloneUrl makes Omni return a URL containing everything needed, e.g. the credentials or token
  // in the URL itself, instead of returning them in the headers.
  //
  // Such a URL is as sensitive as the credentials it carries. Anything it is written into, like an
  // iPXE script or a task log, exposes them to whoever can read it.
  //
  // When this is set, Headers on the response will be empty. It is implied for PXE assets, which are
  // fetched by firmware that cannot send headers.
  bool standalone_url = 3;
```

client/pkg/infra/provision/context.go
```diff
@@ -79,6 +76,13 @@ func WithOverlay(overlay schematic.Overlay) SchematicOption {
 	}
 }
 
+// WithEmbeddedMachineConfig adds embedded machine config to the schematic.
+func WithEmbeddedMachineConfig(config string) SchematicOption {
```
utkuozdemir, line 80: Nothing in this PR uses this option. Is it groundwork for the provider updates?

internal/backend/grpc/schematics.go
```diff
@@ -176,6 +177,54 @@ func (s *managementServer) CreateSchematicFromRaw(ctx context.Context, request *
...
+	management.BootAssetURLRequest_BOOT_ASSET_KIND_PXE:  imagefactory.BootAssetKindPXE,
+	management.BootAssetURLRequest_BOOT_ASSET_KIND_ISO:  imagefactory.BootAssetKindISO,
+	management.BootAssetURLRequest_BOOT_ASSET_KIND_DISK: imagefactory.BootAssetKindDisk,
+}
+
+// GetBootAssetURL implements managementServer.
+func (s *managementServer) GetBootAssetURL(ctx context.Context, request *management.BootAssetURLRequest) (*management.BootAssetURLResponse, error) {
+	if _, err := auth.CheckGRPC(ctx, auth.WithExactRoles(role.InfraProvider, role.Operator, role.Admin)); err != nil {
+		return nil, err
+	}
```
utkuozdemir, line 192: Why not keep `ImageFactoryAuth` non-readable to infra providers and do the read here as the internal actor instead? This endpoint reads the state filtered by the caller's role, so the `state_access.go` change exists only to let this read pass. With this, that change can be reverted, and providers get the credentials only through this response instead of also being able to read and watch the resource directly. It'd be more secure. Needs the `actor` import.

```suggestion
	if _, err := auth.CheckGRPC(ctx, auth.WithExactRoles(role.InfraProvider, role.Operator, role.Admin)); err != nil {
		return nil, err
	}

	// The role check above is the authorization, so read the image factory credentials as Omni itself
	// rather than requiring the caller to be able to read the ImageFactoryAuth resource.
	ctx = actor.MarkContextAsInternalActor(ctx)
```

client/pkg/imagefactory/bootasset.go
```diff
@@ -0,0 +1,304 @@
...
+	}
+
+	return ""
+}
+
+// BootAsset is a fetchable boot asset.
+//
+// Fetch URL, sending Headers when they are not empty. The URL is opaque and can carry secrets, whether
+// as credentials or as a factory token, so it does not belong in a log line.
+type BootAsset struct {
```
utkuozdemir, line 138: Since providers will pass this struct around and log it, what about a `String()` with a value receiver that prints only the host, the schematic ID and the storage key? `zap.Any` and `%v` then stay safe even when the URL carries credentials. Masking just the userinfo would not be enough, a token can travel in the query, so printing only the safe fields seems better. It will not cover reflection or `json.Marshal`, but it covers the common accident.

oguzkilcan: This is a good catch.

client/api/omni/management/management.proto
```diff
@@ -189,6 +189,53 @@ message CreateSchematicResponse {
...
+  // Platform is the Talos platform, e.g. "metal" or "nocloud".
+  string platform = 5;
+  string architecture = 6;
+  // Format is the disk image format of the disk kind, e.g. "raw.xz" or "qcow2.gz", and unused by every
+  // other kind, whose format the kind itself implies.
+  string format = 7;
+  bool secure_boot = 8;
+}
+
+message BootAssetURLResponse {
```
utkuozdemir, line 232: The URL is the only identity the provider gets for the asset here, and it is not a stable one. For example, the Proxmox provider names the ISO it stores after a hash of the URL, and with a standalone URL the credentials are part of that hash:

1. The provider stores the ISO as `7f3a...iso`, a hash of `https://bob:secret1@factory.corp/image/<schematic>/v1.13.8/nocloud-amd64.iso`.
2. The factory password is rotated. The same asset now has `bob:secret2` in the URL, so the hash becomes `c09d...iso`.
3. The provider does not find `c09d...iso` in the storage and downloads the same ISO again. The old copy stays behind forever.

What about returning a stable storage key in this response (e.g., a digest of the factory base URL, the schematic ID, the Talos version and the filename)? Providers would store assets under that name instead. It only changes when the asset itself changes, e.g., when the schematic is different, or a factory serving different builds is configured.

Since Omni computes it, the algorithm can also evolve without the providers updating their client library. The changes to it need to be additive, so that the existing keys stay stable, the same way schematic IDs stay stable when new customization fields are added.

internal/backend/grpc/schematics.go
oguzkilcan: We need to give infra providers ImageFactoryAuth to so, an infra provider using new omni client can connect to an older version of Omni where management endpoint doesn't exist.

oguzkilcan: Come to think of it, you are right. I'll drop state access changes and use internal actor in management api.

internal/backend/runtime/omni/state_access.go
oguzkilcan: Will go away.

utkuozdemir (approved):

PR #3251, opened 2026-08-12
fix: force re-authentication at the IdP so logout takes effect

oguzkilcan:
Logging out of Omni destroys Omni's session but it doesn't destroy IdP's session unless explicit configuration is there (Single Logout for SAML, manually configuring via logoutURL for OIDC). User is then redirected to home page, which starts auth process again, since user's IdP session is still available, user is automatically logged back in. This prevents keeping user logged out as well as being able to switch between accounts.

Now Omni asks for re-authentication after logout. So users logging out of Omni get the same treatment of unauthenticated user opening Omni home page for the first time.

### DIFF
--- a/frontend/e2e/saml/auth.spec.ts
+++ b/frontend/e2e/saml/auth.spec.ts
@@ -34,12 +34,16 @@ test('Auth flow', async ({ page }) => {
   })
 
   await test.step('Logout', async () => {
+    // Keycloak answering the LogoutRequest is what proves Single Logout ran, and it has to be watched for
+    // rather than inferred from the login form: every login now carries ForceAuthn, so that form appears
+    // whether or not the IdP session was ended.
+    const logoutResponse = page.waitForRequest((request) => request.url().includes('/saml/slo'))
+
     await page.getByRole('button', { name: 'user actions' }).click()
     await page.getByRole('menuitem', { name: 'Log Out' }).click()
 
-    // Landing back on the login form means Single Logout ended the Keycloak session too.
-    // If the LogoutRequest had been rejected, Keycloak would answer the next /login
-    // redirect from its own SSO cookie and drop us straight back on the home page.
+    await logoutResponse
+
     await expect(page.getByRole('heading', { name: 'Sign in to your account' })).toBeVisible()
   })
 })
--- a/internal/backend/saml/saml_test.go
+++ b/internal/backend/saml/saml_test.go
@@ -6,20 +6,27 @@
 package saml_test
 
 import (
+	"bytes"
+	"compress/flate"
 	"encoding/base64"
 	"encoding/json"
+	"io"
 	"net/http"
 	"net/http/httptest"
 	"net/url"
 	"strings"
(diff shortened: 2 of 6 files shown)


frontend/e2e/saml/auth.spec.ts
```diff
@@ -34,12 +34,16 @@ test('Auth flow', async ({ page }) => {
   })
 
   await test.step('Logout', async () => {
+    // Keycloak answering the LogoutRequest is what proves Single Logout ran, and it has to be watched for
+    // rather than inferred from the login form: every login now carries ForceAuthn, so that form appears
+    // whether or not the IdP session was ended.
+    const logoutResponse = page.waitForRequest((request) => request.url().includes('/saml/slo'))
```
Slessi, line 40: its not great to assert things like this, we should assert user visible behaviour

why is asserting that we are shown login screen again not enough?

oguzkilcan: Because it's testing whether that SAML provider configured Single Logout. When Single Logout is configured, after user initiates logout Omni cleans up user session and redirects the browser to IdP's SingleLogout endpoint, after that IdP logs out user from itself and redirects the browser to Omni `omni.url/saml/slo`. When Single Logout is not configured this redirect cycle doesn't happen.

internal/backend/saml/saml_test.go
```diff
@@ -89,6 +96,71 @@ func newMiddleware(t *testing.T, sloEndpoint string) *samlsp.Middleware {
...
+	require.Equal(t, http.StatusFound, resp.StatusCode)
+
+	redirectURL, err := url.Parse(resp.Header.Get("Location"))
+	require.NoError(t, err)
+
+	deflated, err := base64.StdEncoding.DecodeString(redirectURL.Query().Get("SAMLRequest"))
+	require.NoError(t, err)
+
+	authnRequest, err := io.ReadAll(flate.NewReader(bytes.NewReader(deflated)))
+	require.NoError(t, err)
```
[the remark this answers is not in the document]
oguzkilcan: Fixed.

frontend/e2e/saml/auth.spec.ts
Slessi: This is testing implementation details, we only need to assert the behaviour:
- Click logout
- Am I logged out?

What SAML flow happened is outside scope of this test IMO.

oguzkilcan: But we have the implementation to do correct redirects in Omni. Shouldn't we test that?

Slessi: Yea but that is unit test territory

Slessi: Discussed in slack, lets keep it

(TL;DR not ideal but it tests only Omni's contract and with `ForceAuthn` this test passes whether SLO is broken or not, without this extra check)

Slessi (approved):

PR #3272, opened 2026-08-14
feat: proxy security reports through omni

Slessi:
Proxy all security report downloads through Omni rather than sending user directly to factory. This allows Omni to handle the factory auth, backed by user's own credentials. Created a new image factory gRPC path for this.

Reorganised code a bit:
- Moved frontend report utility functions to `frontend/src/views/ClusterSecurity/util`
- Moved backend factory artifact functions to `internal/backend/imagefactory/artifacts`

The SBOM/VEX downloads have been moved into the modal, which allows them to be accessed for running clusters as well, not just the installation media wizard.

image

Related #3166


Unix4ever:
I think we could just use the normal gRPC API and put the reports as the blob into `bytes` field in the message there, then we wouldn't have to add an extra middleware for the plain HTTP requests. Then bytes can be added as base64 to a link for a fake download.
Or it's going to be hard to write the frontend that way?

Slessi:
@Unix4ever True I could do that... I just reached straight for HTTP but it didn't cross my mind. Could make like a `factory.proto` or something? And make it work like how downloading support bundles work. I'll take a look at that

Slessi:
@Unix4ever Moved everything to gRPC - should be good now

frontend/src/pages/(authenticated)/machines/installation-media/create/confirmation.mocks.ts
```diff
@@ -152,15 +156,12 @@ export const handlers = (enterprise = true) => [
       downloads,
     })
   }),
-  http.get<{ version: string }>(
-    '/api/vulns/:schematicId/:talosVersion/:arch/report.json',
+  http.post<never, VulnerabilityReportRequest, VulnerabilityReportResponse>(
```
Unix4ever, line 159: As we use gRPC now there should be a generated method for it. Why not use that one?

frontend/src/views/ClusterSecurity/ClusterSecurity.stories.ts
```diff
@@ -158,20 +162,26 @@ const talosVersionsHandler = http.post<never, ListRequest, ListResponse>(
...
 )
 
 // Resolves a vulnerability report for any (schematic, version, arch) the page asks for.
-const scanHandler = http.get('/api/vulns/:schematic/:version/:arch/report.json', ({ params }) => {
-  return HttpResponse.json({
-    status: 'done',
-    report: { matches: matchesForVersion(params.version as string) },
-  })
-})
+const scanHandler = http.post<never, VulnerabilityReportRequest, VulnerabilityReportResponse>(
```
Unix4ever, line 173: And same here

frontend/src/pages/(authenticated)/machines/installation-media/create/confirmation.mocks.ts
Slessi: this and the other comment are storybook mocks, code uses the `ImageFactoryService.VulnerabilityReport` in the actual runtime code

Unix4ever (approved):

PR #3275, opened 2026-08-14
test: run talemu suite in both chrome and firefox

Slessi:
Run the `e2e-talemu` suite twice, once in chrome and once in Firefox.

As part of this change, changed the CI host from `my-instance.omni.localhost` to `omni.ci.test` as Firefox (like Chrome) has special handling of `*.localhost` addresses which binds to its own loopback, rather than the host's loopback, which breaks address resolution for them both. Chrome was working OK due to a slightly different implementation and a special `chromiumArgs` helper, but Firefox wouldn't load any pages at all.

Instead of adding more workaround code, just changed the host and dropped the custom chromium code.


utkuozdemir:
Good that you changed the domain name, I didn't like `*.localhost`. This is better for workload proxy. Just make sure that workload proxy tests are not broken - I'm adding the label.

utkuozdemir (approved):

PR #3287, opened 2026-08-18
fix: properly clean up pods of the health check jobs

### LINKED ISSUE #3282: [bug] cluster health check job pods are not cleaned up properly
As job deletion doesn't guarantee pod cleanup, we need a foreground way to clean them up, or need to use a unique name for the created healthcheck jobs.

Unix4ever:
Also improve the UI of the healthchecks: show the existing jobs, save
their last failure in a resource. Display the last failure reason in the UI.

Fixes: https://github.com/siderolabs/omni/issues/3282

### DIFF
--- a/client/api/omni/specs/omni.proto
+++ b/client/api/omni/specs/omni.proto
@@ -1849,6 +1849,33 @@ message KubernetesHealthCheckSpec {
   google.protobuf.Duration interval = 2;
 }
 
+// KubernetesHealthCheckStatusSpec is the outcome of the last run of a KubernetesHealthCheck.
+//
+// The runner job is deleted as soon as it reaches a terminal state, so the failure output is captured here
+// while the job's pod still exists: it stays readable after the pod, the job and the upgrade itself are gone.
+message KubernetesHealthCheckStatusSpec {
+  enum State {
+    UNKNOWN = 0;
+    // RUNNING means the runner job is executing and Omni is waiting for it to finish.
+    RUNNING = 1;
+    // FAILED means the last run of the healthcheck failed, holding the upgrade.
+    FAILED = 2;
+    // PASSED means the last run of the healthcheck succeeded.
+    PASSED = 3;
+  }
+
+  State state = 1;
+  // Error is a short, single-line description of the last failure, suitable for a table cell.
+  string error = 2;
+  // Output is the full output of the last failed run: the tail of the failing container's output, captured by
+  // the kubelet as the pod's termination message. Empty when the check has never failed or produced no output.
+  string output = 3;
+  // JobNamespace and JobName locate the runner job in the workload cluster, so a running check's progress can
+  // be inspected with `kubectl logs -n <job_namespace> job/<job_name>`.
+  string job_namespace = 4;
+  string job_name = 5;
+}
+
 // MachineConfigExtractionStatusSpec tracks whether the config a machine arrived with has been extracted into a config patch.
 message MachineConfigExtractionStatusSpec {
   // Initialized is set once the machine's incoming config has been extracted (or determined to be empty / not preservable).
--- a/frontend/src/views/Overview/components/ClusterHealthChecks.vue
+++ b/frontend/src/views/Overview/components/ClusterHealthChecks.vue
@@ -1,0 +1,127 @@
+<!--
+Copyright (c) 2026 Sidero Labs, Inc.
+
+Use of this software is governed by the Business Source License
+included in the LICENSE file.
+-->
+<script setup lang="ts">
+import { computed, ref } from 'vue'
+
+import { Runtime } from '@/api/common/omni.pb'
+import type { KubernetesHealthCheckStatusSpec } from '@/api/omni/specs/omni.pb'
+import { KubernetesHealthCheckStatusSpecState } from '@/api/omni/specs/omni.pb'
+import { EphemeralNamespace, KubernetesHealthCheckStatusType, LabelCluster } from '@/api/resources'
+import TButton from '@/components/Button/TButton.vue'
+import CodeBlock from '@/components/CodeBlock/CodeBlock.vue'
+import type { IconType } from '@/components/Icon/TIcon.vue'
+import TIcon from '@/components/Icon/TIcon.vue'
+import LogViewer from '@/components/LogViewer/LogViewer.vue'
+import Modal from '@/components/Modals/Modal.vue'
+import TInput from '@/components/TInput/TInput.vue'
+import type { LogLine } from '@/methods/logs'
+import { useResourceWatch } from '@/methods/useResourceWatch'
+
+const { clusterId } = defineProps<{ clusterId: string }>()
+
+const { data: healthChecks } = useResourceWatch<KubernetesHealthCheckStatusSpec>(() => ({
+  runtime: Runtime.Omni,
+  resource: {
+    namespace: EphemeralNamespace,
+    type: KubernetesHealthCheckStatusType,
+  },
+  selectors: [`${LabelCluster}=${clusterId}`],
+}))
+
+const State = KubernetesHealthCheckStatusSpecState
+
+const stateProps: Record<KubernetesHealthCheckStatusSpecState, { icon: IconType; class: string }> =
+  {
+    [State.PASSED]: { icon: 'check-in-circle', class: 'text-green-g1' },
+    [State.FAILED]: { icon: 'error', class: 'text-red-r1' },
+    [State.RUNNING]: { icon: 'loading', class: 'animate-spin text-yellow-y1' },
+    [State.UNKNOWN]: { icon: 'time', class: 'text-naturals-n10' },
+  }
+
+const stateFor = (spec: KubernetesHealthCheckStatusSpec) => stateProps[spec.state ?? State.UNKNOWN]
+
+// the healthcheck whose output is shown in the modal. The ID is kept while the modal animates closed,
+// so its content does not flash as it disappears
+const checkModal = ref<{ id?: string; open: boolean }>({ open: false })
+const searchOutput = ref('')
+
+const openedCheck = computed(() =>
+  healthChecks.value.find((check) => check.metadata.id === checkModal.value.id),
+)
+
+// the output is the tail of the failing container's own output, captured by Omni before the runner
+// job was deleted - it carries no timestamps of its own
+const outputLines = computed<LogLine[]>(() =>
+  (openedCheck.value?.spec.output ?? '').split('\n').map((msg) => ({ msg })),
+)
+
+const kubectlLogs = (spec: KubernetesHealthCheckStatusSpec) =>
+  `kubectl logs -n ${spec.job_namespace} job/${spec.job_name}`
+</script>
+
+<template>
+  <div v-if="healthChecks.length" class="mb-5 rounded bg-naturals-n2 pt-5">
+    <div class="flex items-center gap-1 px-6 pb-4">
+      <span class="flex-1 text-sm text-naturals-n13">Health Checks</span>
+    </div>
+
+    <div class="flex flex-col gap-2 border-t-8 border-naturals-n4 p-4 text-xs">
+      <div
+        v-for="check in healthChecks"
+        :key="check.metadata.id"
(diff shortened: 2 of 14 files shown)


Unix4ever:
New UI elements:

Screenshot 2026-08-18 at 19 06 33
Screenshot 2026-08-18 at 19 06 43

frontend/src/views/Overview/components/ClusterHealthChecks.vue
```diff
@@ -0,0 +1,130 @@
...
+      <span class="min-w-0 flex-1 truncate text-naturals-n13">{{ check.metadata.id }}</span>
+      <TButton
+        v-if="check.spec.output"
+        variant="secondary"
+        icon="log"
+        class="shrink-0"
+        @click="openedCheck = check.metadata.id"
+      >
+        View output
+      </TButton>
```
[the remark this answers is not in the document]
Unix4ever: last pod logs are shown. It should be alright

frontend/src/views/Overview/components/OverviewContent.vue
```diff
@@ -405,7 +406,7 @@ const machineLockedForSecretRotation = computed(() => {
             class="h-6 w-6"
           />
           <TIcon v-else icon="loading" class="h-6 w-6 animate-spin text-yellow-y1" />
-          <div class="flex-1">
+          <div class="min-w-0 flex-1 break-words">
```
Slessi: > The class `break-words` can be written as `wrap-break-word`

Using a deprecated class there

frontend/src/views/Overview/components/ClusterHealthChecks.vue
```diff
@@ -0,0 +1,130 @@
...
+const opened = computed(() =>
+  healthChecks.value.find((check) => check.metadata.id === openedCheck.value),
+)
+
+const modalOpen = computed({
+  get: () => !!opened.value,
+  set: (value: boolean) => {
+    if (!value) openedCheck.value = undefined
+  },
+})
```
Slessi: More confusing than it needs to be IMO. `opened` sounds like a boolean but is a health check. `modalOpen` is a boolean for `get` but `set` edits the selected health check. Editing the `openedCheck` to close might cause a content flash during the closing animation (didnt confirm) when `outputLines` change.

There is IMO a simpler way:

```ts
// ID / open combo object
const checkModal = ref<{ id?: string; open: boolean }>({ open: false })

// Note the rename from opened -> openedCheck
const openedCheck = computed(() =>
  healthChecks.value.find((check) => check.metadata.id === checkModal.value.id),
)

const outputLines = computed<LogLine[]>(() =>
  (openedCheck.value?.spec.output ?? '').split('\n').map((msg) => ({ msg })),
)

// Set entire object at once
<TButton
  @click="checkModal = { open: true, id: check.metadata.id }"
/>
// Assign child values for v-model / params
<Modal
  v-model:open="checkModal.open"
  :title="`Health Check: ${checkModal.id}`"
/>
```

frontend/src/views/Overview/components/ClusterHealthChecks.vue
```diff
@@ -0,0 +1,130 @@
...
+      >
+        View output
+      </TButton>
+    </div>
+
+    <Modal
+      v-model:open="modalOpen"
+      :title="`Health Check: ${openedCheck}`"
+      cancel-label="Close"
+      content-class="flex w-3xl max-w-screen flex-col gap-4 overflow-hidden"
```
Slessi: This will cause the content to overflow off screen when less than `w-3xl` is available for the content.

The inner content has `max-w-screen` but its surrounded by padding, so the entire modal will expand to `w-screen` + padding.

Drop `w-3xl max-w-screen` and use `max-w-3xl` - that is probably what you want

Slessi (commented):
[frontend review]

client/api/omni/specs/omni.proto
```diff
@@ -1819,6 +1819,33 @@ message KubernetesHealthCheckSpec {
...
+    // RUNNING means the runner job is executing and Omni is waiting for it to finish.
+    RUNNING = 1;
+    // FAILED means the last run of the healthcheck failed, holding the upgrade.
+    FAILED = 2;
+    // PASSED means the last run of the healthcheck succeeded.
+    PASSED = 3;
+  }
+
+  State state = 1;
+  // error is a short, single-line description of the last failure, suitable for a table cell.
```
utkuozdemir: nit: these fields will be capitalized in go, so by the convention, this would start like `Error is ...`. Also applies to fields below.

utkuozdemir (approved):
lgtm overall, but pls see the comments

internal/backend/runtime/omni/controllers/omni/talosupgrade/status.go
```diff
@@ -611,12 +641,16 @@ type healthCheckRunnerStatus struct {
...
 // Each healthcheck is executed in a runner pod that is created fresh on each attempt and deleted once it reaches a terminal state,
 // so the next attempt runs a fresh check. A non-nil error is only returned for unexpected Kubernetes API or manifest-parsing errors.
-func (ctrl *TalosUpgradeStatusController) runHealthChecks(ctx context.Context, r controller.Reader, logger *zap.Logger, cluster *omni.Cluster) (healthCheckResult, error) {
+func (ctrl *TalosUpgradeStatusController) runHealthChecks(ctx context.Context, r controller.ReaderWriter, logger *zap.Logger, cluster *omni.Cluster) (healthCheckResult, error) {
 	healthchecks, err := safe.ReaderListAll[*omni.KubernetesHealthCheck](ctx, r, state.WithLabelQuery(resource.LabelEqual(omni.LabelCluster, cluster.Metadata().ID())))
 	if err != nil {
 		return healthCheckResult{}, err
 	}
 
+	if err = ctrl.cleanupHealthCheckStatuses(ctx, r, cluster, healthchecks); err != nil {
```
utkuozdemir: This is the only path that cleans up a `KubernetesHealthCheckStatus` whose healthcheck is gone, and it is only reached while the upgrade is actively gated on the healthchecks. Because of this:

1. A cluster finishes an upgrade, phase goes to `Done`.
2. The user deletes a `KubernetesHealthCheck`. The controller reconciles, but `runHealthChecks` is not called, so the status stays behind indefinitely.
3. Similarly, canceling an upgrade while a check is mid-run freezes its status at `RUNNING`, and the UI shows a spinner through the whole revert.

A reused healthcheck ID would also inherit the stale error and output on its first `RUNNING` report, since those are intentionally preserved across the running state.

Should we move the cleanup call into the transform itself, next to `reconcileTalosVersions` which already does the same diff-and-teardown for `ClusterMachineTalosVersion`?

frontend/src/views/Overview/components/OverviewContent.vue
```diff
@@ -426,6 +427,7 @@ const machineLockedForSecretRotation = computed(() => {
             Cancel
           </TButton>
         </div>
+        <ClusterHealthChecks :cluster-id="clusterId" />
```
utkuozdemir: This block is behind `v-if="talosUpgradeStatus && talosUpgradeStatus.spec.status"`, and the status field is cleared when the phase goes to `Done`. So the healthcheck list and the "View output" button disappear the moment the upgrade finishes or is canceled, which is exactly when the captured failure output is needed to understand what happened.

The status resource still holds the output, but the UI can no longer show it. Should we show this component whenever statuses exist, outside the upgrade block?

internal/backend/runtime/omni/controllers/omni/talosupgrade/status.go
```diff
@@ -931,9 +1068,30 @@ func deleteHealthCheckJob(ctx context.Context, client *kubernetes.Client, namesp
 		return fmt.Errorf("failed to delete healthcheck job: %w", err)
 	}
 
+	// Background propagation only queues the pods for garbage collection, so delete them explicitly: the job is
+	// already gone at this point, so nothing re-creates them. Their own grace period applies - force deleting
+	// would not help, as it's the job tracking finalizer, not the grace period, that keeps a pod around.
+	if err := client.Clientset().CoreV1().Pods(namespace).DeleteCollection(ctx, v1.DeleteOptions{}, v1.ListOptions{
```
utkuozdemir, line 1137: This makes a failed pod cleanup fatal on the passed-check path too: `JobComplete` -> `deleteHealthCheckJob` -> error -> the whole transform errors, the passing result is dropped and the upgrade status is not written at all. e.g., an admission webhook on pod DELETE failing closed would hold the upgrade with no hint of the cause.

`deleteHealthCheckRunnerObjects` above states the policy as "cleanup must not block the upgrade", so it is probably better to log and continue here, keeping only the job deletion error fatal.

internal/backend/runtime/omni/controllers/omni/talosupgrade/status.go
```diff
@@ -917,10 +1039,25 @@ func deleteHealthCheckRunnerObjects(ctx context.Context, logger *zap.Logger, cli
 				zap.String("namespace", job.Namespace), zap.String("job", job.Name), zap.Error(err))
 		}
 	}
+
+	// pick up any runner pods left behind by a job that is already gone (e.g. deleted before the pod label was
```
utkuozdemir: Pods created before this change never carried this label (only the job did), and the existing jobs are not recreated to gain it, since the config hash is computed over the raw manifest which did not change. So this sweep only catches the pods left behind by the new code, not the ones that piled up before the upgrade to this version.

Worth tightening the comment, or listing the pods that have the `batch.kubernetes.io/job-name` label and filtering by the `omni-healthcheck-` prefix, which the old pods do carry.

frontend/src/views/Overview/components/ClusterHealthChecks.vue
```diff
@@ -0,0 +1,124 @@
...
+      :title="`Health Check: ${checkModal.id}`"
+      cancel-label="Close"
+      content-class="flex max-w-3xl flex-col gap-4 overflow-hidden"
+    >
+      <template #description>
+        Output of the last failed run. Omni captures it before deleting the runner job, so it stays
+        available after the job's pod is gone.
+      </template>
+
+      <div v-if="openedCheck?.spec.job_name" class="flex flex-col gap-1">
```
utkuozdemir: The modal only opens when there is failure output, and at that point the runner job was just deleted (recreated on the next attempt, never again once the upgrade ends). So the copied command often answers with `jobs.batch ... not found`.

Gating this hint on `state === RUNNING` instead of `job_name` would make it show only when it works.

internal/integration/talos_test.go
```diff
@@ -627,6 +627,15 @@ func assertUpgradeHeldByHealthCheck(ctx context.Context, t *testing.T, st state.
 		// the failing healthcheck container's stderr must bubble up into the upgrade status step
 		assert.Contains(r.TypedSpec().Value.Step, upgradeGateClosedMessage, resourceDetails(r))
 	})
+
+	// the failure output must also be captured in the healthcheck's own status, so it stays readable in the UI
+	// after the runner job and its pod are deleted
+	rtestutils.AssertResources(ctx, t, st, []resource.ID{healthCheckID}, func(r *omni.KubernetesHealthCheckStatus, assert *assert.Assertions) {
+		assert.Equal(specs.KubernetesHealthCheckStatusSpec_FAILED, r.TypedSpec().Value.State, resourceDetails(r))
```
utkuozdemir, line 639: Since the PR is about the pods being cleaned up, this test could also assert that no runner pods are left in the workload cluster once the upgrade completes: `kubeClient` is already in scope in `AssertTalosUpgradeFlow`, and listing the pods with the `app.kubernetes.io/name=omni-health-checker` selector after the upgrade should come back empty.

That would be the actual regression test for the fix.

frontend/src/views/Overview/components/OverviewContent.vue
Unix4ever: Good idea

internal/backend/runtime/omni/controllers/omni/talosupgrade/status.go
Unix4ever: Yeah... Good catch

PR #3297, opened 2026-08-20
chore: bump deps

oguzkilcan:
Bump dependencies


smira (approved):
probably update config schema for rc.1 ? or we don't have it yet?

PR #3298, opened 2026-08-20
fix: stop a link to an exposed service from starting a login flow

oguzkilcan:
The workload proxy auth cookies were Strict, so a browser withheld them from any top-level navigation starting outside the instance domain. Someone already logged in to Omni who followed a link to an exposed service was pushed through the whole login flow, and since every AuthnRequest carries ForceAuthn, that flow now stops at an interactive prompt from the IdP.

Lax attaches the cookies to that navigation, so the click reaches the service. Strict was not denying those requests in the first place: it redirected them to login, signed the caller's own URL into the redirect parameter, and served them once a session existed. The handler now makes that decision itself, reading fetch metadata to refuse cross-site subresource loads, frames and unsafe methods, and answering an unauthenticated non-navigation with 401 rather than sending an asset request to a login page.

### DIFF
--- a/internal/backend/grpc/auth.go
+++ b/internal/backend/grpc/auth.go
@@ -47,6 +47,13 @@ const (
 	awaitPublicKeyConfirmationTimeout = 5 * time.Minute
 )
 
+// workloadProxyCookieSameSite is Lax so that a link to an exposed service still works when someone
+// follows it from outside the instance domain. Strict withholds these cookies on that navigation and
+// turns the click into a full login round trip for a person who is already logged in. The proxy
+// handler reads fetch metadata to refuse cross-site subresource loads and unsafe methods, which is
+// the part SameSite cannot express on its own.
+const workloadProxyCookieSameSite = http.SameSiteLaxMode
+
 type authServer struct {
 	authpb.UnimplementedAuthServiceServer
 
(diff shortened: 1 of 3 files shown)


Unix4ever (commented):
I noticed that sometimes if you open exposed service directly by the URL it constantly redirects you to the home page instead. Is this change fixing that issue too?

oguzkilcan:
> I noticed that sometimes if you open exposed service directly by the URL it constantly redirects you to the home page instead. Is this change fixing that issue too?

I can't think of a scenario opening an exposed service redirecting to Omni homepage. I wonder if it was trying get you to login. What do you mean exactly by redirecting to the homepage? To ask you to login again or at least click the login button?

Slessi (approved):

Unix4ever:
> > I noticed that sometimes if you open exposed service directly by the URL it constantly redirects you to the home page instead. Is this change fixing that issue too?
> 
> I can't think of a scenario opening an exposed service redirecting to Omni homepage. I wonder if it was trying get you to login. What do you mean exactly by redirecting to the homepage? To ask you to login again or at least click the login button?

The bug I see pretty often now: I open the exposed service via it's full domain directly. If I'm logged out, I get through login flow and end up on the Omni homepage incorrectly. And to make it worse, the domain doesn't even change to the Omni one, it keeps being that `abcdefg-account.proxy-us.omni.siderolabs.io` while showing the homepage.
It's reproduced on my mobile Firefox all the time. If you reload the page it keeps opening the homepage instead of the exposed service.

Looks like this, pay attention to the URL on top

PR #3299, opened 2026-08-20
fix: provide machine id consistently in provision API logs

### LINKED ISSUE #3273: siderolink provision API logs should have a machine ID

smira:
Fixes #3273

### DIFF
--- a/internal/pkg/siderolink/provision.go
+++ b/internal/pkg/siderolink/provision.go
@@ -113,7 +113,7 @@ func (h *ProvisionHandler) runCleanup(ctx context.Context) error {
 			for machine := range pendingMachines.All() {
 				if time.Since(machine.Metadata().Updated()) > time.Second*30 || machine.Metadata().Phase() == resource.PhaseTearingDown {
 					if err = h.removePendingMachine(ctx, machine); err != nil {
-						h.logger.Error("failed to remove pending machine", zap.Error(err), zap.String("id", machine.Metadata().ID()))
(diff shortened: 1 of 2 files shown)


oguzkilcan (approved):

internal/pkg/siderolink/provision.go
```diff
@@ -396,14 +397,25 @@ func (h *ProvisionHandler) removePendingMachine(ctx context.Context, pendingMach
...
 
-func establishLink[T res](ctx context.Context, h *ProvisionHandler, provisionContext *provisionContext, annotationsToAdd []string, annotationsToRemove []string,
+// pendingMachineFields describes a pending machine in the logs: it is keyed by the node public key,
+// so the machine ID is only available as a label.
+func pendingMachineFields(pendingMachine *siderolinkres.PendingMachine) []zap.Field {
+	machineID, _ := pendingMachine.Metadata().Labels().Get(omni.MachineUUID)
+
+	return []zap.Field{
+		zap.String("machine", machineID),
+		zap.String("id", pendingMachine.Metadata().ID()),
```
Unix4ever: This id has the public key for the pending machines I think. So it might look a bit weird in the logs. But it's mostly a nit.

Unix4ever (approved):

internal/pkg/siderolink/provision.go
smira: would you like to change it to some other label?

Unix4ever: maybe `peer_key`? I've checked again and it's identified by that

smira: done!

PR #3307, opened 2026-08-24
fix: drop manifest apply-order weighting and fix stale status tracking

kevintijssen:
Remove the weight-based prefix from Kubernetes manifest group resource IDs, since apply order does not depend on declaration order and the weight only added churn to IDs across renames/reorders. Update the export sort comment to reflect that IDs are now purely name-based.

Also fix ClusterManifestsStatusController to track which manifest identities are currently owned by each group. Previously, when a group was torn down and recreated under a new ID, its manifests could be incorrectly marked as deleting and left waiting on a live-resource check that would never resolve, since the identity still existed under a different live group. Now stale entries are purged immediately when an identity is found to have moved to another live group.

### DIFF
--- a/client/pkg/template/operations/export.go
+++ b/client/pkg/template/operations/export.go
@@ -584,7 +584,8 @@ func transformManifestsToModels(manifests []*omni.KubernetesManifestGroup) (mode
 		return nil, nil
 	}
 
-	// sort by resource ID to preserve the original weight ordering
+	// sort by resource ID (which is name-based) for a deterministic, stable export order;
+	// manifest apply order doesn't depend on declaration order, so this doesn't need to match the original list order
 	slices.SortFunc(manifests, func(a, b *omni.KubernetesManifestGroup) int {
 		return strings.Compare(a.Metadata().ID(), b.Metadata().ID())
 	})
--- a/internal/backend/runtime/omni/controllers/omni/kubernetes/cluster_manifest_status_test.go
+++ b/internal/backend/runtime/omni/controllers/omni/kubernetes/cluster_manifest_status_test.go
@@ -13,6 +13,8 @@ import (
 	"github.com/stretchr/testify/require"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+
+	"github.com/siderolabs/omni/client/api/omni/specs"
 )
 
 func TestManifestGroupForClusterScopedObjectWithNamespace(t *testing.T) {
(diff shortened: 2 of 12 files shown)


Unix4ever:
I _think_ order actually matters for SSA. Like if you have manifests from one group and the other one, you have some overrides in the second one, it should override the first group.

Or you observed a different behavior?
What if you try to change the same deployment from two groups?
Not sure if that kind of usage makes any sense though.

kevintijssen:
> I _think_ order actually matters for SSA. Like if you have manifests from one group and the other one, you have some overrides in the second one, it should override the first group.
> 
> Or you observed a different behavior? What if you try to change the same deployment from two groups? Not sure if that kind of usage makes any sense though.

The SSA order matters in the group but all groups are applied simultaneously. I have seen this behavior continuously. That should be the default behavior or both are fighting over the same resource (like ArgoCD conflicting resources).
This issue I'm trying to solve here is that if you delete a manifest (group) the order is changed and all are re-applied if missing. The "one-time" mode also runs again if the order is changed, because it’s losing the weight number

This solves all of that and the weight number is also remove from the name.

client/pkg/template/operations/export.go
```diff
@@ -584,7 +584,8 @@ func transformManifestsToModels(manifests []*omni.KubernetesManifestGroup) (mode
 		return nil, nil
 	}
 
-	// sort by resource ID to preserve the original weight ordering
+	// sort by resource ID (which is name-based) for a deterministic, stable export order;
+	// manifest apply order doesn't depend on declaration order, so this doesn't need to match the original list order
 	slices.SortFunc(manifests, func(a, b *omni.KubernetesManifestGroup) int {
```
Unix4ever, line 589: I wonder if we should also have migration code in the cluster export code.
Like if you have the legacy manifests in the cluster and do export I think it will end up having the wrong names (with weight encoded into the names).

kevintijssen: > I wonder if we should also have migration code in the cluster export code. Like if you have the legacy manifests in the cluster and do export I think it will end up having the wrong names (with weight encoded into the names).

The original names will remain, because if the weight was exported as well, an import we create new resources with a double name convention.

internal/backend/runtime/omni/controllers/omni/kubernetes/cluster_manifest_status_test.go
```diff
@@ -78,3 +80,45 @@ func TestManifestGroupForObjectMatchesGroupKind(t *testing.T) {
...
 }
+
+func TestIsLegacyGroupID(t *testing.T) {
+	for _, tt := range []struct {
+		name         string
+		oldID, newID string
+		want         bool
+	}{
+		{"weighted match", "200-cluster-edge-migration", "cluster-edge-migration", true},
+		{"different padding width", "7-cluster-edge-migration", "cluster-edge-migration", true},
```
Unix4ever: "7-" is not a valid weight. We always pad it to be 3 digits.

internal/backend/runtime/omni/controllers/omni/kubernetes/cluster_manifest_status_test.go
```diff
@@ -78,3 +80,45 @@ func TestManifestGroupForObjectMatchesGroupKind(t *testing.T) {
...
+		{"identical strings are not legacy", "cluster-edge-migration", "cluster-edge-migration", false},
+		{"unrelated id", "cluster-edge-migration", "cluster-other", false},
+		{"non-numeric prefix", "abc-cluster-edge-migration", "cluster-edge-migration", false},
+		{"missing separating hyphen", "200cluster-edge-migration", "cluster-edge-migration", false},
+		{"digits embedded in the manifest name still match", "200-cluster-x-7-foo", "cluster-x-7-foo", true},
```
Unix4ever: I think the right format is always `<cluster-name>-<weight>-<name>` actually. Except 7 will be 007.

internal/backend/runtime/omni/controllers/omni/kubernetes/cluster_manifest_status_test.go
kevintijssen: I will remove that one

Unix4ever (approved):

internal/backend/runtime/omni/controllers/omni/kubernetes/cluster_manifest_status_test.go
kevintijssen: I will remove all except these.
* Weighted match
* Identical

Reason: Others are and will never be used

PR #3309, opened 2026-08-24
feat: authenticate boot asset downloads with image factory tokens

oguzkilcan:
Omni authenticates boot asset downloads for infra providers with a short-lived token from the image factory, scoped to one identity and accepted only on image downloads, so a provider holds only what the download needs. A provision step that passes the URL to something else, instead of downloading immediately, asks for the lifetime it needs with download_token_ttl and learns when the URL stops working from expires_at. PXE keeps basic auth, since the iPXE script the factory serves passes on credentials alone, and a factory that issues no tokens keeps working unchanged.

Related to: #3109

### DIFF
--- a/client/pkg/imagefactory/bootasset.go
+++ b/client/pkg/imagefactory/bootasset.go
@@ -14,11 +14,14 @@ import (
 	"net/http"
 	"net/url"
 	"strings"
+	"time"
 
 	"github.com/blang/semver/v4"
 	"github.com/cosi-project/runtime/pkg/safe"
 	"github.com/cosi-project/runtime/pkg/state"
+	"github.com/siderolabs/image-factory/pkg/client"
 	"github.com/siderolabs/talos/pkg/machinery/platforms"
+	"go.uber.org/zap"
 
 	"github.com/siderolabs/omni/client/pkg/constants"
 	"github.com/siderolabs/omni/client/pkg/omni/resources/omni"
@@ -140,6 +143,15 @@ type BootAsset struct {
 	// authentication, or when it travels inside the URL instead.
 	Headers http.Header
 
+	// ExpiresAt is when URL stops working, for a caller that does not perform the fetch itself and has
+	// to know how long the URL it passed on stays good. Zero when the URL does not expire: an anonymous
+	// factory, or one authenticated with credentials rather than a token.
+	//
+	// It is an estimate. The factory confirms only that it granted the lifetime Omni asked for, never the
+	// instant it started counting, and it allows another 30s of clock leeway when verifying. Treat it as
+	// the earliest moment the URL may stop working.
+	ExpiresAt time.Time
+
 	// URL of the boot asset.
 	URL string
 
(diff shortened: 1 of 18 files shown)


client/pkg/imagefactory/bootasset.go
```diff
@@ -185,17 +203,132 @@ func storageKey(factoryBaseURL, pathPrefix, schematicID, version, filename strin
...
+// instead of the basic auth credentials Omni holds.
+func WithDownloadTokens(opts DownloadTokenOptions) ResolveOption {
+	return func(o *resolveOptions) {
+		o.downloadTokens = &opts
+	}
+}
+
+// requestDownloadToken returns a token authenticating a download from the factory at baseURL and the
+// moment it stops working.
+func requestDownloadToken(ctx context.Context, baseURL string, opts *DownloadTokenOptions) (string, time.Time, bool, error) {
```
smira, line 248: we can re-use the token for the same schematic/... , should we do that ? I'm not sure how it would look like in the frontend if we request 20 tokens if we need just one (on a page which lists e.g. 20 different kinds of assets)

oguzkilcan: We do not provide schematic info while requesting the token. How will we know that it can be used for the same schematic? Shouldn't it be used for everything under `/image` while the token is valid?

We can sure cache it and use it before it expires but Omni is the middleman here, we do not know when the token will be used by the infra provider. Since TTL is settable by the client, wouldn't it be weird if we return a token with a expiry date way less then requested TTL duration or way more?

smira: Sorry, schematic ID should be my wrong thinking - owner if the key.

But I was thinking about the frontend showing 10 URLs to different assets - if we could save a bit by re-using the token, but I'm not sure if your PR covers the frontend even

oguzkilcan: No, this only covers the management endpoint to provide boot assets to clients (mainly infra provider clients). Frontend would directly use backend. I guess we can add caching per token/owner but in the end we don't know when the url will be used for media downloads.

IMHO adding caching at this step is premature optimization. We can evaluate the usage after this ships and act on it.

client/pkg/imagefactory/bootasset.go
```diff
@@ -185,17 +203,132 @@ func storageKey(factoryBaseURL, pathPrefix, schematicID, version, filename strin
...
+
+		return "", time.Time{}, false, nil
+	case err == nil:
+		return token, requestedAt.Add(ttl), true, nil
+	case client.IsHTTPErrorCode(err, http.StatusNotFound), client.IsHTTPErrorCode(err, http.StatusMethodNotAllowed):
+		logger.Debug("the image factory does not issue download tokens, authenticating the download with credentials",
+			zap.String("factory_url", baseURL), zap.Error(err))
+
+		return "", time.Time{}, false, nil
+	case client.IsInvalidSchematicError(err):
```
Unix4ever, line 289: This thing looks weird for DownloadToken responses. But I guess the problem on the IF client side 🫠 

I guess here it should be generic, something like IsBadRequest error

client/pkg/imagefactory/bootasset.go
```diff
@@ -185,17 +203,132 @@ func storageKey(factoryBaseURL, pathPrefix, schematicID, version, filename strin
...
+
+	logger := opts.Logger
+	if logger == nil {
+		logger = zap.NewNop()
+	}
+
+	factory := opts.Factories.ForURL(baseURL)
+	if factory == nil {
+		logger.Warn("no image factory client is configured for the factory serving this asset, authenticating the download with credentials",
+			zap.String("factory_url", baseURL))
```
Slessi: i think this message (and similar ones below) should remove the "authenticating the download with credentials" section, and that section should live next to the part that actually injects credentials, as a separate log line

at this stage you dont know what the consumer will be doing with the missing token, you only can log why there is no token

Unix4ever (approved):

client/pkg/imagefactory/bootasset.go
oguzkilcan: Yes, it is on the IF client side.

PR #3314, opened 2026-08-25
feat(frontend): hide machine creation commands not supported by enterprise

### LINKED ISSUE #3139: Enterprise: present Image Factory credentials as customer ID + API keys, never expose secrets
### Problem Description

Enterprise Image Factory credentials are presented as a "username" and "password" — shared-credential semantics, with the secret handled like a human login. Enterprise customers expect API-key semantics: a per-customer identifier plus a revocable secret.

The original driver for this ticket — plaintext credentials embedded in the download/boot URLs of the install-media dialog — is now owned elsewhere: image download URLs are replaced by signed download tokens (#3138, mechanism shipped in siderolabs/image-factory#509), and security artifacts (SBOM/VEX/scan reports) will be served through Omni server-side so factory credentials never reach the browser (#3166).

What neither of those covers is the credential the customer holds and uses **outside Omni**:

- The cluster create command on the install-media confirmation screen renders `--image-factory-auth='username:password'` — the local cluster pulls the installer from the factory and must authenticate to it.
- The installer image in the machine config is pulled from the enterprise factory's registry, so machines authenticate with registry credentials.
- Self-hosted Omni is itself configured with factory credentials (`registries.factories.primary`).

These need a durable credential — a short-lived link can't serve them. The question this ticket answers is what shape that credential takes and how it is presented.

The shape is already half-decided: the factory accepts Auth0 bearer tokens in the Basic auth password field (siderolabs/image-factory#514), because OCI and Talos registry clients only speak Basic. So what a customer holds is an identifier plus a token — customer ID + API key. The rename describes what the credential now actually is.

### Solution

- Rename throughout the enterprise UI: "username" → **customer ID**, "password" → **API key**.
- **Deliberate, never incidental:** a credential appears only where the user explicitly asked for it (e.g. the cluster-create command they're about to copy), never baked into rendered URLs, download links, or page content as a side effect. Surfaces that just need an authenticated fetch use #3138 tokens (images) or the #3166 server-side path (artifacts).
- Align rotation language with the Auth0 model: revocation/rotation happens on the identity side; when Omni's own factory credential rotates, Omni reconciles the change out to machine configs.

### Out of scope

- Removing credentials from download URLs — #3138 (images) and #3166 (security artifacts).
- The auth backend itself (Auth0 provider rollout, machine-account/API-key semantics) — ops track, siderolabs/image-factory#514.
- **Show-once-at-creation presentation — parked.** Omni has no credential creation flow to hang it on; it returns with the RFD 38 provisioning work.

### Notes

- The credential is already redacted in the Talos machine config; this ticket is about the UI surfaces.
- Related: #3138 (signed download URLs), #3166 (artifacts through Omni), #3248 (consolidated the current embedded-credential path into `useResolvedFactory` — the code the rename lands in), siderolabs/image-factory#514 (tokens in the Basic password field).

Slessi:
When using enterprise image factory, the commands in the machine quick start do not currently work, and need changes in other areas for them to work. Will hide them from frontend until support is added. With regards to the installation media talosctl cluster create command in particular, it won't work at all regardless of enterprise or not currently.

Unrelated: Update scrollbar styling with cross-browser versions added in tailwindcss 4.3

Closes #3139

### DIFF
--- a/frontend/src/views/InstallationMedia/Confirmation.vue
+++ b/frontend/src/views/InstallationMedia/Confirmation.vue
@@ -5,7 +5,6 @@ Use of this software is governed by the Business Source License
 included in the LICENSE file.
 -->
 <script setup lang="ts">
-import { gte } from 'semver'
 import { computed } from 'vue'
 
 import { Runtime } from '@/api/common/omni.pb'
@@ -75,9 +74,7 @@ const { data: talosVersion } = useResourceGet<TalosVersionSpec>(() => ({
   },
 }))
 
-const { url: factoryUrl, credentials: imageFactoryAuth } = useResolvedFactory(
-  () => talosVersion.value?.spec.image_factory_url,
-)
+const { url: factoryUrl } = useResolvedFactory(() => talosVersion.value?.spec.image_factory_url)
 
 const isEnterpriseFactory = computed(() => talosVersion.value?.spec.is_enterprise)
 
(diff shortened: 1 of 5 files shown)


frontend/src/views/InstallationMedia/Confirmation.vue
```diff
@@ -154,38 +151,6 @@ const installerImage = computed(() => {
...
-
-const clusterCreateCommand = computed(() => {
-  if (!factoryUrl.value) return
-
-  const parts = [
-    'talosctl cluster create qemu',
-    `--image-factory-url=${factoryUrl.value}`,
-    `--schematic-id=${schematicId.value}`,
-    `--talos-version=v${resolvedTalosVersion.value}`,
-  ]
```
Unix4ever, line 178: Well, for the community factory this one should work. Unless I missed something.

Slessi: I will try this out locally but my understanding is that we need `--skip-injecting-config` like in `create dev` to prevent Talos creating its own k8s cluster, and `create qemu` does not support that.

Slessi: The `talosctl` command hangs on applying config, eventually timing out with:
```
applying configuration to the cluster nodes
3 error(s) occurred:
        rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing: dial tcp 10.5.0.2:50000: connect: connection refused"
        rpc error: code = DeadlineExceeded desc = context deadline exceeded while waiting for connections to become ready
        timeout
```

But the machines do show up in Omni and if you want for the create command to timeout you can create a cluster with them. Machines don't start installing until command times out though.

IDK, UX looks bad for me here. I don't think we should have the command if its in this state. I can add it back for community if we think it is still valuable.

Unix4ever: Ok, lets remove it for now

Unix4ever (approved):

PR #3315, opened 2026-08-25
feat!: remove deprecated factory params from omni config

Slessi:
Remove the deprecated flat config image factory parameters from Omni config. Users should now use exclusively the `registries.factories.primary` config.

Related:
- https://github.com/siderolabs/omni-controller/pull/422

### DIFF
--- a/cmd/omni/cmd/cmd.go
+++ b/cmd/omni/cmd/cmd.go
@@ -341,17 +341,6 @@ func defineRegistriesFlags(b *FlagBinder, flagConfig *config.Params) {
 	b.StringVar("registries.talos", &flagConfig.Registries.Talos)
 	b.StringVar("registries.kubernetes", &flagConfig.Registries.Kubernetes)
 
-	// Deprecated image factory flags, superseded by the --primary-factory-* flags below.
-	// These intentionally target the deprecated fields to preserve backward compatibility.
-	//nolint:staticcheck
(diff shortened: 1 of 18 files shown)


cmd/omni/cmd/cmd.go
```diff
@@ -341,17 +341,6 @@ func defineRegistriesFlags(b *FlagBinder, flagConfig *config.Params) {
 	b.StringVar("registries.talos", &flagConfig.Registries.Talos)
 	b.StringVar("registries.kubernetes", &flagConfig.Registries.Kubernetes)
 
-	// Deprecated image factory flags, superseded by the --primary-factory-* flags below.
-	// These intentionally target the deprecated fields to preserve backward compatibility.
-	//nolint:staticcheck
-	b.deprecatedStringAlias("image-factory-address", "use --primary-factory-url", &flagConfig.Registries.ImageFactoryBaseURL)
```
Unix4ever, line 347: I would maybe keep this one for a little longer. Maybe one more release.

Slessi: Mmmm it might get messy with the new params I have to add but ok

oguzkilcan: Why do we want to keep it longer?

Unix4ever: I think it was deprecated in the previous release, that's why I thought we might want to wait a bit more to give more time for people to react on the change.

PR #3317, opened 2026-08-26
feat: add access tokens to omni to IF auth

Slessi:
- Add new `auth0` config to factories
- Update image factory auth resource
- Add new controller to reconcile tokens as needed, and also the username/password auth

### DIFF
--- a/client/api/omni/specs/omni.proto
+++ b/client/api/omni/specs/omni.proto
@@ -1887,9 +1887,40 @@ message MachineConfigExtractionStatusSpec {
   string error = 2;
 }
 
+// ImageFactoryAuthSpec holds the credentials Omni uses to authenticate against a single image factory.
+//
+// The resource ID is the factory's base URL, stripped of any trailing slash.
 message ImageFactoryAuthSpec {
   string username = 1;
   string password = 2;
+  // Token is the access token Omni presents to the factory when that factory is
+  // configured with OAuth2 client credentials.
+  //
+  // A factory accepts one authentication mechanism, not both, so exactly one of this and
+  // username/password is set. Omni's configuration rejects a factory that specifies both.
+  AccessToken token = 3;
+
+  // AccessToken is an access token obtained through the OAuth2 client
+  // credentials grant.
+  message AccessToken {
+    // Token is the token to send to the factory in the Authorization header.
+    string token = 1;
+    // TokenType is the token type the issuer returned, normally "Bearer".
+    string token_type = 2;
+    // IssuedAt is when Omni obtained the token. Together with ExpiresAt it gives the token's
+    // lifetime, which is what the rotation schedule is derived from.
+    google.protobuf.Timestamp issued_at = 3;
+    // ExpiresAt is when the token stops being accepted by the factory.
+    google.protobuf.Timestamp expires_at = 4;
+    // ClientId is the OAuth2 client ID the token was issued to.
+    //
+    // It is recorded so that reconfiguring the factory's credentials forces a new token instead of
+    // leaving the one issued to the previous client in place until it expires on its own.
+    string client_id = 5;
+    // Audience is the OAuth2 API identifier the token was issued for, recorded for the same reason
+    // as ClientId.
+    string audience = 6;
+  }
 }
 
 // MachineInstallDiskConfigSpec describes the user's install disk selection for a machine.
--- a/internal/backend/runtime/omni/controllers/omni/imagefactory/auth.go
+++ b/internal/backend/runtime/omni/controllers/omni/imagefactory/auth.go
@@ -1,0 +1,321 @@
+// Copyright (c) 2026 Sidero Labs, Inc.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+
+package imagefactory
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"slices"
+	"time"
+
+	"github.com/cosi-project/runtime/pkg/controller"
+	"github.com/cosi-project/runtime/pkg/safe"
+	"github.com/cosi-project/runtime/pkg/state"
+	"go.uber.org/zap"
+	"google.golang.org/protobuf/types/known/timestamppb"
+
+	"github.com/siderolabs/omni/client/api/omni/specs"
+	"github.com/siderolabs/omni/client/pkg/omni/resources/omni"
+	"github.com/siderolabs/omni/internal/backend/runtime/omni/controllers/helpers"
+	"github.com/siderolabs/omni/internal/pkg/config"
+	"github.com/siderolabs/omni/internal/pkg/imagefactory/tokens"
+)
+
+// AuthControllerName is the name of AuthController.
+//
+// It is exported because the migration that adopts the resources previously written by the Omni
+// startup path has to stamp this exact owner on them.
+const AuthControllerName = "ImageFactoryAuthController"
+
+const (
+	// DefaultAuthRetryInterval is how long AuthController waits before
+	// retrying work it could not finish: a token request that failed, or a teardown still held up by
+	// another controller's finalizer.
+	//
+	// A failed token request is not urgent: tokens are replaced halfway through their lifetime, so
+	// the token in state is still valid for about as long as it has already lived.
+	DefaultAuthRetryInterval = 30 * time.Second
+)
+
+// AuthController maintains the credentials Omni uses to authenticate against the
+// configured image factories: the basic auth taken straight from the configuration, and the
+// access tokens issued and rotated against the OAuth2 client.
+//
+// The set of factories comes from Omni's configuration rather than from a resource, so the
+// controller has no inputs: it reconciles on its own schedule, derived from the lifetimes of the
+// tokens it has issued.
+type AuthController struct {
+	issuers   *tokens.Issuers
+	factories []config.Factory
+}
+
+// NewAuthController creates a new AuthController.
+func NewAuthController(registries *config.Registries, issuers *tokens.Issuers) *AuthController {
+	return &AuthController{
+		issuers:   issuers,
+		factories: registries.AllFactories(),
+	}
+}
+
+// Name implements controller.Controller interface.
+func (ctrl *AuthController) Name() string {
+	return AuthControllerName
+}
+
+// Inputs implements controller.Controller interface.
+func (ctrl *AuthController) Inputs() []controller.Input {
+	return nil
+}
+
+// Outputs implements controller.Controller interface.
+func (ctrl *AuthController) Outputs() []controller.Output {
+	return []controller.Output{
+		{
+			Type: omni.ImageFactoryAuthType,
+			Kind: controller.OutputExclusive,
+		},
+	}
+}
+
+// Run implements controller.Controller interface.
+func (ctrl *AuthController) Run(ctx context.Context, r controller.Runtime, logger *zap.Logger) error {
+	timer := time.NewTimer(0)
+	defer timer.Stop()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return nil
+		case <-r.EventCh():
+		case <-timer.C:
+		}
+
+		sleep, err := ctrl.reconcile(ctx, r, logger)
+		if err != nil {
+			// A failing token endpoint must not take the controller down: the credentials already in
+			// state stay usable, and the next attempt may well succeed.
+			logger.Error("failed to reconcile image factory auth", zap.Error(err))
+
+			return err
+		}
+
+		if sleep <= 0 {
+			// Nothing on record expires and nothing is left unfinished. The factories come from
+			// static configuration, so there is nothing left to wake up for: leaving the timer unset
+			// parks the loop on the select above until the runtime shuts down. An installation with
+			// no OAuth2 configured factory reconciles exactly once.
+			logger.Debug("image factory auth is fully reconciled, no further work is scheduled")
+
+			continue
+		}
+
+		timer.Reset(sleep)
+	}
+}
+
+// reconcile brings the ImageFactoryAuth resources in line with the configured factories, and returns
+// how long to wait before the next reconciliation.
+func (ctrl *AuthController) reconcile(ctx context.Context, r controller.Runtime, logger *zap.Logger) (time.Duration, error) {
+	authenticated := make([]string, 0, len(ctrl.factories))
+
+	// A zero sleep means nothing needs scheduling; only a token that expires, or work left
+	// unfinished, puts a wake-up on the clock.
+	var (
+		sleep time.Duration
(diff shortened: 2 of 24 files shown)


client/api/omni/specs/omni.proto
```diff
@@ -1889,6 +1889,28 @@ message ImageFactoryAuthSpec {
   string password = 2;
 }
 
+// ImageFactoryTokenSpec holds the machine-to-machine access token Omni presents to an image factory.
+//
+// The resource ID is the factory's base URL, stripped of any trailing slash, matching ImageFactoryAuth.
+message ImageFactoryTokenSpec {
```
utkuozdemir: Talked with Oguz, and maybe this should indeed be part of ImageFactoryAuth resource. As a dedicated `Token token` field maybe. They both are about the same concern - image factory auth. Same ACLs and rules would apply to them.

Slessi: Yep just draft still, gonna look into that

internal/backend/runtime/omni/controllers/omni/image_factory_token.go
```diff
@@ -0,0 +1,239 @@
...
+}
+
+// reconcile brings the ImageFactoryToken resources in line with the configured factories, and
+// returns how long to wait before the next reconciliation.
+func (ctrl *ImageFactoryTokenController) reconcile(ctx context.Context, r controller.Runtime, logger *zap.Logger) (time.Duration, error) {
+	factoryURLs := ctrl.issuers.FactoryURLs()
+
+	if err := ctrl.pruneTokens(ctx, r, factoryURLs); err != nil {
+		return 0, err
+	}
```
oguzkilcan: Failing to prune shouldn't prevent other tokens from being refreshed, it's important yes but shouldn't be a blocker.

internal/backend/runtime/omni/controllers/omni/image_factory_token_test.go
```diff
@@ -0,0 +1,232 @@
...
+)
+
+const (
+	testFactoryURL          = "https://factory.example.com"
+	testFactoryClientID     = "client-id"
+	testFactoryAudience     = "https://image-factory.example.com"
+	testFactoryTokenTimeout = 10 * time.Second
+
+	// imageFactoryTokenControllerName is the owner stamped on the resources the controller writes.
+	imageFactoryTokenControllerName = "ImageFactoryTokenController"
```
oguzkilcan: nit: you can export the controller name to a constant on the controller and use that in the test

internal/backend/runtime/omni/controllers/omni/image_factory_token.go
```diff
@@ -0,0 +1,239 @@
...
+
+	sleep := imageFactoryTokenMaxSleep
+
+	var errs []error
+
+	for _, factoryURL := range factoryURLs {
```
oguzkilcan: If we don't have any factory that was configured with auth0, this controller will still reconcile every hour. Yes, it's cheap but it's also unnecessary.

internal/backend/runtime/omni/controllers/omni/image_factory_token.go
```diff
@@ -0,0 +1,239 @@
...
+	return sleep, nil
+}
+
+// reconcileFactory makes sure the factory has a usable token, and returns how long that token stays
+// usable before it should be replaced.
+func (ctrl *ImageFactoryTokenController) reconcileFactory(ctx context.Context, r controller.Runtime, logger *zap.Logger, factoryURL string) (time.Duration, error) {
+	issuer := ctrl.issuers.ForURL(factoryURL)
+	if issuer == nil {
+		return 0, errors.New("no token issuer is configured")
+	}
```
oguzkilcan: nit: maybe mention the factory url here so its clear in the logs

internal/backend/runtime/omni/controllers/omni/image_factory_auth.go
```diff
@@ -0,0 +1,325 @@
...
+	if err != nil {
+		return err
+	}
+
+	for auth := range existing.All() {
+		if slices.Contains(authenticated, auth.Metadata().ID()) {
+			continue
+		}
+
+		if err = r.Destroy(ctx, auth.Metadata()); err != nil && !state.IsNotFoundError(err) {
```
oguzkilcan: This might be risky, not sure if we can destroy this resource directly if it has finalizers.

internal/backend/runtime/omni/controllers/omni/image_factory_auth_test.go
```diff
@@ -0,0 +1,311 @@
...
+	if username != "" {
+		primary.SetUsername(username)
+		primary.SetPassword(password)
+	}
+
+	return &config.Registries{Factories: config.Factories{Primary: primary}}
+}
+
+type ImageFactoryAuthSuite struct {
+	OmniSuite
```
oguzkilcan: You are using `testify.Suite` for the controller tests, I'd recommend using `testutils.WithRuntime` instead.

client/api/omni/specs/omni.proto
```diff
@@ -1884,9 +1884,40 @@ message MachineConfigExtractionStatusSpec {
...
+  // username/password is set. Omni's configuration rejects a factory that specifies both.
+  AccessToken token = 3;
+
+  // AccessToken is an access token obtained through the OAuth2 client
+  // credentials grant.
+  message AccessToken {
+    // Token is the token to send to the factory in the Authorization header.
+    string token = 1;
+    // TokenType is the token type the issuer returned, normally "Bearer".
+    string token_type = 2;
```
Unix4ever, line 1909: Maybe we could use enum instead of a string here? Or there's a specific case for using string?

Slessi: IDK what all the possible values could be, its gonna be whatever the token issuer returns

internal/backend/runtime/omni/controllers/omni/image_factory_auth.go
```diff
@@ -0,0 +1,343 @@
...
+			return nil
+		case <-r.EventCh():
+		case <-timer.C:
+		}
+
+		sleep, err := ctrl.reconcile(ctx, r, logger)
+		if err != nil {
+			if ctx.Err() != nil {
+				return nil
+			}
```
utkuozdemir: I think we should drop this. The context can only be canceled when Omni receives a sigterm, and in that case, everything is shutting down anyway.

The only benefit of this would be not propagating a `context.Canceled` error to upstream (and potentially Omni terminating with a zero exit code than a non-zero) - but even in that case, handling context.Canceled should be done at a higher level (not sure if appliable, and even if it is, is a separate concern from this PR)

utkuozdemir: And when you actually need this logic, I'd suggest `if errors.Is(err, context.Canceled)` instead.

Slessi: Sure, can drop. Didn't use `errors.Is` in case it got wrapped somewhere along the way in a way that loses the `context.Canceled` - but maybe not possible

internal/backend/runtime/omni/controllers/omni/imagefactory/auth.go
utkuozdemir, line 1: We try to add all new controllers to their respective packages instead of `omni` package - can you move it? You can create a new `imagefactory` package, or find one that already fits (there's the `image` but maybe `imagefactory` would be a better package).

internal/backend/runtime/omni/controllers/omni/image_factory_auth.go
```diff
@@ -0,0 +1,329 @@
...
+	return min(current, candidate)
+}
+
+// tokenRefreshDelay returns how long the token stays in use: half of its lifetime, which leaves the
+// remaining half as the window to recover from a failing token endpoint. A non-positive result means
+// the token is due to be replaced now, as is a token whose lifetime was not fully recorded.
+func tokenRefreshDelay(token *specs.ImageFactoryAuthSpec_AccessToken) time.Duration {
+	if token.GetToken() == "" || token.GetIssuedAt() == nil || token.GetExpiresAt() == nil {
+		return 0
+	}
```
utkuozdemir: If this is a condition which should not happen, I think we should simply return an error here.

internal/backend/runtime/omni/controllers/omni/image_factory_auth.go
```diff
@@ -0,0 +1,329 @@
...
+			allRemoved = false
+		}
+	}
+
+	return allRemoved, errors.Join(errs...)
+}
+
+// soonest folds a candidate wake-up into the earliest one scheduled so far, where a zero value on
+// either side means "nothing scheduled".
+func soonest(current, candidate time.Duration) time.Duration {
```
utkuozdemir: Maybe this and `tokenRefreshDelay` should also be methods, to have a narrower scope for them?

internal/backend/runtime/omni/controllers/omni/image_factory_auth.go
```diff
@@ -0,0 +1,329 @@
...
+			// Keep the token already on record if issuing a new one fails: it may still be valid,
+			// and replacing a usable token with nothing would be strictly worse.
+			issued, issueErr := issueToken(ctx, logger, issuer, factoryURL)
+			if issueErr != nil {
+				tokenErr = issueErr
+			} else {
+				token = issued
+				refreshIn = tokenRefreshDelay(issued)
+			}
+		}
```
utkuozdemir: Not sure how, but I feel like this code can be simplified (extracting to a method and using early returns maybe? The whole `case` can be extracted, maybe).

internal/backend/runtime/omni/controllers/omni/image_factory_auth.go
```diff
@@ -0,0 +1,329 @@
...
+		return nil
+	}); err != nil {
+		return 0, err
+	}
+
+	return refreshIn, tokenErr
+}
+
+// issueToken obtains a fresh token from the issuer.
+func issueToken(ctx context.Context, logger *zap.Logger, issuer tokens.Issuer, factoryURL string) (*specs.ImageFactoryAuthSpec_AccessToken, error) {
```
utkuozdemir: Can also be a method

internal/backend/runtime/omni/controllers/omni/image_factory_auth.go
```diff
@@ -0,0 +1,343 @@
...
+			issued, issueErr := issueToken(ctx, logger, issuer, factoryURL)
+			if issueErr != nil {
+				tokenErr = issueErr
+			} else {
+				token = issued
+				refreshIn = tokenRefreshDelay(issued)
+			}
+		}
+
+		if refreshIn <= 0 && tokenErr == nil {
```
oguzkilcan: I don't think this `if block` is reachable. `tokenRefreshDelay` may return `0` if issued token is invalid but that shouldn't be the case if the token was issued successfully. 
So either `refreshIn` is 0 or less or token doesn't come from current issuer and we attempt to issue a token:
 -> on success: set token and new refresh date - no error was returned so **we can assume** token is healthy therefore refresh date is in the future: (refreshIn <= 0) = false
-> on failure: set tokenErr - now tokenErr is not nil and refreshIn is still <= 0: (tokenErr == nil) = false
Or refreshIn is positive and token is coming from current issuer, so we don't attempt to issue a new token: (refreshIn <= 0) = false

On both cases this `if statement` is never true.

internal/backend/runtime/omni/controllers/omni/image_factory_auth.go
Slessi: it can happen, i will try rewrite for clarity

internal/backend/runtime/omni/controllers/omni/image_factory_auth.go
Slessi: methods on `(ctrl *ImageFactoryAuthController)`? I could, but I don't see what that really changes

oguzkilcan: No need to expose them on package level.

internal/backend/runtime/omni/controllers/omni/image_factory_auth.go
Slessi: Yea, was checking something else originally and during refactoring became useless. Removed.

internal/backend/runtime/omni/controllers/omni/image_factory_auth.go
Slessi: Made it a method

internal/backend/runtime/omni/controllers/omni/image_factory_auth.go
Slessi: Made it a method

internal/backend/runtime/omni/controllers/omni/image_factory_auth.go
Slessi: Inlined it, which makes it more clear now

internal/backend/runtime/omni/controllers/omni/imagefactory/auth.go
Slessi: Moved

internal/backend/runtime/omni/controllers/omni/image_factory_auth.go
Slessi: I changed things a bit here see if you are ok with how it is? If not I'll extract

utkuozdemir: it's just a nit, your call in the end. and yes it looks better

internal/backend/runtime/omni/controllers/omni/imagefactory/auth.go
```diff
@@ -0,0 +1,321 @@
...
+		// error never reaches here.
+		destroyed, err := helpers.TeardownAndDestroy(ctx, r, auth.Metadata())
+		if err != nil {
+			// One factory's credentials failing to go must not strand the rest: a single resource
+			// the controller cannot touch — one the ownership migration missed, say — would
+			// otherwise wedge every removal behind it for good.
+			errs = append(errs, fmt.Errorf("failed to remove the credentials of image factory %q: %w", auth.Metadata().ID(), err))
+		}
+
+		if err != nil || !destroyed {
```
utkuozdemir, line 301: Nit: err check is not needed here I think

oguzkilcan (approved):

internal/backend/runtime/omni/controllers/omni/imagefactory/auth.go
Slessi: either its needed here or i have to duplicate `allRemoved = false` in the previous err check

PR #3322, opened 2026-08-26
fix: record the image factory host for machines already at target

oguzkilcan:
A machine that reaches its target version without an upgrade records the version and the schematic but not the image factory host, so the field stays empty for good while every other path that records an installed image writes all three together.


Unix4ever (approved):
Good catch 👍

PR #3329, opened 2026-08-27
fix: apply pending updates to unhealthy machines so they can recover

### LINKED ISSUE #3262: Machine cannot recover from a kernel args or extensions change that makes it unhealthy
When a kernel args or extensions update (both result in a new schematic) breaks a machine, reverting the change has no effect. The machine stays broken, and without console access to the machine, the practical way out is to remove it and provision it again.

Observed on Omni 1.10 with a Talos 1.14-beta.1 machine:

1. On a machine in a cluster, update the kernel args in a way that breaks the node. For example, add `nosmt` on a machine where kubelet is configured with `reservedSystemCPUs` referencing cores that disappear with SMT off. Omni upgrades the machine to the new schematic and reboots it.
2. Kubelet fails validation and crashloops:
   ```
   failed to run Kubelet: reserved-cpus: 0,2,16,18 is not a subset of online-cpus: 0-15
   ```
3. The machine stays in the booting stage, connected to Omni but not ready.
4. Revert the kernel args change.
5. Nothing happens. `KernelArgsStatus` keeps showing the diff between the intended and the current args, with no unmet conditions and no hint about why the change is not applied.

Omni should apply the reverted args and reboot the machine instead. Manual warm and cold reboots do not help either, since the args are baked into the installed UKI, so every reboot comes back with the same cmdline.

Schematic changes are applied through the Talos upgrade flow, and that flow refuses to touch machines that are not healthy, on two independent levels:

- On the machine level, a pending install/upgrade on the Talos 1.13+ lifecycle path runs only when the machine stage is `RUNNING`: https://github.com/siderolabs/omni/blob/71b3a9e4/internal/backend/runtime/omni/controllers/omni/machineconfig/status.go#L725-L727. A machine stuck in the booting stage never qualifies, and this is what blocks the case above. For comparison, the config apply path in the same controller accepts booting machines (https://github.com/siderolabs/omni/blob/71b3a9e4/internal/backend/runtime/omni/controllers/omni/machineconfig/status.go#L972-L973), and so does the legacy upgrade path taken when either side is older than Talos 1.13 (https://github.com/siderolabs/omni/blob/71b3a9e4/internal/backend/runtime/omni/controllers/omni/machineconfig/status.go#L528-L535).
- On the cluster level, we subtract the number of not-ready machines in the whole cluster from the upgrade quota of every machine set, unless the upgrade phase is `Reverting`: https://github.com/siderolabs/omni/blob/71b3a9e4/internal/backend/runtime/omni/controllers/omni/talosupgrade/status.go#L375-L378. With the default parallelism of 1, a single not-ready machine drives the quota of all machine sets in the cluster to 0. In the case above this gate is not even reached, because the machine still holds its upgrade lock from the failed update and skips the quota check. But it blocks the same recovery whenever the machine does not hold the lock, e.g., when the machine went unhealthy for an unrelated reason and the fixing schematic change comes later. It also blocks schematic updates for every other machine in the cluster, including machines that are running but not ready, which would pass the machine-level gate.

The Talos version revert flow already has an exception for the cluster-level gate: going back to the last successful version puts the upgrade into the `Reverting` phase, which skips the health checks and the not-ready subtraction (https://github.com/siderolabs/omni/blob/71b3a9e4/internal/backend/runtime/omni/controllers/omni/talosupgrade/status.go#L344-L351). But there is no such exception for the machine-level gate, so even a version revert is blocked when the machine is stuck in the booting stage and both versions are 1.13+ (https://github.com/siderolabs/omni/blob/71b3a9e4/internal/backend/talos/lifecycle/decide.go#L42-L62). The schematic flow has no exception on either level.

Config changes are better off: a pending config change can be corrected while the machine is booting. However, a pending schematic change blocks the config path too, because the lifecycle operation is reconciled before the config apply (https://github.com/siderolabs/omni/blob/71b3a9e4/internal/backend/runtime/omni/controllers/omni/machineconfig/status.go#L431-L435). So once the revert is in place, a corrective config change (e.g., fixing the kubelet CPU config) does not reach the machine either.

Proposed direction: a machine's own unhealthiness should not gate a pending update to that same machine, since the update is the recovery path.

- On the machine level, allow the lifecycle operation on a booting machine, the same way config apply and the legacy path already do. Open question: this path cordons and drains the node before the reboot, we need to check how that behaves when the node is broken.
- On the cluster level, exclude the machine being updated from the not-ready subtraction.

This covers both the schematic case and the version revert gap with the same rule, without introducing a new revert concept for schematics.

Additionally, we could expose the Talos `Rollback` API (equivalent to picking the previous installation in the bootloader A/B menu) as an explicit recovery action in Omni. This gives a way out without console access even when the pending update cannot run at all. Most pieces are already in place: the Talos API proxy already routes `Rollback` to the machine, but stamps it with the `os:operator` role (https://github.com/siderolabs/omni/blob/71b3a9e4/internal/backend/grpc/router/talos_backend.go#L48-L59), while Talos requires `os:admin` for it (https://github.com/siderolabs/talos/blob/v1.14.0-beta.1/internal/app/machined/pkg/system/services/machined.go#L104). Note that no tests reference this API on the Talos side atm.

utkuozdemir:
When a kernel args or extensions change broke a machine, reverting it had no effect: updates run through the upgrade flow, which skipped machines that are not healthy. The machine could not become healthy without the update, and the update never ran while it was unhealthy, so the only way out was the bootloader console or a reprovision.

A machine's own unhealthiness no longer blocks its own pending update: a booting machine is now eligible for the upgrade like it already was for config apply, a not-ready machine is guaranteed one upgrade slot, and the cordon and drain of a recovery are best-effort. Healthy machines still wait while the cluster is degraded, broken machines recover one at a time, and Talos version reverts on booting machines are unblocked the same way.

On machines older than Talos 1.13, the regular upgrade drains the node first, and a broken machine may have no working Kubernetes API to drain through. The upgrade then failed before installing anything, and the machine rebooted into the same broken image over and over. Recovery on these machines now uses a staged upgrade followed by an explicit reboot, which skips the drain and applies the new image on the next boot.

Closes siderolabs/omni#3262.

### DIFF
--- a/internal/backend/talos/lifecycle/manager.go
+++ b/internal/backend/talos/lifecycle/manager.go
@@ -89,10 +89,11 @@ func (m *Manager) clientset(ctx context.Context, cluster string) (k8s.Interface,
 // PreRebootHook runs before the node is drained. An error aborts the Run.
 type PreRebootHook func(ctx context.Context, talosClient *talosclient.Client) error
 
-// nodeRef names a cluster node.
+// nodeRef names a cluster node. With bestEffort, cordon/drain failures do not abort the operation.
 type nodeRef struct {
 	clusterName string
 	nodeName    string
+	bestEffort  bool
 }
 
 // runConfig holds the per-call options resolved from Option values.
@@ -108,10 +109,18 @@ type runConfig struct {
 type Option func(*runConfig)
 
 // WithCordonDrain makes Run cordon and drain the node before reboot. Set for cluster members.
-func WithCordonDrain(clusterName, nodeName string) Option {
+//
+// With bestEffort, cordon and drain failures do not abort the operation. Set it when the operation
+// is the recovery path of a machine that is itself unhealthy: nothing may block its reboot then,
+// and the workload Kubernetes API may be served by the very machine being recovered. The drain is
+// still attempted, because an unhealthy machine does not imply an unhealthy node: a machine can
+// stay Talos-not-ready (e.g. an extension service missing its configuration) while its node keeps
+// serving pods, and those deserve an eviction before the reboot. On a node that cannot finish
+// evictions the attempt costs at most the drain timeout, acceptable for a recovery.
+func WithCordonDrain(clusterName, nodeName string, bestEffort bool) Option {
 	return func(cfg *runConfig) {
 		if clusterName != "" && nodeName != "" {
-			cfg.cordonDrain = &nodeRef{clusterName: clusterName, nodeName: nodeName}
+			cfg.cordonDrain = &nodeRef{clusterName: clusterName, nodeName: nodeName, bestEffort: bestEffort}
 		}
 	}
 }
@@ -283,7 +292,15 @@ func (m *Manager) runCordonAndDrain(ctx context.Context, cfg runConfig) error {
 
 	clientset, err := m.clientset(ctx, target.clusterName)
 	if err != nil {
-		return WrapErr(err, "failed to get kubernetes client to cordon/drain node")
+		err = WrapErr(err, "failed to get kubernetes client to cordon/drain node")
+
+		if target.bestEffort {
+			emitf(cfg.progress, "[omni] skipping cordon and drain of node %s: %s", target.nodeName, err)
+
+			return nil
+		}
+
+		return err
 	}
 
 	cordonCtx, cancel := context.WithTimeout(ctx, CordonTimeout)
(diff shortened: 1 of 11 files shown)


internal/backend/talos/lifecycle/manager.go
```diff
@@ -292,19 +309,38 @@ func (m *Manager) runCordonAndDrain(ctx context.Context, cfg runConfig) error {
...
+
+			return nil
+		}
+
 		return err
 	}
 
-	emitf(cfg.progress, "[omni] draining node %s", target.nodeName)
+	// See WithCordonDrain for why a best-effort operation tolerates a failed drain.
+	if target.bestEffort {
```
oguzkilcan: Nit: Why not implement this like above? No need to call drain two times in that case.

oguzkilcan (approved):

PR #3333, opened 2026-08-28
feat: use tokens for authenticating with factory

Slessi:
Use the stored access tokens for authenticating Omni with factory. Also noticed secondary factory was missing from CSP so added it there.

### DIFF
--- a/client/pkg/imagefactory/client.go
+++ b/client/pkg/imagefactory/client.go
@@ -75,7 +75,7 @@ type Client struct {
 // The base URL is canonicalized by stripping any trailing slash, so that the URL reported by
 // [Client.URL] can be compared to a factory URL from any other source (a configured factory, a
 // TalosVersion resource, a client request) without each comparison having to normalize first.
-func NewClient(imageFactoryBaseURL, username, password string) (*Client, error) {
+func NewClient(imageFactoryBaseURL, username, password string, extraOpts ...client.Option) (*Client, error) {
 	imageFactoryBaseURL = normalizeFactoryURL(imageFactoryBaseURL)
 
 	sniffer := &serverSnifferTransport{wrapped: http.DefaultTransport}
@@ -87,6 +87,8 @@ func NewClient(imageFactoryBaseURL, username, password string) (*Client, error) 
 	if username != "" && password != "" {
 		clientOptions = append(clientOptions, client.WithBasicAuth(username, password))
 	}
+
+	clientOptions = append(clientOptions, extraOpts...)
 
 	factoryClient, err := client.New(imageFactoryBaseURL, clientOptions...)
 	if err != nil {
--- a/internal/backend/runtime/omni/controllers/omni/internal/imagefactoryauth/imagefactoryauth.go
+++ b/internal/backend/runtime/omni/controllers/omni/internal/imagefactoryauth/imagefactoryauth.go
@@ -17,10 +17,11 @@ import (
 )
 
 func buildDoc(auth *omni.ImageFactoryAuth) (*cri.RegistryAuthConfigV1Alpha1, error) {
+	token := auth.TypedSpec().Value.GetToken().GetToken()
 	username := auth.TypedSpec().Value.Username
 	password := auth.TypedSpec().Value.Password
 
-	if username == "" || password == "" {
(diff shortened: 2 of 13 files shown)


internal/backend/runtime/omni/controllers/omni/internal/imagefactoryauth/imagefactoryauth.go
```diff
@@ -30,8 +31,13 @@ func buildDoc(auth *omni.ImageFactoryAuth) (*cri.RegistryAuthConfigV1Alpha1, err
...
 	doc := cri.NewRegistryAuthConfigV1Alpha1(u.Host)
-	doc.RegistryUsername = username
-	doc.RegistryPassword = password
+	if token != "" {
+		doc.RegistryUsername = "token"
+		doc.RegistryPassword = token
+	} else {
+		doc.RegistryUsername = username
+		doc.RegistryPassword = password
+	}
```
Slessi, line 40: This piece i am not sure about, I guess it is where the node tokens should go, but I thought maybe useful as a temporary stopgap for testing?

client/pkg/imagefactory/client.go
```diff
@@ -75,7 +75,7 @@ type Client struct {
 // The base URL is canonicalized by stripping any trailing slash, so that the URL reported by
 // [Client.URL] can be compared to a factory URL from any other source (a configured factory, a
 // TalosVersion resource, a client request) without each comparison having to normalize first.
-func NewClient(imageFactoryBaseURL, username, password string) (*Client, error) {
+func NewClient(imageFactoryBaseURL, username, password string, extraOpts ...client.Option) (*Client, error) {
```
oguzkilcan, line 78: Shouldn't we also move username password into the extraOpts?

Slessi: I thought about it, feels cleaner to me too but would have to duplicate the `!=""` checks everywhere though, WDYT?

internal/backend/runtime/omni/controllers/omni/internal/imagefactoryauth/imagefactoryauth_test.go
```diff
@@ -92,4 +101,40 @@ func TestBuildDocs(t *testing.T) {
...
+		docs, err := imagefactoryauth.BuildDocs(creds)
+		require.NoError(t, err)
+		require.Len(t, docs, 1)
+
+		assert.Equal(t, "factory.example.org", docs[0].Name())
+		assert.Equal(t, "token", docs[0].Username())
+		assert.Equal(t, "the-token", docs[0].Password())
+	})
+
+	t.Run("token takes precedence over username/password when both are set", func(t *testing.T) {
```
oguzkilcan: I don't think both shouldn't be set ever together.

Slessi: yea, removed, needless test

client/pkg/imagefactory/client.go
oguzkilcan: What if you check it in the beginning and if both are not present don't set the option?

Slessi: inside `WithBasicAuth`? Would have to be a factory change, but its possible

utkuozdemir: This is why I don't like the functional options pattern. IMO, should have been a config / options struct in the first place. One example: https://github.com/siderolabs/omni-infra-provider-bare-metal/blob/main/internal/qemu/machines.go#L57

But this is a public package, moving username/password would break older clients.

An approach could be a new constructor:
```go
func NewClientWithOptions(imageFactoryBaseURL string, options ClientOptions) {
...
}
```

Define the `ClientOptions` properly (_make zero value useful_), update the call sites, make `NewClient` call `NewClientWithOptions` with the right options. That's probably what I would do.

utkuozdemir: You can also see this exact transition in many Go libraries out there.

utkuozdemir:
We are not going with this approach, closing this for hygiene, feel free to reopen if you think some parts of it should land.

PR #3334, opened 2026-08-28
feat: download installation media through the boot asset API

oguzkilcan:
omnictl assembled the image factory URL itself and read the factory's credentials out of Omni to authenticate the download, so every operator who downloaded an image ended up holding them. It now asks Omni for the URL and for whatever the fetch has to carry, the same way an infrastructure provider does, so the download carries a short-lived token instead of the credentials..

Related: #3138

### DIFF
--- a/client/pkg/omnictl/internal/download/export_test.go
+++ b/client/pkg/omnictl/internal/download/export_test.go
@@ -1,0 +1,13 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+package download
+
+import "github.com/siderolabs/omni/client/pkg/imagefactory"
+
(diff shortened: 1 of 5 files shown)


client/pkg/omnictl/internal/download/export_test.go
```diff
@@ -0,0 +1,13 @@
...
+
+package download
+
+import "github.com/siderolabs/omni/client/pkg/imagefactory"
+
+// Exported for testing: the mapping from an image and its download parameters onto the boot asset the
+// image factory serves, so it can be checked without an Omni to answer the RPC.
+func AssetSpec(image ImageInfo, params Params) imagefactory.AssetSpec {
+	return assetSpec(image, params)
+}
```
utkuozdemir: This in an internal, narrow package - I'd say simply make the `assetSpec` function exported.

Unix4ever (approved):

oguzkilcan:
I'll do `BootAssetURL` to `InstallationMediaURL` rename PR first before this.

PR #3337, opened 2026-08-31
release(v1.11.0-beta.0): prepare release

Unix4ever:
This is the official v1.11.0-beta.0 release.


hack/release.toml
```diff
@@ -6,24 +6,70 @@ github_repo = "siderolabs/omni"
...
+lists. Downloaded images keep their proper filename again. Against an Image Factory Enterprise, the machine quick start no longer offers the creation \
+commands that do not work with it.
+"""
+
+[notes.image-factory-oauth2]
+title = "OAuth2 Client Credentials for the Image Factory"
+description = """\
+An Image Factory Enterprise can now be authenticated with an OAuth2 client in addition to a username and password, set under the factory's `oAuth2` \
+section or through the `--primary-factory-oauth2-*` and `--secondary-factory-oauth2-*` flags. The two ways are mutually exclusive per factory.
+"""
```
utkuozdemir: Should this be here? cc @oguzkilcan

hack/release.toml
```diff
@@ -6,24 +6,70 @@ github_repo = "siderolabs/omni"
...
 
-Each URL also carries a fresh token, so it is never the same twice. Infrastructure providers should compare the storage key rather than the URL when \
-deciding whether something has already been fetched, and ask for a lifetime covering any delay.
+[notes.kubernetes-signing-keys]
+title = "Kubernetes Token Signing Keys"
+description = """\
+Admins can now list the keys that sign Kubernetes access tokens and delete them. Deleting one immediately invalidates every kubeconfig and OIDC token \
+it signed, which is what it takes to pull a long-lived service account kubeconfig out of circulation before it expires.
+"""
```
utkuozdemir, line 31: Here we can put a link to https://docs.siderolabs.com/omni/security-and-authentication/revoke-kubernetes-access-tokens - the page wasn't ready back then, but now is.

hack/release.toml
smira: let's not have it imho

hack/release.toml
```diff
@@ -6,24 +6,71 @@ github_repo = "siderolabs/omni"
...
+section or through the `--primary-factory-oauth2-*` and `--secondary-factory-oauth2-*` flags. The two ways are mutually exclusive per factory.
+"""
 
 [notes.infra-provider-installation-media]
 title = "Infrastructure Providers Use the Image Factory Omni Is Configured With"
 description = """\
-Every infrastructure provider reached the image factory itself, so an on-prem or authenticated one had to be configured provider by provider, each \
-with its own copy of the factory's URL layout. A provision step now states which medium it wants through Context.EnsureInstallationMedia, and Omni \
-creates the schematic, picks the factory for that Talos version, works out the filename, and answers with a URL and any headers.
+Infrastructure providers no longer need an image factory configured on their side. They ask Omni for the installation media they want, and get it \
```
smira: this is misleading until providers are updated imho

Unix4ever: We're updating them now.

hack/release.toml
utkuozdemir: I feel like this shouldn't get in, like, we might want to revert the oauth part of [this](https://github.com/siderolabs/omni/pull/3317), in the controller/config, before doing the relase. This is user-facing.

smira (commented):
imho the release notes are badly written, there are not written not from a user perspective - rather describing low-level details.

everything around Image Factory Enterprise is of no use for public release notes, or should be compressed down to a single entry of better support for EIF.

hack/release.toml
smira: but from user perspective, there is nothing that they get with Omni release itself - the providers don't get anything unless they are upgraded, which might be in the release notes of the providers, not Omni

Unix4ever: Yeah, I guess writing it for each provider will make more sense.

Unix4ever:
> imho the release notes are badly written, there are not written not from a user perspective - rather describing low-level details.
> 
> everything around Image Factory Enterprise is of no use for public release notes, or should be compressed down to a single entry of better support for EIF.

Yeah, I think to drop all EIF related stuff for now actually.
I can't really imagine how to write that in a good way as it's not really finished yet.

hack/release.toml
Unix4ever: yeah, I dropped all EIF related stuff for now. It's too low level

Unix4ever:
Ok, I rewrote it a couple of times, I guess it should be better now.

smira (approved):

PR #3345, opened 2026-09-01
fix: don't give a UUID-conflicting machine the live machine's address

### LINKED ISSUE #3341: [bug] duplicate machine UUID resolution kicks out original machine
Need to confirm that, but it looks like in some cases joining a machine with a conflicting uuid might kick out another already connected machine.

Unix4ever:
This PR closes another gap in the conflicting UUID resolution that would make the existing machine disconnect from Omni.

Fixes: https://github.com/siderolabs/omni/issues/3341


internal/pkg/siderolink/tunnel_handoff_repro_test.go
utkuozdemir, line 1: Sorry 🙈

Unix4ever: I merged everything into provision_test.go :)

utkuozdemir (approved):

Issue #817, opened 2025-01-08
[bug] pi memory and cpu machine stats not showing


mdiskin:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

![image](https://github.com/user-attachments/assets/f1868108-a951-4ff6-ae70-917f739528dc)

### Expected Behavior

The monitor tab is able to show cpu and memory realtime data so woul expect the machine status to not show CPU as Not Defected and Memory as 0 Bytes. This also is an issue on the machines view :

Processors - No processors detected
Memory - No memory modules detected

### Steps To Reproduce

talos 1.9.1 image on pi cm4/rk1 modules

### What browsers are you seeing the problem on?

Chrome

Unix4ever:
Talos reads this information from SMBIOS, which is not available on the RPI4 unfortunately.

mdiskin:
Maybe the pi's dont follow this https://developer.arm.com/documentation/102677/0400/SMBIOS-requirements

Would it be worth adding a feature request to maybe set these as part of the image creation

Unix4ever:
SBC support is mostly community driven. We would be happy to accept any contributions if you find the way to fix that.
In any case this issue should be addressed to the https://github.com/siderolabs/sbc-raspberrypi repo.

PR #3348, opened 2026-09-02
feat: use the Linux view of CPU cores and report machines by core count

### LINKED ISSUE #3224: use new CPUCores resource
Talos now provides CPUCore resource which is based on Linux view of things vs. the old Processor which was based only on SMBIOS data.

The new one should be more accurate, and it supports hotplug events (e.g. for the VM).

utkuozdemir:
Talos 1.14 reports the CPU cores the way the Linux kernel sees them, one resource per online core. This is more accurate than the SMBIOS processor table and follows CPU hotplug. Omni now takes the core and thread counts from that view when a machine provides it, and keeps using SMBIOS on older Talos. The count reflects the cores currently available to workloads rather than the installed hardware, so offlined or hot-unplugged cores lower it. The kernel does not know everything about a CPU: on arm64 it has no vendor or model name, and it never knows the rated clock speed. Those fields still come from SMBIOS, so the vendor label, the CPU description and the frequency shown in the UI stay as they were.

The total core count of an instance says nothing about how the cores are distributed across machines. A new metric counts the machines in the instance for each distinct core count, so the distribution of machine sizes can be read off directly and size classes can be derived at query time without Omni having to know their thresholds. Machines whose core count is not known yet are reported under zero cores, so the series always add up to the total number of machines.

Firmware commonly reports the processor manufacturer as "Advanced Micro Devices, Inc.", which the CPU label never mapped to amd, so AMD machines carried the raw vendor string as their label. The label now recognizes that spelling next to the short one and the kernel's vendor identifier.

Closes siderolabs/omni#3224.

The new metric, for an instance with twelve 4-core machines, three 16-core machines and one 64-core machine:

```
omni_machine_cores{cores="4"} 12
omni_machine_cores{cores="16"} 3
omni_machine_cores{cores="64"} 1
```

### DIFF
--- a/client/pkg/omni/resources/omni/machine_status.go
+++ b/client/pkg/omni/resources/omni/machine_status.go
@@ -148,7 +148,7 @@ func ReconcileMachineStatusLabels(machineStatus *MachineStatus) {
 		switch {
 		case strings.Contains(cpuManufacturer, "intel"):
 			return "intel"
-		case strings.Contains(cpuManufacturer, "amd"):
+		case strings.Contains(cpuManufacturer, "amd"), strings.Contains(cpuManufacturer, "advanced micro devices"):
 			return "amd"
 		default:
 			return cpuManufacturer
--- a/client/pkg/omni/resources/omni/machine_status.go
+++ b/client/pkg/omni/resources/omni/machine_status.go
@@ -146,9 +146,9 @@ func ReconcileMachineStatusLabels(machineStatus *MachineStatus) {
 		cpuManufacturer = strings.ToLower(strings.TrimSpace(cpuManufacturer))
 
 		switch {
-		case strings.HasPrefix(cpuManufacturer, "intel"):
+		case strings.Contains(cpuManufacturer, "intel"):
 			return "intel"
-		case strings.HasPrefix(cpuManufacturer, "amd"):
+		case strings.Contains(cpuManufacturer, "amd"):
 			return "amd"
 		default:
 			return cpuManufacturer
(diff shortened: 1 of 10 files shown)


client/pkg/omni/resources/omni/machine_status.go
utkuozdemir, line 1: Changed here were done to cover `GenuineIntel` and `AuthenticAMD` as well. I checked if calling `.Contains` would be safe with other common manufacturers: it is safe.

frontend/src/views/Clusters/Management/ClusterMachineItem.vue
```diff
@@ -413,8 +413,9 @@ const onSavePatchConfig = (config: string) => {
         <div class="mt-4 mb-2">Network Interfaces</div>
         <div>
           <div v-for="(processor, index) in item?.spec?.hardware?.processors" :key="index">
-            {{ (processor.frequency ?? 0) / 1000 }} GHz, {{ processor.core_count }}
-            {{ pluralize('core', processor.core_count) }}, {{ processor.description }}
+            <template v-if="processor.frequency">{{ processor.frequency / 1000 }} GHz,</template>
+            {{ processor.core_count }} {{ pluralize('core', processor.core_count) }},
```
Unix4ever: was the count missing here?
`pluralize` was probably called with the wrong args. I think it can include the number in the generated string.

Unix4ever (approved):

frontend/src/views/Clusters/Management/ClusterMachineItem.vue
utkuozdemir: No, the count was there before, only rewrapped. What I changed is making the frequency conditional: processor entries can now come from the kernel view (CPUCores) when SMBIOS has no usable CPU entry, for example on a Raspberry Pi (#817), and that view has no rated frequency, so without the v-if it would render as "0 GHz, 4 cores". SMBIOS still supplies the frequency whenever it has one.

I now switched it to pass `pluralize: true` as a 3rd arg, so `{{ processor.core_count }}` because unnecessary - it'll render `8 cores` directly.

PR #3351, opened 2026-09-02
fix: validate the kubeconfig received from the Omni API

utkuozdemir:
Ensure that any kubeconfig received from the Omni management API conforms to one of the shapes actually produced by Omni, and it doesn't contain any extra unexpected fields.

The check lives in the client library, so omnictl and every other consumer of the client library get it without changes.

This is the Omni counterpart of the same change in talosctl.


client/pkg/kubeconfig/validate.go
utkuozdemir, line 1: I made this function to be part of the public client API, because it was the only clean way I could find to make the unit tests in the root package (omni itself) to be able to import it.

Those tests are at the locations where we generate kubeconfigs, so we are testing if the kubeconfigs server generates are valid for the same exact version of the client.

client/pkg/kubeconfig/validate.go
```diff
@@ -0,0 +1,387 @@
...
+		"password":           authInfo.Password != "",
+		"tokenFile":          authInfo.TokenFile != "",
+		"username":           authInfo.Username != "",
+	})...)
+}
+
+const (
+	execCommand     = "kubectl"
+	execAPIVersion  = "client.authentication.k8s.io/v1"
+	execRedirectArg = "--oidc-redirect-url=urn:ietf:wg:oauth:2.0:oob"
```
Unix4ever: I wonder if that `urn:ietf:wg:oauth:2.0:oob` is some constant that we can reuse

client/pkg/kubeconfig/validate.go
```diff
@@ -0,0 +1,387 @@
...
+	return append(multiErr, disallowed("exec", map[string]bool{
+		"env":                exec.Env != nil,
+		"installHint":        exec.InstallHint != "",
+		"provideClusterInfo": exec.ProvideClusterInfo,
+	})...)
+}
+
+// validateExecArgs checks the exec args. They must be the fixed prefix, then an optional grant type with the fixed redirect URL,
+// then an optional token cache directory, and nothing else.
+func validateExecArgs(args []string) []error {
```
Unix4ever, line 246: We could maybe just verify that the command executed is the right one. Or do you see any ways to exploit oidc-login through the args list?

client/pkg/kubeconfig/validate.go
utkuozdemir: Yep, good one, there was one on the server side, in an internal package. Now I moved it to the client package and reused it.

client/pkg/kubeconfig/validate.go
utkuozdemir: Seems that's not enough, because it has flags like `--browser-command` which runs an arbitrary command to "open the browser", and `--token-cache-dir` which writes the token cache to any path - so, args matter quite a bit too.

Unix4ever (approved):

PR #3358, opened 2026-09-03
test: retry the maintenance upgrade right after the install reboot

utkuozdemir:
The maintenance lifecycle test upgrades the machine as soon as Omni sees its new boot after the install. On Talos 1.14 the machine API and its DNS resolver come up a few seconds later than that, so the first upgrade attempt can fail with a refused connection. Retry the upgrade call for a bounded time instead of failing on the first attempt.

### DIFF
--- a/internal/integration/talos_test.go
+++ b/internal/integration/talos_test.go
@@ -1239,6 +1239,8 @@ func AssertMachineShouldBeInstalledInMaintenanceMode(testCtx context.Context, op
 		})
 
 		rtestutils.AssertResource[*omni.MachineStatusSnapshot](ctx, t, omniState, machineID, func(snapshot *omni.MachineStatusSnapshot, assertion *assert.Assertions) {
+			// An empty boot ID means the collect task could not read it yet, so it does not prove a reboot.
+			assertion.NotEmpty(snapshot.TypedSpec().Value.BootId, "machine %q has no boot ID after the maintenance install", machineID)
 			assertion.NotEqual(bootID, snapshot.TypedSpec().Value.BootId, "machine %q has not rebooted after the maintenance install", machineID)
 		})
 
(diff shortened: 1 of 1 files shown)


oguzkilcan (approved):

internal/integration/talos_test.go
```diff
@@ -1291,21 +1293,33 @@ func AssertMachineShouldBeUpgradedViaMaintenanceLifecycle(
...
-			ctx,
-			*machineID,
-			management.MaintenanceLifecycleRequest_OPERATION_UPGRADE,
-			target,
-			"",
-		)
+		// The machine has just rebooted into the installed Talos. Its API and its DNS resolver come up a few
+		// seconds after Omni first sees the new boot, so the first attempts can fail with a refused connection.
+		// The budget also has to cover a successful attempt, which includes the image factory building the installer image.
+		err = retry.Constant(10*time.Minute, retry.WithUnits(5*time.Second)).RetryWithContext(ctx, func(ctx context.Context) error {
```
frezbo, line 1299: doesn't the talos api itself retry internally?

utkuozdemir: No, it seems there's a small Talos issue, I'll try to see if I can understand and fix it.

utkuozdemir: It's caused by this one @shanduur has fixed: https://github.com/siderolabs/talos/pull/14007
Then was reverted by @smira: https://github.com/siderolabs/talos/pull/14009

@smira, shouldn't we actually do something about this?

With a partial config in maintenance mode, host DNS is enabled at boot (no config yet) and gets disabled when the partial config becomes active. machined keeps the cached `resolv.conf` pointing at `127.0.0.53` for up to 5 seconds, so DNS lookups from machined fail in that window. Omni pulls the installer image through the machine API when it installs or upgrades in maintenance mode, and the pull fails when it lands in that window.

Wouldn't it be better if a partial config without a host DNS setting did not change the host DNS state, as in siderolabs/talos#14007?

smira: The problem is that it breaks backwards compatibility - the HostDNS is _not_ enabled by default.

But all of this is actually caused by another change in 1.14 - the ImagePull API on Talos side no longer does automatic retries and honors deadline from the caller.

So the proper fix imho is to either add retries when Omni pulls the image, or let the caller (like in your PR) retry the whole maintenance upgrade Omni API.

PR #3364, opened 2026-09-04
fix: retry the installer image pull in the maintenance lifecycle API

utkuozdemir:
Since Talos 1.14, the image pull API does not retry or time out on its own anymore and leaves both to the caller. Omni pulls the installer image through this API when it installs or upgrades Talos on a machine in maintenance mode, and the pull can fail for transient reasons, e.g., when it is requested right after the machine rebooted and the machine is not fully up yet. This made the maintenance lifecycle integration test fail on Talos 1.14.

Retry the pull for a bounded time, as Talos did internally before. The workaround of retrying the whole maintenance lifecycle call in the integration test is removed, only the check for a non-empty boot ID after the install reboot stays.

### DIFF
--- a/internal/backend/talos/lifecycle/client.go
+++ b/internal/backend/talos/lifecycle/client.go
@@ -12,6 +12,7 @@ import (
 	"io"
 	"strings"
 
+	"github.com/siderolabs/go-retry/retry"
 	machineapi "github.com/siderolabs/talos/pkg/machinery/api/machine"
 	talosclient "github.com/siderolabs/talos/pkg/machinery/client"
 	"github.com/siderolabs/talos/pkg/machinery/constants"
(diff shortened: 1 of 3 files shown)


internal/backend/talos/lifecycle/client.go
```diff
@@ -90,10 +94,44 @@ func (m *Manager) pullInstallerImage(
...
+
+		var pullErr error
+
+		resolved, pullErr = m.pullImage(ctx, talosClient, imageRef)
+		if pullErr == nil {
+			return nil
+		}
+
+		switch status.Code(pullErr) { //nolint:exhaustive
+		case codes.NotFound, codes.PermissionDenied, codes.InvalidArgument:
```
smira: there is one weird thing with NotFound - it might return this when there are two mirrors:

* the first one failed for some transient/network reason
* the second one really doesn't have the image (not found)

containerd pull might return 404 in this case, while the retry might fix it.. 

not sure if we want to do some (bounded) 404 retries

smira (approved):

internal/backend/talos/lifecycle/client.go
utkuozdemir: Good point, changed it to do some 404 retries

Unix4ever (approved):

PR #3368, opened 2026-09-04
fix: manage the lifetime of the cached Talos clients with explicit leases

utkuozdemir:
The cached Talos clients were closed when their cache wrapper was garbage collected. Nothing told the callers to keep that wrapper referenced, so a caller working through the embedded client or through a stream lost its connection as soon as the cache dropped the entry. The cache expired the entries a fixed time after their creation, in use or not. This caused an installer image pull to die in the middle of an install.

Every lookup now returns a lease on the shared connection, and the caller closes it when done. The connection is closed only when the last lease and the cache have released it.

Additionally:
- The entries expire when idle instead of on a timer from their creation, so a connection in use is never torn down.
- The cache is a plain map with a periodic sweep in the cache manager loop instead of the LRU library. The client creation is serialized with the cache invalidation, so a client built from a stale state never survives in the cache.
- A lease which is garbage collected without being closed is reported in the log.
- The cache keeps a size limit. When it is full, the clients idle for the longest are closed to make room. A client with open leases is never chosen, so the cache can stay above the limit until those clients are released and the next lookup or the idle sweep trims it.
- The interfaces representing a returned Talos client gained a close method, and the callers which hand the clients over to a library keep them open until the library is done.

Related to siderolabs/omni#3364.

### DIFF
--- a/internal/backend/runtime/talos/clients.go
+++ b/internal/backend/runtime/talos/clients.go
@@ -12,18 +12,17 @@ import (
 	"fmt"
 	"runtime"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/cosi-project/runtime/pkg/controller"
 	"github.com/cosi-project/runtime/pkg/resource"
 	"github.com/cosi-project/runtime/pkg/safe"
 	"github.com/cosi-project/runtime/pkg/state"
-	"github.com/hashicorp/golang-lru/v2/expirable"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/siderolabs/talos/pkg/machinery/client"
 	clientconfig "github.com/siderolabs/talos/pkg/machinery/client/config"
 	"go.uber.org/zap"
-	"golang.org/x/sync/singleflight"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials/insecure"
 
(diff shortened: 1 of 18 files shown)


internal/backend/runtime/talos/clients.go
```diff
@@ -156,23 +182,34 @@ func GetSocketOptions(address string) []client.OptionFunc {
...
 // ClientFactory creates client based on the resource state.
+//
+// See Client for the lifetime of the returned clients.
 type ClientFactory struct {
 	omniState state.State
 	logger    *zap.Logger
 
-	cache *expirable.LRU[string, *Client]
-	sf    singleflight.Group
+	entries map[string]*entry
```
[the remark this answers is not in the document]
utkuozdemir: addressed

smira:
I think we need a way now to monitor for unclosed clients. 

I think there is a way to do this via a custom pprof profile (to capture where the unclosed client was created), and we should assert this in the integration test (?)

internal/backend/runtime/talos/clients.go
```diff
@@ -263,88 +304,223 @@ func (factory *ClientFactory) connectionOptions(ctx context.Context, id string,
...
+	// the leak guard reports a lease which is garbage collected without being closed. It never releases the lease itself,
+	// as the connection might still be in use through a stream. Its argument must not reference the lease.
+	cli.cleanup = runtime.AddCleanup(cli, factory.reportLeak, e.key)
+
+	return cli
 }
 
-// releaseForCluster evicts all cached clients for the given cluster (cluster-wide and per-machine).
+func (factory *ClientFactory) reportLeak(key string) {
+	factory.logger.Warn("Talos client was garbage collected without being closed", zap.String("key", key))
```
Unix4ever, line 441: Does it make sense to have a prometheus metric for that?

utkuozdemir:
> I think we need a way now to monitor for unclosed clients.
> 
> I think there is a way to do this via a custom pprof profile (to capture where the unclosed client was created), and we should assert this in the integration test (?)

Ah, great idea. I didn't know about the custom pprof profiles. Will look into doing that.

internal/backend/runtime/talos/clients.go
utkuozdemir: I think it probably does, a counter. And also, Andrey's suggestion.

utkuozdemir: added the counter.

utkuozdemir:
> I think we need a way now to monitor for unclosed clients.
> 
> I think there is a way to do this via a custom pprof profile (to capture where the unclosed client was created), and we should assert this in the integration test (?)

Added the custom pprof profile, leaked client metrics, and the no leak assertion in the tests.

Unix4ever (approved):

PR #3369, opened 2026-09-04
DRAFT / DO NOT MERGE: e2e suite IO and CPU wins

### LINKED ISSUE #3327: [bug] Omni experimental base params set synchronous mode incorrectly by default
We migrated to zombiezen.com/go/sqlite driver, but the defaults used in the base params are not fully supported by it.
We might need to add support for setting synchronous mode in a different way.

Maybe this kind of approach should work:

```go
package main

import (
	"log"

	"zombiezen.com/go/sqlite"
	"zombiezen.com/go/sqlite/sqlitex"
)

func main() {
	dbpool, err := sqlitex.NewPool("app.db", sqlitex.PoolOptions{
		PrepareConn: func(conn *sqlite.Conn) error {
			return sqlitex.Execute(conn, "PRAGMA synchronous = NORMAL;", nil)
		},
	})
	if err != nil {
		log.Fatal(err)
	}
	defer dbpool.Close()
}
```

utkuozdemir:
Draft branch with the e2e suite IO and CPU wins from the CI analysis. One commit per item, to be split into separate pull requests. Opened as a draft to get a full e2e run. Do not merge.

On purpose in this draft:
- the three "temporarily" commits carry replace directives and a talosctl image from a fork
- the commits are unsigned

What the commits do:
- Do not preallocate the QEMU disks. `talosctl` preallocates every VM disk by default, so a QEMU job reserved 60 GiB on the runner disk before a single machine booted.
- Wait for Vault readiness instead of sleeping a fixed 15 seconds, like the Dex, Keycloak and MinIO setups already do.
- Print the host pressure stall information, the disk counters and the disk usage at the start and at the end of every test script, so that a slow runner shows up in the job log.
- Run the Omni server with normal priority. The e2e scripts started it under `nice -n 10` since the initial commit.
- Skip the etcd fsync in the Helm e2e embedded etcd phase, like the other suites do.
- Run the sqlite database with the NORMAL synchronous mode. The pragmas in the connection string were ignored by the current driver, so every machine log line and every audit event was an fsync. Closes siderolabs/omni#3327.
- Retry a failed machine notification instead of dropping it. A dropped notification left the machine without its hostname and hardware information for good.
- Use ECDSA service account keys for the clusters in the tests, through a development environment variable honored in debug builds only.
- Run the QEMU machine disks with the unsafe cache mode, so that the guest writes stay in the host page cache. Requires the disk cache mode option of `talosctl cluster create`.
- Keep the compiler optimizations in the debug builds of the tests. The tests need the debug build tag, not the disabled optimizations.

Related branches on the forks, to be opened as pull requests: siderolabs/talos (disk cache mode option, ECDSA bundle option), cosi-project/state-sqlite (connection prepare hook), siderolabs/kres (overridable debug gcflags).

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -249,6 +249,7 @@ jobs:
       - name: run-e2e-auth0
         if: github.event_name == 'pull_request'
         env:
+          GO_DEBUG_GCFLAGS: ""
           TEST_CLASS: e2e-auth0
           WITH_DEBUG: "true"
           WITH_RACE: "true"
@@ -327,6 +328,7 @@ jobs:
           sops -d .secrets.yaml | yq -e '.secrets | to_entries[] | .key + "=" + .value' >> "$GITHUB_ENV"
       - name: run-integration-qemu
         env:
+          GO_DEBUG_GCFLAGS: ""
           INTEGRATION_TEST_ARGS: --test.run TestIntegration/Suites/(CleanState|EtcdBackupAndRestore)$
           TEST_CLASS: integration-qemu
           WITH_DEBUG: "true"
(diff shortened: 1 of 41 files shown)


utkuozdemir:
Superseded: the changes land one PR at a time, starting with #3375. The remaining pieces follow after it merges.

PR #3370, opened 2026-09-06
fix: reuse the node unique token already written to the machine

### LINKED ISSUE #3350: [bug] Omni might still incorrectly write node unique tokens to the machines
1. We see lots of logs from Omni that machines are rejected with the wrong node unique tokens.
2. We observed the new node getting PermissionDenied after adding it to Omni 1.9.3 (should be recent enough to be not affected by https://github.com/siderolabs/omni/pull/2218)

Need to check the code one more time to see if there are any racy places left that could write the wrong token to the node.

utkuozdemir:
The pending machine controller writes a node unique token to the META partition of the machine and records it in the pending machine status. Talos re-joins with the token as soon as the write is applied, before the call returns, and the provision handler registers the token carried by the join request.

The pending machine status is ephemeral and is not saved when the reconcile fails, so the controller could forget a token it had already written and generate a new one. Omni then kept the first token and the machine the second one, and the machine was rejected on every later join with "the machine has the correct fingerprint, but the random token part doesn't match". This happens when:

1. The META write is applied by Talos but the call fails, e.g., the reconcile deadline expires while META is flushed to the disk.
2. A machine with a duplicate UUID already holds a token, and Talos re-joins with the new UUID and the existing token before the new token is written.
3. A registered machine without a token reboots with a new WireGuard key, so the new boot and the legacy token migration create two pending machines for the same machine.

Read the token back from the machine and reuse it, generate a new one only when there is none. The token is written even when it is reused, so that a failed flush is retried and the machine re-joins. Concurrent writes to the same machine address share a single write.

Additionally, reuse the UUID override already written to the machine the same way, so that a failed reconcile does not make the machine re-join under yet another UUID.

Affected machines are recovered with `omnictl configure machine <id> --reset-node-unique-token`.

Closes siderolabs/omni#3350.

### DIFF
--- a/internal/backend/runtime/omni/controllers/omni/pending_machine_status.go
+++ b/internal/backend/runtime/omni/controllers/omni/pending_machine_status.go
@@ -20,7 +20,9 @@ import (
 	"github.com/google/uuid"
 	"github.com/siderolabs/gen/xerrors"
 	"github.com/siderolabs/talos/pkg/machinery/client"
+	talosruntime "github.com/siderolabs/talos/pkg/machinery/resources/runtime"
 	"go.uber.org/zap"
+	"golang.org/x/sync/singleflight"
 
 	"github.com/siderolabs/omni/client/pkg/jointoken"
 	"github.com/siderolabs/omni/client/pkg/meta"
@@ -65,7 +67,9 @@ func NewPendingMachineStatusController() *PendingMachineStatusController {
 	)
 }
 
-type pendingMachineStatusHandler struct{}
+type pendingMachineStatusHandler struct {
+	tokenWrites singleflight.Group
+}
 
 func (handler *pendingMachineStatusHandler) reconcileRunning(
 	ctx context.Context,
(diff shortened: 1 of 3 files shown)


internal/backend/runtime/omni/controllers/omni/pending_machine_status.go
```diff
@@ -178,30 +213,45 @@ func (handler *pendingMachineStatusHandler) handleUUIDConflicts(
...
-	// creates duplicate Link resources for a single physical machine (same public key and node subnet).
-	// Mirror the idempotency guard used by generateUniqueNodeToken: issue the override exactly once.
-	if existing, ok := pendingMachineStatus.Metadata().Annotations().Get(omni.MachineUUID); ok && existing != machineUUID {
-		return nil
+	// The pending machine keeps the conflict annotation until it is cleaned up, so this handler reconciles several times for the same machine,
+	// and the pending machine status is not saved when a reconcile fails. Reuse the override UUID already written to the machine:
+	// generating a new one on each attempt makes the machine re-join under multiple UUIDs, which creates duplicate links for a single physical machine.
+	// An override equal to the conflicting UUID is from an earlier conflict and does not resolve this one.
+	// Unlike the token, an existing override is not written again: a lost flush only causes another conflict resolution after a reboot.
+	id, err := readMetaKey(ctx, c, meta.UUIDOverride)
```
Unix4ever, line 221: Ohh good catch.

Unix4ever (approved):
💯

Issue #3346, opened 2026-09-02
Adapt Omni to the unified Image Factory API tokens


utkuozdemir:
Image Factory is unifying download tokens, node tokens and service account credentials into one factory-issued token with scopes (`download`, `pull`, `schematic`, `token`), see siderolabs/image-factory#566. Once that lands, Omni should be able to authenticate to the factory with a single long-lived, org-scoped token passed to it from outside (omni-controller for SaaS, config for self-hosted). We removed the OAuth2 client credentials path before 1.11 for the same reason. Username and password stay supported for self-hosted factories using htpasswd.

This is the list of what needs to change on the Omni side, so that we do not lose anything when we pick it up after 1.11.

What changes:

- Factory config: a token per factory, next to the existing username and password. Omni sends it as a bearer token on every factory request (the factory client already has `WithBearerToken` and `WithTokenSource`).
- `ImageFactoryAuth` holds the token as well. Everything that reads the credentials (the clients wrapper, `imagefactoryauth.BuildDocs`, `RequiresAuth()`) learns about it.
- Machine credentials: Omni creates one stored `pull` token per Omni instance via `POST /tokens` (client `TokenCreate`) and puts it into the `RegistryAuthConfig` document instead of the org credentials. Omni has to rotate it before it expires and re-generate machine configs. The maintenance mode registry auth patch is computed once per process and does not react to `ImageFactoryAuth` changes today, this needs fixing for the rotation to reach maintenance mode machines.
- Download tokens: the client `DownloadToken` now posts `/tokens` with `{"scopes":["download"],"stored":false}`, so bumping the client dependency covers it. Additionally, remove the fallback in `requestDownloadToken` that treats a 404 as "this factory does not issue download tokens" and quietly falls back to putting the credentials into media URLs and PXE URLs. Against an authenticated factory, failing to get a download token must be an error.
- Security reports: Omni proxies `/spdx/`, `/vex/` and `/scans/` with its factory credential. No token scope reaches those today, asked on the factory PR. If they end up under `schematic`, nothing changes on the Omni side.

What becomes obsolete:

- The deprecated flat `imageFactory*` config params. Their removal (siderolabs/omni#3315) was postponed by one release, this is the moment to do it.

Unchanged: enterprise detection via the `Server` header, per Talos version factory routing, the installation media URL API.

[...]

PR #3374, opened 2026-09-07
feat: authenticate to the image factory with an API token

utkuozdemir:
The image factory now issues API tokens, factory-signed JWTs with scopes, replacing its download tokens and node tokens. Omni can now authenticate to a factory with one. Basic auth stays supported for self-hosted factories using htpasswd.

The token is taken as a file only (`registries.factories.<primary|secondary>.tokenFile` or `--primary-factory-token-file`), never as a value. Omni follows the file with a directory watch and a one minute poll, so the token can be rotated without a restart.

For such a factory, `ImageFactoryAuthController` creates a stored `image:read` machine token named `omni-machines-<account name>-<yyyy-mm-dd>` and puts it into the `RegistryAuthConfig` document of every machine config. It is replaced at half of its lifetime (6 months for the factory's default of 1 year) or when Omni's own token changes. Replaced tokens are not revoked, they expire on their own. Until the machine token exists, no machine config is generated for that factory.

Installation media downloads from an authenticated factory are authenticated with a download token only. The fallback of putting credentials into the URL is gone: a factory that issues no token has to be upgraded.

Part of siderolabs/omni#3346.


internal/pkg/config/schema.json
```diff
@@ -257,6 +259,16 @@
...
         "password": ["username"]
       },
+      "if": {
+        "title": "Check Image Factory token file",
+        "required": ["tokenFile"]
+      },
+      "then": {
+        "properties": {
+          "username": { "not": {} },
+          "password": { "not": {} }
```
Unix4ever, line 269: Do we have the similar mutually exclusive rule for username and password, that blocks tokenFile? I couldn't find it, but maybe I missed it

internal/pkg/imagefactory/tokenfile/tokenfile.go
```diff
@@ -0,0 +1,254 @@
...
+		<-ctx.Done()
+
+		return nil
+	}
+
+	eg, ctx := panichandler.ErrGroupWithContext(ctx)
+
+	for _, token := range s.tokens {
+		eg.Go(func() error { return token.Run(ctx) })
+	}
```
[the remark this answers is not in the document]
utkuozdemir: no more an issue

internal/backend/runtime/omni/controllers/omni/imagefactory/auth.go
```diff
@@ -71,7 +105,11 @@ func (ctrl *AuthController) Outputs() []controller.Output {
 
 // Run implements controller.Controller interface.
 func (ctrl *AuthController) Run(ctx context.Context, r controller.Runtime, logger *zap.Logger) error {
+	// The runtime queues the first reconcile itself, so the timer starts stopped: nothing is scheduled
+	// until a reconcile asks for it.
 	timer := time.NewTimer(0)
+	timer.Stop()
+
```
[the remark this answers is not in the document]
utkuozdemir: the timer starts stopped

internal/backend/runtime/omni/controllers/omni/maintenance_config_status.go
```diff
@@ -128,6 +128,18 @@ func NewMaintenanceConfigStatusController(
...
+		qtransform.WithExtraMappedInput[*omni.ImageFactoryAuth](
+			func(ctx context.Context, _ *zap.Logger, r controller.QRuntime, _ controller.ReducedResourceMetadata) ([]resource.Pointer, error) {
+				links, err := safe.ReaderListAll[*siderolinkres.Link](ctx, r)
+				if err != nil {
+					return nil, err
+				}
+
+				return slices.Collect(links.Pointers()), nil
+			},
+		),
```
Slessi, line 142: just for my education - is this a zero-downtime change? it shouldn't matter for machines until upgrade time, right?

internal/backend/runtime/omni/controllers/omni/internal/imagefactoryauth/imagefactoryauth.go
```diff
@@ -16,9 +16,28 @@ import (
...
 func buildDoc(auth *omni.ImageFactoryAuth) (*cri.RegistryAuthConfigV1Alpha1, error) {
-	username := auth.TypedSpec().Value.Username
-	password := auth.TypedSpec().Value.Password
+	spec := auth.TypedSpec().Value
+
+	// A factory Omni authenticates to with a token has no credentials for the machines until the
+	// machine token exists: a config generated without them would only fail the installer pull.
+	if spec.ApiToken != "" && spec.MachineToken == "" {
+		return nil, fmt.Errorf("the machine token for the image factory %q is not available yet", auth.Metadata().ID())
+	}
```
Slessi: comment is kinda worded funny here, i think this is easier to understand if correct:

> When using an authenticated factory Omni has no credentials for the machines until ...

internal/pkg/imagefactory/tokenfile/tokenfile.go
```diff
@@ -0,0 +1,254 @@
...
+
+	ticker := time.NewTicker(t.pollInterval)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return nil
+		case <-ticker.C:
```
Unix4ever, line 121: I think we could completely omit the timer if fsnotify works, but it's mostly a nit.

cmd/omni/pkg/app/app.go
```diff
@@ -290,19 +306,36 @@ func setupImageFactoryClients(cfg *config.Params, state *omni.State) (*imagefact
...
+}
+
+// factoryAuth is what the configuration says Omni authenticates to the factory with. A token file is
+// loaded here, so that a factory Omni cannot authenticate to fails at startup.
+func factoryAuth(factory config.Factory, tokenFiles *tokenfile.Set, logger *zap.Logger) (imagefactory.Auth, error) {
+	if path := factory.GetTokenFile(); path != "" {
+		token, err := tokenFiles.Load(imagefactory.NormalizeFactoryURL(factory.GetUrl()), path, logger)
+		if err != nil {
+			return imagefactory.Auth{}, fmt.Errorf("failed to load the token of image factory %q: %w", factory.GetUrl(), err)
+		}
```
Slessi: its a bit ugly imo that we are mutating `tokenFiles` in here to populate it at the call site above but i guess its ok for now

client/pkg/imagefactory/client.go
```diff
@@ -70,22 +71,54 @@ type Client struct {
...
+				// not point at the token.
+				return "", errors.New("the image factory token is empty")
+			}
+
+			return token, nil
+		}))
+	case auth.Token != "":
+		clientOptions = append(clientOptions, client.WithBearerToken(auth.Token))
+	case auth.Username != "" && auth.Password != "":
+		clientOptions = append(clientOptions, client.WithBasicAuth(auth.Username, auth.Password))
```
Slessi, line 118: Don't we only need `TokenSource`? Is `Token` for omnictl? 🤔 

I would've thought omnictl should only be talking to backend for things, not factory directly

internal/backend/runtime/omni/controllers/omni/imagefactory/auth.go
```diff
@@ -135,17 +178,192 @@ func (ctrl *AuthController) reconcile(ctx context.Context, r controller.Runtime,
...
+	ctx, cancel := context.WithTimeout(ctx, machineTokenRequestTimeout)
+	defer cancel()
+
+	_, token, err := factoryClient.TokenCreate(ctx, client.TokenCreateOptions{
+		Name:   fmt.Sprintf("omni-machines-%s-%s", ctrl.accountName, now.UTC().Format(time.DateOnly)),
+		Scopes: []string{machineTokenScope},
+	})
+	if err != nil {
+		return "", err
+	}
```
Slessi, line 313: AI reviewer seemed concerned about this part potentially failing repeatedly until we hit the `maxPerOrg` token limit on factory side but I'm not sure how realistic that could be, just sharing

Slessi (commented):
mostly LGTM I'm mostly interested in my comment about `Token` vs `TokenSource` for (if i understood) omnictl usage

Unix4ever (approved):

internal/pkg/imagefactory/tokenfile/tokenfile.go
utkuozdemir: I did it this way since AFAIK inotify is not always reliable, so the state-of-the-art is "use but not trust". Also checked how Talos watches files (many examples!) and brought the concept as-is here. :)

internal/backend/runtime/omni/controllers/omni/maintenance_config_status.go
utkuozdemir: Not sure if I get the question, but let me explain: imagine 6 months has passed and Omni renewed the token. It means, `ImageFactoryAuth` resource got updated. At that time, we should apply the new `RegistryAuthConfig` resource to all machines, including the maintenance mode ones.

For that, we need to convert an `ImageFactoryAuth` update event to "all machines" update event - so, for this controller, all `Link` resources.

The RegistryAuthConfig is a partial document, and applying that is immediate (does not cause reboot).

You could ask, why not defer applying it until the install/upgrade time, but why complicate things? The moment we know the new value, we reflect it to the maintenance machines.

cmd/omni/pkg/app/app.go
utkuozdemir: Actually you are right, better to just construct and return it - let me see what I can do.

internal/backend/runtime/omni/controllers/omni/maintenance_config_status.go
Slessi: > applying that is immediate (does not cause reboot)

Yea that was kinda the question, was just wondering if it would have any unwanted impact to workloads, or any such thing

internal/pkg/config/schema.json
utkuozdemir: This check already blocks in both directions, I just verified.

cmd/omni/pkg/app/app.go
utkuozdemir: done

client/pkg/imagefactory/client.go
utkuozdemir: Token was yes, mainly for the clients, omnictl and client library and so on.

But since a token could already be expressed using `TokenSource`, it's now gone.

internal/backend/runtime/omni/controllers/omni/imagefactory/auth.go
utkuozdemir: That'd basically most likely mean a broken factory - our code doesn't combine it with many other operations after creation,  it directly proceeds with storing it, so if that fails, I think a retry is warranted / most reasonable thing to do.

internal/backend/runtime/omni/controllers/omni/internal/imagefactoryauth/imagefactoryauth.go
utkuozdemir: Fair, replaced with yours.

PR #3375, opened 2026-09-07
test: reduce the IO load of the e2e test scripts

utkuozdemir:
### test: do not preallocate the QEMU disks in the integration tests

`talosctl` preallocates every VM disk by default. Because of this, a QEMU integration job reserved 60 GiB on the runner disk before a single machine booted, and the cluster import job another 36 GiB. The runners keep the VM disks on the node disk next to the container images, so the jobs were getting evicted on disk pressure.

The disks are throwaway. Let them grow with what the machines actually write.

### test: wait for Vault readiness instead of sleeping in the tests

Previously, the Vault setup slept for a fixed 10 seconds before loading the key, and for another 5 seconds after it. This happened in all but one of the integration and e2e jobs.

Poll the health endpoint until the dev server is unsealed instead, like the Dex, Keycloak and MinIO setups already do. Also drop the sleep after the key write, as the write is synchronous anyway.

### test: print the host pressure and disk usage in the test logs

The self-hosted runners share their disk and CPUs with other jobs. When a run fails, there is nothing in the log to tell a slow runner from a real failure.

Print the following when a test script starts and again in its cleanup, so that every job log carries the numbers:
- the kernel pressure stall information for IO, CPU and memory
- the disk counters
- the disk usage of the test directories
- the Docker disk usage

### test: run the Omni server with normal priority in the e2e tests

The e2e scripts started the Omni server under `nice -n 10` since the initial commit. On the CPU starved runners, this gives Omni a tenth of the scheduler weight of the browser, Docker and the QEMU machines it competes with. At the same time, Omni is the process the UI tests are waiting on.

Run it with the normal priority.

### test: skip the etcd fsync in the Helm e2e embedded etcd phase

All other e2e suites run the embedded etcd of the Omni under test without fsync, because the state is throwaway. The Helm suite did not. Its embedded etcd phase runs with persistence disabled inside the QEMU node, so every fsync went through the virtual disk to the runner disk.

Enable the unsafe fsync setting in the test values.

### test: give the node label check the same budget as the node list check

The Kubernetes API test waits 30 seconds for every ready node to carry the label the test injects through a config patch. The label is applied by the node's own controller through its local apiserver. On an IO starved runner, this took minutes twice, once on the enterprise backports run and once on the IO wins draft, while the node list check right before it already waits up to 3 minutes.

Give the label check the same budget.

### DIFF
--- a/hack/test/common.sh
+++ b/hack/test/common.sh
@@ -131,7 +131,30 @@ export RUN_DIR
 LOCAL_IP=$(ip -o route get to 8.8.8.8 | sed -n 's/.*src \([0-9.]\+\).*/\1/p')
 export LOCAL_IP
 
+# Print the host IO and CPU pressure, the disk counters and the disk usage, so that a slow runner shows up in the job log.
+function print_host_pressure() {
+  # the trace output would interleave with the snapshot, so it is turned off for the duration of the function
+  local -
+  set +x
+
+  echo "--- host pressure snapshot: $1 ($(date -u +%FT%TZ)) ---"
+  {
+    local f
+    for f in io cpu memory; do
+      echo "pressure/${f}:"
+      cat "/proc/pressure/${f}"
+    done
+    echo "diskstats:"
+    awk '$3 ~ /^(nvme[0-9]+n[0-9]+|sd[a-z]+|vd[a-z]+)$/' /proc/diskstats
+    df -h "${TEST_OUTPUTS_DIR}" "${HOME}" "${ARTIFACTS}" 2>/dev/null
+    docker system df
+  } 2>&1 || true
+  echo "--- end of host pressure snapshot ---"
+}
+
 mkdir -p "$TEST_OUTPUTS_DIR"
+
+print_host_pressure "start"
 
 export ENABLE_TALOS_PRERELEASE_VERSIONS=true
 VAULT_DOCKER_IMAGE=hashicorp/vault:1.18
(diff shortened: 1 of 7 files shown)


smira:
> talosctl preallocates every VM disk by default. Because of this, a QEMU integration job reserved 60 GiB on the runner disk before a single machine booted, and the cluster import job another 36 GiB. The runners keep the VM disks on the node disk next to the container images, so the jobs were getting evicted on disk pressure.

This is no longer true:

1. There was a provisioning bug when 2 nodes joined with small EPHEMERAL - fixed now.
2. QEMU disks are now mapped to a separate disk (non-EPHEMERAL).

utkuozdemir:
> > talosctl preallocates every VM disk by default. Because of this, a QEMU integration job reserved 60 GiB on the runner disk before a single machine booted, and the cluster import job another 36 GiB. The runners keep the VM disks on the node disk next to the container images, so the jobs were getting evicted on disk pressure.
> 
> This is no longer true:
> 
> 1. There was a provisioning bug when 2 nodes joined with small EPHEMERAL - fixed now.
> 2. QEMU disks are now mapped to a separate disk (non-EPHEMERAL).

yes, I'm gonna make this smaller PRs, this is now obsolete.

utkuozdemir:
Closing, this goes in as single commit PRs one at a time. The preallocation commit is dropped, it is obsolete with the VM disks on a separate disk.

PR #3381, opened 2026-09-07
fix: remove incomplete installation media downloads

immanuwell:
## What

If the image server drops an HTTP 200 response mid body, `omnictl` leaves a partial image at the final path. Retrying then fails with `destination already exists`

Downloads now go to a temporary file beside the destination. The file is renamed only after the download and close succeed, and cleaned up on errors

## Repro

1. Return HTTP 200 with `Content-Length: 100`
2. Write `partial response`, then close the connection
3. Download to `metal.iso`
4. `unexpected EOF` is returned, but `metal.iso` remains and blocks the next try

## Tests

`go test ./...`

`golangci-lint run ./pkg/omnictl/internal/download`


client/pkg/omnictl/internal/download/download.go
```diff
@@ -834,11 +835,18 @@ func downloadResponseTo(dest string, resp *http.Response) error {
 		return err
 	}
 
-	defer checkCloser(f)
+	_, copyErr := io.Copy(f, resp.Body)
```
smira: this is wrong/incomplete fix - the download should go to a temporary file, and removed to a final location on success.

temporary file should be always removed.

immanuwell: fixed with atomic temp-file writes and cleanup tests

immanuwell:
@smira fixed with atomic temp-file writes and cleanup tests, please take a look

utkuozdemir (approved):

utkuozdemir:
@immanuwell Thanks for the contribution!

PR #3382, opened 2026-09-07
fix: build install images with the factory that issued the schematic

utkuozdemir:
A schematic ID is only meaningful together with the factory that issued it: another factory does not know the ID, and an Enterprise factory stamps an owner into the schematic and issues a different ID for the same content. Install images were built with the factory serving the target Talos version instead, which is the primary. After a factory switch (an Enterprise factory becomes the primary, the public one the secondary), a machine in maintenance mode that is added to a cluster keeps its public schematic ID, because nothing about it changed, and gets an install image naming the Enterprise host. The installer pull fails with 404 and the machine never joins.

`SchematicConfiguration` now records the base URL of the factory that issued its schematic ID, in the same write. Resources written before the field existed get it backfilled by asking the configured factories which one knows the ID, without changing the ID. `MachineConfigGenOptionsController` takes the install image host from that record when it covers the schematic ID being published, keeps the host already recorded in the install image otherwise, and publishes nothing rather than a guess when it has neither. The maintenance schematic upgrade takes the host from the record as well and waits when it is not known. The guessing fallbacks in the cluster machine config generation and in the lifecycle client are gone: an install image with a schematic and no host is not built.

A running machine that does not change keeps its schematic ID and its factory host. When its extensions, kernel args or Talos version change, Omni ensures the schematic on the factory that serves the target Talos version. That factory issues the new ID, so the ID and the host move together.

This way, a machine enrolled before the switch keeps installing from the public factory until something on it changes. Because of this, the public factory has to stay configured as the secondary until no schematic names it.

Part of siderolabs/omni#3346.

### DIFF
--- a/client/api/omni/specs/omni.pb.go
+++ b/client/api/omni/specs/omni.pb.go
@@ -6628,11 +6628,15 @@ func (x *TalosExtensionsSpec) GetItems() []*TalosExtensionsSpec_Info {
 
 // SchematicConfigurationSpec is the desired Image Factory schematic for a machine, machine set or a cluster.
 type SchematicConfigurationSpec struct {
-	state         protoimpl.MessageState `protogen:"open.v1"`
-	SchematicId   string                 `protobuf:"bytes,1,opt,name=schematic_id,json=schematicId,proto3" json:"schematic_id,omitempty"`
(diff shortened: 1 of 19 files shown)


utkuozdemir:
Closing, superseded by a much smaller change. The only broken path was the install of a maintenance machine, whose schematic was never re-ensured on the factory serving the target version, while upgrades and customization changes already re-ensure it there. The schematic controller now ensures the schematic at that install moment as well, no record of the issuing factory is needed.

PR #3386, opened 2026-09-08
chore: bump delve version in compose

Slessi:
Bump go-delve version to 1.27.1 in docker-compose.yml to support go 1.27 (version match is only a coindence). Also set the image to use TOOLCHAIN.

### DIFF
--- a/hack/compose/docker-compose.yml
+++ b/hack/compose/docker-compose.yml
@@ -77,7 +77,7 @@ services:
 
   # The launcher is needed regardless of WITH_DEBUG.
   runtime-tools-init:
-    image: golang:1.26-alpine
+    image: ${TOOLCHAIN}
     restart: on-failure
     volumes:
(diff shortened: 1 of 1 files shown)


hack/compose/docker-compose.yml
```diff
@@ -77,7 +77,7 @@ services:
 
   # The launcher is needed regardless of WITH_DEBUG.
   runtime-tools-init:
-    image: golang:1.26-alpine
```
smira, line 80: is is in the environment already as a variable? 

if it is, no problem

smira (approved):

hack/compose/docker-compose.yml
Slessi: yea this is `TOOLCHAIN ?= docker.io/golang:1.27.1-alpine` from the makefile (kres)

PR #3398, opened 2026-09-09
fix: stop using the plain Talos installer image for installs

### LINKED ISSUE #3388: Installing a machine with an invalid schematic leaves it disconnected on Talos 1.13+
When a machine has an invalid schematic (it reports extensions but no schematic ID, e.g. a custom Talos build or media built with the imager), Omni installs it with the plain `ghcr.io/siderolabs/installer` image. On Talos 1.13 and later, the machine never reconnects after that install: the plain installer carries no SideroLink kernel args, and the install writes no machine config either. The machine comes back in maintenance mode from disk, joined to nothing, and the cluster stays in the installing state forever.

Reproduced on 2026-09-08 with QEMU:

1. Build an ISO with the imager: Talos 1.13.10, `siderolabs/hello-world-service`, the Omni join kernel args on the command line (`sidero-dev boot-media iso --extension ...`).
2. Boot a machine from it. It joins Omni in maintenance mode. Omni marks its schematic as invalid, since the machine reports extensions but no schematic ID.
3. Add the machine to a 1.13.10 cluster. Omni runs the maintenance install through the lifecycle API with `ghcr.io/siderolabs/installer:v1.13.10`.
4. The install completes, the machine reboots into the installed Talos, maintenance mode, `apid` is up.
5. The machine never connects again: no `siderolink.api` on the kernel command line, no config on the disk. Omni shows it as disconnected, the cluster machine stays in the installing stage.

It should either come back and join the cluster, or Omni should refuse the install with a clear reason before touching the disk.

## Why it happens

- The install image falls back to `<talos registry>:<version>` when the schematic ID is empty (`internal/backend/installimage/installimage.go`). Invalid schematics are considered ready on purpose, so that such machines can be allocated (`SchematicReady()` in `client/api/omni/specs/machine_status.go`), and the schematic controller publishes an empty ID for them.
- Below 1.13 the install is a config apply, so the installed machine boots with the full config on its disk, including the SideroLink join document, and connects through it. The plain installer worked because of this, regardless of what the kernel command line contained. The lifecycle install on 1.13 and later sends only the image and the disk (`internal/backend/talos/lifecycle/client.go`), the config comes later over SideroLink. So the installed machine has to reconnect on its own first, and it can only do that with join args on the kernel command line.
- The plain installer has no join args in it. On the Talos side, the args of the running system are only carried over on the legacy path: `internal/pkg/install/install.go` passes `siderolink.api`, `talos.events.sink`, `talos.logging.kernel` etc. from `/proc/cmdline` to the installer, but the installer only uses them in the code marked "legacy code which is only used when running in GRUB mode with GrubUseUKICmdline set to false" (`cmd/installer/pkg/install/install.go`). With UKI, or with grub and `grubUseUKICmdline` (the default for configs generated with a 1.12 or later contract, forced by Omni for clusters created at 1.14), the command line is the one embedded in the installer image.

So a machine with an invalid schematic breaks on 1.13 and later, unless it is one of the older grub machines that still carry their kernel args over.

The upgrade path is not affected in practice: a machine that installed from a factory image keeps its published schematic ID, the fallback needs an empty one.

## What we could do

a. Fail closed: when the schematic is invalid and the machine will not carry its kernel args over (booted with UKI, or a 1.12+ config contract), do not start the install, and put the reason into the machine status. Same for the maintenance upgrade path.

b. Let the user provide the installer image for such machines. Custom builds have their own installer image with the extensions and the kernel args baked in. Today a `machine.install.image` patch is only honored by the legacy install, the lifecycle install builds the image from the config generation options, and 1.14 has no `machine.install` section at all.

c. Keep the plain installer fallback only for the machines that carry their kernel args over (the legacy grub population), which is where it was working before.

a. is the minimal fix and stops the silent breakage. b. is what makes custom builds actually usable on 1.13+. Question: do we want to keep supporting custom builds (invalid schematics) on 1.13+ at all, or document them as unsupported?

utkuozdemir:
A machine that reports system extensions but no image factory schematic has an *invalid schematic* in Omni, e.g., a machine booted from an imager-built ISO with an extension. Omni installed such machines with the plain installer image `ghcr.io/siderolabs/installer`. On Talos 1.13 and later this leaves the machine dead:

1. The machine joins Omni in maintenance mode and is added to a cluster.
2. Omni installs it through the lifecycle API. The install sends only the installer image and the disk, no machine config is written to the disk.
3. The plain installer carries no SideroLink kernel args, so the installed machine has nothing to connect back with. It comes back in maintenance mode joined to nothing, and the cluster stays in the installing stage forever.

Talos 1.14 and later do not publish the plain installer image at all, so there the pull fails and the install is stuck.

Before Talos 1.13 this was not a problem: the install ran through apply-config, and the applied config carried the SideroLink documents to the disk, so the machine connected back through them regardless of the kernel args.

With this PR:

- **The plain installer fallback is removed.** An install image always comes from an image factory, an install image without a schematic ID is an error.
- **An invalid machine gets a schematic**, the same way a machine without any extensions already does: built from the machine's own join kernel args, with no extensions (the custom extensions of the machine are not carried over), and registered with the image factory. In the case above the install image becomes `factory.talos.dev/metal-installer/<id>:v1.13.10`, the installed machine connects back through the baked-in join args, and it is a normal factory machine from then on.
- **The invalid flag stays.** It still means that the extensions of the machine are not managed and that a schematic difference alone never triggers an upgrade. Only a Talos version change does. Machines with Talos already installed are left alone by this change, and their next upgrade goes through the factory image the same way.
- The maintenance mode schematic sync skips invalid machines explicitly instead of relying on their missing raw schematic.
- The Talos installer registry setting and its flag are deprecated as unused.

Closes siderolabs/omni#3388.


client/pkg/omni/resources/omni/machine_config_gen_options.go
```diff
@@ -14,25 +14,19 @@ import (
...
 		SchematicId:          schematicID,
 		SchematicInitialized: schematicInitialized,
 		SecurityState:        spec.SecurityState,
 		Platform:             spec.GetPlatformMetadata().GetPlatform(),
 		ImageFactoryHost:     imageFactoryHost,
 	}
-
-	if schematicInitialized {
-		installImage.SchematicInvalid = spec.GetSchematic().GetInvalid()
-	}
```
utkuozdemir, line 33: Review note: the flag had no reader left. Its only consumer was the install image builder, which used it to drop the schematic ID and fall back to the plain installer. The fallback is gone, so the field goes with it: removed from the proto, its number reserved.

internal/backend/grpc/management.go
```diff
@@ -776,7 +776,7 @@ func (s *managementServer) ensureSchematic(ctx context.Context, talosVersion str
 
 	id := schematicSpec.GetFullId()
 
-	if schematicSpec.GetFullId() != "" && schematicSpec.GetRaw() != "" && !schematicSpec.GetInvalid() && !schematicSpec.GetInAgentMode() {
```
utkuozdemir, line 779: Review note: an invalid machine has a raw schematic now (synthesized from its join kernel args), so a maintenance upgrade registers it with the factory serving the target version like for any other machine. Skipping it would build the image from the ID the machine computed for itself, which differs from the factory's on an Enterprise factory (it stamps its owner into the schematic).

internal/backend/grpc/management.go
```diff
@@ -861,13 +861,12 @@ func (s *managementServer) MaintenanceUpgrade(ctx context.Context, req *manageme
 		TalosVersion:         req.Version,
 		SchematicId:          schematicID,
 		SchematicInitialized: true,
-		SchematicInvalid:     machineStatus.TypedSpec().Value.Schematic.Invalid,
```
utkuozdemir, line 864: Review note: nothing reads this flag anymore, see `NewInstallImage`.

internal/backend/installimage/installimage.go
```diff
@@ -65,13 +67,5 @@ func Build(resID resource.ID, installImage *specs.MachineConfigGenOptionsSpec_In
...
 
-	if schematicID != "" {
-		if installImage.ImageFactoryHost == "" {
-			return "", fmt.Errorf("machine %q has no image factory host set", resID)
-		}
-
-		return installImage.ImageFactoryHost + "/" + installerName + "/" + schematicID + ":" + desiredTalosVersion, nil
-	}
-
-	return talosRegistry + ":" + desiredTalosVersion, nil
```
utkuozdemir, line 76: Review note: the plain installer fallback is removed. The image is always `<factory host>/<installer name>/<schematic id>:<version>`, and a missing schematic ID or factory host is an error instead of a silent fallback to `ghcr.io/siderolabs/installer`. The `SchematicInvalid` override that forced the fallback for invalid machines is gone too. The factory host check moved up front, next to the schematic ID check. The rest (platform prefix, secure boot name, `v` prefix) is unchanged.

internal/backend/runtime/omni/controllers/omni/cluster_machine_config.go
```diff
@@ -307,11 +306,11 @@ func reconcileClusterMachineConfig(
 		return xerrors.NewTagged[qtransform.SkipReconcileTag](errors.New("install image is not set yet"))
 	}
 
-	// skip if the machine schematic information is not yet detected
-	if !installImage.SchematicInitialized {
+	// skip if the machine schematic is not yet detected or not yet published
+	if !installImage.SchematicInitialized || installImage.SchematicId == "" {
```
utkuozdemir, line 310: Review note: `installimage.Build` errors on an empty schematic ID now. The config generator already waited while the machine schematic was not detected yet, it now also waits while no schematic ID is published yet, instead of failing the reconcile. E.g., a machine that an older Omni recorded as invalid (so it has no raw schematic yet) and that has not reconnected since the upgrade.

internal/backend/runtime/omni/controllers/omni/cluster_machine_config.go
```diff
@@ -307,11 +306,11 @@ func reconcileClusterMachineConfig(
...
 		logger.Error("machine schematic is not set, skip reconcile")
 
-		return xerrors.NewTagged[qtransform.SkipReconcileTag](errors.New("machine schematic is not set detected"))
+		return xerrors.NewTagged[qtransform.SkipReconcileTag](errors.New("machine schematic is not detected or not published yet"))
```
utkuozdemir, line 313: Review note: published = the ID `SchematicConfiguration` carries for the machine, the one the schematic controller got back from the image factory or, for a machine that already runs what Omni wants, the one the machine reports. Before this PR an invalid machine had an empty ID published forever, and downstream an empty ID meant "use the plain installer", so the generator never had to treat it as a wait state. Now it is one.

internal/backend/runtime/omni/controllers/omni/internal/talos/schematic.go
```diff
@@ -19,15 +19,14 @@ import (
...
-	FullID      string
-	Raw         string
-	Extensions  []string
-	KernelArgs  []string
+	FullID     string
+	Raw        string
+	Extensions []string
+	KernelArgs []string
+	// Invalid is set when the machine has extensions installed bypassing the image factory. They cannot be managed, so none are reported.
+	Invalid     bool
```
utkuozdemir, line 29: Review note: moved here from the machine task's wrapper around this struct (`machine.SchematicInfo`), which existed only to carry the flag next to the error. The detection sets the flag itself now, and the wrapper is gone.

internal/backend/runtime/omni/controllers/omni/internal/talos/schematic.go
```diff
@@ -99,11 +99,12 @@ func GetSchematicInfo(ctx context.Context, talosState state.CoreState, fallbackK
...
 
-	if fullID == "" && len(exts) > 0 {
-		return SchematicInfo{}, ErrInvalidSchematic
-	}
-
 	if fullID == "" { // we could not find the full ID, so we fall back to synthesizing it (and the raw YAML) using the default args
+		invalid := len(exts) > 0
+		if invalid {
+			exts = nil
+		}
```
utkuozdemir, line 106: Review note: the flag keeps its meaning, the machine has extensions installed bypassing the image factory, so Omni cannot manage them. What changed is what such a machine gets. Before, the detection returned an error and the machine ended up with no schematic at all: no ID, no raw, no kernel args. Now it gets the same synthesized schematic as a machine without extensions, i.e., its join kernel args and no extensions, plus the flag. The flag still guards everything that must not happen for these machines: extension and kernel args management, and an upgrade because of a schematic difference.

internal/backend/runtime/omni/controllers/omni/internal/task/machine/machine.go
```diff
@@ -38,13 +38,6 @@ import (
 	"github.com/siderolabs/omni/internal/backend/runtime/talos"
 )
 
-// SchematicInfo contains all collected schematic information.
-type SchematicInfo struct {
-	talosschematic.SchematicInfo
-
-	Invalid bool
-}
-
```
utkuozdemir, line 47: Review note: the wrapper's only field was the flag, which moved into the detection result, so it became a struct embedding a struct and is removed.

internal/backend/runtime/omni/controllers/omni/internal/task/machine/poll.go
```diff
@@ -573,20 +572,10 @@ func pollExtensions(ctx context.Context, c *client.Client, info *Info) error {
...
 	schematicInfo, err := talos.GetSchematicInfo(ctx, c.COSI, info.DefaultKernelArgs)
 	if err != nil {
-		if errors.Is(err, talos.ErrInvalidSchematic) {
-			info.Schematic = &SchematicInfo{
-				Invalid: true,
-			}
-
-			return nil
-		}
-
```
utkuozdemir, line 583: Review note: the detection returns the schematic info with the flag set instead of an error, so there is no error to handle here anymore. The info is stored as is.

internal/backend/runtime/omni/controllers/omni/machineconfig/status.go
```diff
@@ -873,33 +872,27 @@ func (ctrl *StatusController) checkInstalledImage(
...
+		if err != nil {
+			return installedImage{}, err
 		}
 
-		return installedImage{}, err
+		invalid = schematicInfo.Invalid
+	}
+
+	// the schematic plays no role in the checks for the machines running extensions installed bypassing the image factory
+	if invalid {
```
utkuozdemir, line 891: Review note: this checks what the machine runs against the target install image, for every install and upgrade path (legacy and lifecycle). The invalid case was handled twice with an identical block: once from the machine status, once from the live check on the machine, for when the status has not caught up yet (that one used to come back as an error). Merged into one: the live check runs only when the status does not already say invalid, and one block handles both. Same behavior, the schematic plays no role for such machines.

internal/backend/runtime/omni/controllers/omni/machineupgrade/status.go
```diff
@@ -85,6 +84,16 @@ func (ctrl *StatusController) transform(ctx context.Context, r controller.Reader
...
 
+	// This controller upgrades maintenance mode machines when the schematic Omni wants for them differs from the one they run.
+	// The extensions and kernel args of an invalid machine are not managed, so there is never such a difference to act on.
+	if schematicSpec.Invalid {
+		status.TypedSpec().Value.Phase = specs.MachineUpgradeStatusSpec_Unknown
+		status.TypedSpec().Value.Status = "machine was not provisioned using an image factory image, schematic changes are not managed"
+		status.TypedSpec().Value.Error = ""
+
+		return nil
+	}
```
utkuozdemir, line 95: Review note: this controller upgrades maintenance mode machines when the schematic Omni wants for them differs from what they run (cluster machines are upgraded by the cluster machine config status controller). For an invalid machine that difference must never trigger anything: its extensions and kernel args are not managed, and after this PR its published ID can differ from the one it reports, e.g., on an Enterprise factory. Before, the controller stopped by accident, on the missing raw schematic. Now it stops here explicitly.

internal/backend/runtime/omni/controllers/omni/schematic/configuration.go
```diff
@@ -196,18 +196,6 @@ func (ctrl *ConfigurationController) transform(ctx context.Context, r controller
...
-	// resource to exist so the install image / config generation pipeline runs. Emit a minimal one and
-	// let the downstream's existing Invalid handling (installimage.Build falls back to talosRegistry:version,
-	// ReconciliationContext skips schematic mismatch) take over.
-	if ms.TypedSpec().Value.Schematic.Invalid {
-		schematicConfiguration.TypedSpec().Value.TalosVersion = talosVersion
-		schematicConfiguration.TypedSpec().Value.SchematicId = ""
-
-		return ctrl.saveMachineExtensionStatus(ctx, r, machineExtensionsStatus)
-	}
-
```
utkuozdemir, line 210: Review note: this controller computes the schematic Omni wants for a machine (the machine's extensions and kernel args applied to the schematic it reports), registers it with the image factory and publishes the ID in `SchematicConfiguration`, which the install image is built from.

- Before, invalid machines took a special branch here: publish an empty ID, which downstream meant "use the plain installer".
- Now they go through the normal path. The schematic they report is the synthesized one (join kernel args, no extensions), so the published ID is a real factory ID and the install image comes from the factory.
- The "the machine runs what it should, keep its own ID" case is skipped for them, see the note below.

internal/backend/runtime/omni/controllers/omni/schematic/configuration.go
```diff
@@ -264,7 +252,8 @@ func (ctrl *ConfigurationController) publishSchematicID(ctx context.Context, log
 	talosInstalled := omni.GetMachineStatusSystemDisk(ms) != ""
 
 	switch {
-	case runsDesired && talosInstalled && !installPending:
+	// an invalid machine reports an id it computed itself, no factory issued it, so its published id must come from the factory, see the next case
+	case runsDesired && talosInstalled && !installPending && !ms.TypedSpec().Value.Schematic.Invalid:
```
utkuozdemir, line 257: Review note: the first case keeps the ID the machine reports when it already runs what Omni wants. That ID normally comes from the factory meta extension, so some factory issued it. An invalid machine has no meta extension, its ID is computed locally from the synthesized schematic, and on an Enterprise factory it differs from what the factory answers for the same content. So this case is skipped for invalid machines and they fall through to the next one, which asks the factory. Otherwise the published ID would flip to an ID nobody serves once the machine leaves maintenance mode.

internal/backend/talos/lifecycle/client.go
```diff
@@ -46,37 +46,30 @@ func (m *Manager) buildInstallImage(ctx context.Context, machineID string, ms *o
 		version = strings.TrimPrefix(spec.TalosVersion, "v")
 	}
 
-	if target != nil {
-		// The target is the spec of a cached resource, so patch a copy of it instead of writing through into the cache.
-		target = target.CloneVT()
+	if target == nil {
+		return "", fmt.Errorf("machine %q has no target install image", machineID)
+	}
```
utkuozdemir, line 51: Review note: the target install image is required now. All four callers pass one, so this is only a guard.

internal/backend/talos/lifecycle/client.go
```diff
@@ -46,37 +46,30 @@ func (m *Manager) buildInstallImage(ctx context.Context, machineID string, ms *o
...
 	}
 
-	imageFactoryClient, err := m.imageFactoryClients.ForTalosVersion(ctx, version)
-	if err != nil {
-		return "", fmt.Errorf("failed to get image factory client for Talos version %q: %w", version, err)
-	}
-
-	installImage := omni.NewInstallImage(ms, version, spec.Schematic.FullId, imageFactoryClient.Host(), true)
-
-	return installimage.Build(machineID, installImage, m.talosRegistry)
```
utkuozdemir, line 79: Review note: this was the fallback for a nil target, building the image from the schematic ID the machine reports. Nothing ever passed nil, and after this PR that ID can be a locally computed one for an invalid machine, which the factory does not know. Removed rather than left as a trap.

Unix4ever (approved):

PR #3399, opened 2026-09-10
fix: correct handling of Wireguard peers and tunnel mode

smira:
Imports https://github.com/siderolabs/siderolink/pull/55

This reverts commit 194c629cf40e468d9b52cec321ea5954fc99021e.


internal/backend/runtime/omni/controllers/omni/link_status.go
```diff
@@ -75,31 +72,31 @@ type linkStatusHandler[T siderolinkSpecWrapper] struct {
...
-		// will recreate the peer using the old mode, and the machine will become reachable again.
-		if err := handler.peers.Remove(ctx, siderolink.GetPeerID(linkStatus.TypedSpec().Value), link.Metadata()); err != nil {
-			return err
-		}
-
-		if (link.TypedSpec().Value.VirtualAddrport != "") != grpcTunnelConfig.TypedSpec().Value.Enabled {
-			return xerrors.NewTaggedf[qtransform.DestroyOutputTag]("removed peer")
-		}
-	}
+	// grpcTunnelConfig, err := safe.ReaderGetByID[*siderolinkres.GRPCTunnelConfig](ctx, r, link.Metadata().ID())
```
utkuozdemir: Unintentional?

smira: intentional, but not finished, I think this code doesn't make any sense (I don't even understand why it did in the first place)

smira: ok, I figured what it does, but I don't think it's correct code still.

So the problem it's trying to solve (which I think is not a real one) is the following:

* someone flips the tunnel setting on/off per machine
* if Talos is disconnected, it will connect back next time on new settings
* if Talos is still connected, Omni can't force it go new way, as long as the old setting works (tunnel/no-tunnel), Talos will keep a happy path connection until the next reconnect
* this code is trying to _force_ disconnect by removing the peer, but I don't think it's correct the way conditions are placed (it might over-disconnect)

So if the Talos current path doesn't work, flipping the setting works already without this code block - Talos will follow the new setting on next Provision call, we don't need anything else to be done here.

smira: reworked this piece

internal/backend/runtime/omni/controllers/omni/link_status.go
```diff
@@ -92,11 +89,11 @@ func (handler *linkStatusHandler[T]) reconcileRunning(ctx context.Context, r con
 		//
 		// If grpcTunnelConfig is reverted within that 4 minute 35 second window, the controller
 		// will recreate the peer using the old mode, and the machine will become reachable again.
-		if err := handler.peers.Remove(ctx, siderolink.GetPeerID(linkStatus.TypedSpec().Value), link.Metadata()); err != nil {
-			return err
-		}
-
 		if (link.TypedSpec().Value.VirtualAddrport != "") != grpcTunnelConfig.TypedSpec().Value.Enabled {
```
smira, line 92: reworked this a bit to drop the peer _only if_ if the requested change is incompatible with the link

internal/backend/runtime/omni/controllers/omni/link_status.go
```diff
@@ -128,10 +125,6 @@ func (handler *linkStatusHandler[T]) reconcileTearingDown(ctx context.Context, _
 }
 
 func (handler *linkStatusHandler[T]) needsPeerUpdate(oldSpec, newSpec linkSpec) bool {
-	if oldSpec.GetVirtualAddrport() != "" && oldSpec.GetVirtualAddrport() != newSpec.GetVirtualAddrport() {
```
smira, line 131: this is not needed - we never need to update (actually recreate) the peer on VA change

internal/backend/runtime/omni/controllers/omni/link_status_test.go
```diff
@@ -54,10 +53,6 @@ func (h *fakeWireguardHandler) PeerEvent(_ context.Context, spec *specs.Sideroli
 	if deleted {
 		delete(h.peers, spec.NodePublicKey)
 	} else {
-		if _, ok := h.peers[spec.NodePublicKey]; ok {
```
smira, line 57: wrong assertion - library accepts perfectly peer updates

internal/pkg/siderolink/peers_pool.go
```diff
@@ -16,10 +16,12 @@ import (
...
-type PeerID struct {
-	key             string
-	virtualAddrport string
-}
+//
+// PeerID matches the public key of the Wireguard peer, the same key Wireguard uses
+// to identify the peer.
+//
+// This is used to deduplicate peers in the pool and to remove them when they are no longer needed.
+type PeerID string
```
smira, line 24: the root cause of all problems - the PeerID should match Wireguard's primary key (public key ID)

internal/pkg/siderolink/peers_pool.go
```diff
@@ -32,6 +34,10 @@ type ownerID struct {
 	namespace    string
 }
 
+func (o ownerID) String() string {
```
smira, line 37: for proper logging

internal/pkg/siderolink/peers_pool.go
```diff
@@ -78,24 +80,32 @@ func (pool *PeersPool) Add(ctx context.Context, spec *specs.SiderolinkSpec, owne
 
 	oid := getOwnerID(owner)
 
+	if err := pool.wgHandler.PeerEvent(ctx, spec, false); err != nil {
```
smira, line 83: important: push the update to the library always; library can handle dedup, no-op updates, etc.; we don't know and don't track what exectly changes (e.g. new VA address or lack of)

internal/pkg/siderolink/provision_test.go
```diff
@@ -1104,6 +1158,217 @@ func retryDestroy(ctx context.Context, st state.State, md *resource.Metadata) er
...
 }
 
+// Regression test for the WireGuard-over-gRPC pending-to-link handoff.
+//
+// The pending machine and the link it hands off to share a public key, and the WireGuard device
+// tracks peers by that key alone. If the link were to get its own virtual address-port, the peers
+// pool would track two entries over one device peer, and the pending machine cleanup would delete
+// the live link's device peer and revoke its tunnel token. The test drives the full handoff with
+// the real controllers and asserts that the peer and the token survive the cleanup.
+func TestGrpcTunnelPendingHandoffKeepsLinkPeer(t *testing.T) {
```
smira, line 1166: restored Utku's test with adjustments

smira:
There's some flakiness in the unit-test - will try to address that on Monday

utkuozdemir:
> There's some flakiness in the unit-test - will try to address that on Monday

Can be related to `time.Sleep` it does - I was about to suggest removing it.

utkuozdemir:
Looked into the flakiness, it is not the `time.Sleep`. It fails on the token assertion right after the handoff:

```
go test -race -count=30 -cpu=2 -run TestGrpcTunnelPendingHandoffKeepsLinkPeer ./internal/pkg/siderolink/
--- FAIL: TestGrpcTunnelPendingHandoffKeepsLinkPeer
    provision_test.go:1231: the link tunnel token must be allowed
```

If I understand correctly, the pending machine and the link share the key but have different virtual addresses, and `PeersPool.Add` now pushes the spec to the device on every add. The pending machine's controller reconciles a second time after adding its finalizer, and when that lands after the link's add, it installs the pending address as the token again and revokes the link's:

```go
pool.Add(ctx, pendingSpec, pending.Metadata()) // token = A
pool.Add(ctx, linkSpec, link.Metadata())       // token = B, A revoked
pool.Add(ctx, pendingSpec, pending.Metadata()) // token = A again, B revoked
```

In the test the two provisions are back to back, so it races. In prod the handoff takes seconds, so probably only the test is affected. Still, the two owners disagreeing on the spec is order dependent - should the pool prefer the link's spec while both exist, or is fixing the test enough?

internal/backend/runtime/omni/controllers/omni/link_status.go
```diff
@@ -92,11 +89,11 @@ func (handler *linkStatusHandler[T]) reconcileRunning(ctx context.Context, r con
 		//
 		// If grpcTunnelConfig is reverted within that 4 minute 35 second window, the controller
 		// will recreate the peer using the old mode, and the machine will become reachable again.
-		if err := handler.peers.Remove(ctx, siderolink.GetPeerID(linkStatus.TypedSpec().Value), link.Metadata()); err != nil {
-			return err
-		}
-
 		if (link.TypedSpec().Value.VirtualAddrport != "") != grpcTunnelConfig.TypedSpec().Value.Enabled {
+			if err := handler.peers.Remove(ctx, siderolink.GetPeerID(link.TypedSpec().Value), link.Metadata()); err != nil {
```
utkuozdemir: The registered key is the one recorded in `linkStatus`, same as in the `Remove` below. If the key rotated in the same reconcile, this removes the new key, which is not registered yet, and the old one stays in the pool.

```suggestion
			if err := handler.peers.Remove(ctx, siderolink.GetPeerID(linkStatus.TypedSpec().Value), link.Metadata()); err != nil {
```

utkuozdemir:
Might want to update the commit message, as it is stale now (the breaking commit is already reverted).

smira:
I came to the same conclusion, but I think this code is broken, it just doesn't make any sense what is going on there. 

This idea of "sharing" the peer across PendingMachine & Link is bad one, I don't even understand why it even makes any sense to do so. They should not co-exist at any given time, or they should actually be the same peer.

internal/pkg/siderolink/provision_test.go
```diff
@@ -1104,6 +1158,215 @@ func retryDestroy(ctx context.Context, st state.State, md *resource.Metadata) er
...
+	assert.True(t, device.hasPeer(pubKey))
+	assert.True(t, device.allowed.CheckToken(tokenOf(t, linkVA)))
+
+	// Across handoff, cleanup, and same-key re-provision the device must have
+	// seen exactly ONE add and NO removal for the key: the peer is never torn
+	// down and re-created. The settle delay lets any in-flight reconcile land
+	// before the negative assertion.
+	time.Sleep(250 * time.Millisecond)
+
+	assert.GreaterOrEqual(t, device.countEvents("add", pubKey), 1, "one ore more device adds for the key through handoff, cleanup and re-provision")
```
utkuozdemir, line 1294: This is always true now (there are five adds in the event log), can be dropped.

smira: I'm actually reworking this code path... I think there might be some assertion we can make here still

Unix4ever:
> I came to the same conclusion, but I think this code is broken, it just doesn't make any sense what is going on there.
> 
> This idea of "sharing" the peer across PendingMachine & Link is bad one, I don't even understand why it even makes any sense to do so. They should not co-exist at any given time, or they should actually be the same peer.

The idea is not to "share" it, but rather promote PendingMachine to a Link without recreating the peer.

smira:
> > I came to the same conclusion, but I think this code is broken, it just doesn't make any sense what is going on there.
> > This idea of "sharing" the peer across PendingMachine & Link is bad one, I don't even understand why it even makes any sense to do so. They should not co-exist at any given time, or they should actually be the same peer.
> 
> The idea is not to "share" it, but rather promote PendingMachine to a Link without recreating the peer.

The idea is understood, but recreating the peer is no problem - it will not do anything to the Wireguard connection (if done quickly enough).

The problem is that at some point two connection definitions exist for one Wireguard peer, and they do not agree on each other, and that is unresolvable. 

So at any point in time only one Wireguard peer definition might exist for a single public key, and this assumption is broken with concurrent PendingMachine & Link, that's why this design is not the correct one today. I will find a workaround, but it doesn't help to address the root cause.

smira:
I have the updated flow up for the review - the only change is keeping the VA across pending machine -> link transition.

utkuozdemir (approved):

PR #3402, opened 2026-09-10
fix: revert stabilizing the tunnel peer address across provisions

utkuozdemir:
Commit 194c629c kept a machine's tunnel peer address stable across provisions. This turned out to leave machines on the WireGuard-over-gRPC tunnel stuck after an Omni restart, where they fail to reconnect until they reboot.

Revert it to restore the previous behavior and work around the issue for the time being.

This reverts commit 194c629cf40e468d9b52cec321ea5954fc99021e. The regression test added with it was later moved into the provisioning tests, so it is removed here as well.

### DIFF
--- a/internal/pkg/siderolink/provision.go
+++ b/internal/pkg/siderolink/provision.go
@@ -275,13 +275,12 @@ func updateResource[T res](ctx context.Context,
 
 		var err error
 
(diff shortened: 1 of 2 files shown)


smira (approved):
#3399 should bring back no delay join fixed up

Unix4ever (approved):

PR #3406, opened 2026-09-11
feat: support adding user labels through join tokens

### LINKED ISSUE #3392: Support setting machine labels through Join token
V2 join tokens already support encoding extra data into them, but the scope of available things you can set there is too narrow, so it can be used only for the infra providers flow now.

Extend the available things that can be set there to arbitrary machine user labels. So that Terraform provider or other similar automation tool will be able to map the connected machines back to the internal resource state.

Unix4ever:
A join token can now carry user labels, set with `omnictl jointoken machine-config|kernel-args --set-labels key=value`. A machine joining with one is seeded with an ownerless MachineLabels resource, so the labels are the user's to change from then on; a machine that already has one keeps it.

Labels have to be signed, which introduces version 3. Such a token also carries a fingerprint of the join token that signed it, so that any token can be resolved without the secret being transmitted, not just the default one. Provider tokens keep resolving by the provider ID.

Fixes: https://github.com/siderolabs/omni/issues/3392


client/pkg/jointoken/jointoken.go
```diff
@@ -100,12 +185,122 @@ func NewWithExtraData(token, version string, extraData map[string]string) (JoinT
...
+				ErrInvalidLabels, key, omni.SystemLabelPrefix)
+		}
+
+		if err := validateLabel(key, value); err != nil {
+			return fmt.Errorf("%w: %w", ErrInvalidLabels, err)
+		}
+	}
+
+	return nil
+}
```
utkuozdemir, line 251: We have some label validation code already I think, in state validations. Maybe let's unify them: choose a common place and call it from both locations?

client/pkg/client/management/management.go
```diff
@@ -714,12 +714,37 @@ func (client *Client) CreateJoinToken(ctx context.Context, name string, ttl time
...
+
+// WithMachineLabels sets the initial labels to assign to any machine joining with the config.
+//
+// The labels are signed into the join token, and behave like the ones read from the Talos META
+// partition: the user can override them on the machine afterwards.
+func WithMachineLabels(labels map[string]string) MachineJoinConfigOption {
+	return func(req *management.GetMachineJoinConfigRequest) {
+		if len(labels) == 0 {
+			return
+		}
```
utkuozdemir: Is this necessary - is there any semantics anywhere on not nil but empty slice?

client/pkg/siderolink/siderolink.go
```diff
@@ -76,6 +78,26 @@ func WithMachine(machine *omni.Machine) JoinConfigOption {
...
+//
+// A machine joining with them is seeded with a MachineLabels resource, which the user owns and can
+// change afterwards. They are only applied to a machine Omni does not know yet: a machine that
+// already has MachineLabels keeps what it has. Keys carrying the omni.sidero.dev/ prefix are
+// rejected.
+func WithMachineLabels(labels map[string]string) JoinConfigOption {
+	return func(opts *JoinConfigOptions) {
+		if len(labels) == 0 {
+			return
+		}
```
utkuozdemir: Same question as above for here.

internal/pkg/siderolink/provision.go
```diff
@@ -468,19 +473,76 @@ func establishLink[T res](ctx context.Context, h *ProvisionHandler, logger *zap.
...
+	if err := st.Create(ctx, machineLabels); err != nil {
+		if state.IsConflictError(err) {
+			// the machine already has its labels, they are the user's to change now
+			return nil
+		}
+
+		return err
+	}
+
+	logger.Info("seeded the machine labels from the join token", zap.Strings("labels", keys))
```
utkuozdemir: Don't we want to log values as well?

client/pkg/client/management/management.go
Unix4ever: yeah, not necessary, nice catch

client/pkg/jointoken/jointoken.go
Unix4ever: merged into a single common module

Slessi (approved):

PR #3421, opened 2026-09-16
fix: honor context override in omnictl config info

immanuwell:
`config info --context production` shows production's URL and identity but labels them with the saved context name. A bit misleading

Use the resolved context name for both lookup and output. 
The saved selection stays untouched

Repro:
```sh
export OMNICONFIG="$(mktemp -d)/omniconfig"
omnictl config new
omnictl config add production --url https://prod.example.com
omnictl --context production config info
```

Before: `Current context: default` beside the production URL. 
After: `Current context: production`

Added CLI regression coverage. All tests under `client/pkg/omnictl/...` pass, lint is clean


client/pkg/omnictl/config.go
```diff
@@ -260,15 +260,20 @@ var configInfoCmd = &cobra.Command{
 
 		var result string
 
-		context, err := conf.GetContext(access.CmdFlags.Context)
+		contextName := access.CmdFlags.Context
+		if contextName == "" {
+			contextName = conf.Context
+		}
+
+		context, err := conf.GetContext(contextName)
```
Slessi, line 266: IMO the root cause of the issue is `GetContext` doing `contextName` resolution without conveying that resolution to the caller, and then callers like here have to duplicate that if they also need to know the resolved name.

I think we can fix it for all call sites by moving the resolution code out of `GetContext` and into say a `ResolveContextName` function, and callers that are reading from flags can call it before calling `GetContext`.

`omnictl config contexts` also seems incorrect for example, so could be fixed the same way:
```
❯ omnictl config contexts --context=production
CURRENT   NAME         URL
*         default      https://dev.example.com
          production   https://prod.example.com
```

client/pkg/omnictl/config_test.go
```diff
@@ -0,0 +1,98 @@
...
+
+	"github.com/siderolabs/omni/client/pkg/omnictl"
+)
+
+func TestConfigInfo(t *testing.T) {
+	t.Parallel()
+
+	// Run the command in a subprocess to capture its filtered stdout and isolate global CLI flags.
+	if os.Getenv("OMNI_TEST_CONFIG_INFO") == "1" {
+		omnictl.RootCmd.SetArgs(os.Args[3:])
```
Slessi: nit: works but maybe `slices.Index(os.Args, "--")+1` or something is nicer, instead of hardcoding the position

Slessi (commented):
Bug looks legit to me, thanks for reporting

client/pkg/omnictl/config.go
immanuwell: done, moved name resolution into ResolveContextName and fixed the marker in config contexts

client/pkg/omnictl/config_test.go
immanuwell: done, test now finds `--` instead of relying on its position

immanuwell:
@Slessi everything is addressed, please take a look

Slessi (approved):
LGTM!

